PR #222, opened 2025-04-22
Remove custom nightly pipeline

pentschev:
RAPIDSMPF has been added to the RAPIDS nightly workflows in https://github.com/rapidsai/workflows/pull/82 and as of today have been confirmed to run as expected, thus the custom nightly pipeline may be removed. The `workflow_call` was required to make the workflow reusable and is not needed anymore as well.


pentschev:
Thanks @gforsyth !

PR #231, opened 2025-04-28
Adding `ChunkBatch` interface - Updated 

nirandaperera:
This PR adds the `ChunkBatch` interface. 

It serializes each chunk data into a metadata buffer and a payload buffer. Additional information such as chunk batch ID, etc are also injected at the front of the metadata buffer.

Metadata buffer format:
```
| BatchHeader | [[MetadataMessageHeader, Metadata], ...] |
 ```
 Payload buffer format:
 ```
| [[Data, ...] |
```

waiting on #178 

Closes https://github.com/rapidsai/rapids-multi-gpu/issues/175


cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -122,6 +124,48 @@ class Buffer {
...
+     * @param dest Destination buffer.
+     * @param offset Offset of the destination buffer.
+     * @param stream CUDA stream to use for the copy.
+     * @returns Number of bytes written to the destination buffer.
+     *
+     * @throws std::logic_error if copy violates the bounds of the destination buffer.
+     */
+    [[nodiscard]] std::ptrdiff_t copy_to(
+        Buffer& dest, std::ptrdiff_t offset, rmm::cuda_stream_view stream
+    ) const;
```
wence-, line 167: Are these slice-based interfaces the best approach, or should we accept `span`-like things. That is, we could have a `slice` method that returns a `Span` over a `Buffer` and then `copy_to` could accept a `Span &dest`, this would push validation of offsets to the `Span` construction. etc...

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -94,6 +99,17 @@ class Chunk {
...
 
+    /**
+     * @brief Serializes this chunk into a given metadata message buffer.
+     *
+     * @param msg The metadata message as a serialized byte vector.
+     * @param offset The offset in the message buffer to start writing.
+     * @returns The number of bytes written to the message buffer.
+     */
+    [[nodiscard]] std::ptrdiff_t to_metadata_message(
+        std::vector<uint8_t>& msg, std::ptrdiff_t offset
```
wence-, line 110: Here if we accepted a `cuda::std::span` (C++20 but backported to C++14) over the message we wouldn't need offset argument separately, I think.

cpp/include/rapidsmpf/shuffler/chunk_batch.hpp
```diff
@@ -0,0 +1,390 @@
...
+        for (size_t i = 0; i < header()->num_chunks; ++i) {
+            assert(
+                std::ptrdiff_t(metadata_buffer_->size())
+                >= metadata_offset + chunk_metadata_header_size
+            );
+
+            auto const* chunk_header =
+                reinterpret_cast<Chunk::MetadataMessageHeader const*>(
+                    metadata_buffer_->data() + metadata_offset
+                );
```
wence-, line 180: *question*: I think accessing the metadata through this pointer _might_ be undefined behaviour if `metadata_offset % alignof(Chunk::MetadataMessageHeader)` is not zero. Can you assuage my doubts?

cpp/include/rapidsmpf/shuffler/chunk_batch.hpp
```diff
@@ -0,0 +1,390 @@
...
+  public:
+    using iterator = ChunkForwardIterator;  ///< Chunk iterator type
+
+    static constexpr std::ptrdiff_t chunk_metadata_header_size =
+        sizeof(Chunk::MetadataMessageHeader
+        );  ///< The size of the chunk metadata header in bytes.
+
+    /**
+     * @brief The structure of the batch header.
+     * @note This is allocated at the front of the the metadata buffer.
```
madsbk, line 45:
```suggestion
     * @note This is allocated at the front of the metadata buffer.
```

cpp/include/rapidsmpf/shuffler/chunk_batch.hpp
```diff
@@ -0,0 +1,390 @@
...
+     * signature:
+     * void(Chunk::MetadataMessageHeader const* chunk_header,
+     *      std::vector<uint8_t> const& metadata_buf,
+     *      std::ptrdiff_t metadata_offset,
+     *      Buffer const& payload_buf,
+     *      std::ptrdiff_t payload_offset)
+     * @param visitor visitor function
+     */
+    template <typename VisitorFn>
+    void visit_chunk_data(VisitorFn visitor) const {
```
wence-, line 164: I think I don't really like this interface and would prefer to only have the iterator-like access. WDYT?

cpp/include/rapidsmpf/shuffler/chunk_batch.hpp
```diff
@@ -0,0 +1,390 @@
...
+    mutable std::unique_ptr<iterator>
+        cached_begin_iter_;  ///< cached begin iterator, as creating an iterator object is
+                             ///< relatively expensive. This allows calling `begin()`
+                             ///< trivially multiple times.
+};
+
+/**
+ * @brief Forward iterator of chunks in the chunk batch.
+ */
+class ChunkForwardIterator {
```
wence-, line 253: I think I would prefer to implement `operator()[]` and `.at` on the `ChunkBatch` class so that we could just have a `RandomAccessIterator`. I think this might require a few changes in the metadata we store in a `ChunkBatch` (basically we need to keep the exclusive scan of the offsets around).

cpp/src/buffer/buffer.cpp
```diff
@@ -99,4 +103,120 @@ std::unique_ptr<Buffer> Buffer::copy(MemoryType target, rmm::cuda_stream_view st
...
+    RAPIDSMPF_EXPECTS(
+        offset <= std::ptrdiff_t(size), "offset can't be greater than size"
+    );
+    RAPIDSMPF_EXPECTS(
+        offset + length <= std::ptrdiff_t(size),
+        "offset + length can't be greater than size"
+    );
+
+    if (mem_type() == target) {
+        return copy_slice(offset, length, stream);
```
wence-, line 157: *nit*: This is why I don't really like these "C-like" APIs. It's very easy to switch up `offset` and `length` since they are the same type.

wence- (commented):
Partial review

cpp/src/shuffler/chunk_batch.cpp
```diff
@@ -0,0 +1,262 @@
...
+        metadata->size() >= batch_header_size,
+        "metadata buffer size is less than the header size"
+    );
+    // first create a chunk batch
+    ChunkBatch batch{std::move(metadata), std::move(payload_data)};
+
+    // visit chunk data and verify if the given buffers adhere to the format
+    size_t visited_metadata_size = batch_header_size;
+    size_t visited_payload_size = 0;
+    batch.visit_chunk_data([&](Chunk::MetadataMessageHeader const* chunk_header,
```
madsbk, line 126: Could we use `ChunkForwardIterator` here?

madsbk (commented):
Consider implementing the Buffer slicing (or `operator()[]`/`.at`) in a standalone PR with its own testing?

cpp/src/shuffler/postbox.cpp
```diff
@@ -31,7 +31,10 @@ std::unordered_map<ChunkID, Chunk> PostBox::extract(PartID pid) {
 std::vector<Chunk> PostBox::extract_all() {
     std::lock_guard const lock(mutex_);
     std::vector<Chunk> ret;
+
     for (auto& [_, chunks] : pigeonhole_) {
+        // reserve for the chunks size
+        ret.reserve(ret.size() + chunks.size());
```
wence-, line 37: This is quadratic since you're not growing geometrically so you have to do `num_pigeonhole_entries * num_chunks_per_pigeonhole` allocations.

cpp/tests/test_chunk_batch.cu
```diff
@@ -0,0 +1,251 @@
...
+/**
+ * Types of chunks
+ * 1. Chunks with control messages
+ * 2. Chunks with no data
+ * 3. Chunks with data
+ *    a. Device data
+ *    b. Host data
+ */
+
+// Parametarized test for MemoryType and types of chunks
```
wence-, line 78:
```suggestion
// Parametrized test for MemoryType and types of chunks
```

cpp/tests/test_chunk_batch.cu
```diff
@@ -0,0 +1,251 @@
...
+        } else if (chunks_type == "single_data_middle") {
+            chunks.emplace_back(3, 3, 100);
+            chunks.emplace_back(4, 4, 0, len, copy_metadata(), copy_data());
+            chunks.emplace_back(5, 5, 101);
+        } else if (chunks_type == "single_data_end") {
+            chunks.emplace_back(3, 3, 100);
+            chunks.emplace_back(5, 5, 101);
+            chunks.emplace_back(4, 4, 0, len, copy_metadata(), copy_data());
+        } else {
+            RAPIDSMPF_EXPECTS(chunks_type == "empty", "unkown chunk type " + chunks_type);
```
wence-, line 158:
```suggestion
            RAPIDSMPF_EXPECTS(chunks_type == "empty", "unknown chunk type " + chunks_type);
```

nirandaperera:
@wence- thanks. Let's close this. #291 succeeds this.

PR #236, opened 2025-04-30
Organize chunks by destination rank in the `PostBox`

nirandaperera:
As discussed [here](https://docs.google.com/document/d/1lFbAAbwFJRvZ5ie3Svg66aUIc5TacJ5oT7M3VYRnHw0/edit?tab=t.0#heading=h.qexm2qchktwi), we have decided to organize `Chunk`s by the destination rank in the `PostBox`. This PR adds this functionality. 
There could be some penalty to preserve the current API functionality. A new `unordered_set` was required to track the `ChunkID`s in the PostBox, and it needs to be updated correctly as we extract Chunks. 

This PR also renames `inbox_` and `outbox_` in the `Shuffler`.


cpp/src/shuffler/postbox.cpp
```diff
@@ -48,10 +114,10 @@ std::vector<std::tuple<PartID, ChunkID, std::size_t>> PostBox::search(MemoryType
 ) const {
     std::lock_guard const lock(mutex_);
     std::vector<std::tuple<PartID, ChunkID, std::size_t>> ret;
-    for (auto& [pid, chunks] : pigeonhole_) {
-        for (auto& [cid, chunk] : chunks) {
+    for (auto const& [pid, chunks] : pigeonhole_) {
+        for (auto const& chunk : chunks) {
             if (chunk.gpu_data && chunk.gpu_data->mem_type() == mem_type) {
-                ret.emplace_back(pid, cid, chunk.gpu_data->size);
+                ret.emplace_back(pid, chunk.cid, chunk.gpu_data->size);
```
wence-, line 120: I think this whole changeset seems much more complicated than it needs to be.

For example `PostBox::search` is only used in `postbox_spilling`. It doesn't need to return `PartID, ChunkID` pairs, it could just as easily return `Rank, ChunkID` pairs. I think.

cpp/src/shuffler/postbox.cpp
```diff
@@ -12,34 +12,100 @@
...
+    );
+
+    // insert into the rank-based pigeonhole
+    pigeonhole_[owner].emplace_back(std::move(chunk));
 }
 
 Chunk PostBox::extract(PartID pid, ChunkID cid) {
     std::lock_guard const lock(mutex_);
-    return extract_item(pigeonhole_.at(pid), cid).second;
+    std::cout << "extracting chunk " << cid << " pid " << pid << std::endl;
```
wence-, line 34: Debugging left in.

cpp/src/shuffler/postbox.cpp
```diff
@@ -12,34 +12,100 @@
...
+    auto [_, inserted] = chunk_ids_.emplace(chunk.cid);
+    RAPIDSMPF_EXPECTS(
+        inserted, "PostBox.insert(): chunk already exist " + std::to_string(chunk.cid)
+    );
+
+    // insert into the rank-based pigeonhole
+    pigeonhole_[owner].emplace_back(std::move(chunk));
 }
 
 Chunk PostBox::extract(PartID pid, ChunkID cid) {
```
wence-, line 32: Again, this function is only called by the `postbox_spilling` so if `PostBox::search` were adapted to return rank, id pairs, we wouldn't need to change this definition really.

cpp/include/rapidsmpf/shuffler/postbox.hpp
```diff
@@ -88,8 +104,16 @@ class PostBox {
   private:
     // TODO: more fine-grained locking e.g. by locking each partition individually.
     mutable std::mutex mutex_;
-    std::unordered_map<PartID, std::unordered_map<ChunkID, Chunk>>
-        pigeonhole_;  ///< Storage for chunks, organized by partition and chunk ID.
```
wence-, line 92: It seems like most of the changes are required because you change the internal storage here.

How about making the `PostBox` templated over the key type instead like this: https://github.com/wence-/rapidsmpf/commit/04e32dc44200b3641f73054112a92fe552f8df11

Then we don't need any changes in the shuffler or these implementations other than to provide a key mapper when constructing the postboxes and inlining the definitions in the header file.

wence- (changes requested):
I wonder if we can simplify things by templating the postbox over the key type, see comments and linked commit.

cpp/include/rapidsmpf/shuffler/postbox.hpp
nirandaperera: I like the idea. I was under the impression that we wanted to change the internal storage as well.

wence-: I think not in the first round. Not least because where we end up might not want the internal storage change that is used here.

nirandaperera: Okay. Let me add the changes

nirandaperera: @wence- I added the suggested templated impl now

cpp/include/rapidsmpf/shuffler/postbox.hpp
```diff
@@ -88,7 +116,9 @@ class PostBox {
   private:
     // TODO: more fine-grained locking e.g. by locking each partition individually.
     mutable std::mutex mutex_;
-    std::unordered_map<PartID, std::unordered_map<ChunkID, Chunk>>
+    std::function<key_type(PartID)>
+        key_map_fn_;  ///< Function to map partition IDs to keys.
+    std::unordered_map<key_type, std::unordered_map<ChunkID, Chunk>>
```
madsbk, line 121: Update comment

cpp/src/shuffler/shuffler.cpp
```diff
@@ -581,7 +594,7 @@ detail::ChunkID Shuffler::get_new_cid() {
 
 std::string Shuffler::str() const {
     std::stringstream ss;
-    ss << "Shuffler(inbox=" << inbox_ << ", outbox=" << outbox_ << ", "
+    ss << "Shuffler(inbox=" << outgoing_chunks_ << ", outbox=" << received_chunks_ << ", "
```
madsbk, line 597:
```suggestion
    ss << "Shuffler(outgoing=" << outgoing_chunks_ << ", received=" << received_chunks_ << ", "
```

cpp/include/rapidsmpf/shuffler/shuffler.hpp
```diff
@@ -252,8 +252,8 @@ class Shuffler {
     rmm::cuda_stream_view stream_;
     BufferResource* br_;
     bool active_{true};
-    detail::PostBox inbox_;
-    detail::PostBox outbox_;
+    detail::PostBox<Rank> outgoing_chunks_;
+    detail::PostBox<PartID> received_chunks_;
```
madsbk, line 256: I would be good with some comments here that clearly explains the terminology of outgoing and receiving :)

madsbk (approved):
Nice work @nirandaperera, I only have some minor suggestions

cpp/src/shuffler/postbox.cpp
```diff
@@ -80,5 +95,8 @@ std::string PostBox::str() const {
     return ss.str();
 }
 
+// Explicit instantiation for PartID and Rank
+template class PostBox<PartID>;
+template class PostBox<Rank>;
```
wence-, line 100: *nit*: This split of the template class into definition in header and implementation in cpp file, along with this explicit instantiation, means that to use a hypothetical newly templated `PostBox` we must remember to explicitly instantiate the template type _somewhere_. 

This is in contrast to the usual case where the definition of the template class in the header also contains the implementation.

Is this a deliberate choice?

cpp/src/shuffler/shuffler.cpp
```diff
@@ -381,6 +383,16 @@ Shuffler::Shuffler(
...
       stream_{stream},
       br_{br},
+      outgoing_chunks_{
+          [this](PartID pid) -> Rank {
+              return this->partition_owner(this->comm_, pid);
+          },  // extract Rank from pid
+          static_cast<std::size_t>(comm->nranks())
+      },
+      received_chunks_{
+          [](PartID pid) -> PartID { return pid; },  // identity mapping
```
wence-, line 393: Reminder to use `std::identity` once we're on C++20.

wence- (approved):
Tiny nit, looks good.

cpp/src/shuffler/postbox.cpp
madsbk: I have no strong opinion,  but I think the implementation is small/simple enough to include everything in `postbox.hpp`.

nirandaperera: @wence- It was a deliberate choice TBH. I wanted to make sure that we know explicitly how this PostBox templates are being used. In the impl cpp, we know the will be a PostBox with partition ID as the key as well as Rank as the key.

nirandaperera:
\merge

PR #237, opened 2025-05-02
[Refactor] Dask worker context

madsbk:
Clean up the RapidsMPF specific attributes associated a Dask worker.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
+++ b/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
@@ -13,7 +13,6 @@ from rmm.pylibrmm.stream import DEFAULT_STREAM
 from rapidsmpf.integrations.dask.core import (
     DataFrameT,
     get_dask_client,
-    get_progress_thread,
     get_worker_context,
     get_worker_rank,
     global_rmpf_barrier,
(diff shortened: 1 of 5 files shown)


python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -70,29 +68,32 @@ def get_shuffler(
...
-                    f" Shufflers: {dask_worker._rmp_shufflers}"
+                    f" Shufflers: {ctx.shufflers}"
                 )
-            dask_worker._rmp_shufflers[shuffle_id] = Shuffler(
-                get_comm(dask_worker),
-                get_progress_thread(dask_worker),
+            assert ctx.br is not None
+            assert ctx.comm is not None
+            assert ctx.progress_thread is not None
+            ctx.shufflers[shuffle_id] = Shuffler(
```
wence-, line 87: *question*: I see this is not handled at the moment anyway, but whose job is it to clean up these created shufflers once we are done with them?

I note that shuffle ids are just a continuously incrementing integer, but `op_id` in the shuffler is an `unsigned char` so there can only be 256 concurrent shuffles (which is probably fine). But we _should_ release ids once we are done with a shuffle.

And ideally we would restrict this `ctx.shufflers` to only have 256 slots so we can detect and produce an error if we ever go over the 256 concurrent shuffle limit

madsbk: I don't think we even support 256 consecutive shuffles.
Let's fix this in a new PR: https://github.com/rapidsai/rapidsmpf/issues/239

pentschev (approved):
Minor docs suggestions, otherwise LGTM.

wence- (approved):

PR #238, opened 2025-05-02
Move Buffer::overloaded dispatch helper to top level

wence-:
GCC versions prior to v12.0 had a bug in class template argument deduction rules: they were incorrectly required to live at namespace scope.

For example, the following definition

    struct Foo {
      template <typename... Ts> struct overloaded : Ts... {
        using Ts::operator()...;
      };
      template <typename... Ts>
      overloaded(Ts...) -> overloaded<Ts...>;
    };

Produces the compile error

    error: deduction guide 'Foo::overloaded(Ts ...) -> Foo::overloaded<Ts ...>' must be declared at namespace scope
        6 |   overloaded(Ts...) -> overloaded<Ts...>;
          |   ^~~~~~~~~~

In contrast, both clang, and later versions of GCC, correctly accept this definition. See
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79501 and https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100983 for details.

Since we still advertise support for GCC 11, and pip-based devcontainer builds use GCC 11, work around this bug by moving the overloaded dispatch helper (used in methods that visit the variant storage) to a top-level anonymous namespace.

### DIFF
--- a/cpp/include/rapidsmpf/buffer/buffer.hpp
+++ b/cpp/include/rapidsmpf/buffer/buffer.hpp
@@ -26,6 +26,18 @@ enum class MemoryType : int {
 /// @brief Array of all the different memory types.
 constexpr std::array<MemoryType, 2> MEMORY_TYPES{{MemoryType::DEVICE, MemoryType::HOST}};
 
+namespace {
+/// @brief Helper for overloaded lambdas using std::visit.
+template <class... Ts>
+struct overloaded : Ts... {
+    using Ts::operator()...;
+};
+/// @brief Explicit deduction guide
+template <class... Ts>
+overloaded(Ts...) -> overloaded<Ts...>;
+
+}  // namespace
+
 /**
  * @brief Buffer representing device or host memory.
  *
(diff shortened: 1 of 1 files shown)


wence-:
Noticed as part of https://github.com/rapidsai/devcontainers/pull/497

pentschev (commented):
I don't oppose this change, but IIUC this is _only_ required for CUDA 11.x which we explicitly don't support. See conditions where [gcc 11.x is required in cuDF](https://github.com/rapidsai/cudf/blob/b62f24b3dbe213d6a8474cc3c8c8af8091714f0e/dependencies.yaml#L430-L450) versus the only condition [rapidsmpf satisfies](https://github.com/rapidsai/rapidsmpf/blob/74142e341d40cf4bb1fddc956c76ca837c9f7aad/dependencies.yaml#L70-L81). To me it seems that the correct solution is to not try to compile rapidsmpf with CUDA 11.x in devcontainers.

nirandaperera (approved):
LGTM

wence-:
> I don't oppose this change, but IIUC this is _only_ required for CUDA 11.x which we explicitly don't support. See conditions where [gcc 11.x is required in cuDF](https://github.com/rapidsai/cudf/blob/b62f24b3dbe213d6a8474cc3c8c8af8091714f0e/dependencies.yaml#L430-L450) versus the only condition [rapidsmpf satisfies](https://github.com/rapidsai/rapidsmpf/blob/74142e341d40cf4bb1fddc956c76ca837c9f7aad/dependencies.yaml#L70-L81). To me it seems that the correct solution is to not try to compile rapidsmpf with CUDA 11.x in devcontainers.

This is required for the pip devcontainer which has GCC 11 as a host. The nvcc host compiler requirement for cuda 12.8 is GCC >= 10; cudf needs GCC >= 11.4.

PR #242, opened 2025-05-02
Ensure data are ready for comms libraries with events

wence-:
Rather than using device synchronisation, use events to track work on streams for buffers and only hand over the buffer data pointer to the comms library when the event is ready.

### DIFF
--- a/cpp/include/rapidsmpf/buffer/buffer.hpp
+++ b/cpp/include/rapidsmpf/buffer/buffer.hpp
@@ -9,8 +9,10 @@
 #include <variant>
 #include <vector>
 
+#include <rmm/cuda_stream_view.hpp>
 #include <rmm/device_buffer.hpp>
 
+#include <rapidsmpf/cuda_event.hpp>
 #include <rapidsmpf/error.hpp>
 
 namespace rapidsmpf {
@@ -32,6 +34,9 @@ constexpr std::array<MemoryType, 2> MEMORY_TYPES{{MemoryType::DEVICE, MemoryType
  * @note The constructors are private, use `BufferResource` to construct buffers.
  * @note The memory type (e.g., host or device) is constant and cannot change during
  * the buffer's lifetime.
+ * @note A buffer is a stream-ordered object, when passing to a
+ * library which is not stream-aware one must ensure that `is_ready`
+ * returns `true` otherwise behaviour is undefined.
  */
 class Buffer {
     friend class BufferResource;
@@ -106,6 +111,14 @@ class Buffer {
     [[nodiscard]] void const* data() const;
 
     /**
+     * @brief Check if the buffer is ready to be used.
+     *
+     * @return true If all stream-ordered work on the buffer is
+     * complete.
+     */
+    [[nodiscard]] bool is_ready() const;
+
+    /**
      * @brief Get the memory type of the buffer.
      *
      * @return The memory type of the buffer.
@@ -134,22 +147,34 @@ class Buffer {
      *
      * @param host_buffer A unique pointer to a vector containing host memory.
      * @param br Buffer resource for memory allocation.
+     * @param event Event tracking all stream-ordered work populating
+     * `host_buffer`. Discarded if the `host_buffer` is zero-sized.
      *
      * @throws std::invalid_argument if `host_buffer` is null.
      */
-    Buffer(std::unique_ptr<std::vector<uint8_t>> host_buffer, BufferResource* br);
+    Buffer(
+        std::unique_ptr<std::vector<uint8_t>> host_buffer,
+        BufferResource* br,
+        std::shared_ptr<Event> event
+    );
 
     /**
      * @brief Construct a Buffer from device memory.
      *
      * @param device_buffer A unique pointer to a device buffer.
      * @param br Buffer resource for memory allocation.
+     * @param event Event tracking all stream-ordered work populating
+     * `device_buffer`. Discarded if the `device_buffer` is zero-sized.
      *
      * @throws std::invalid_argument if `device_buffer` is null.
      * @throws std::invalid_argument if `stream` or `br->mr` isn't the same used by
      * `device_buffer`.
      */
-    Buffer(std::unique_ptr<rmm::device_buffer> device_buffer, BufferResource* br);
+    Buffer(
+        std::unique_ptr<rmm::device_buffer> device_buffer,
+        BufferResource* br,
+        std::shared_ptr<Event> event
+    );
 
     /**
      * @brief Access the underlying host memory buffer.
@@ -208,6 +233,9 @@ class Buffer {
     /// @brief The underlying storage host memory or device memory buffer (where
     /// applicable).
     StorageT storage_;
+
+    std::shared_ptr<Event> event_{nullptr
+    };  ///< Event used to ensure work to populate gpu data is complete.
 };
 
 }  // namespace rapidsmpf
--- a/cpp/src/buffer/buffer.cpp
+++ b/cpp/src/buffer/buffer.cpp
@@ -2,10 +2,15 @@
  * SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION & AFFILIATES.
  * SPDX-License-Identifier: Apache-2.0
  */
+#include <memory>
 #include <stdexcept>
+#include <utility>
+
+#include <rmm/cuda_stream_view.hpp>
 
 #include <rapidsmpf/buffer/buffer.hpp>
 #include <rapidsmpf/buffer/resource.hpp>
+#include <rapidsmpf/cuda_event.hpp>
 
 namespace rapidsmpf {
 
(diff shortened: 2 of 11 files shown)


wence-:
An alternative approach (borrowing a bunch of the same ideas) to #225.

cpp/src/cuda_event.cpp
```diff
@@ -0,0 +1,60 @@
...
+    }
+}
+
+void Event::add_as_dependency_to(rmm::cuda_stream_view stream) const {
+    if (!query()) {
+        RAPIDSMPF_CUDA_TRY(
+            cudaStreamWaitEvent(stream.value(), event_, cudaEventWaitDefault)
+        );
+    }
+}
```
wence-, line 36: This is preparing for if we start using an internal stream inside the shuffler that is distinct from the user stream. In that case, when we copy a user buffer on the internal stream we must add a dependency on the work on the user's stream.

cpp/src/shuffler/shuffler.cpp
```diff
@@ -485,7 +502,10 @@ void Shuffler::insert(std::unordered_map<PartID, PackedData>&& chunks) {
                 continue;
             }
             auto chunk = create_chunk(
-                pid, std::move(packed_data.metadata), std::move(packed_data.gpu_data)
+                pid,
+                std::move(packed_data.metadata),
+                std::move(packed_data.gpu_data),
+                event
```
wence-, line 508: Chunk creation takes an event, and forwards it to the buffer.

cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -105,6 +111,28 @@ class Buffer {
      */
     [[nodiscard]] void const* data() const;
 
+    /**
+     * @brief ensure all stream-ordered work to populate the buffer is completed.
+     */
+    void synchronize() const;
```
pentschev, line 117: Ok, but how do we prevent abuse then? In my mind, providing an interface like this is recipe to "go do the easy thing".

cpp/src/buffer/resource.cpp
```diff
@@ -88,15 +90,20 @@ std::unique_ptr<Buffer> BufferResource::allocate(
...
-        break;
+        {
+            auto event = std::make_shared<Event>();
+            auto buf = std::make_unique<rmm::device_buffer>(size, stream, device_mr_);
+            event->record(stream);
+            ret =
+                std::unique_ptr<Buffer>(new Buffer{std::move(buf), this, std::move(event)}
+                );
+            break;
+        }
```
wence-, line 106: Device allocations do, so make a new event and record the allocation.

cpp/src/buffer/buffer.cpp
```diff
@@ -18,24 +23,37 @@ template <typename T>
...
+)
     : br{br},
       size{host_buffer ? host_buffer->size() : 0},
-      storage_{std::move(host_buffer)} {
+      storage_{std::move(host_buffer)},
+      event_{size > 0 ? std::move(event) : nullptr} {
     RAPIDSMPF_EXPECTS(
         std::get<HostStorageT>(storage_) != nullptr, "the host_buffer cannot be NULL"
     );
     RAPIDSMPF_EXPECTS(br != nullptr, "the BufferResource cannot be NULL");
```
wence-, line 38: Can't add the check for `event_` being non-null here if the `size` is non-zero because it might not be necessary.

cpp/src/buffer/buffer.cpp
```diff
@@ -73,30 +96,55 @@ std::unique_ptr<Buffer> Buffer::copy(MemoryType target, rmm::cuda_stream_view st
...
         overloaded{
             [&](const HostStorageT& storage) -> std::unique_ptr<Buffer> {
-                return std::unique_ptr<Buffer>(new Buffer{
-                    std::make_unique<rmm::device_buffer>(
-                        storage->data(), storage->size(), stream, br->device_mr()
-                    ),
-                    br
-                });
+                // Have to ensure that any async work is complete.
+                synchronize();
```
pentschev, line 104: Why?

cpp/src/buffer/buffer.cpp
```diff
@@ -50,17 +68,22 @@ std::unique_ptr<Buffer> Buffer::copy(rmm::cuda_stream_view stream) const {
     return std::visit(
         overloaded{
             [&](const HostStorageT& storage) -> std::unique_ptr<Buffer> {
-                return std::unique_ptr<Buffer>(
-                    new Buffer{std::make_unique<std::vector<uint8_t>>(*storage), br}
-                );
-            },
-            [&](const DeviceStorageT& storage) -> std::unique_ptr<Buffer> {
+                // Have to ensure that any async work is complete.
+                synchronize();
```
pentschev, line 72: Why do you have to ensure async work is complete? Your source is host memory, seems like anything that has been done with it should have been synchronized before, it should never be up to the consumer to magically figure out how the host buffer has been processed, how do you know it is even the right stream then?

cpp/src/buffer/buffer.cpp
```diff
@@ -73,30 +96,55 @@ std::unique_ptr<Buffer> Buffer::copy(MemoryType target, rmm::cuda_stream_view st
...
+                auto buf = std::make_unique<rmm::device_buffer>(
+                    storage->data(), storage->size(), stream, br->device_mr()
+                );
+                event->record(stream);
+
+                return std::unique_ptr<Buffer>(
+                    new Buffer{std::move(buf), br, std::move(event)}
+                );
```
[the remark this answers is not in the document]
wence-: Actually, this one can add a stream dependency as well...

cpp/src/buffer/buffer.cpp
wence-: Suppose you spill a device side buffer to `Buffer` as host memory. You did that with `cudaMemcpyAsync`. Now you have a `Buffer` with host memory, but it might not be `ready`. So if you then copy it to a new `Buffer` with host memory you can either fail if it's not `is_ready` or you synchronize.

cpp/src/buffer/buffer.cpp
wence-: This one was wrong, so I fixed it with a stream dependency.

wence-:
> and add_as_dependency_to which after a high-level look I still fail to understand how it adds any important information that cannot be capture by recording the event.

Suppose the application stream is `streamA` and the shuffler is internally using a stream `streamB`.

When we take ownership of the application buffer we record the state of `streamA` in an `event`.

Then suppose we spill that data, but we're using `streamB` to call the `cudaMemcpyAsync`. Unless we insert a dependency between `streamA` and `streamB` we will have an invalid result. The way to do this is to call `cudaStreamWaitEvent(streamB, event)` which is what `add_as_dependency_to`.

cpp/src/buffer/buffer.cpp
pentschev: Yeah, ok, but when would that happen in practice? I'd argue the consumer needs to checking the `Buffer` (or whatever other object) that delivered the host memory should be checked if `is_ready()`, otherwise, to use your words, we are now putting a "big hammer" (although slightly smaller than device sync) here. What is the argument for whoever called `Buffer::copy()` not to have ensured that in the first place?

pentschev:
> Suppose the application stream is `streamA` and the shuffler is internally using a stream `streamB`.
> 
> When we take ownership of the application buffer we record the state of `streamA` in an `event`.
> 
> Then suppose we spill that data, but we're using `streamB` to call the `cudaMemcpyAsync`. Unless we insert a dependency between `streamA` and `streamB` we will have an invalid result. The way to do this is to call `cudaStreamWaitEvent(streamB, event)` which is what `add_as_dependency_to` does.

Fair, but to me this is beyond the scope of this change, although since it's already here it's fine, otherwise I would have argued it should be part of a follow-up PR where `Shuffler` uses another stream. I mean we could just do everything we can think of are good changes in one PR, but I feel there's value in keeping things minimal so we don't risk introducing more complexity than necessary and simultaneously risking introducing another bug, but that's just me.

cpp/src/buffer/buffer.cpp
wence-: If we don't do the synchronisation then every time you do a buffer host-to-host copy you must check that that is what you're doing and then `sync`. Or else check `is_ready` and then if you're not ready, push it back into a copy queue?

That seems like a very difficult to manage API.

wence-:
> > Suppose the application stream is `streamA` and the shuffler is internally using a stream `streamB`.
> > When we take ownership of the application buffer we record the state of `streamA` in an `event`.
> > Then suppose we spill that data, but we're using `streamB` to call the `cudaMemcpyAsync`. Unless we insert a dependency between `streamA` and `streamB` we will have an invalid result. The way to do this is to call `cudaStreamWaitEvent(streamB, event)` which is what `add_as_dependency_to` does.
> 
> Fair, but to me this is beyond the scope of this change, although since it's already here it's fine, otherwise I would have argued it should be part of a follow-up PR where `Shuffler` uses another stream. I mean we could just do everything we can think of are good changes in one PR, but I feel there's value in keeping things minimal so we don't risk introducing more complexity than necessary and simultaneously risking introducing another bug, but that's just me.


I can certainly rip it out, but we still need to think about what to do about host-to-host copies (given that the source buffer in that scenario might have stream-queued work).

wence-:
> I can certainly rip it out, but we still need to think about what to do about host-to-host copies (given that the source buffer in that scenario might have stream-queued work).

Added a `RAPIDSMPF_EXPECTS(is_ready())` instead and removed all the `synchronize` APIs and `add_as_dependency_to` calls.

cpp/src/buffer/buffer.cpp
wence-: In almost all cases the source host buffer will have no event and thus the sync is a noop. But sometimes it might

madsbk:

PR #248, opened 2025-05-06
Python spilling: use chunked packing with a staging buffer

madsbk:
First really useful when https://github.com/rapidsai/cudf/pull/18614 is merged.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/integrations/dask/core.py
+++ b/python/rapidsmpf/rapidsmpf/integrations/dask/core.py
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+import functools
 import logging
 import threading
 import weakref
@@ -68,6 +69,7 @@ class DaskWorkerContext:
     spill_collection: SpillCollection = field(default_factory=SpillCollection)
     statistics: Statistics | None = None
     shufflers: dict[int, Shuffler] = field(default_factory=dict)
+    spill_staging_buffer: rmm.DeviceBuffer = None
 
 
 def get_worker_context(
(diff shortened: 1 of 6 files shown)


python/rapidsmpf/rapidsmpf/integrations/dask/spilling.py
```diff
@@ -130,7 +148,14 @@ def spill(self, amount: int) -> int:
...
             if on_device is not None and self._on_host is None:
                 ret = self.approx_spillable_amount()
-                self._on_host = dask_dumps(on_device)
+                self._on_host = dask_dumps(
+                    on_device,
+                    context={
+                        "stream": stream,
+                        "device_mr": device_mr,
+                        "staging_device_buffer": staging_device_buffer,
+                    },
```
pentschev, line 157: Are all these attributes serializable?

pentschev (approved):
Left a question, but otherwise LGTM. Thanks @madsbk .

python/rapidsmpf/rapidsmpf/integrations/dask/spilling.py
madsbk: no, but they don't have to be. AFAICT, `context` is only passed down to the local serialization function.

pentschev: Ok, I think you're right. I wasn't familiar with the purpose of `context`, thanks for confirming.

madsbk:
Thanks @pentschev

wence- (commented):
Just to note, to fix in a followup now. We need to sync the stream before unspilling

python/rapidsmpf/rapidsmpf/integrations/dask/spilling.py
```diff
@@ -130,7 +148,14 @@ def spill(self, amount: int) -> int:
             on_device = self._on_device
             if on_device is not None and self._on_host is None:
                 ret = self.approx_spillable_amount()
-                self._on_host = dask_dumps(on_device)
+                self._on_host = dask_dumps(
+                    on_device,
+                    context={
+                        "stream": stream,
```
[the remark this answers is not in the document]
madsbk: Currently, we are using `DEFAULT_STREAM` so no immediately action is needed. But you are right, we need a unified stream (and rmm) strategy for both the C++ and Python API!

python/rapidsmpf/rapidsmpf/integrations/dask/core.py
```diff
@@ -250,12 +251,51 @@ def rmpf_worker_setup(
...
+                finally:
+                    spill_staging_buffer_lock.release()
+            return ctx.spill_collection.spill(
+                amount, stream=DEFAULT_STREAM, device_mr=mr
+            )
+
+        # Add the spill function using a negative priority (-10) such that spilling
+        # of internal shuffle buffers (non-python objects) have higher priority than
+        # spilling of the Python objects in the collection.
+        ctx.br.spill_manager.add_spill_function(func=spill_func, priority=-10)
```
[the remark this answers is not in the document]
madsbk: Querying the spill manager for info such as priorities is a good idea!

PR #251, opened 2025-05-06
Chunk with multiple messages

nirandaperera:
Chunk with multiple messages. This PR only moves the existing `Chunk` class to the new impl and it would only have 1 message in it. 

This class has two buffers:
 - `metadata_`: The metadata buffer that contains information about the messages in the  chunks and the concatenated metadata of the messages.
 - `data_`: The data buffer that contains the concatenateddata of the messages in the  chunk.

All the chunk information will be encoded to the metadata_ buffer as follows. 
The metadata_ buffer uses the following format:
 - `chunk_id`: uint64_t, ID of the chunk
 - `n_elements`: size_t, Number of messages in the chunk
 - `[partition_ids]`: vector<PartID>, Partition IDs of the messages, size = n_elements
 - `[expected_num_chunks]`: vector<size_t>, Expected number of chunks of the messages, size = n_elements
 - `[meta_offsets]`: vector<uint32_t>, Offsets (excluding 0) of the metadata sizes of the messages, size = n_elements
 - `[data_offsets]`: vector<uint64_t>, Offsets (excluding 0) of the data sizes of the messages, size = n_elements
 - `[concat_metadata]`: vector<uint8_t>, Concatenated metadata of the messages, size = meta_offsets[n_elements - 1]

For a chunk with N messages with M bytes of concat metadata the size of metadata_ buffer is `sizeof(ChunkID) + sizeof(size_t) + N (sizeof(PartID) + sizeof(size_t) + sizeof(uint32_t) + sizeof(uint64_t)) + M = 16 + N 24 + M` bytes.

For a chunk with a single control message, the size of the metadata_ buffer is `sizeof(ChunkID) + sizeof(PartID)+ 2*sizeof(size_t) + sizeof(uint32_t) + sizeof(uint64_t) = 40` bytes.

For a chunk with a single message with M bytes of metadata, the size of the metadata_ buffer is `sizeof(ChunkID) + sizeof(PartID) + sizeof(size_t) + sizeof(uint32_t) + sizeof(ChunkID) + sizeof(PartID) + sizeof(size_t) + sizeof(uint32_t) + sizeof(uint64_t) + M = 40 + M` bytes.


pentschev:
If you're merging branch-25.08 into your PR you need to retarget it to branch-25.08 as well. Was that intended @nirandaperera ?

nirandaperera:
> If you're merging branch-25.08 into your PR you need to retarget it to branch-25.08 as well. Was that intended @nirandaperera ?

Thanks @pentschev. It was a mistake. I force pushed the changes now

nirandaperera:
@wence- @madsbk this PR has the "new" Chunk API, with scaffolding for housing multiple messages. I didnt rename it to `Chunk` ATM, because I felt like the API is cleaner to review like this. I will replace `Chunk` once this comes out of draft.

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -24,6 +26,291 @@ namespace rapidsmpf::shuffler::detail {
...
+        return *reinterpret_cast<ChunkID*>(metadata_->data());
+    }
+
+    /**
+     * @brief The number of messages in the chunk.
+     *
+     * @return The number of messages in the chunk.
+     */
+    inline size_t n_messages() const {
+        return *reinterpret_cast<size_t*>(metadata_->data() + sizeof(ChunkID));
```
wence-, line 97: All these reinterpret_cast type-punning approaches break strict-aliasing rules unfortunately.

What we should do is:

1. (C++ 20) use `std::bit_cast` (but it's messy because we're carrying around `std::byte` and `bit_cast` takes values, not pointer + size).
2. Use `memcpy` (the compiler will optimise this)

So, this would be (for example):

```
inline size_t n_messages() const {
    size_t result;
    memcpy(&result, metadata_->data() + sizeof(ChunkID), sizeof(result));
    return result;
}
```

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -25,145 +27,335 @@ namespace rapidsmpf::shuffler::detail {
...
+
+    /**
+     * @brief The expected number of chunks of the i-th message.
+     *
+     * @param i The index of the message.
+     * @return The expected number of chunks for the message. Non-zero when the message
+     * is a control message, otherwise zero (data message).
+     */
+    inline size_t expected_num_chunks(size_t i) const {
+        return expected_num_chunks_[i];
```
madsbk, line 111:
```suggestion
        return expected_num_chunks_.at(i);
```

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -25,145 +27,335 @@ namespace rapidsmpf::shuffler::detail {
...
-    /// If known, the size of the GPU data buffer (in bytes).
-    std::size_t const gpu_data_size;
+    /**
+     * @brief Whether the i-th message is a control message.
+     *
+     * @param i The index of the message.
+     * @return True if the message is a control message, false otherwise.
+     */
+    inline bool is_control_message(size_t i) const {
+        return expected_num_chunks_[i] > 0;
```
madsbk, line 121:
```suggestion
        return expected_num_chunks(i) > 0;
```

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -25,145 +27,335 @@ namespace rapidsmpf::shuffler::detail {
...
-     * @returns A `cudf::table` that represents the chunk data.
+     * @param msg The metadata message received from another rank.
+     * @param validate Whether to validate the metadata buffer.
+     * @return The ChunkBatch.
+     *
+     * @throws std::runtime_error if the metadata buffer does not follow the expected
+     * format and `validate` is true.
      */
-    [[nodiscard]] std::unique_ptr<cudf::table> unpack(rmm::cuda_stream_view stream) const;
+    static Chunk from_serialized_buf(
```
madsbk, line 285:
```suggestion
    static Chunk from_serialized_buffer(
```

cpp/src/shuffler/chunk.cpp
```diff
@@ -13,126 +13,243 @@
...
+    };
+}
+
+Chunk Chunk::from_finished_partition(
+    ChunkID chunk_id, PartID part_id, size_t expected_num_chunks
+) {
+    return {chunk_id, 1, {part_id}, {expected_num_chunks}, {0}, {0}};
+}
+
+Chunk Chunk::from_serialized_buf(std::vector<uint8_t> const& msg, bool validate) {
```
madsbk, line 101:
```suggestion
Chunk Chunk::from_serialized_buffer(std::vector<uint8_t> const& msg, bool validate) {
```

madsbk (commented):
Overall looks good!
@nirandaperera, let's prioritize getting full chunk support. I think it will have a significant impact!

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -25,145 +27,335 @@ namespace rapidsmpf::shuffler::detail {
...
+     * @brief Get the size of the metadata of the i-th message.
+     *
+     * @param i The index of the message.
+     * @return The size of the metadata of the message. Zero when the message is a
+     * control message, otherwise the size of `PackedData::metadata`.
+     */
+    inline uint32_t metadata_size(size_t i) const {
+        assert(i < n_messages());
+        assert(!meta_offsets_.empty());
+        assert(!is_control_message(i));
```
madsbk, line 148: I

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -25,145 +27,335 @@ namespace rapidsmpf::shuffler::detail {
...
+        return i == 0 ? meta_offsets_[0] : meta_offsets_[i] - meta_offsets_[i - 1];
+    }
```
madsbk, line 150: I think it would ggod to be bit more defensive here 
```suggestion
    inline uint32_t metadata_size(size_t i) const {
        RAPIDSMPF_EXPECTS(i < n_messages());
        RAPIDSMPF_EXPECTS(!is_control_message(i));
        return i == 0 ? meta_offsets_.at(0) : meta_offsets_.at(i) - meta_offsets_.at(i - 1);
    }
```

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -25,145 +27,335 @@ namespace rapidsmpf::shuffler::detail {
...
-        std::size_t gpu_data_size,
-        std::unique_ptr<std::vector<uint8_t>> metadata,
-        std::unique_ptr<Buffer> gpu_data
-    );
+    inline size_t data_size(size_t i) const {
+        assert(i < n_messages());
+        assert(!data_offsets_.empty());
+        assert(!is_control_message(i));
+        return i == 0 ? data_offsets_[0] : data_offsets_[i] - data_offsets_[i - 1];
+    }
```
madsbk, line 164: And defensive here as well

cpp/src/shuffler/postbox.cpp
```diff
@@ -15,8 +15,14 @@ namespace rapidsmpf::shuffler::detail {
...
-    auto [_, inserted] =
-        pigeonhole_[key_map_fn_(chunk.pid)].insert({chunk.cid, std::move(chunk)});
+    // TODO: this is a questionable change!
+    // now the outgoing_chunks_ postbox in shuffler is rank-based. If there are multiple
+    // messages in the chunk now, they SHOULD have the same target rank! Otherwise, we
+    // would have to slice the chunk, which is not what we want.
+    // This is why I think we should have a Chunk class and ChunkBatch class separately.
+    auto [_, inserted] = pigeonhole_[key_map_fn_(chunk.part_id(0))].insert(
+        {chunk.chunk_id(), std::move(chunk)}
+    );
```
madsbk, line 25: > If there are multiple messages in the chunk now, they SHOULD have the same target rank! 

But this is always true, right? I mean, we only concatenate chunks with the same target rank?

cpp/src/shuffler/shuffler.cpp
```diff
@@ -200,10 +200,10 @@ class Shuffler::Progress {
         while (true) {
             auto const [msg, src] = shuffler_.comm_->recv_any(metadata_tag);
             if (msg) {
-                auto chunk = Chunk::from_metadata_message(msg);
+                auto chunk = Chunk::deserialize(*msg, false);
                 log.trace("recv_any from ", src, ": ", chunk);
                 RAPIDSMPF_EXPECTS(
-                    shuffler_.partition_owner(shuffler_.comm_, chunk.pid)
+                    shuffler_.partition_owner(shuffler_.comm_, chunk.part_id(0))
```
madsbk, line 206: I guess we only handle singleton chunks here right?
It would be good with TODO and maybe a constexpr instead of `0`.

madsbk (changes requested):

cpp/src/shuffler/shuffler.cpp
nirandaperera: I added a comment here. If the `outgoing_postbox_` guarantees that when it accepts a chunk, all partitions IDs in the chunk maps to the same key, we can safely use the first partition ID.

madsbk: But in the future when we receive a `ChunkBatch`, it might contain different partitions IDs right? So we could use a TODO saying that we will need some code to split `ChunkBatch` based on their partition IDs here?

nirandaperera: I see what you mean. Let me add some todos

nirandaperera: @madsbk but coming to think of it, at this point, if the `outgoing_postbox_` gurantees that it accepts only `Chunk`s that has partition IDs map to the same rank, then we only have to check the first part ID here, isnt it?

madsbk: That is true. I guess we need the TODO where we insert into `outgoing_postbox_`

cpp/src/shuffler/shuffler.cpp
```diff
@@ -174,6 +174,9 @@ class Shuffler::Progress {
         // Check for new chunks in the inbox and send off their metadata.
         auto const t0_send_metadata = Clock::now();
         for (auto&& chunk : shuffler_.outgoing_postbox_.extract_all_ready()) {
+            // TODO: all messages in the chunk must map to the same key. This is
+            // guaranteed by the PostBox. Therefore, we can use the partition
+            // ID of the first message in the chunk to determine the source rank.
```
madsbk, line 179:
```suggestion
            // All messages in the chunk maps to the same key (checked by the PostBox)
            // thus we can use the partition ID of the first message in the chunk to 
            // determine the source rank of all of them.
```

cpp/src/shuffler/shuffler.cpp
```diff
@@ -200,10 +203,13 @@ class Shuffler::Progress {
         while (true) {
             auto const [msg, src] = shuffler_.comm_->recv_any(metadata_tag);
             if (msg) {
-                auto chunk = Chunk::from_metadata_message(msg);
+                auto chunk = Chunk::deserialize(*msg, false);
                 log.trace("recv_any from ", src, ": ", chunk);
+                // TODO: all messages in the chunk must map to the same key. This is
+                // guaranteed by the PostBox. Therefore, we can use the partition
+                // ID of the first message in the chunk to determine the source rank.
```
madsbk, line 210:
```suggestion
            // All messages in the chunk maps to the same key (checked by the PostBox)
            // thus we can use the partition ID of the first message in the chunk to 
            // determine the source rank of all of them.
```

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -25,145 +27,335 @@ namespace rapidsmpf::shuffler::detail {
...
-    Chunk(
-        PartID pid,
-        ChunkID cid,
-        std::size_t gpu_data_size,
-        std::unique_ptr<std::vector<uint8_t>> metadata,
-        std::unique_ptr<Buffer> gpu_data
-    );
+    inline size_t data_size(size_t i) const {
+        assert(!data_offsets_.empty());
+        assert(!is_control_message(i));
```
wence-, line 161: *question*: Should we use `RAPIDSMPF_EXPECTS` here?

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -25,145 +27,335 @@ namespace rapidsmpf::shuffler::detail {
...
+
+    /**
+     * @brief Get the size of the concatenated metadata.
+     *
+     * @return The size of the concatenated metadata.
+     */
+    inline size_t concat_metadata_size() const {
+        assert(metadata_);
+        assert(!meta_offsets_.empty());
+        assert(meta_offsets_[n_messages() - 1] == metadata_->size());
```
wence-, line 221: These invariants are (or should be) enforced by the constructor, so why do we need to assert them here?

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -25,145 +27,335 @@ namespace rapidsmpf::shuffler::detail {
...
+     * @return True if the chunk is ready, false otherwise.
+     * @note ChunkBatch is ready if it has no data or if the data is ready. data_ buffer
+     * could be set later, so we need to check if it is non-null.
+     */
+    [[nodiscard]] inline bool is_ready() const {
+        // data_offsets_[-1] contains the size of the data buffer. If it is 0, the chunk
+        // has no data messages, so it is ready. Else, the chunk is ready if the data
+        // buffer is non-null and the data buffer is ready.
+        return !data_offsets_.empty()
+               && (data_offsets_[n_messages() - 1] == 0 || (data_ && data_->is_ready()));
```
wence-, line 309: *nit*: if `data_offsets_` is not empty, under what circumstances can `data_` be `nullptr`?

i.e. why is this not:

```
return !data_offsets_.empty() && (data_.size() == 0 || data_->is_ready());
```

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -25,145 +27,335 @@ namespace rapidsmpf::shuffler::detail {
...
+        std::vector<PartID> part_ids,
+        std::vector<size_t> expected_num_chunks,
+        std::vector<uint32_t> meta_offsets,
+        std::vector<uint64_t> data_offsets,
+        std::unique_ptr<std::vector<uint8_t>> metadata = nullptr,
+        std::unique_ptr<Buffer> data = nullptr
     );
+
+    ChunkID const chunk_id_;  ///< The ID of the chunk.
+    size_t const n_messages_;  ///< The number of messages in the chunk.
```
wence-, line 344: *question*: this information needs to be sent over the wire, but I think it redundantly encodes `meta_offsets_.size()`, is that right? If yes, should we remove it?

cpp/src/shuffler/chunk.cpp
```diff
@@ -13,126 +13,243 @@
...
-      metadata{std::move(metadata)},
-      gpu_data{std::move(gpu_data)} {}
+    : chunk_id_{chunk_id},
+      n_messages_{n_messages},
+      part_ids_{std::move(part_ids)},
+      expected_num_chunks_{std::move(expected_num_chunks)},
+      meta_offsets_{std::move(meta_offsets)},
+      data_offsets_{std::move(data_offsets)},
+      metadata_{std::move(metadata)},
+      data_{std::move(data)} {}
```
wence-, line 32: *suggestions*: Let us do all the validation of the format of the `Chunk` here. Then we don't need assertions scattered around the rest of the code, since it will not be possible to construct an invalid chunk.

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -25,145 +27,335 @@ namespace rapidsmpf::shuffler::detail {
...
+    /**
+     * @brief Get the data of the i-th message, as a new ChunkBatch.
+     *
+     * @param new_chunk_id The ID of the new chunk.
+     * @param i The index of the message.
+     * @param stream The CUDA stream.
+     * @return A new ChunkBatch containing the data of the i-th message.
+     * @note This will create a copy of the packed data. If there is only one message and
+     * the message is a data message, the buffers will be moved to the new ChunkBatch.
+     * Otherwise a new ChunkBatch will be created by copying data.
```
wence-, line 132: *nit*: Not a `ChunkBatch` any more.

cpp/src/shuffler/chunk.cpp
```diff
@@ -13,126 +13,243 @@
...
-    auto br = gpu_data->br;
+bool Chunk::validate_format(std::vector<uint8_t> const& serialized_buf) {
+    // Check if buffer is large enough to contain at least the header
+    if (serialized_buf.size() < sizeof(ChunkID) + sizeof(size_t)) {
+        return false;
+    }
+
+    // Get number of messages
+    size_t n_messages =
+        *reinterpret_cast<size_t const*>(serialized_buf.data() + sizeof(ChunkID));
```
wence-, line 159: As above, this should be `std::memcpy`.

cpp/src/shuffler/chunk.cpp
```diff
@@ -13,126 +13,243 @@
...
+    // Check if buffer is large enough to contain all the messages' metadata
+    size_t header_size = metadata_message_header_size(n_messages);
+    if (serialized_buf.size() < header_size) {
+        return false;
+    }
+
+    // For each message, validate the metadata and data sizes
+    auto const* psum_meta = reinterpret_cast<uint32_t const*>(
+        serialized_buf.data() + sizeof(ChunkID) + sizeof(size_t)
+        + n_messages * (sizeof(PartID) + sizeof(size_t))
```
wence-, line 174: And this.

cpp/src/shuffler/chunk.cpp
```diff
@@ -13,126 +13,243 @@
...
     );
+    auto const* psum_data = reinterpret_cast<uint64_t const*>(psum_meta + n_messages);
```
wence-, line 176: And this.

cpp/src/shuffler/postbox.cpp
```diff
@@ -14,9 +14,16 @@ namespace rapidsmpf::shuffler::detail {
 
 template <typename KeyType>
 void PostBox<KeyType>::insert(Chunk&& chunk) {
+    // check if all partition IDs in the chunk map to the same key
+    KeyType key = key_map_fn_(chunk.part_id(0));
```
wence-, line 18: *question*: I guess by construction a chunk always contains at least one piece?

cpp/src/shuffler/shuffler.cpp
```diff
@@ -251,21 +257,24 @@ class Shuffler::Progress {
...
+                    in_transit_chunks_.insert({chunk.chunk_id(), std::move(chunk)})
+                        .second,
                     "in transit chunk already exist"
                 );
                 shuffler_.statistics_->add_bytes_stat(
-                    "shuffle-payload-recv", chunk.gpu_data_size
+                    "shuffle-payload-recv", chunk.concat_data_size()
                 );
                 // Tell the source of the chunk that we are ready to receive it.
+                // TODO: all partition IDs in the chunk must map to the same key (rank).
```
wence-, line 275: This is true by construction no? This chunk is extracted from `incoming_chunks_` which are populated by the deserialized chunks we received on line 206. The sender guaranteed on construction of that chunk that all the partitions mapped to our rank.

cpp/src/shuffler/shuffler.cpp
```diff
@@ -457,27 +468,46 @@ void Shuffler::shutdown() {
...
-        ++outbound_chunk_counter_[chunk.pid];
+        // TODO: There are multiple partitions in the chunk. So, do this for each
+        // partition.
+        ++outbound_chunk_counter_[chunk.part_id(0)];
     }
-    if (partition_owner(comm_, chunk.pid) == comm_->rank()) {
-        if (chunk.gpu_data) {
-            statistics_->add_bytes_stat("shuffle-payload-send", chunk.gpu_data->size);
-            statistics_->add_bytes_stat("shuffle-payload-recv", chunk.gpu_data->size);
+    // TODO: Guarantee that all messages in the chunk map to the same key (rank).
```
wence-, line 506: As above, do this in the constructor of `Chunk` then you don't need to litter the rest of the code with checks.

wence- (changes requested):
I think we can tidy some things up and remove TODOs by ensuring `Chunk`s obey their invariants by construction.

cpp/include/rapidsmpf/shuffler/chunk.hpp
nirandaperera: I was previously thinking about only populate `data_offsets_` only when there are data messages. But ended up populating `data_offsets_` and `meta_offsets_` for all messages. So, we can remove these asserts.

cpp/include/rapidsmpf/shuffler/chunk.hpp
nirandaperera: fair point. I was mostly using the asserts to prevent me from _shooting myself in the foot_ :wink: I'll remove these

cpp/include/rapidsmpf/shuffler/chunk.hpp
nirandaperera: When we receive messages, we first receive concat metadata buffer, and then we deserialize it to form a valid `Chunk` that knows data buffer size, but it's not allocated. https://github.com/nirandaperera/rapidsmpf/blob/multi-packed-data-chunk/cpp/src/shuffler/shuffler.cpp#L206
Allocation happens later, 
https://github.com/nirandaperera/rapidsmpf/blob/multi-packed-data-chunk/cpp/src/shuffler/shuffler.cpp#L237-L238

cpp/include/rapidsmpf/shuffler/chunk.hpp
nirandaperera: I agree. We can remove `Chunk::n_messges_` and use the size of either on these vectors 
```c++
       std::vector<PartID> part_ids,
        std::vector<size_t> expected_num_chunks,
        std::vector<uint32_t> meta_offsets,
        std::vector<uint64_t> data_offsets,
```

cpp/src/shuffler/postbox.cpp
nirandaperera: yes

cpp/src/shuffler/shuffler.cpp
nirandaperera: Yes, that is true. So, it's guaranteed by the PostBox.

cpp/src/shuffler/shuffler.cpp
nirandaperera: Are you suggesting that we add a mapping function to the Chunk ctr?

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -25,145 +27,325 @@ namespace rapidsmpf::shuffler::detail {
...
      *
-     * @param pid The ID of the partition this chunk is part of.
-     * @param cid The ID of the chunk.
-     * @param expected_num_chunks If not zero, the number of chunks of the partition
-     * expected to get from the sending rank. Ignored when it is zero.
+     * @param data The data buffer.
      */
-    Chunk(PartID pid, ChunkID cid, std::size_t expected_num_chunks);
+    void set_data_buffer(std::unique_ptr<Buffer> data) {
+        data_ = std::move(data);
```
madsbk, line 168:
```suggestion
        RAPIDSMPF_EXPECTS(data_, "buffer is already set");
        data_ = std::move(data);
```

cpp/tests/test_shuffler.cpp
```diff
@@ -627,13 +639,11 @@ TEST_F(PostBoxTest, ThreadSafety) {
...
-                    rapidsmpf::shuffler::PartID{j / chunks_per_partition},
+                auto chunk = rapidsmpf::shuffler::detail::make_dummy_chunk(
                     rapidsmpf::shuffler::detail::ChunkID{i * chunks_per_thread + j},
-                    0,  // gpu_data_size
-                    nullptr,  // metadata
-                    nullptr  // gpu_data
-                };
+                    rapidsmpf::shuffler::PartID{j / chunks_per_partition}
+                );
+
```
madsbk, line 646:
```suggestion
```

cpp/tests/test_shuffler.cpp
```diff
@@ -581,13 +595,11 @@ TEST_F(PostBoxTest, InsertAndExtractMultipleChunks) {
...
-            rapidsmpf::shuffler::PartID{i % num_partitions},
+        auto chunk = rapidsmpf::shuffler::detail::make_dummy_chunk(
             rapidsmpf::shuffler::detail::ChunkID{i},
-            0,  // gpu_data_size
-            nullptr,  // metadata
-            nullptr  // gpu_data
-        };
+            rapidsmpf::shuffler::PartID{i % num_partitions}
+        );
+
```
madsbk, line 602:
```suggestion
```

madsbk (changes requested):
Final suggestions

cpp/include/rapidsmpf/shuffler/chunk.hpp
wence-: Did you mean that, or `RAPIDSMPF_EXPECTS(!data_, ...)`?

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -25,145 +27,326 @@ namespace rapidsmpf::shuffler::detail {
...
      *
-     * @param pid The ID of the partition this chunk is part of.
-     * @param cid The ID of the chunk.
-     * @param expected_num_chunks If not zero, the number of chunks of the partition
-     * expected to get from the sending rank. Ignored when it is zero.
+     * @param data The data buffer.
      */
-    Chunk(PartID pid, ChunkID cid, std::size_t expected_num_chunks);
+    void set_data_buffer(std::unique_ptr<Buffer> data) {
+        RAPIDSMPF_EXPECTS(data_, "buffer is already set");
```
madsbk, line 168: Sorry, had a type in my last suggestion, should have been:
```suggestion
        RAPIDSMPF_EXPECTS(!data_, "buffer is already set");
```

madsbk (approved):

madsbk:
@wence- do you have anything else?

PR #253, opened 2025-05-07
ProgressThread: add sleep argument

madsbk:
While working on setups with may shuffle instances I noticed that `ProgressThread::add_function()` can block for a long time when we have many (>30) shuffler instances.

Instead of introducing advance starvation avoidance in `ProgressThread`, I think a 1us sleep suffice. Let's also wait with the Python API.

### DIFF
--- a/cpp/include/rapidsmpf/progress_thread.hpp
+++ b/cpp/include/rapidsmpf/progress_thread.hpp
@@ -132,10 +132,15 @@ class ProgressThread {
      *
      * @param logger The logger instance to use.
      * @param statistics The statistics instance to use (disabled by default).
+     * @param sleep The duration to sleep between each progress loop iteration.
+     * If 0, the thread yields execution instead of sleeping. Anecdotally, a 1 us
+     * sleep time (the default) is sufficient to avoid starvation and get smooth
+     * progress.
      */
     ProgressThread(
         Communicator::Logger& logger,
-        std::shared_ptr<Statistics> statistics = std::make_shared<Statistics>(false)
+        std::shared_ptr<Statistics> statistics = std::make_shared<Statistics>(false),
+        Duration sleep = std::chrono::microseconds{1}
     );
 
     ~ProgressThread();
--- a/cpp/src/progress_thread.cpp
+++ b/cpp/src/progress_thread.cpp
@@ -37,17 +37,20 @@ void ProgressThread::FunctionState::operator()() {
 }
 
 ProgressThread::ProgressThread(
-    Communicator::Logger& logger, std::shared_ptr<Statistics> statistics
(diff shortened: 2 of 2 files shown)


wence-:
> While working on setups with may shuffle instances I noticed that ProgressThread::add_function() can block for a long time when we have many (>30) shuffler instances.

Is this because everyone needs to grab the lock to mutate the `functions_` list. Would it suffice to pause the thread before adding the function? Like this:

```
diff --git a/cpp/src/progress_thread.cpp b/cpp/src/progress_thread.cpp
index 5d6b6bf..e429ddd 100644
--- a/cpp/src/progress_thread.cpp
+++ b/cpp/src/progress_thread.cpp
@@ -64,6 +64,7 @@ void ProgressThread::stop() {
 }
 
 ProgressThread::FunctionID ProgressThread::add_function(Function&& function) {
+    thread_.pause();
     std::lock_guard lock(mutex_);
     // We can use `this` as the thread address only because `ProgressThread` isn't
     // moveable or copyable.
```

madsbk:
> Would it suffice to pause the thread before adding the function? Like this

It helps `add_function` but not `remove_function` :/

wence-:
> It helps `add_function` but not `remove_function` :/

```diff
diff --git a/cpp/src/progress_thread.cpp b/cpp/src/progress_thread.cpp
index 5d6b6bf..90eb026 100644
--- a/cpp/src/progress_thread.cpp
+++ b/cpp/src/progress_thread.cpp
@@ -64,6 +64,7 @@ void ProgressThread::stop() {
 }
 
 ProgressThread::FunctionID ProgressThread::add_function(Function&& function) {
+    thread_.pause();
     std::lock_guard lock(mutex_);
     // We can use `this` as the thread address only because `ProgressThread` isn't
     // moveable or copyable.
@@ -81,6 +82,7 @@ void ProgressThread::remove_function(FunctionID function_id) {
         "Function was not registered with this ProgressThread"
     );
 
+    thread_.pause();
     std::unique_lock lock(mutex_);
     auto it = functions_.find(function_id.function_index);
     RAPIDSMPF_EXPECTS(
@@ -97,6 +99,8 @@ void ProgressThread::remove_function(FunctionID function_id) {
 
     if (functions_.empty()) {
         thread_.pause();
+    } else {
+        thread_.resume();
     }
 }
 
```
?

madsbk:
But then the thread is paused while we wait in:
```
cv_.wait(lock, [&, f_idx = it->first]() { return functions_.at(f_idx).is_done; });
```

wence-:
Ah, ok. In which case, carry on.

We should consider at more leisure introducing more lock-free data structures. But I would rather we did not implement them.

wence- (approved):

pentschev (approved):
LGTM as well. Thanks @madsbk .

Issue #239, opened 2025-05-02
Dask-shuffle: handle shuffle IDs and cleanup


madsbk:
Currently, Dask never clean up `Shuffler` instances. Even worse, we do not support more than 256 consecutive shuffle operations because `op_id` in the shuffler is an `unsigned char` and [`get_shuffle_id()`](
https://github.com/rapidsai/rapidsmpf/blob/74142e341d40cf4bb1fddc956c76ca837c9f7aad/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py#L33) never reuse IDs (even if `op_id` is wrapping, the order of use isn't guaranteed, I think).

PR #254, opened 2025-05-07
Dask shuffler cleanup

madsbk:
~Depend on https://github.com/rapidsai/rapidsmpf/pull/253~
Closes https://github.com/rapidsai/rapidsmpf/issues/239

Currently, Dask never cleanup shuffler instances.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
+++ b/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
@@ -27,7 +27,7 @@ if TYPE_CHECKING:
 
 
 # Set of available shuffle IDs
-_shuffle_id_vacancy: set[int] = set(range(256))
+_shuffle_id_vacancy: set[int] = set(range(Shuffler.max_concurrent_shuffles))
 _shuffle_id_vacancy_lock: threading.Lock = threading.Lock()
 
 
@@ -35,12 +35,11 @@ def _get_new_shuffle_id(client: Client) -> int:
     """
     Get a new available shuffle ID.
 
-    Since RapidsMPF only supports 256 shuffler instances at any given time,
-    this function maintains a shared pool of 256 shuffle IDs to ensure the
-    returned shuffle ID is not currently in use on any Dask worker.
+    Since RapidsMPF only supports a limited number of shuffler instances at
+    any given time, this function maintains a shared pool of shuffle IDs.
 
     If no IDs are available locally, it queries all workers for IDs in use,
-    updates the vacancy set accordingly, and retries. If all 256 IDs are in use
+    updates the vacancy set accordingly, and retries. If all IDs are in use
     across the cluster, an error is raised.
 
     Parameters
@@ -55,7 +54,7 @@ def _get_new_shuffle_id(client: Client) -> int:
     Raises
     ------
     ValueError
-        If all 256 shuffle IDs are currently in use across the cluster.
+        If all shuffle IDs are currently in use across the cluster.
     """
     with _shuffle_id_vacancy_lock:
         if not _shuffle_id_vacancy:
@@ -65,12 +64,13 @@ def _get_new_shuffle_id(client: Client) -> int:
                 with ctx.lock:
                     return set(ctx.shufflers.keys())
 
-            _shuffle_id_vacancy.update(range(256))
+            _shuffle_id_vacancy.update(range(Shuffler.max_concurrent_shuffles))
             _shuffle_id_vacancy.difference_update(*client.run(get_vacancy_ids).values())
 
             if not _shuffle_id_vacancy:
                 raise ValueError(
-                    "Cannot shuffle more than 256 times in a single Dask compute."
+                    f"Cannot shuffle more than {Shuffler.max_concurrent_shuffles} "
+                    "times in a single Dask compute."
                 )
 
         return _shuffle_id_vacancy.pop()
--- a/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
+++ b/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
@@ -68,10 +68,10 @@ def _get_new_shuffle_id(client: Client) -> int:
             _shuffle_id_vacancy.update(range(256))
             _shuffle_id_vacancy.difference_update(*client.run(get_vacancy_ids).values())
 
-        if not _shuffle_id_vacancy:
-            raise ValueError(
-                "Cannot shuffle more than 256 times in a single Dask compute."
-            )
+            if not _shuffle_id_vacancy:
+                raise ValueError(
+                    "Cannot shuffle more than 256 times in a single Dask compute."
+                )
 
         return _shuffle_id_vacancy.pop()
 
--- a/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
+++ b/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
@@ -314,18 +314,17 @@ def _extract_partition(
     -------
     Extracted DataFrame partition.
     """
-    if callback is None:
-        raise ValueError("Missing callback in _extract_partition.")
     shuffler = get_shuffler(shuffle_id)
-    ret = callback(
-        partition_id,
-        column_names,
-        shuffler,
-    )
-    if shuffler.finished():
-        ctx = get_worker_context()
-        del ctx.shufflers[shuffle_id]
-    return ret
+    try:
+        return callback(
+            partition_id,
+            column_names,
+            shuffler,
+        )
+    finally:
+        if shuffler.finished():
+            ctx = get_worker_context()
+            del ctx.shufflers[shuffle_id]
 
 
 def rapidsmpf_shuffle_graph(
--- a/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
+++ b/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
@@ -56,17 +56,22 @@ def _get_new_shuffle_id(client: Client) -> int:
     ValueError
         If all shuffle IDs are currently in use across the cluster.
     """
+    global _shuffle_id_vacancy  # noqa: PLW0603
+
     with _shuffle_id_vacancy_lock:
         if not _shuffle_id_vacancy:
 
-            def get_vacancy_ids(dask_worker: Worker) -> set[int]:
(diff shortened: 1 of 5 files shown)


rjzamora:
Thanks for submitting this @madsbk ! You have good timing - I spent this morning investigating a performance regression when I sweep over all queries (or many iterations of the same query). Turns out that this is exactly the cause. The time it takes to stage a new `Shuffler` increases with the number of Shufflers we are keeping alive.

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -282,11 +316,16 @@ def _extract_partition(
     """
     if callback is None:
         raise ValueError("Missing callback in _extract_partition.")
-    return callback(
+    shuffler = get_shuffler(shuffle_id)
+    ret = callback(
```
TomAugspurger, line 320: What do we want to do if `callback` errors? Maybe this should instead be

```python
try:
    ret = callback(...)
finally:
    ctx = get_worker_context()
    del ctx.shufflers[shuffle_id]
else:
    return ret
```

madsbk: The problem is that to call `shuffler.finished()`, we need to call `get_shuffler()`, which might create a new `Shuffler`.

I don't think we should guard against calling `_extract_partition` with the wrong types. But it is a good idea to guard against errors in `callback`. It now looks like:
```python
    shuffler = get_shuffler(shuffle_id)
    try:
        return callback(
            partition_id,
            column_names,
            shuffler,
        )
    finally:
        if shuffler.finished():
            ctx = get_worker_context()
            del ctx.shufflers[shuffle_id]
```

TomAugspurger:
Do you know roughly how long this new test takes to run?

rjzamora:
@madsbk - Is this ready to be removed from "Draft"?

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -22,24 +23,57 @@
 if TYPE_CHECKING:
     from collections.abc import Callable, MutableMapping, Sequence
 
-    from distributed import Worker
+    from distributed import Client, Worker
 
 
-_shuffle_counter: int = 0
+# Set of available shuffle IDs
+_shuffle_id_vacancy: set[int] = set(range(256))
```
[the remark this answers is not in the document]
madsbk: Added `Shuffler.max_concurrent_shuffles`

python/rapidsmpf/rapidsmpf/tests/test_dask.py
```diff
@@ -110,3 +113,76 @@ def test_bootstrap_dask_cluster_idempotent() -> None:
...
+        )
+
+    with LocalCUDACluster(n_workers=1, loop=loop) as cluster:  # noqa: SIM117
+        with Client(cluster) as client:
+            bootstrap_dask_cluster(client, spill_device=0.1)
+
+            # We can shuffle 256 consecutive times.
+            do_shuffle(seed=1, num_shuffles=256)
+            # And more times after a compute.
+            do_shuffle(seed=2, num_shuffles=10)
```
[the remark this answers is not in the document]
madsbk: Yeah, if CI accept the current run we can remove it

madsbk:
> Do you know roughly how long this new test takes to run? Looking at the total runtimes

Yeah, it might be too expensive. Let's remove the full 256 shuffler test

python/rapidsmpf/rapidsmpf/shuffler.pyx
```diff
@@ -281,7 +290,7 @@ cdef class Shuffler:
                 progress_thread._handle,
                 op_id,
                 total_num_partitions,
-                self._stream.view(),
+                _stream,
```
madsbk, line 293: Drive by fix

madsbk:
Reducing the number of shuffles didn't do much to the CI runtime, `conda-python-tests` took 4min. I think it is still reasonable?

python/rapidsmpf/rapidsmpf/tests/test_dask.py
```diff
@@ -110,3 +114,77 @@ def test_bootstrap_dask_cluster_idempotent() -> None:
...
+            # running time, we just do 5+10 shuffles.
+            do_shuffle(seed=1, num_shuffles=5)
+            do_shuffle(seed=2, num_shuffles=10)
+
+            # Check that all shufflers has been cleaned up.
+            def check_worker(dask_worker: Worker) -> None:
+                ctx = get_worker_context(dask_worker)
+                assert len(ctx.shufflers) == 0
+
+            client.run(check_worker)
```
pentschev, line 183: Would it be worth to add a similar check after the block below to ensure the exception didn't unintentionally leave anything behind?

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -280,13 +314,17 @@ def _extract_partition(
     -------
     Extracted DataFrame partition.
     """
-    if callback is None:
-        raise ValueError("Missing callback in _extract_partition.")
```
pentschev, line 284: Looks like this check is not needed anymore because the callback that is passed is `DaskIntegration.extract_partition`, and for an object to qualify for the `DaskIntegration` the `extract_partition` must exist properly implemented as the method in the protocol, is that right?

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -22,24 +23,57 @@
...
-    global _shuffle_counter  # noqa: PLW0603
+    with _shuffle_id_vacancy_lock:
+        if not _shuffle_id_vacancy:
+
+            def get_vacancy_ids(dask_worker: Worker) -> set[int]:
+                ctx = get_worker_context(dask_worker)
+                with ctx.lock:
+                    return set(ctx.shufflers.keys())
+
+            _shuffle_id_vacancy.update(range(Shuffler.max_concurrent_shuffles))
```
pentschev, line 67: Is there a reason to use `update` instead of `_shuffle_id_vacancy = set(range(Shuffler.max_concurrent_shuffles))`? Not a big deal, it just took me a while to understand what was being done here (assuming my understanding is even correct that the two are equivalent).

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -22,24 +23,57 @@
...
 
-    _shuffle_counter += 1
-    return _shuffle_counter
+            if not _shuffle_id_vacancy:
+                raise ValueError(
+                    f"Cannot shuffle more than {Shuffler.max_concurrent_shuffles} "
+                    "times in a single Dask compute."
+                )
+
+        return _shuffle_id_vacancy.pop()
```
pentschev, line 76: Do we ever (want to) deregister shuffler IDs after they're done?

pentschev (approved):
Left a suggestion and a couple of questions, otherwise LGTM. Thanks @madsbk .

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
madsbk: Exactly

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
madsbk: Yes, to support more than 256 shuffles in total, we need to release IDs.

python/rapidsmpf/rapidsmpf/tests/test_dask.py
madsbk: Very good point, some instances will actually be left behind!

I think this will require some new design to do gracefully, added an issue: https://github.com/rapidsai/rapidsmpf/issues/258 (low priority).

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
madsbk: Only to avoid to declare `_shuffle_id_vacancy` global :)
Changed to:
```python
_shuffle_id_vacancy = set(range(Shuffler.max_concurrent_shuffles))
```

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
pentschev: Sorry, I think I wasn't clear, I mean that I _think_ they are not released _after use_, here the ID is "given" to the caller, but when the caller doesn't use it anymore should it be put back into `_shuffle_id_vacancy` somehow? I didn't see anything of that sort, how is that handled currently besides having to ask workers for [occupied IDs](https://github.com/rapidsai/rapidsmpf/pull/254/files#diff-f6d20cdd9996b100c878b44518f46cfcda72fec240c6420adfe590c66ea96cfaR64-R67)? Is there an additional mechanism for that, and if not, should we have that?

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
pentschev: I wasn't proposing making it global, I'm fine with either the old or new implementation, TBH I didn't realize that's what you were avoiding thus I asked because it didn't look the most natural way to do it.

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
madsbk: No, we don't have a mechanism for that. This is way the client has to ask the workers. Notice, all workers must have completed the shuffle before we can reuse that ID. 

Since 256 is a high number, I think it is fine to call the workers once all of the 256 ID has been used.

pentschev: Makes sense, thanks for clarifying.

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
madsbk: Sure, but I think using global make the code more clear, in this case.

pentschev: Sounds good to me. Thanks!

pentschev (approved):
LGTM still, thanks @madsbk !

madsbk:
Thanks @pentschev, let's merge

PR #256, opened 2025-05-07
Basic sorting support with Dask

rjzamora:
Follow up to https://github.com/rapidsai/rapidsmpf/pull/249

- Adds minimal changes to support sorting with a `DaskIntegration` protocol.
- Does **not** bother to implement/demonstrate anything beyond an ascending sort by a single column.
- Adds `sort_boundaries` and `options` arguments to `DaskIntegration.insert_partition` and adds an `options` argument to `DaskIntegration.extract_partition`
    - The `sort_boundaries` must be a separate positional argument so that it can be generated dynamically as the output of a separate task. All other sorting options can be passed through with `options` (e.g. ascending vs descending and null precedence).

This technically breaks existing cudf-polars code, so we will want to update the `DaskIntegration` protocol defined in cudf asap after this is finished/merged.


python/rapidsmpf/rapidsmpf/examples/dask.py
```diff
@@ -59,22 +64,39 @@ def insert_partition(
             Number of output partitions for the current shuffle.
         shuffler
             The RapidsMPF Shuffler object to extract from.
+        sort_boundaries
+            Output partition boundaries for sorting.
+        options
+            Optional key-work arguments.
```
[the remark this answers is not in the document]
TomAugspurger: And we use "Additional options." in a few places below. Let's pick one description and copy it through.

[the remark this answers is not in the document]
rjzamora: Ah yeah, missed this one - Let's do the simple `"Additional options."` for now.

python/rapidsmpf/rapidsmpf/examples/dask.py
```diff
@@ -59,22 +64,39 @@ def insert_partition(
...
-        columns_to_hash = tuple(list(df.columns).index(val) for val in on)
-        packed_inputs = partition_and_pack(
-            df.to_pylibcudf()[0],
-            columns_to_hash=columns_to_hash,
-            num_partitions=partition_count,
-            stream=DEFAULT_STREAM,
-            device_mr=rmm.mr.get_current_device_resource(),
-        )
+        if options:
+            raise ValueError(f"Unsupported options: {options}")
```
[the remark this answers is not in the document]
rjzamora: Yeah, additional options aren't used/supported here, but the cudf-polars implementation will need to account for ascending/descending and nulls, etc.

python/rapidsmpf/rapidsmpf/tests/test_dask.py
```diff
@@ -63,7 +63,12 @@ def get_rank(dask_worker: Worker) -> int:
 
 
 @pytest.mark.parametrize("partition_count", [None, 3])
-def test_dask_cudf_integration(loop: pytest.FixtureDef, partition_count: int) -> None:  # noqa: F811
+@pytest.mark.parametrize("sort", [True, False])
+def test_dask_cudf_integration(
+    loop: pytest.FixtureDef,  # noqa: F811
+    partition_count: int,
+    sort: bool,  # noqa: FBT001
```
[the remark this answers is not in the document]
rjzamora: Interesting. I didn't know you could have a keyword-only test parameter.

seberg (approved):
Looks good to me, although my Dask eye isn't keen.  Two small comments that may or may not be relevant.

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -125,13 +127,19 @@ def insert_partition(
             Number of output partitions for the current shuffle.
         shuffler
             The RapidsMPF Shuffler object to extract from.
+        sort_boundaries
+            Output partition boundaries for sorting. If None,
+            hashing will be used to calculate output partitions.
+        options
+            Additional options.
         """
```
wence-, line 135: I can imagine that we might eventually want more all-to-all-like patterns. Would it make more sense to change this interface such that `insert_partition` just takes the `list[PackedData]` and the shuffler and we provide separate functions for hash and sort-based partitioning (and the user can bring their own).

So something like:


```
def insert_partition(
    shuffler: Shuffler,
    chunks: Sequence[PackedData], # Or whatever it is
) -> None:
```

And we provide two builtin functions

```
def hash_partition(df, partition_count, *, on) -> list[PackedData]:
    ...
def sort_partition(df, partition_count, *, by) -> list[PackedData]:
    ...
```

rjzamora: I don't think that helps us generalize at all. We already have `Shuffler.insert_chunks`, which is essentially the `insert_partition` function you are proposing. The purpose of `DaskIntegration.insert_partition` is to aviod the need for various Dask shuffling applications to write their own task graph.

We want `insert_partition`/`extract_partition` to include the minimal necessary arguments to construct a "general" shuffling task graph. Since we are revising things, this may be:

```
    @staticmethod
    def insert_partition(
        df: DataFrameT,  # Partition to insert
        partition_count: int,   # Output partition count
        shuffler: Shuffler,   # Shuffler object
        options: dict[str, Any] | None,   # Arbitrary keyword arguments
        *other: Any,   # "Other" task-output data (e.g. sorting boundaries/quantiles)
    ) -> None:

    @staticmethod
    def extract_partition(
        partition_id: int,  # Partition ID to extract
        shuffler: Shuffler,   # Shuffler object
        options: dict[str, Any] | None,   # Arbitrary keyword arguments
    ) -> DataFrameT:
```

I think the `options` argument can be used to control most variation of a shuffle, and the `*other` positional argument could be used to pass in information that must be calculated dynamically at execution time.

python/rapidsmpf/rapidsmpf/examples/dask.py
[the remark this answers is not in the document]
seberg: I _think_ it would be good to pass on the `pid` of `df` here for sorting.  That is needed to find the right `splits` if you want to balance the result partition sizes for degenerate case (such as all equal).
And I believe we don't have another way to pass it in.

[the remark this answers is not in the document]
rjzamora: @seberg - I updated/generalized the protocol a bit. I didn't include the input partition id as a required argument, but we *can* add that now that we are changing things. Can you explain how having the input partition id would help you handle degenerate values?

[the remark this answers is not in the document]
seberg: Basically, the idea is that the split_boundary values know which partition ID they came from (and ideally their local row).

For example, we split `(1, 1, 1, 1)`, distributed as `(1, 1)` and `(1, 1)`.
If you the `pid` and `row`, then the split boundary will be `(value, pid=1, row=0)`.

With that `pid` information, you can figure out now here that `pid=0` should send it's data to 0 (split _after_ the boundary) and `pid=1` should send it all to 1 (split _before_ at boundary here).

Without the additional information, there is no choice but for both pids to send all data to 0.

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -147,9 +147,11 @@ class DaskIntegration(Protocol[DataFrameT]):
     @staticmethod
     def insert_partition(
         df: DataFrameT,
-        on: Sequence[str],
+        partition_id: int,
         partition_count: int,
         shuffler: Shuffler,
+        options: dict[str, Any],
```
TomAugspurger, line 153: Sorry for the cross-posting, but https://github.com/rapidsai/cudf/pull/18720/files/94554a885f83021f72ce91ca04b96f0d67c6be28#diff-2d46acc0f845c678bb9acbbe8d6c14aaef1e85b63df306b87737a4f1cb3ece78 has a discussion about changing this type to `Any`, and then each implementation can (maybe) require a specific type like `TypedDict` with fields `on: Sequence[str]` for the implementation in cudf-polars.

rjzamora, line 153:
```suggestion
        options: Any,
```

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -158,19 +160,24 @@ def insert_partition(
...
+            Other data needed for partitioning. For example,
+            this may be boundary values needed for sorting.
         """
 
     @staticmethod
     def extract_partition(
         partition_id: int,
-        column_names: list[str],
         shuffler: Shuffler,
+        options: dict[str, Any],
```
rjzamora, line 180:
```suggestion
        options: Any,
```

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -253,11 +260,23 @@ def _stage_shuffler(
...
 
 def _insert_partition(
-    callback: Callable[[DataFrameT, Sequence[str], int, Shuffler], None],
+    callback: Callable[
+        [
+            DataFrameT,
+            int,
+            int,
+            Shuffler,
+            dict[str, Any],
```
rjzamora, line 269:
```suggestion
            Any,
```

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -253,11 +260,23 @@ def _stage_shuffler(
...
+            *tuple[str | tuple[str, int], ...],
+        ],
+        None,
+    ],
     df: DataFrameT,
-    on: Sequence[str],
+    partition_id: int,
     partition_count: int,
     shuffle_id: int,
+    options: dict[str, Any],
```
rjzamora, line 278:
```suggestion
    options: Any,
```

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -270,31 +289,40 @@ def _insert_partition(
...
         get_shuffler(shuffle_id),
+        options,
+        *other_keys,
     )
 
 
 def _extract_partition(
-    callback: Callable[[int, Sequence[str], Shuffler], DataFrameT],
+    callback: Callable[
+        [int, Shuffler, dict[str, Any]],
```
rjzamora, line 319:
```suggestion
        [int, Shuffler, Any],
```

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -270,31 +289,40 @@ def _insert_partition(
...
+        DataFrameT,
+    ],
     shuffle_id: int,
     partition_id: int,
-    column_names: list[str],
     worker_barrier: tuple[int, ...],
+    options: dict[str, Any],
```
rjzamora, line 325:
```suggestion
    options: Any,
```

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -335,11 +363,11 @@ def _extract_partition(
 def rapidsmpf_shuffle_graph(
     input_name: str,
     output_name: str,
-    column_names: Sequence[str],
-    shuffle_on: Sequence[str],
     partition_count_in: int,
     partition_count_out: int,
     integration: DaskIntegration,
+    options: dict[str, Any],
```
rjzamora, line 369:
```suggestion
    options: Any,
```

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
rjzamora: Thanks for the review @TomAugspurger - Do you think this PR is ready once https://github.com/rapidsai/cudf/pull/18720 is green?

python/rapidsmpf/rapidsmpf/examples/dask.py
```diff
@@ -106,8 +128,9 @@ def extract_partition(
 
 def dask_cudf_shuffle(
     df: dask_cudf.DataFrame,
-    shuffle_on: list[str],
+    on: list[str],
```
TomAugspurger, line 131: https://github.com/rapidsai/rapidsmpf/blob/branch-25.06/docs/source/quickstart.md?plain=1#L29 will need to be updated from `shuffle_on` to `on` I think.

TomAugspurger (approved):
I think this should be good. I think our quickstart example will need to be updated for the new keyword.

@wence-'s comment in https://github.com/rapidsai/rapidsmpf/pull/256/files#r2079496979 is worth resolving one way or another. Aside from churn, I think it can be handled later, but I don't have a strong opinion on it.

TomAugspurger:
Yep that makes sense to me. Let's go ahead and merge this and open a followup issue if @wence- has anything to add.

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
wence-: Ah, ok, carry on then

PR #261, opened 2025-05-08
Change get_gtest to link statically and remove tests from package

seberg:
Currently, the `librapidsmpf` packages ship gtest/gmock library and include files, which they shouldn't.
Simplifying this down to the current version also in libcudf should fix that (i.e. by linking statically).

I would just do this to keep similar to stay similar to other rapids libraries, unless this is somehow different?

~(I think there is another small issue with `mpi_tests` and `ucxx_tests` getting shipped in `librapidsmpf` rather than `librapidsmpf-tests`, but I'll have a look at it seperately.)~

### DIFF
--- a/conda/recipes/librapidsmpf/install_librapidsmpf.sh
+++ b/conda/recipes/librapidsmpf/install_librapidsmpf.sh
@@ -2,4 +2,4 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION & AFFILIATES.
 # SPDX-License-Identifier: Apache-2.0
 
-cmake --install cpp/build
+cmake --install cpp/build --component rapidsmpf
--- a/conda/recipes/librapidsmpf/install_librapidsmpf.sh
+++ b/conda/recipes/librapidsmpf/install_librapidsmpf.sh
@@ -2,4 +2,5 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION & AFFILIATES.
 # SPDX-License-Identifier: Apache-2.0
 
-cmake --install cpp/build --component rapidsmpf
+cmake --install cpp/build
+cmake --install cpp/build --component=benchmarking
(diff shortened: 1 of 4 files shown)


seberg:
Sorry for the noise, but trying close/reopen once, because this error seems too early and I can't repro locally...

conda/recipes/librapidsmpf/install_librapidsmpf.sh
```diff
@@ -2,4 +2,4 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION & AFFILIATES.
 # SPDX-License-Identifier: Apache-2.0
 
-cmake --install cpp/build
+cmake --install cpp/build --component rapidsmpf
```
seberg, line 5: I think this may be more of a hot-fix.

The [`cpp/tests/CMakeLists.txt`](https://github.com/rapidsai/rapidsmpf/blob/branch-25.06/cpp/tests/CMakeLists.txt) installs it's components also without `--component=testing` right now.  Not sure exactly why, maybe as it doesn't use `rapids_test_add`/`rapids_test_install_relocatable` (but it's a bit more complicated).

On the plus side, excluding the tests almost halves the size of the `librapidsmpf` package :).

madsbk (approved):
Thanks @seberg

conda/recipes/librapidsmpf/install_librapidsmpf.sh
pentschev: I agree about tests, but I think it's useful to have benchmarks in the package. How about adding benchmarks to a separate component (currently they're part of `testing`) and including those in the package as well?

seberg: OK, that also made me realize that all that was missing from testing was `EXCLUDE_FROM_ALL`, so now benchmarking is its own component, but explicitly included in the package.

pentschev (approved):
LGTM as well. Thanks @seberg !

PR #262, opened 2025-05-08
Add support for Python 3.13

gforsyth:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/120

This PR adds support for Python 3.13.

## Notes for Reviewers

This is part of ongoing work to add Python 3.13 support across RAPIDS.
It temporarily introduces a build/test matrix including Python 3.13, from https://github.com/rapidsai/shared-workflows/pull/268.

A follow-up PR will revert back to pointing at the `branch-25.06` branch of `shared-workflows` once all
RAPIDS repos have added Python 3.13 support.

### This will fail until all dependencies have been updated to Python 3.13

CI here is expected to fail until all of this project's upstream dependencies support Python 3.13.

This can be merged whenever all CI jobs are passing.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   conda-cpp-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@branch-25.06
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@python-3.13
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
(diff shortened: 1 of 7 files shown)


gforsyth:
Looks like `ray-default` doesn't build for Python 3.13 yet, not sure where they are in the conda-forge migration path.  I'm going to revert the workflow changes on that testing branch because it seems like `ray` is only being used for testing and isn't a runtime dependency.

gforsyth:
Depending on how slowly `ray` adds Python 3.13 support, we could annotate all the `ray` tests so they only run in a Python 3.12 environment, so we can run the rest of the tests under Python 3.13, but that can be done separately.

bdice (approved):

PR #265, opened 2025-05-08
Make `ProgressThread::remove_function` less blocking

pentschev:
Track completed functions to make `ProgressThread::remove_function` less blocking.

### DIFF
--- a/cpp/src/progress_thread.cpp
+++ b/cpp/src/progress_thread.cpp
@@ -84,36 +84,41 @@ void ProgressThread::remove_function(FunctionID function_id) {
         "Function was not registered with this ProgressThread"
     );
 
-    std::unique_lock lock(mutex_);
-    auto it = functions_.find(function_id.function_index);
-    RAPIDSMPF_EXPECTS(
-        it != functions_.end(), "Function not registered or already removed"
-    );
+    auto function_idx = function_id.function_index;
 
-    // Wait for the function to complete.
-    // iterator it can get invalidated, if some other thread erases a function. So, query
-    // functions_ instead
-    cv_.wait(lock, [&, f_idx = it->first]() { return functions_.at(f_idx).is_done; });
-
-    // Waiting done. Now, mutex_ is locked again
-    functions_.erase(function_id.function_index);
-
-    if (functions_.empty()) {
-        thread_.pause();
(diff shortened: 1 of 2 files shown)


pentschev:
@nirandaperera just to follow-up on our brief offline discussion, this is a way we could make the existing `ProgressThread::remove_function` less blocking. However, it doesn't make the idea around implementation much different than yours, and if we were to make `add_function` less blocking as well we would end up with a staging structure here as well. I don't believe also what's in this PR is necessarily the right approach for us nor that it implies being superior to your implementation, just pushed this to have a counterpoint to the discussion. Also note that I didn't cleanup other things that now have become unnecessary, but we can do that if our discussion leads for us to keep on working on the current implementation instead of going with your reimplementation of `ProgressThread`.

cpp/src/progress_thread.cpp
```diff
@@ -84,36 +84,41 @@ void ProgressThread::remove_function(FunctionID function_id) {
...
+    std::vector<FunctionIndex> completed_functions;
     {
         std::lock_guard<std::mutex> lock(mutex_);
         for (auto& [id, function] : functions_) {
             function();
+            if (function.is_done) {
+                completed_functions.push_back(id);
+                completion_tracker_.mark_completed(id);
+            }
+        }
```
nirandaperera, line 111: @pentschev do you think we could traverse the `functions_` map without holding the lock? I think that is the bottleneck in the current impl. And suspect that's why @madsbk 's [PR of adding a small sleep](https://github.com/rapidsai/rapidsmpf/pull/253) helped.

pentschev: Yes, that is in essence what you've implemented as well, done in https://github.com/rapidsai/rapidsmpf/pull/265/commits/595647568e16dbac46c6ec4ef7112bf05e68a31a . (Again, not cleaned up for any resources that are now redundant)

nirandaperera: Let me go through this

nirandaperera:
@pentschev I think we are on the same page here. I also sent a PR https://github.com/rapidsai/rapidsmpf/pull/266 which I think is a little more simpler. LMK what you think

madsbk:

Issue #219, opened 2025-04-22
Make MPI an optional build and runtime dependency


TomAugspurger:
MPI is currently a required build and runtime dependency. We'd like to make it an optional dependency in both places:

- Most of our usage will probably be via UCXX, so the MPI dependency will be unneeded
- A required runtime dependency on MPI would complicate packaging for PyPI

This issue tracks the progress to making MPI an optional dependency. We'll need to

- [ ] Update our cmake build scripts to add a flag for including / excluding MPI
- [ ] Update our C++ code to make only include MPI when the build flag is provided
- [ ] Update any Python code relying on MPI to make it optional (not sure if we have any of this)
- [ ] Determine whether our conda packages should include a MPI as a build / runtime dependency

PR #269, opened 2025-05-12
feat: add wheel build of `rapidsmpf`

gforsyth:
xref #219

Part of rapidsai/build-planning#170

Adds wheels for `rapidsmpf`.

Currently the `librapidsmpf` and `rapidsmpf` wheels are combined into a single file.

My CMake isn't great, and for now I'm hacking in some paths with an environment variable, just so I can start testing this in CI, but I'll work on something more idiomatic.


python/rapidsmpf/rapidsmpf/tests/conftest.py
```diff
@@ -14,11 +14,34 @@
...
     from collections.abc import Generator
 
+    from mpi4py import MPI
+
     from rmm.pylibrmm.stream import Stream
 
     from rapidsmpf.communicator.communicator import Communicator
 
 
+def _get_mpi_module_or_skip() -> MPI:
```
TomAugspurger, line 24: I'm not sure if it's 100% equivalent, but we could also use `pytest.importorskip("mpi4py")` in any test that needs to import mpi4py. LMK if you want me to look into that.

gforsyth: yeah, I think that's a good idea.  I just fixed the import path to get builds passing locally, but I think `importorskip` makes good sense

gforsyth:
wheel tests were building `cupy` from source which is suboptimal.  I'll try to sort that out tomorrow

ci/build_wheel_rapidsmpf.sh
```diff
@@ -0,0 +1,50 @@
...
+    --exclude libucp.so.0 \
+    --exclude libucxx.so \
+    --exclude libucxx_python.so \
+    --exclude librapidsmpf.so \
+    -w "${RAPIDS_WHEEL_BLD_OUTPUT_DIR}" \
+    ${package_dir}/dist/*
+
+./ci/validate_wheel.sh "${package_dir}" "${RAPIDS_WHEEL_BLD_OUTPUT_DIR}"
+
+RAPIDS_PY_WHEEL_NAME="${package_name}_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 cpp "${RAPIDS_WHEEL_BLD_OUTPUT_DIR}"
```
pentschev, line 50: Should we be switching to GH artifacts, like in https://github.com/rapidsai/rapidsmpf/pull/274?

python/rapidsmpf/rapidsmpf/tests/conftest.py
pentschev: Agree with Mads, if `mpi4py` isn't available we can't do much on the Python end.

ci/build_wheel_rapidsmpf.sh
```diff
@@ -0,0 +1,50 @@
...
+    -r /tmp/requirements-build.txt
+
+# build with '--no-build-isolation', for better sccache hit rate
+# 0 really means "add --no-build-isolation" (ref: https://github.com/pypa/pip/issues/5735)
+export PIP_NO_BUILD_ISOLATION=0
+
+export SKBUILD_CMAKE_ARGS="-DBUILD_MPI_SUPPORT=OFF;-DBUILD_TESTS=OFF;-DBUILD_BENCHMARKS=OFF;-DBUILD_EXAMPLES=OFF;-DBUILD_IN_CONDA=OFF"
+
+SITE_PACKAGES=$(python -c "import site; print(site.getsitepackages()[0])")
+export SITE_PACKAGES
```
gforsyth, line 33: This is being set for path resolution in `python/rapidsmpf/CMakeLists.txt` (suboptimal)

cpp/CMakeLists.txt
```diff
@@ -241,11 +242,20 @@ include(GNUInstallDirs)
...
+    EXPORT rapidsmpf-exports
+  )
+else()
+  install(
+    TARGETS rapidsmpf
+    DESTINATION ${lib_dir}
+    EXPORT rapidsmpf-exports
+  )
+endif()
+
```
gforsyth, line 258: I'm sure there's a better way to do this, but without it, when we build a Python wheel, `librapidsmpf.so` gets installed into `./lib64/librapidsmpf.so` instead of `./rapidsmpf/lib64/librapidsmpf.so` and then we cannot load the shared object.

Conversely, if we install it into `./rapidsmpf/lib64/` exclusively, this prevents `conda-cpp` packages from locating the shared object.

python/rapidsmpf/CMakeLists.txt
```diff
@@ -10,16 +10,33 @@ include(rapids-cpm)
...
 
-find_package(rapidsmpf REQUIRED "${RAPIDS_VERSION}")
-find_package(ucxx-python REQUIRED "${UCXX_VERSION}")
+if(NOT BUILD_IN_CONDA)
+  set(RELOCATE_LIBRAPIDSMPF ON)
+  set(CUDA_STATIC_RUNTIME ON)
+  add_subdirectory(../../cpp librapidsmpf)
+
+  # TODO: don't use env-vars for this
+  include_directories($ENV{SITE_PACKAGES}/ucxx/include)
```
gforsyth, line 30: Without this (or similar) we fail at `rapids_find_package(ucxx)` because of complains about not being able to find `ucxx-config.cmake` or similar (but that file IS present, just not where it's looking)

python/rapidsmpf/CMakeLists.txt
```diff
@@ -10,16 +10,33 @@ include(rapids-cpm)
...
+  set(ucxx-python_DIR $ENV{SITE_PACKAGES}/ucxx/lib64/cmake/ucxx-python/)
```
gforsyth, line 31: Without this, compilation fails looking for `ucxx/python/exceptions.h`

gforsyth (commented):
Flagging some of my more questionable CMake decisions for @KyleFromNVIDIA to review

dependencies.yaml
```diff
@@ -101,10 +132,8 @@ dependencies:
       - output_types: [conda, pyproject, requirements]
         packages:
           - cython>=3.0.3
-          - *rmm_unsuffixed
-          - pylibcudf==25.6.*,>=0.0.0a0
```
wence-, line 105: We definitely need pylibcudf to build the cython.

gforsyth: Yep, it's just been moved to https://github.com/rapidsai/rapidsmpf/pull/269/files#diff-5475a6d76de4c506ee92cf6f941bba30a6a07b5881cfea531d42a6ec035095a6R280

cpp/CMakeLists.txt
KyleFromNVIDIA: You can use `wheel.install-dir` in `pyproject.toml` to do this. See https://github.com/rapidsai/cudf/blob/5671b501ba9b4443a161d55a3b029bac565faf79/python/libcudf/pyproject.toml#L69 for an example.

Since the Conda build of the library doesn't build as a wheel, this setting gets ignored, and it gets installed in the correct directory.

python/rapidsmpf/CMakeLists.txt
```diff
@@ -10,16 +10,33 @@ include(rapids-cpm)
...
+
+  rapids_find_package(ucx REQUIRED)
+  rapids_find_package(ucxx REQUIRED "${UCXX_VERSION}")
+else()
+  find_package(rapidsmpf REQUIRED "${RAPIDS_VERSION}")
```
KyleFromNVIDIA, line 36: See https://github.com/rapidsai/kvikio/blob/5ab1a8ac4d1100b69fb57699a292a8ad4816e4fe/python/libkvikio/CMakeLists.txt#L31-L55 for an example of how we usually handle these library wheel builds.

python/rapidsmpf/CMakeLists.txt
KyleFromNVIDIA: I *think* that once you've reorganized the build to be more like kvikio and cudf, this shouldn't be needed. If you're still getting errors, we can take another look.

KyleFromNVIDIA (changes requested):

cpp/CMakeLists.txt
gforsyth: The `conda-python-build` _does_ build as a wheel, which is why I did this.

https://github.com/rapidsai/rapidsmpf/actions/runs/15006819191/job/42167765062?pr=269#step:9:1318

This, I think, is an artifact of not splitting this into c++ and Python wheels, which seems like more work, but would avoid some of these headaches.

gforsyth:
closing to reset the workflow ID because of artifact name collisions

pentschev:
> Hey @pentschev ! Do you have any thoughts on the log-level test failures here? I can't figure out why those warnings are getting thrown.

I see that our test is matching a warning from UCX, what is curious is that you're not expected to have changed anything in the conda path. However, I'm suspicious of `load_library()` making some changes to the import order and causing that, specifically in https://github.com/rapidsai/rapidsmpf/pull/269/files#diff-cd358590c5bd8d2361909674dac2e1fbc83507be07652f476d2af6ceb206ad5dR64, I think this is not executed at all in conda, but that's the only change I'm seeing that could have changed execution path/order at the moment.

One other thing I've noticed is that we're perhaps not seeing that error in the wheels tests because [now `test_communicator.py` is completely skipped if mpi4py isn't available](https://github.com/rapidsai/rapidsmpf/pull/269/files#diff-d0c14d90780b0e67cbfd7dc6438b57595f8ada40f94a8e1a173c7ae49f469693R10), so the neither `test_log_level` nor `test_ucxx` in that file run, only `test_mpi` should be skipped. The same is done in `test_example.py` as well: https://github.com/rapidsai/rapidsmpf/pull/269/files#diff-0be558322c3e2111feeb64c7921a0214e00f701b21796f4d9bb50978aae87c5cR16-R17 . Could you fix that so we can see whether wheels will fail too?

pentschev:
> ~I don't understand why `test_log_level` is running at all, since it relies on `MPI.COMM_WORLD`, so it should be skipped: https://github.com/rapidsai/rapidsmpf/pull/269/files#diff-d0c14d90780b0e67cbfd7dc6438b57595f8ada40f94a8e1a173c7ae49f469693R21~
> 
> nevermind, I was thinking of the wheels
>
> `test_example.py` only has one test in it, and the `importorskip` is required before `bulk_mpi_shuffle` is imported.
> 
> All of the tests in `test_communicator.py` make use of `MPI.COMM_WORLD` and the imports also require `mpi4py` at collection time.

Ah sorry, you're correct in both cases. We made a bad choice of relying on MPI to always bootstrap our testing environment and now got stuck being unable to run several of them for wheels. In that case we really can't test those on wheels at the moment. Do you have a sense whether the loading could be the cause for any of that on conda?

gforsyth:
> Do you have a sense whether the loading could be the cause for any of that on conda?

I don't think so -- for the `conda` builds (and installs, etc) there isn't a `librapidsmpf` (Python) module, so the attempt to import should fail and those loads never happen.

pentschev:
Not entirely sure why, but its seems there's an issue with import ordering. The following resolves the failure for me locally:

```
diff --git a/python/rapidsmpf/rapidsmpf/tests/test_communicator.py b/python/rapidsmpf/rapidsmpf/tests/test_communicator.py
index 2180c0d..4bb0315 100644
--- a/python/rapidsmpf/rapidsmpf/tests/test_communicator.py
+++ b/python/rapidsmpf/rapidsmpf/tests/test_communicator.py
@@ -4,12 +4,12 @@ from __future__ import annotations

 import pytest

-from rapidsmpf.communicator.communicator import LOG_LEVEL
-from rapidsmpf.communicator.testing import initialize_ucxx, ucxx_mpi_setup
-
 MPI = pytest.importorskip("mpi4py.MPI")
 from rapidsmpf.communicator.mpi import new_communicator  # noqa: E402

+from rapidsmpf.communicator.communicator import LOG_LEVEL
+from rapidsmpf.communicator.testing import initialize_ucxx, ucxx_mpi_setup
+

 @pytest.mark.parametrize(
     "level", list(LOG_LEVEL), ids=[level.name for level in LOG_LEVEL]
```

Do you see any obvious issues with the patch above that I'm overlooking? If not, could you try applying that?

gforsyth:
woohoo!  that gets the test passing again, thanks @pentschev !  

Obviously we don't want a test that's that sensitive to import order, but that can be handled separately.

jameslamb:
@gforsyth as part of this, could you also adopt the patterns I'm rolling out to the rest of the wheel scripts, similar to https://github.com/rapidsai/cudf/pull/18902?

* using `source rapids-init-pip` near the top of scripts using `pip`
* using `${PIP_CONSTRAINT}` instead of setting up a new file like `/tmp/constraints.txt`

Or if you'd prefer, let me know and I'll push those changes directly to your branch.

python/rapidsmpf/CMakeLists.txt
```diff
@@ -18,8 +19,13 @@ project(
   LANGUAGES CXX CUDA
 )
 
+set(rapidsmpf_DIR $ENV{SITE_PACKAGES}/librapidsmpf/lib64/cmake/rapidsmpf/)
```
KyleFromNVIDIA, line 22: The [kvikio build](https://github.com/rapidsai/kvikio/blob/ad66c30aaa11603ab2749c49fe7cb8ae69fe70ee/python/kvikio/CMakeLists.txt#L31) does not require this to be set for `libkvikio`. Let's try taking it out and see if this was fixed by the `install-dir` option.

bdice:
#301 (adding devcontainers) will conflict with this PR. Let's land this PR first, and then #301.

python/rapidsmpf/CMakeLists.txt
gforsyth: Still fails: https://github.com/rapidsai/rapidsmpf/actions/runs/15276337429/job/42964318262?pr=269

I'm going to revert that change for now

bdice (approved):

python/rapidsmpf/CMakeLists.txt
```diff
@@ -18,8 +19,12 @@ project(
   LANGUAGES CXX CUDA
 )
 
+set(ucxx-python_DIR $ENV{SITE_PACKAGES}/ucxx/lib64/cmake/ucxx-python/)
+include_directories($ENV{SITE_PACKAGES}/ucxx/include)
```
KyleFromNVIDIA, line 23: It's too late to add `[project.entry-points."cmake.prefix"]` to ucxx, since that is now in code freeze for 25.06, so I will approve. We will add it to ucxx for 25.08 and then remove these lines.

KyleFromNVIDIA (approved):

PR #270, opened 2025-05-12
Buffer copy utils

nirandaperera:
This PR adds the following utils to the `Buffer` class. These will be used in the Chunk in a follow up PR. 

```c++ 
public: 
    /**
     * @brief Copy data from this buffer to a destination buffer with a given offset.
     *
     * @param dest Destination buffer.
     * @param dest_offset Offset of the destination buffer.
     * @param stream CUDA stream to use for the copy.
     * @returns Number of bytes written to the destination buffer.
     * @throws std::invalid_argument if copy violates the bounds of the destination
     * buffer.
     */
    [[nodiscard]] std::ptrdiff_t copy_to(
        Buffer& dest, std::ptrdiff_t dest_offset, rmm::cuda_stream_view stream
    ) const;

private:
    /**
     * @brief Copy a slice of the buffer to a new buffer.
     *
     * @param offset Offset in bytes from the start of the buffer.
     * @param length Length in bytes of the slice.
     * @param stream CUDA stream to use for the copy.
     * @returns A new buffer containing the copied slice.
     */
    [[nodiscard]] std::unique_ptr<Buffer> copy_slice(
        std::ptrdiff_t offset, std::ptrdiff_t length, rmm::cuda_stream_view stream
    ) const;

    /**
     * @brief Copy a slice of the buffer to a new buffer.
     *
     * @param target Memory type of the new buffer.
     * @param offset Offset in bytes from the start of the buffer.
     * @param length Length in bytes of the slice.
     * @param stream CUDA stream to use for the copy.
     * @returns A new buffer containing the copied slice.
     */
    [[nodiscard]] std::unique_ptr<Buffer> copy_slice(
        MemoryType target,
        std::ptrdiff_t offset,
        std::ptrdiff_t length,
        rmm::cuda_stream_view stream
    ) const;
```

### DIFF
--- a/cpp/include/rapidsmpf/buffer/buffer.hpp
+++ b/cpp/include/rapidsmpf/buffer/buffer.hpp
@@ -124,7 +124,7 @@ class Buffer {
      *
      * @throws std::logic_error if the buffer does not manage host memory.
      */
-    [[nodiscard]] constexpr std::unique_ptr<std::vector<uint8_t>> const& host() const {
+    [[nodiscard]] constexpr HostStorageT const& host() const {
         if (const auto* ref = std::get_if<HostStorageT>(&storage_)) {
             return *ref;
         } else {
@@ -139,7 +139,7 @@ class Buffer {
      *
      * @throws std::logic_error if the buffer does not manage device memory.
      */
-    [[nodiscard]] constexpr std::unique_ptr<rmm::device_buffer> const& device() const {
+    [[nodiscard]] constexpr DeviceStorageT const& device() const {
         if (const auto* ref = std::get_if<DeviceStorageT>(&storage_)) {
             return *ref;
         } else {
@@ -180,6 +180,15 @@ class Buffer {
             },
             storage_
         );
+    }
+
+    /**
+     * @brief Set the event for the buffer.
+     *
+     * @param event The event to set.
+     */
+    inline void override_event(std::shared_ptr<Event> event) {
+        event_ = std::move(event);
     }
 
     /**
@@ -233,7 +242,7 @@ class Buffer {
      *
      * @throws std::logic_error if the buffer does not manage host memory.
      */
-    [[nodiscard]] std::unique_ptr<std::vector<uint8_t>>& host() {
+    [[nodiscard]] HostStorageT& host() {
         if (auto ref = std::get_if<HostStorageT>(&storage_)) {
             return *ref;
         } else {
@@ -248,7 +257,7 @@ class Buffer {
      *
      * @throws std::logic_error if the buffer does not manage device memory.
      */
-    [[nodiscard]] std::unique_ptr<rmm::device_buffer>& device() {
+    [[nodiscard]] DeviceStorageT& device() {
         if (auto ref = std::get_if<DeviceStorageT>(&storage_)) {
             return *ref;
         } else {
@@ -273,6 +282,34 @@ class Buffer {
      */
     [[nodiscard]] std::unique_ptr<Buffer> copy(
         MemoryType target, rmm::cuda_stream_view stream
+    ) const;
+
+    /**
+     * @brief Copy a slice of the buffer to a new buffer.
+     *
+     * @param offset Offset in bytes from the start of the buffer.
+     * @param length Length in bytes of the slice.
+     * @param stream CUDA stream to use for the copy.
+     * @returns A new buffer containing the copied slice.
+     */
+    [[nodiscard]] std::unique_ptr<Buffer> copy_slice(
+        std::ptrdiff_t offset, std::ptrdiff_t length, rmm::cuda_stream_view stream
+    ) const;
+
+    /**
+     * @brief Copy a slice of the buffer to a new buffer.
+     *
+     * @param target Memory type of the new buffer.
+     * @param offset Offset in bytes from the start of the buffer.
+     * @param length Length in bytes of the slice.
+     * @param stream CUDA stream to use for the copy.
+     * @returns A new buffer containing the copied slice.
+     */
+    [[nodiscard]] std::unique_ptr<Buffer> copy_slice(
+        MemoryType target,
+        std::ptrdiff_t offset,
+        std::ptrdiff_t length,
+        rmm::cuda_stream_view stream
     ) const;
 
   public:
--- a/cpp/include/rapidsmpf/buffer/buffer.hpp
+++ b/cpp/include/rapidsmpf/buffer/buffer.hpp
@@ -199,6 +199,20 @@ class Buffer {
      */
     [[nodiscard]] bool is_ready() const;
 
+    /**
+     * @brief Copy data from this buffer to a destination buffer with a given offset.
+     *
+     * @param dest Destination buffer.
+     * @param dest_offset Offset of the destination buffer.
+     * @param stream CUDA stream to use for the copy.
+     * @returns Number of bytes written to the destination buffer.
+     * @throws std::invalid_argument if copy violates the bounds of the destination
+     * buffer.
+     */
+    [[nodiscard]] std::ptrdiff_t copy_to(
+        Buffer& dest, std::ptrdiff_t dest_offset, rmm::cuda_stream_view stream
+    ) const;
+
     /// @brief Delete move and copy constructors and assignment operators.
     Buffer(Buffer&&) = delete;
     Buffer(Buffer const&) = delete;
--- a/cpp/include/rapidsmpf/buffer/resource.hpp
+++ b/cpp/include/rapidsmpf/buffer/resource.hpp
@@ -47,8 +47,8 @@ class MemoryReservation {
      */
     MemoryReservation(MemoryReservation&& o)
         : MemoryReservation{
-            o.mem_type_, std::exchange(o.br_, nullptr), std::exchange(o.size_, 0)
-        } {}
+              o.mem_type_, std::exchange(o.br_, nullptr), std::exchange(o.size_, 0)
+          } {}
 
     /**
      * @brief Move assignment operator for MemoryReservation.
@@ -344,6 +344,26 @@ class BufferResource {
     std::unique_ptr<Buffer> copy(
         MemoryType target,
         std::unique_ptr<Buffer> const& buffer,
+        rmm::cuda_stream_view stream,
+        MemoryReservation& reservation
+    );
+
+    /**
+     * @brief Copy a slice of the buffer to a new buffer.
+     *
+     * @param target Memory type of the new buffer.
+     * @param buffer The buffer to copy from.
+     * @param offset Offset in bytes from the start of the buffer.
+     * @param length Length in bytes of the slice.
+     * @param stream CUDA stream to use for the copy.
+     * @param reservation The reservation to use for memory allocations.
+     * @returns A new buffer containing the copied slice.
+     */
+    [[nodiscard]] std::unique_ptr<Buffer> copy_slice(
+        MemoryType target,
+        std::unique_ptr<Buffer> const& buffer,
+        std::ptrdiff_t offset,
+        std::ptrdiff_t length,
         rmm::cuda_stream_view stream,
         MemoryReservation& reservation
     );
--- a/cpp/include/rapidsmpf/buffer/resource.hpp
+++ b/cpp/include/rapidsmpf/buffer/resource.hpp
@@ -47,8 +47,8 @@ class MemoryReservation {
      */
     MemoryReservation(MemoryReservation&& o)
         : MemoryReservation{
-              o.mem_type_, std::exchange(o.br_, nullptr), std::exchange(o.size_, 0)
-          } {}
+            o.mem_type_, std::exchange(o.br_, nullptr), std::exchange(o.size_, 0)
+        } {}
 
     /**
      * @brief Move assignment operator for MemoryReservation.
--- a/cpp/src/buffer/buffer.cpp
+++ b/cpp/src/buffer/buffer.cpp
@@ -227,6 +227,43 @@ std::unique_ptr<Buffer> Buffer::copy_slice(
     );
 }
 
+std::ptrdiff_t Buffer::copy_to(
+    Buffer& dest, std::ptrdiff_t dest_offset, rmm::cuda_stream_view stream
+) const {
+    RAPIDSMPF_EXPECTS(
+        dest.size - size_t(dest_offset) >= size,
+        "destination buffer is too small",
+        std::invalid_argument
+    );
+
+    if (size == 0) {  // empty buffer, nothing to do
+        return 0;
+    }
+
+    cudaMemcpyKind kind;
+
+    // if both buffers are on the host, use memcpy, otherwise, use cudaMemcpyAsync
+    if (mem_type() == MemoryType::HOST && dest.mem_type() == MemoryType::HOST) {
+        std::memcpy(static_cast<uint8_t*>(dest.data()) + dest_offset, data(), size);
+        return std::ptrdiff_t(size);
+    } else if (mem_type() == MemoryType::HOST) {
+        kind = cudaMemcpyHostToDevice;
+    } else if (dest.mem_type() == MemoryType::HOST) {
+        kind = cudaMemcpyDeviceToHost;
+    } else {
+        kind = cudaMemcpyDeviceToDevice;
+    }
+
+    RAPIDSMPF_CUDA_TRY_ALLOC(cudaMemcpyAsync(
+        static_cast<cuda::std::byte*>(dest.data()) + dest_offset,
+        data(),
+        size,
+        kind,
+        stream
+    ));
+    return std::ptrdiff_t(size);
+}
+
 bool Buffer::is_ready() const {
     return !event_ || event_->is_ready();
 }
--- a/cpp/src/buffer/buffer.cpp
+++ b/cpp/src/buffer/buffer.cpp
@@ -139,6 +139,94 @@ std::unique_ptr<Buffer> Buffer::copy(MemoryType target, rmm::cuda_stream_view st
     );
 }
 
+std::unique_ptr<Buffer> Buffer::copy_slice(
+    std::ptrdiff_t offset, std::ptrdiff_t length, rmm::cuda_stream_view stream
+) const {
+    RAPIDSMPF_EXPECTS(offset <= std::ptrdiff_t(size), "offset can't be more than size");
+    RAPIDSMPF_EXPECTS(
+        offset + length <= std::ptrdiff_t(size), "offset + length can't be more than size"
+    );
+    return std::visit(
+        overloaded{
+            [&](const HostStorageT& storage) {
+                auto host_buf = std::unique_ptr<Buffer>(new Buffer{
+                    std::make_unique<std::vector<uint8_t>>(
+                        storage->begin() + offset, storage->begin() + offset + length
+                    ),
+                    br
+                });
+                host_buf->override_event(event_);  // if there was an event, use it
+                return host_buf;
+            },
+            [&](DeviceStorageT const& storage) {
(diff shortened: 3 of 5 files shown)


cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -182,6 +182,15 @@ class Buffer {
         );
     }
 
+    /**
+     * @brief Set the event for the buffer.
+     *
```
madsbk, line 187: Add a note about the consequence of overwriting the exist event.

cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -190,6 +199,20 @@ class Buffer {
      */
     [[nodiscard]] bool is_ready() const;
 
+    /**
+     * @brief Copy data from this buffer to a destination buffer with a given offset.
+     *
+     * @param dest Destination buffer.
+     * @param dest_offset Offset of the destination buffer.
```
madsbk, line 206:
```suggestion
     * @param dest_offset Offset of the destination buffer (in bytes).
```

cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -275,6 +298,34 @@ class Buffer {
         MemoryType target, rmm::cuda_stream_view stream
     ) const;
 
+    /**
+     * @brief Copy a slice of the buffer to a new buffer.
+     *
+     * @param offset Offset in bytes from the start of the buffer.
```
madsbk, line 304:
```suggestion
     * @param offset Offset from the start of the buffer (in bytes).
```

cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -275,6 +298,34 @@ class Buffer {
...
+     */
+    [[nodiscard]] std::unique_ptr<Buffer> copy_slice(
+        std::ptrdiff_t offset, std::ptrdiff_t length, rmm::cuda_stream_view stream
+    ) const;
+
+    /**
+     * @brief Copy a slice of the buffer to a new buffer.
+     *
+     * @param target Memory type of the new buffer.
+     * @param offset Offset in bytes from the start of the buffer.
```
madsbk, line 317:
```suggestion
     * @param offset Offset from the start of the buffer (in bytes).
```

cpp/include/rapidsmpf/buffer/resource.hpp
```diff
@@ -348,6 +348,26 @@ class BufferResource {
         MemoryReservation& reservation
     );
 
+    /**
+     * @brief Copy a slice of the buffer to a new buffer.
+     *
+     * @param target Memory type of the new buffer.
+     * @param buffer The buffer to copy from.
+     * @param offset Offset in bytes from the start of the buffer.
```
madsbk, line 356:
```suggestion
     * @param offset Offset from the start of the buffer (in bytes).
```

cpp/include/rapidsmpf/buffer/resource.hpp
```diff
@@ -348,6 +348,26 @@ class BufferResource {
...
+     * @param length Length in bytes of the slice.
```
madsbk, line 357:
```suggestion
     * @param length Length of the slice (in bytes).
```

cpp/src/buffer/buffer.cpp
```diff
@@ -139,6 +145,133 @@ std::unique_ptr<Buffer> Buffer::copy(MemoryType target, rmm::cuda_stream_view st
...
+    return std::visit(
+        overloaded{
+            [&](const HostStorageT& storage) {  // host -> host
+                auto host_buf = std::unique_ptr<Buffer>(new Buffer{
+                    std::make_unique<std::vector<uint8_t>>(
+                        storage->begin() + offset, storage->begin() + offset + length
+                    ),
+                    br
+                });
+                host_buf->override_event(event_);  // if there was an event, use it
```
madsbk, line 164: Can you give `event_` to the Buffer's ctor instead?

cpp/src/buffer/buffer.cpp
```diff
@@ -139,6 +145,133 @@ std::unique_ptr<Buffer> Buffer::copy(MemoryType target, rmm::cuda_stream_view st
...
+                        static_cast<cuda::std::byte const*>(storage->data()) + offset,
+                        size_t(length),
+                        cudaMemcpyDeviceToHost,
+                        stream
+                    ));
+                    auto host_buf =
+                        std::unique_ptr<Buffer>(new Buffer{std::move(ret), br});
+                    // Create a new event to track the async copy only if length > 0
+                    if (length > 0) {
+                        host_buf->override_event(std::make_shared<Event>(stream));
```
madsbk, line 228: Again, give it to the Buffer's ctor

madsbk (commented):
Overall look good

cpp/src/buffer/buffer.cpp
nirandaperera: @madsbk HostBuffer constructor doesnt have an event arg :slightly_frowning_face:

madsbk: So what happens if the data in `this` buffer hasn't been written yet? Since we do not wait on `event_`, what will happen in:
```c++
                   std::make_unique<std::vector<uint8_t>>(
                        storage->begin() + offset, storage->begin() + offset + length
                    ),
```

nirandaperera: @madsbk Ah! that's a good point. `throw` if its not ready, maybe? or return `nullptr` so that it can be attempted later on?

madsbk: Or wait on the event, `cudaEventSynchronize` ?

madsbk: True, but if the alternative is a crash? Maybe we should have wrapper function of `cudaEventSynchronize` that times the call and adds it to the statistics. My guess is that we almost never would block here.

nirandaperera: I think that's true. Okay, let me add that.

cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -95,15 +95,15 @@ class Buffer {
...
+
       private:
         cudaEvent_t event_;  ///< CUDA event used to track device memory allocation
         std::atomic<bool> done_{false
         };  ///< Cache of the event status to avoid unnecessary queries.
-        mutable std::mutex mutex_;  ///< Protects access to event_
-        std::atomic<bool> destroying_{false
-        };  ///< Flag to indicate destruction in progress
-        std::atomic<int> active_readers_{0
-        };  ///< Number of threads currently executing is_ready()
```
nirandaperera, line 106: tombstones

cpp/src/buffer/buffer.cpp
nirandaperera: @madsbk I have added the changes now

cpp/src/buffer/buffer.cpp
```diff
@@ -33,14 +33,21 @@ Buffer::Event::~Event() {
 
 [[nodiscard]] bool Buffer::Event::is_ready() {
     if (!done_.load(std::memory_order_relaxed)) {
-        bool result = cudaEventQuery(event_) == cudaSuccess;
+        bool result = (cudaEventQuery(event_) == cudaSuccess);
```
madsbk, line 36: Check for other errors like:
```suggestion
        ​cudaError_t err = cudaEventQuery(event_);
        if(err != cudaSuccess && err != cudaErrorNotReady) {
            RAPIDSMPF_CUDA_TRY(err);
        }
        bool result = (cudaEventQuery(event_) == cudaSuccess);
```

cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -275,6 +319,34 @@ class Buffer {
...
+    /**
+     * @brief Copy a slice of the buffer to a new buffer.
+     *
+     * @param offset Offset from the start of the buffer (in bytes).
+     * @param length Length of the slice (in bytes).
+     * @param stream CUDA stream to use for the copy.
+     * @returns A new buffer containing the copied slice.
+     */
+    [[nodiscard]] std::unique_ptr<Buffer> copy_slice(
+        std::ptrdiff_t offset, std::ptrdiff_t length, rmm::cuda_stream_view stream
```
wence-, line 331: *nit*: length should probably be `std::size_t`.

cpp/src/buffer/buffer.cpp
```diff
@@ -139,8 +152,148 @@ std::unique_ptr<Buffer> Buffer::copy(MemoryType target, rmm::cuda_stream_view st
...
+) const {
+    RAPIDSMPF_EXPECTS(offset <= std::ptrdiff_t(size), "offset can't be more than size");
+    RAPIDSMPF_EXPECTS(
+        offset + length <= std::ptrdiff_t(size), "offset + length can't be more than size"
+    );
+    return std::visit(
+        overloaded{
+            [&](const HostStorageT& storage) {  // host -> host
+                // wait for this buffer to be ready, as host-to-host copy is synchronous
+                wait_for_ready();
```
pentschev, line 166: Where is `copy_slice` intended to be called from in practice? Many of the `Buffer` operations occur during the shuffler's progress thread, we don't want to block there. If it needs to happen as part of the shuffle loop maybe we should use `cudaMemcpyAsync(..., cudaMemcpyHostToHost, stream)`. I'd even argue we should heavily prefer that, because this may cause unintended/hidden synchronizations and make the caller decide when it's really required to synchronize.

pentschev: In fact, I think we could do the same in https://github.com/rapidsai/rapidsmpf/blob/120f13d164c0e1d10a4e0be82f25eb76947f177f/cpp/src/buffer/buffer.cpp#L83-L85, where we previously debated if there should exist some synchronization. This way we would make all of `Buffer` non-blocking but would still ensure safety w.r.t. CUDA streams (we may need to revisit that `copy()` isn't called somewhere where we potentially ignored H2H stream synchronization).

cpp/src/buffer/buffer.cpp
pentschev: Good catch, I previously assumed non-success means the event isn't done, but that's not always true.

cpp/src/buffer/buffer.cpp
```diff
@@ -139,8 +152,148 @@ std::unique_ptr<Buffer> Buffer::copy(MemoryType target, rmm::cuda_stream_view st
...
+std::unique_ptr<Buffer> Buffer::copy_slice(
+    MemoryType target,
+    std::ptrdiff_t offset,
+    std::ptrdiff_t length,
+    rmm::cuda_stream_view stream
+) const {
+    RAPIDSMPF_EXPECTS(offset <= std::ptrdiff_t(size), "offset can't be more than size");
+    RAPIDSMPF_EXPECTS(
+        offset + length <= std::ptrdiff_t(size), "offset + length can't be more than size"
+    );
```
pentschev, line 202:
```suggestion
    RAPIDSMPF_EXPECTS(offset <= std::ptrdiff_t(size), "offset can't be greater than size");
    RAPIDSMPF_EXPECTS(
        offset + length <= std::ptrdiff_t(size), "offset + length can't be greater than size"
    );
```

cpp/src/buffer/buffer.cpp
```diff
@@ -139,8 +152,148 @@ std::unique_ptr<Buffer> Buffer::copy(MemoryType target, rmm::cuda_stream_view st
...
+
+    // if both buffers are on the host, use memcpy, otherwise, use cudaMemcpyAsync
+    if (mem_type() == MemoryType::HOST && dest.mem_type() == MemoryType::HOST) {
+        // if the source buffer is not ready, return 0
+        if (!is_ready()) {
+            return 0;
+        }
+        // src buffer is ready, copy the data
+        std::memcpy(static_cast<uint8_t*>(dest.data()) + dest_offset, data(), size);
+        return std::ptrdiff_t(size);
```
pentschev, line 270: Same here, maybe we should consider `cudaMemcpyAsync(..., cudaMemcpyHostToHost, stream)`.

nirandaperera: Ah. yes, that would simplify things a lot.

nirandaperera: @pentschev after reading the docs, I feel like it is not async. 
https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g85073372f776b4c4d5f89f7124b7bf79

nirandaperera: Please check this guide 
https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html

cpp/src/buffer/buffer.cpp
nirandaperera: @pentschev I raised the same issue with @madsbk. https://github.com/rapidsai/rapidsmpf/pull/270#discussion_r2087487972
Our conclusion was to add a wait until the source buffer was ready. 

But now looking at the cudaMemcpyAsync host to host option, I am also more inclined to use that. 
https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html
However, I think we might not achieve the full "async" semantics, because the driver seems to choose sync copies based on pageable memory, pinned memory etc.

pentschev: > However, I think we might not achieve the full "async" semantics, because the driver seems to choose sync copies based on pageable memory, pinned memory etc.

Yes, but that is the case for any transfer, including D2H/H2D, if pageable memory is involved it will not be async, but it still makes sense to use async calls and "try" to be async. However, that raises an interesting point: any such copies _may_ block the shuffler's progress thread if they're called there.

In any case, I think the explicit wait is not really necessary if we use `cudaMemcpyAsync(..., cudaMemcpyHostToHost, stream)`, in such case we should have an implicit synchronization then, and in the future if something changes where CUDA can now make H2H copies async we're covered. Alternatively, we could explore something like I said, write our own H2H copy that is async.

Another thing we could explore in the future is using pinned host memory if we want to try and squeeze every single bit of performance. This won't help to make H2H transfers with `cudaMemcpyAsync` async, but it may to ensure all transfers to/from device are async.

cpp/src/buffer/buffer.cpp
```diff
@@ -139,8 +155,135 @@ std::unique_ptr<Buffer> Buffer::copy(MemoryType target, rmm::cuda_stream_view st
...
+            static_cast<cuda::std::uint8_t*>(out_buf->data()),
+            static_cast<cuda::std::uint8_t*>(storage->data()) + offset,
+            length,
+            Kind,
+            stream
+        ));
+        // override the event to track the async copy, if the copy is not host-to-host
+        if constexpr (Kind != cudaMemcpyHostToHost) {
+            out_buf->override_event(std::make_shared<Buffer::Event>(stream));
+        }
```
wence-, line 182: *question*: If the copy is host-to-host why do we not need an event? I thought that `cudaMemcpyAsync` is "asynchronous with respect to the host"?

Answer: https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html

Maybe reference that here.

cpp/src/buffer/buffer.cpp
wence-: That seems to suggest we could just use `cudaMemcpyHostToHost` for this case too. It might be blocking but the `std::memcpy` is "blocking" too.

cpp/src/buffer/buffer.cpp
```diff
@@ -139,8 +155,135 @@ std::unique_ptr<Buffer> Buffer::copy(MemoryType target, rmm::cuda_stream_view st
...
+    }
+
+    RAPIDSMPF_CUDA_TRY(cudaMemcpyAsync(
+        static_cast<cuda::std::byte*>(dest.data()) + dest_offset,
+        data(),
+        size,
+        kind,
+        stream
+    ));
+    return std::ptrdiff_t(size);
```
wence-, line 276: Needs to `cudaEventRecord(stream)` for the destination buffer.

wence- (changes requested):
Looking good, I think we are missing one or two events (see comments)

cpp/src/buffer/buffer.cpp
wence-: Still relevant.

cpp/src/buffer/buffer.cpp
```diff
@@ -139,8 +155,135 @@ std::unique_ptr<Buffer> Buffer::copy(MemoryType target, rmm::cuda_stream_view st
...
+    const T& storage,
+    std::ptrdiff_t offset,
+    std::size_t length,
+    MemoryReservation& target_reserv,
+    rmm::cuda_stream_view stream
+) {
+    // allocate buffer using the target reservation
+    auto out_buf = target_reserv.br()->allocate(
+        target_reserv.mem_type(), length, stream, target_reserv
+    );
```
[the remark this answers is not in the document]
nirandaperera: Sure. I overlooked this. I'll remove the storage template. `cudaMemcpyKind Kind` was templated mainly to use `if constexpr` in L180

cpp/src/buffer/buffer.cpp
nirandaperera: @wence- My idea was, in the case of `Buffer::copy_to`, we can return `0` if the source buffer is not ready and then the caller can call `copy_to` again later on. But I also agree that using `cudaMemcpyHostToHost` is a safer option. 
Any preferences?

cpp/src/buffer/buffer.cpp
nirandaperera: @wence- Yes, but this was intentional. Say we want to concat a list of buffers, to a destination buffer. So, we can differ event creation after calling all `copy_to`s.

cpp/src/buffer/buffer.cpp
pentschev: I think for this PR I'd prefer we switch to `cudaMemcpyHostToHost` and try to address a more efficient pipeline where we can make all non-blocking in a follow-up PR, it's best to avoid addressing multiple issues at once.

cpp/src/buffer/buffer.cpp
wence-: ah sorry, I meant cudastreamewaitevent

wence-: ie the copy on this stream has to be ordered wrt the work captured in the event of the source buffer

cpp/src/buffer/buffer.cpp
```diff
@@ -139,8 +155,179 @@ std::unique_ptr<Buffer> Buffer::copy(MemoryType target, rmm::cuda_stream_view st
...
+        auto out_buf = target_reserv.br()->allocate(
+            target_reserv.mem_type(), length, stream, target_reserv
+        );
+
+        if (length > 0) {
+            // if this buffer has an event, ask the current stream to wait for it, before
+            // performing the memcpy
+            if (event_) {
+                RAPIDSMPF_CUDA_TRY(cudaStreamWaitEvent(stream, event_->event()));
+            }
```
madsbk, line 191: Now that `cudaMemcpyAsync` is used, do we still need to wait?

nirandaperera: @madsbk Based on what @wence- suggested, this would be more of a marker put on the `stream` to wait on the event asynchronously. This would be more relevant if the `event` was recorded on a different stream than `stream`.

madsbk (approved):
Looks good, thanks @nirandaperera

wence- (approved):

nirandaperera:
\merge

PR #275, opened 2025-05-13
Turn off uploads of nightly conda packages

wence-:
They rely on ABI stability of pylibcudf's cython bindings, which doesn't exist.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,8 +2,8 @@ name: build
 
 on:
   push:
-    branches:
-      - "branch-*"
+    # branches:
+    #   - "branch-*"
     tags:
       - v[0-9][0-9].[0-9][0-9].[0-9][0-9]
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


wence-:
I suspect this would break the nightly `test.yaml` step (because it won't find the package). But I don't know how to fix that.

This change is the minimal one to stop uploading nightlies that are broken with high probability.

.github/workflows/build.yaml
```diff
@@ -45,15 +45,15 @@ jobs:
...
-      sha: ${{ inputs.sha }}
+  # upload-conda:
+  #   needs: [conda-cpp-build, conda-python-build]
+  #   secrets: inherit
+  #   uses: rapidsai/shared-workflows/.github/workflows/conda-upload-packages.yaml@cuda-12.9.0
+  #   with:
+  #     build_type: ${{ inputs.build_type || 'branch' }}
+  #     branch: ${{ inputs.branch }}
+  #     date: ${{ inputs.date }}
+  #     sha: ${{ inputs.sha }}
```
wence-, line 56: I actually do want to upload the cpp package, but I don't know how to do that. Ideas welcome.

seberg: I think you should be able to add:
```
      skip_upload_pkgs: rapidsmpf
```
(and maybe `librapidsmpf-tests`) instead of deleting the step.  E.g. [used here](https://github.com/rapidsai/cuvs/blob/6bba2ef71560c9a12fbaddcb083f5d80ca90933d/.github/workflows/build.yaml#L98).

madsbk:

PR #284, opened 2025-05-15
dask 2025.5.0 compat

TomAugspurger:


rjzamora:
Thanks @TomAugspurger ! Does this work for `dask<2025.5.0` as well?

TomAugspurger:
It requires dask>=2025.4.0, otherwise we'll get a `TypeError` since the `n_workers` keyword argument didn't exist before then.

Want me to put in a bit of compat code? It'd look identical to the first couple files in https://github.com/rapidsai/cuml/pull/6614/files.

TomAugspurger:
And I'm not able to reproduce the failure with our test suite locally, without this fix. I'll look into seeing making a failing test that passes with this.

TomAugspurger:
I take that back: the test suite does catch the error. I mistakenly had an older version of dask locally.

I added a bit of compat code so this should be good to go.

python/rapidsmpf/rapidsmpf/integrations/dask/_compat.py
```diff
@@ -0,0 +1,23 @@
...
+    return cast(str, distributed.__version__)
+
+
+@functools.lru_cache
+def DISTRIBUTED_2025_4_0() -> bool:
+    return cast(
+        bool,
+        packaging.version.parse(distributed_version())
+        >= packaging.version.parse("2025.4.0"),
+    )
```
rjzamora, line 23: Curious if you have any background motivation for using `functools.lru_cache` here instead of just extracting the version and defining `DISTRIBUTED_2025_4_0` at import time?

rjzamora (approved):
Fix seems good to me. Thanks for taking care of this so fast!

The compat code may be heavier than we "need", but I don't feel too strongly either way.

quasiben:
merged so I could move forward with some testing

pentschev:
> > Want me to put in a bit of compat code?
> 
> Only if we think we will need to support `dask<2025.4.0` (which **may** be a reasonable backup plan)? Could also just comment that the argument requires `dask>2025.4.0` in case we need to add the compat later?
> 
> > And I'm not able to reproduce the failure with our test suite locally, without this fix. I'll look into seeing making a failing test that passes with this.
> 
> Yeah, I had trouble reproducing with a small worker count on a DGX-1, may not be easy to test.

For this I would cite https://github.com/rapidsai/dask-cuda/issues/848 where we've discussed Dask/Distribute version compatibility and eventually realized that there's really no point in supporting multiple versions because the cost for testing them all would be too high. If we don't test them it's safe to assume they don't work, therefore I still don't think it makes sense for us to generally attempt supporting multiple Dask versions.

python/rapidsmpf/rapidsmpf/integrations/dask/_compat.py
TomAugspurger: I'm strongly on team "don't execute code on import", which sometimes leads to overkill like this.

TomAugspurger:
Yeah, the main thing here was not forcing people to create a new environment *right now* with all the benchmarking going on.

Once things settle down a bit we'll rip this out. I don't anticipate it being in the 25.6 release.

pentschev:
> Yeah, the main thing here was not forcing people to create a new environment _right now_ with all the benchmarking going on.
> 
> Once things settle down a bit we'll rip this out. I don't anticipate it being in the 25.6 release.

Yes, that's ok as a short-term migration mechanism, my comment was more of our general support policy. We can't really test all, so there's no point in promising compatibility (even if it's only for ourselves) when in practice we can't tell if that compatibility code will still be working a week into the future.

PR #285, opened 2025-05-16
Configuration System

madsbk:
This PR lays the groundwork for a centralized configuration system in RapidsMPF, allowing options to be passed through the stack without relying on global dictionaries or modifying function signatures.

Currently, only the logging level option is implemented as a proof of concept. Additional options will be introduced in future PRs.

#### Key Features
- Works seamlessly in both Python and C++.
- Efficient: options are initialized only once.
- Supports integration with environment variables while maintaining explicit control — important because environment variable access is not thread-safe.

#### C++ Snippet
```c++
// Initialize configuration options from environment variables.
rapidsmpf::config::Options options{rapidsmpf::config::get_environment_variables()};
// Pass it to the communicator
std::shared_ptr<rapidsmpf::Communicator> comm =
    std::make_shared<rapidsmpf::MPI>(MPI_COMM_WORLD, options);

...

// In its ctor, the communicator can then use `options.get` to retrieve the `LOG_LEVEL` enum.
// `level_from_string` is called once to convert the string representation in `options` to a `LOG_LEVEL`.
Communicator::Logger::Logger(Communicator* comm, config::Options options)
    : comm_{comm}, level_(options.get<LOG_LEVEL>("log", level_from_string)){};
```


python/rapidsmpf/rapidsmpf/config.pyx
```diff
@@ -0,0 +1,195 @@
...
+            `parser_type`.
+        """
+        if issubclass(parser_type, bool):
+            return _options_get_using_parse_string[bool_t](
+                self._handle, str.encode(key), bool(default_value)
+            )
+        elif issubclass(parser_type, int):
+            return _options_get_using_parse_string[int64_t](
+                self._handle, str.encode(key), int(default_value)
+            )
```
pentschev, line 133: Given Python `int` can have arbitrary precision, should we check for its limits and raise an error if it's beyond to avoid over/underflow?

python/rapidsmpf/rapidsmpf/config.pyx
```diff
@@ -0,0 +1,195 @@
...
+    if pattern.groups != 1:
+        raise ValueError(
+            "key_regex must contain exactly one capture group (e.g., 'RAPIDSMPF_(.*)')"
+        )
+    ret = {}
+    for key, value in os.environ.items():
+        match = pattern.fullmatch(key)
+        if match:
+            ret[match.group(1)] = value
+    return ret
```
pentschev, line 195: Why not reuse the C++ implementation?

pentschev (commented):
Left a few comments, but overall looks good.

python/rapidsmpf/rapidsmpf/config.pyx
madsbk: To be on the safe side, I wanted the same semantic as regular Python code:

> This mapping is captured the first time the [os](https://docs.python.org/3/library/os.html#module-os) module is imported, typically during Python startup as part of processing site.py. Changes to the environment made after this time are not reflected in [os.environ](https://docs.python.org/3/library/os.html#os.environ), except for changes made by modifying [os.environ](https://docs.python.org/3/library/os.html#os.environ) directly.

And then it isn't very much code.

cpp/src/config.cpp
```diff
@@ -0,0 +1,83 @@
...
+) {
+    std::unordered_map<std::string, T> ret;
+    ret.reserve(input.size());
+    for (auto&& [key, value] : input) {
+        auto new_key = rapidsmpf::to_lower(rapidsmpf::trim(key));
+        ret.emplace(std::move(new_key), std::move(value));
+    }
+    return ret;
+}
+}  // namespace
```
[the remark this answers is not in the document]
madsbk: Note, values are case-sensitive. It’s up to the component that reads the option to define its meaning and interpret the string representation.

Option keys are not necessarily related to environment variables. In fact, it’s possible to define an option with no string representation at all by assigning its value directly as a `std::any`.

Also, at some point we probably want to support a config file, in which case we likely don’t want all keys to be capitalized?

python/rapidsmpf/rapidsmpf/config.pyx
madsbk: That is already done, added a test: eb63be63d3829546394eacef81c6120891176e57

cpp/include/rapidsmpf/config.hpp
```diff
@@ -0,0 +1,201 @@
...
+            if (options_.find(key) == options_.end()) {
+                if (options_as_strings_.find(key) == options_as_strings_.end()) {
+                    options_[key] = std::make_any<T>(factory(""));
+                } else {
+                    options_[key] = std::make_any<T>(factory(options_as_strings_[key]));
+                }
+            }
+        }
+        try {
+            return std::any_cast<T const&>(options_[key]);
```
nirandaperera, line 80: nit, reduce redundant hash lookup
```suggestion
            auto it = options_.find(key) ;
            if (it == options_.end()) {
                if (options_as_strings_.find(key) == options_as_strings_.end()) {
                    it = options_.emplace(key, std::make_any<T>(factory("")).first;
                } else {
                    it = options_.emplace(key, std::make_any<T>(factory(options_as_strings_[key])).first;
                }
            }
        }
        try {
            return std::any_cast<T const&>(it->second);
```

cpp/include/rapidsmpf/communicator/mpi.hpp
```diff
@@ -103,8 +103,9 @@ class MPI final : public Communicator {
      * @brief Construct an MPI communicator.
      *
      * @param comm The MPI communicator to be used for communication.
+     * @param options Configuration options.
      */
-    MPI(MPI_Comm comm);
+    MPI(MPI_Comm comm, config::Options options);
```
nirandaperera, line 108: @madsbk 
If I'm not mistaken, `Options` is more or less a singleton, and not expected to change during the execution. So, can we use a `shared_ptr` rather than pass by value?

nirandaperera (commented):
Mainly, I feel like its best to use a `shared_ptr` for `Options`.

cpp/src/config.cpp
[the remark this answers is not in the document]
pentschev: Ok, "internal" keys are fine if they are lowercase, but IIUC this is also parsing environment variables as lowercase, i.e., `Rapidsmpf_Log_Level`, `rapidsmpf_log_level` and `RAPIDSMPF_LOG_LEVEL` are all parsed as the same, no? If that's the case, I think we should rewrite to match only the capital `RAPIDSMPF_LOG_LEVEL`.

python/rapidsmpf/rapidsmpf/config.pyx
pentschev: Thanks for adding the test. But where does that happen? I see the test is passing which proves you're right, I don't know/see how this happens.

cpp/include/rapidsmpf/communicator/mpi.hpp
madsbk: Actually, `config::Options` is a shared pointer to `config::OptionsImpl` in order to avoid having to use `shared_ptr` everywhere :)

cpp/include/rapidsmpf/config.hpp
```diff
@@ -0,0 +1,201 @@
...
+    /**
+     * @brief Constructs an `OptionsImpl` instance.
+     *
+     * @param options_as_strings A map of option keys to their string representations.
+     * @param options A map of option keys to their corresponding `Option` objects.
+     */
+    OptionsImpl(
+        std::unordered_map<std::string, std::string> options_as_strings,
+        std::unordered_map<std::string, std::any> options
+    );
```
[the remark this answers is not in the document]
madsbk: Good point. I wanted to avoid another data structure but let me try to see how it would look. It might also be useful if we want to add extra option attributes later.

cpp/src/config.cpp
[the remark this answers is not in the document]
madsbk: Ah no, environment variables are read case sensitive. By default we only read variables starting with`RAPIDSMPF_`:
```c++
get_environment_variables(..., key_regex="RAPIDSMPF_(.*)")
```
And strip the `RAPIDSMPF_` part of the key:
```
* Example:
 *   - Environment variable: RAPIDSMPF_FOO=bar
 *   - key_regex: "RAPIDSMPF_(.*)"
 *   - Resulting map entry: { "FOO", "bar" }
```

cpp/include/rapidsmpf/config.hpp
wence-: That would not be thread-safe, because you've dropped the exclusive-access lock and a different thread calling `emplace` can invalidate your iterator.

cpp/include/rapidsmpf/config.hpp
```diff
@@ -0,0 +1,201 @@
...
+    template <typename T>
+    T const& get(std::string const& key, OptionFactory<T> factory) {
+        {
+            std::lock_guard<std::mutex> lock(mutex_);
+            if (options_.find(key) == options_.end()) {
+                if (options_as_strings_.find(key) == options_as_strings_.end()) {
+                    options_[key] = std::make_any<T>(factory(""));
+                } else {
+                    options_[key] = std::make_any<T>(factory(options_as_strings_[key]));
+                }
```
wence-, line 76: This can avoid multiple lookups in `options_as_strings_` with something like:


```
auto val = options_as_strings_.find(key)
options_[key] = std::make_any<T>(factory(val == options_as_strings_.end() ? "" : *val));
```
if (auto val = options_as_strings_.find(key)

cpp/src/config.cpp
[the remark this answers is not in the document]
pentschev: The prefix is case-sensitive, i.e., `RAPIDSMPF_(.*) != `rapidsmpf_(.*)`, but you still convert everything to lowercase after it has been read, so for the environment variable parser `RAPIDSMPF_LOG_LEVEL == RAPIDSMPF_log_level`, or am I missing something?

python/rapidsmpf/rapidsmpf/config.pyx
madsbk: Cython raises OverflowError, added a test for that as well :)

cpp/src/config.cpp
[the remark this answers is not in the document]
madsbk: Yes, `RAPIDSMPF_LOG_LEVEL` is read in as the `"log_level"` key. So this will only be an issue if we introduce option keys that are case-sensitive.

cpp/include/rapidsmpf/config.hpp
nirandaperera: Ah. Good catch.

cpp/include/rapidsmpf/config.hpp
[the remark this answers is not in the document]
madsbk: Done, we now have a `OptionValue` class.

cpp/src/config.cpp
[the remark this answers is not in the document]
pentschev: Well, there's another edge-case which is the user specifying `RAPIDSMPF_LOG_LEVEL=warn` and `RAPIDSMPF_log_level=info`, which one will be respected? I know this is technically "not supported", but AFAICT nothing prevents the user from doing that.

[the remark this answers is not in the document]
madsbk: Ah OK, added a check: 3c991a51c90b7c085c98eea81c16393894c788f1

madsbk:
Anything else @pentschev ?

cpp/src/config.cpp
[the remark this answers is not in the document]
pentschev: What I had in mind was to only match environment variables that are all caps, e.g., `RAPIDSMPF_LOG_LEVEL` is valid and `RAPIDSMPF_log_level` is discarded, but I think this is ok too.

pentschev (approved):
LGTM, thanks @madsbk .

PR #286, opened 2025-05-19
Install Ray from PyPI to build docs on Python 3.13

pentschev:
RAPIDS' build-docs default has switched to a Python 3.13 environment. However, Ray for Python 3.13 is not available in conda-forge yet, thus we allow installing from PyPI for the time being just to ensure docs can build successfully, even though it may not be usable if installed on from PyPI on a conda environment. See also #272.


pentschev:
Thanks @bdice !

PR #289, opened 2025-05-20
feat: port conda recipes to `rattler-build`

gforsyth:
Ports the conda recipes over to use `rattler-build` instead of `conda-build`


jameslamb (approved):
🎉 yay!

PR #291, opened 2025-05-20
Add multi packed data for the shuffler

### LINKED ISSUE #145: Batch input partitions sent to the same target rank
@wence- @madsbk and myself had a discussion. We discussed that it would be better to batch the input partitions that are sent to the same target rank should be batched to reduce the number of messages we are sending. 

This would require, 
1. Having a mechanism to concat a collection of buffers (metadata and gpu data) and create a concat buffer and a modified metadata struct that has buffer and offsets 
2. Send the modified buffer alongside the metadata control message 
3. Before inserting to the output box, split the concatenated buffer into the original number of buffers

Concatenation and splitting should be pluggable. 

This is further discussed in [this document](https://docs.google.com/document/d/1lFbAAbwFJRvZ5ie3Svg66aUIc5TacJ5oT7M3VYRnHw0/edit?usp=sharing).

nirandaperera:
This PR adds multi-packed data for the shuffler. 

Closes #145 

Depends on #251 #271

## Perf Analysis 

Weak scaling analysis, of 4GB/ rank shuffle (pre-hash partitioned data, in the PDX cluster). The concatenation shows significant performance improvement, amid creating multiple copies of data. 

| out_parts | in_parts | nranks | global throughput GiB/s (new) |        | local throughput GiB/s |       | time (s) |        | time change % | local throughput change % | global throughput change% |
|:---------:|:--------:|:------:|:-----------------------------:|:------:|:----------------------:|:-----:|:--------:|:------:|:-------------:|:-------------------------:|:-------------------------:|
|           |          |        |              new              |   old  |           new          |  old  |    new   |   old  |               |                           |                           |
|    128    |     8    |    2   |             174.20            | 123.78 |          87.10         | 61.89 |  0.0500  | 0.0650 |     23.07     |           40.74           |           40.73           |
|    128    |     8    |    4   |             318.13            | 136.83 |          79.53         | 34.21 |  0.0500  | 0.1175 |     57.44     |           132.49          |           132.49          |
|    128    |     8    |    8   |             440.81            | 271.68 |          55.10         | 33.96 |  0.0737  | 0.1187 |     37.89     |           62.24           |           62.25           |

[notebook link](https://colab.research.google.com/drive/1tgtn-dTw_YB9yfBQNs5RN-EJccAwBIm5?authuser=1#scrollTo=YeM0d1PVQ4UR)

image

image

### DIFF
--- a/cpp/include/rapidsmpf/buffer/resource.hpp
+++ b/cpp/include/rapidsmpf/buffer/resource.hpp
@@ -452,7 +452,7 @@ class LimitAvailableMemory {
  * @param size The size of the buffer to allocate.
  * @param preferred_mem_type The preferred memory type to allocate the buffer from.
  * @return A memory reservation.
- * @throw RAPIDSMPF_FAIL if no memory reservation was made.
+ * @throw std::runtime_error if no memory reservation was made.
  */
 MemoryReservation reserve_or_fail(
     BufferResource* br,


cpp/src/shuffler/chunk.cpp
```diff
@@ -219,6 +264,149 @@ bool Chunk::validate_format(std::vector<uint8_t> const& serialized_buf) {
...
+            std::move(chunks[0].metadata_),
+            std::move(chunks[0].data_)
+        );
+    }
+
+    // Calculate total number of messages and sizes
+    size_t total_messages = 0;
+    size_t total_metadata_size = 0;
+    size_t total_data_size = 0;
+    for (const auto& chunk : chunks) {
```
madsbk, line 292:
```suggestion
    for (auto const& chunk : chunks) {
```

cpp/src/shuffler/shuffler.cpp
```diff
@@ -557,6 +575,78 @@ void Shuffler::insert(std::unordered_map<PartID, PackedData>&& chunks) {
...
+    bool all_groups_built = false;
+
+    // lambda to build all groups and insert the chunks
+    auto build_all_groups_and_insert = [&]() {
+        for (auto&& group : chunk_groups) {
+            if (!group.empty()) {
+                insert(Chunk::concat(std::move(group), get_new_cid(), stream_, br_));
+            }
+        }
+        all_groups_built = true;
```
madsbk, line 611: Remove the side effect of the lambda 
```suggestion
```

cpp/src/shuffler/shuffler.cpp
```diff
@@ -557,6 +575,78 @@ void Shuffler::insert(std::unordered_map<PartID, PackedData>&& chunks) {
...
+    std::vector<std::vector<Chunk>> chunk_groups(size_t(comm_->nranks()));
+    // reserve space for each group assuming an even distribution of chunks
+    for (auto&& group : chunk_groups) {
+        group.reserve(chunks.size() / size_t(comm_->nranks()));
+    }
+
+    // total size of data staged in all builders
+    int64_t total_staged_data_ = 0;
+    auto init_event = std::make_shared<Buffer::Event>(stream_);
+    bool all_groups_built = false;
```
madsbk, line 602:
```suggestion
```

cpp/src/shuffler/shuffler.cpp
```diff
@@ -557,6 +575,78 @@ void Shuffler::insert(std::unordered_map<PartID, PackedData>&& chunks) {
...
+    // lambda to build all groups and insert the chunks
+    auto build_all_groups_and_insert = [&]() {
+        for (auto&& group : chunk_groups) {
+            if (!group.empty()) {
+                insert(Chunk::concat(std::move(group), get_new_cid(), stream_, br_));
+            }
+        }
+        all_groups_built = true;
+    };
+
```
madsbk, line 613:
```suggestion
    bool all_groups_built_flag = false;
```

cpp/src/shuffler/shuffler.cpp
```diff
@@ -557,6 +575,78 @@ void Shuffler::insert(std::unordered_map<PartID, PackedData>&& chunks) {
...
+            insert(create_chunk(pid, std::move(packed_data), init_event));
+            continue;
+        }
+
+        // if the packed data size + total_staged_data_ > headroom, no room to add more
+        // chunks any of the builders. So, call build on all builders.
+        if (!all_groups_built
+            && (int64_t(packed_data.gpu_data->size()) + total_staged_data_ > headroom))
+        {
+            build_all_groups_and_insert();
```
madsbk, line 629:
```suggestion
            build_all_groups_and_insert();
            all_groups_built_flag = true;
```

cpp/src/shuffler/shuffler.cpp
```diff
@@ -557,6 +575,78 @@ void Shuffler::insert(std::unordered_map<PartID, PackedData>&& chunks) {
...
+        }
+    }
+
+    // build any remaining chunks
+    if (!all_groups_built) {
+        build_all_groups_and_insert();
+    }
+}
+
 void Shuffler::insert_finished(PartID pid) {
```
madsbk, line 650: Could this use `Shuffler::insert_finished(std::vector<PartID>&& pids)` ?

madsbk (changes requested):
Overall looks good

cpp/src/shuffler/shuffler.cpp
```diff
@@ -557,6 +575,78 @@ void Shuffler::insert(std::unordered_map<PartID, PackedData>&& chunks) {
...
+                create_chunk(pid, std::move(packed_data), init_event)
+            );
+        }
+    }
+
+    // build any remaining chunks
+    if (!all_groups_built) {
+        build_all_groups_and_insert();
+    }
+}
```
nirandaperera, line 648:
```suggestion
        if (all_groups_built_flag) {
            // insert this chunk without concatenating
            insert(create_chunk(pid, std::move(packed_data), init_event));
        } else {
            // insert this chunk into the builder
            total_staged_data_ += packed_data.gpu_data->ssize();
            chunk_groups[size_t(target_rank)].emplace_back(
                create_chunk(pid, std::move(packed_data), init_event)
            );
        }
    }

    // build any remaining chunks
    if (!all_groups_built_flag) {
        build_all_groups_and_insert();
    }
}
```

cpp/tests/test_shuffler.cpp
```diff
@@ -352,21 +516,204 @@ INSTANTIATE_TEST_SUITE_P(
...
+
+    auto chunks = generate_packed_data();
+    shuffler->insert_grouped(std::move(chunks));
+    shuffler->insert_finished(std::vector<rapidsmpf::shuffler::PartID>(pids));
+
+    ASSERT_NO_FATAL_FAILURE(verify_shuffler_state(*shuffler));
+
+    // resume progress thread - this will guarantee that shuffler progress function is
+    // marked as done. This is important to ensure that the test does not hang.
+    progress_thread->resume();
```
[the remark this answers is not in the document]
madsbk: Good point, `TearDown` saves us here. 
But now I am wondering, what happens for an user run (not testing) if something fails while `ProgressThread` is paused? I guess it will hang?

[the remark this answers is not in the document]
madsbk: Can you submit an issue detailing the problem?

cpp/src/shuffler/chunk.cpp
```diff
@@ -221,6 +278,150 @@ bool Chunk::validate_format(std::vector<uint8_t> const& serialized_buf) {
...
+        concat_metadata = std::make_unique<std::vector<uint8_t>>(total_metadata_size);
+    }
+
+    // Create concatenated data buffer if needed
+    std::unique_ptr<Buffer> concat_data;
+    if (total_data_size > 0) {
+        auto reserve = try_reserve_or_fail(br, total_data_size, preferred_mem_type);
+        concat_data = br->allocate(reserve.mem_type(), total_data_size, stream, reserve);
+    } else {  // no data, allocate an empty host buffer
+        auto [res, ob] = br->reserve(MemoryType::HOST, 0, false);
```
madsbk, line 335:
```suggestion
        auto [res, _] = br->reserve(MemoryType::HOST, 0, false);
```

cpp/src/shuffler/chunk.cpp
```diff
@@ -14,6 +14,37 @@
...
+/**
+ * @brief Try to make a memory reservation.
+ *
+ * @param br The buffer resource.
+ * @param size The size of the buffer to allocate.
+ * @param preferred_mem_type The preferred memory type to allocate the buffer from.
+ * @return A memory reservation.
+ * @throw RAPIDSMPF_FAIL if no memory reservation was made.
+ */
+MemoryReservation try_reserve_or_fail(
```
madsbk, line 27: Maybe move to `resource.hpp`, might be useful other places?

cpp/src/shuffler/chunk.cpp
```diff
@@ -14,6 +14,37 @@
...
 
+
+/**
+ * @brief Try to make a memory reservation.
+ *
+ * @param br The buffer resource.
+ * @param size The size of the buffer to allocate.
+ * @param preferred_mem_type The preferred memory type to allocate the buffer from.
+ * @return A memory reservation.
+ * @throw RAPIDSMPF_FAIL if no memory reservation was made.
```
madsbk, line 25: Update to doc the actually exception thrown

cpp/src/shuffler/chunk.cpp
```diff
@@ -14,6 +14,37 @@
...
+    }
+
+    // try to allocate data buffer from memory types in order [DEVICE, HOST]
+    for (auto mem_type : MEMORY_TYPES) {
+        auto [res, overbooking] = br->reserve(mem_type, size, false);
+        if (res.size() == size) {
+            return std::move(res);
+        }
+    }
+    RAPIDSMPF_FAIL("failed to reserve memory");
```
madsbk, line 45: Throw something other than `std::logic_error`?
```suggestion
    RAPIDSMPF_FAIL("failed to reserve memory", std::runtime_error);
```

madsbk (changes requested):
Looks good, I only some minor stuff

wence- (approved):
Some small nits, but I think this looks OK. It seems more complicated than I thought it would be, but that seems mostly around tracking partition finishes.

cpp/src/shuffler/shuffler.cpp
```diff
@@ -476,38 +481,51 @@ detail::Chunk Shuffler::create_chunk(
...
+
         if (chunk.is_data_buffer_set()) {
             statistics_->add_bytes_stat("shuffle-payload-send", chunk.concat_data_size());
             statistics_->add_bytes_stat("shuffle-payload-recv", chunk.concat_data_size());
         }
         insert_into_ready_postbox(std::move(chunk));
     } else {
+        // this is a remote chunk, so we need to insert it into the outgoing postbox
+        // all messages in the chunk must map to the same key (rank)
+        for (size_t i = 1; i < chunk.n_messages(); ++i) {
```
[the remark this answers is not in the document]
nirandaperera: My idea here was, `Chunk` is not aware of the partition function/ any other restrictions. It is more of a _collection of messages_. So, the caller should apply the restrictions when creating a message collection ([using `Chunk::concat`](https://github.com/rapidsai/rapidsmpf/blob/ba8968fc2229c65012c4a3ffbbe033e8de9faaf3/cpp/src/shuffler/shuffler.cpp#L603)). 
In shuffler::insert, I thought it would be good to verify this in the `Shuffler::insert` (especially, to avoid unintended behavior in the future).

nirandaperera:
Thanks @madsbk and @wence- for the reviews. I have addressed the PR comments now.

PR #294, opened 2025-05-21
Ensure we drop the gil when calling C++ dtors

wence-:
Followup to #208, we need to do this everywhere.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
+++ b/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
@@ -357,7 +357,9 @@ def _extract_partition(
     finally:
         if shuffler.finished():
             ctx = get_worker_context()
-            del ctx.shufflers[shuffle_id]
+            with ctx.lock:
+                if shuffle_id in ctx.shufflers:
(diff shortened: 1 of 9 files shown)


python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -357,7 +357,9 @@ def _extract_partition(
     finally:
         if shuffler.finished():
             ctx = get_worker_context()
-            del ctx.shufflers[shuffle_id]
+            with ctx.lock:
+                if shuffle_id in ctx.shufflers:
+                    del ctx.shufflers[shuffle_id]
```
wence-, line 362: This is not for the gil, but we need this when running with multiple worker threads.

madsbk (approved):
Great catch @wence- !

pentschev (approved):
I was still not able to confirm this resolves the q18 hangs, but since we didn't observe the same hang on another q5@sf1k on DGX-1 as we previously did and we were able to run all queries without any new errors it should be safe to get this in, and it will make further testing easier. Thus approving and merging. Thanks @wence- !

PR #296, opened 2025-05-22
`OptionsValue` compilation error in gcc11

nirandaperera:


madsbk:
Fixed in https://github.com/rapidsai/rapidsmpf/pull/297

PR #297, opened 2025-05-22
Support GCC version 11

madsbk:
Removing `OptionValue`'s ctor that takes `std::any`.

Initially, I used SFINAE to fix GCC version 11's ctor confusion but I think it is better to always require a string representation.


wence- (approved):

pentschev (approved):
LGTM, thanks @madsbk .

PR #303, opened 2025-05-23
Python bindings for `Options.get()`

madsbk:
Removing `Options.get_or_assign()`, which is now replaced by  the more generic `Options.get()`.


python/rapidsmpf/rapidsmpf/config.pyx
```diff
@@ -87,73 +38,130 @@ cdef class Options:
...
+        If the option is not present, it is constructed using the provided
+        factory function, which receives the string representation of the
+        option (or an empty string if unset). The option is cached after
+        the first access.
 
-        The type of the returned value is determined by the `parser_type` argument,
-        which must be one of the supported Python types: bool, int, float, or str.
-        The `default_value` is cast to the corresponding C++ type before being passed
-        to the underlying options system.
+        The option is cast to the specified `return_type`, which must be one
```
TomAugspurger, line 50: This should probably be 

```suggestion
        The option is cast to the specified ``return_type``, which must be one
```

I'm guessing this isn't included in our docs yet. Probably just need to add it to the index at https://github.com/rapidsai/rapidsmpf/blob/branch-25.06/docs/source/index.md?plain=1. LMK if you want to address that here & need help with it.

Likewise on L74 and L126 below.

TomAugspurger (approved):
Looks reasonable. I didn't follow the cython / template stuff completely, but that seems to have been at least partly moved from `config.pyx` so presumably it's OK :)

PR #304, opened 2025-05-23
Add CUDA 12.8 to dependencies.yaml

trxcllnt:
A few weeks back we upgraded most things in RAPIDS to CUDA 12.9, then reverted back to 12.8 due to a compiler bug.

This PR ensures `rapids-dependency-file-generator` will produce the correct conda dependencies for CTK 12.8.


trxcllnt:
Nevermind, this was fixed by updating to a newer rapids-dependency-file-generator version in https://github.com/rapidsai/devcontainers/pull/519

PR #312, opened 2025-05-28
RMM Resource Adaptor

madsbk:
Introduce a new RMM memory resource to replace `statistics_resource_adaptor`, `StatisticsResourceAdaptor`, and `RmmFallbackResource`.

The idea is to use have one unified memory resource to use in RapidsMPF that support all we need for spilling, memory profiling, OOM protection, etc.

Will be used in https://github.com/rapidsai/rapidsmpf/pull/305 to support a more accurate profiling.


bdice (commented):
Can you compare this to RMM's `limiting_resource_adapter`? https://github.com/rapidsai/rmm/blob/e7356abe85949249d79a1c83fb9352c10f5d4c0a/cpp/include/rmm/mr/device/limiting_resource_adaptor.hpp#L49

madsbk:
@bdice, currently we need fallback and a custom statistics to memory profile both the memory of a code block (top of the counter stack) and the total memory used while running a code block (the bottom of the counter stack). 

We introduce an unified resource adapter mainly to avoid having to parse around a stack of memory resources. Since we need a fallback and a custom statistics mr anyways, unifying them seems useful. It also makes it easier for the user, which can use this mr instead of providing a stack of mrs and pointers to each mr in the stack.

nirandaperera (commented):
Had a comment on RAII objects. Other than that LGTM

cpp/src/rmm_resource_adaptor.cpp
```diff
@@ -0,0 +1,125 @@
...
+    } catch (rmm::out_of_memory const& e) {
+        if (fallback_mr_.has_value()) {
+            ret = fallback_mr_->allocate_async(nbytes, stream);
+            std::lock_guard<std::mutex> lock(mutex_);
+            fallback_allocations_.insert(ret);
+            record_.record_allocation(ScopedMemoryRecord::AllocType::FALLBACK, nbytes);
+        } else {
+            throw;
+        }
+    }
```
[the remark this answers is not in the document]
madsbk: What object are you thinking we should wrap?

python/rapidsmpf/rapidsmpf/rmm_resource_adaptor.pyx
```diff
@@ -0,0 +1,206 @@
...
+                                           UpstreamResourceAdaptor)
+
+# Pointer alias used by `dynamic_cast`, which doesn't accept pointer types.
+ctypedef cpp_RmmResourceAdaptor* cpp_RmmResourceAdaptor_ptr
+
+cdef class ScopedMemoryRecord:
+    cdef cpp_ScopedMemoryRecord _handle
+
+    def __cinit__(self):
+        self._handle = cpp_ScopedMemoryRecord()
```
pentschev, line 17: For safety, I think we should also add `__dealloc__` that just creates a new object from the default constructor, like [here](https://github.com/rapidsai/rapidsmpf/blob/b5b669f20762d9368b3a1310af1d5969a91a1e38/python/rapidsmpf/rapidsmpf/config.pyx#L37-L39).

python/rapidsmpf/rapidsmpf/rmm_resource_adaptor.pyx
```diff
@@ -0,0 +1,206 @@
...
+
+        Parameters
+        ----------
+        alloc_type
+            Allocator type to query. Defaults to ALL.
+
+        Returns
+        -------
+        int
+            Number of total allocations.
```
pentschev, line 31:
```suggestion
        Number of total allocations.
```

pentschev: For consistency with the rest.

python/rapidsmpf/rapidsmpf/rmm_resource_adaptor.pyx
```diff
@@ -0,0 +1,206 @@
...
+
+        Parameters
+        ----------
+        alloc_type
+            The allocator that performed the deallocation.
+
+        nbytes
+            The number of bytes deallocated.
+        """
+        self._handle.record_deallocation(alloc_type, nbytes)
```
pentschev, line 126: Does it make sense to expose this? Are we expecting to call those directly from Python when `RmmResourceAdaptor` is the primary (and perhaps should be the only) caller of those? Note that the mutex is also not part of this class and is handled only by `RmmResourceAdaptor`, so calling them is most certainly not safe in a multi-threaded environment.

cpp/include/rapidsmpf/rmm_resource_adaptor.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+    /**
+     * @brief Records a memory deallocation event.
+     *
+     * Reduces the current memory usage for the specified allocator.
+     *
+     * @param alloc_type The allocator that performed the deallocation.
+     * @param nbytes     The number of bytes deallocated.
+     */
+    void record_deallocation(AllocType alloc_type, std::uint64_t nbytes);
```
pentschev, line 118: Maybe make those private and have `RmmResourceAdaptor` as `friend`? There's no mutex protecting these calls here, so arbitrarily calling them outside of `RmmResourceAdaptor` is most certainly unsafe in a multi-threaded context.

cpp/src/rmm_resource_adaptor.cpp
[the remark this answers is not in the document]
nirandaperera: something like this.
```c++
try {
        Closer closer{[&]{
            std::lock_guard lock(mutex_);
            record_.record_allocation(ScopedMemoryRecord::AllocType::PRIMARY, nbytes);
        }};
        ret = primary_mr_.allocate_async(nbytes, stream);
    } catch (rmm::out_of_memory const& e) {
        if (fallback_mr_.has_value()) {
            Closer closer{[&]{
                   fallback_allocations_.insert(ret);
                   record_.record_allocation(ScopedMemoryRecord::AllocType::FALLBACK, nbytes);
            }};
            ret = fallback_mr_->allocate_async(nbytes, stream);
        } else {
            throw;
        }
    }
```

pentschev (commented):
Thanks @madsbk , left some comments and suggestions.

cpp/include/rapidsmpf/rmm_resource_adaptor.hpp
nirandaperera: +1 on this.

madsbk: I don't know, `ScopedMemoryRecord` is just a container of memory counters. I think it is fine if an user want to use `ScopedMemoryRecord` to track some memory allocations manually outside of `RmmResourceAdaptor`?

cpp/tests/test_buffer_resource.cpp
```diff
@@ -592,16 +590,12 @@ TEST(BufferResource, CopySliceDifferentResources) {
 
     // Setup br1 with statistics for its device memory
     rmm::mr::cuda_memory_resource mr_cuda1;
-    rmm::mr::statistics_resource_adaptor<rmm::mr::device_memory_resource> stats_mr1{
-        mr_cuda1
-    };
+    RmmResourceAdaptor stats_mr1{mr_cuda1};
```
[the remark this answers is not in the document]
madsbk: done

cpp/include/rapidsmpf/rmm_resource_adaptor.hpp
pentschev: > I don't know, `ScopedMemoryRecord` is just a container of memory counters. I think it is fine if an user want to use `ScopedMemoryRecord` to track some memory allocations manually outside of `RmmResourceAdaptor`?

Then what does that represent anyway? Likely it's being used for an external RMM resource, the memory counters will then just be garbage, both because the report is not going to be representative due to mixing internal and external numbers and because those methods don't ensure thread-safety given there's no internal mutex.

python/rapidsmpf/rapidsmpf/rmm_resource_adaptor.pyx
madsbk: I guess, we would also need to do:
```python
    def __cinit__(self):
        with nogil:
            self._handle = cpp_ScopedMemoryRecord()
```

But `ScopedMemoryRecord` is just a collection of counters

cpp/include/rapidsmpf/rmm_resource_adaptor.hpp
madsbk: I was thinking it could be used to track memory statistics from multiple ranks if we implement operator+. But generally, I don't think we need to impose restrictions just for the sake of it. Since it's just a POD, there's no need to complicate things. It's also nice that we can test it independently.

I have add some docs: a39cac75e0ee1429bb535a66f3707b8d29ed1e50

But if you insist, we can make them private :)

nirandaperera: I feel like `ScopeMemoryRecord` should be unique to the `RmmResourceAdaptor` and it would be best to make it not copy-able. For an example, 
```c++
RmmResourceAdapter mr ...
auto rec1 = mr.get_record();
mr.allocate(...);
auto rec2 = mr.get_record(); 
```
Now, `rec1 != rec2`, but I feel like this could be problematic. WHat if we only return a ref to the memory record and make it non-copyable?

And also, as @pentschev mentioned, since `record_(de)allocation` methods are public, users can arbitrarily call those methods, but it doesn't really do any (de)allocs on the resource.

madsbk: The point is that my premise assumes `ScopeMemoryRecord` should be a POD (i.e., trivially copyable), so it can be moved between nodes easily and without requiring any locking. However, if we want to support something like @nirandaperera's example, then `ScopeMemoryRecord` would need to include an internal lock. This would also require the Cython bindings to use `with nogil`, and make copying/serialization non-trivial.

madsbk:
Thanks @pentschev. 
What about you @nirandaperera, can you accept my POD insistent? ;)

PR #314, opened 2025-05-28
Use the new config system in the dask integration

madsbk:


python/rapidsmpf/rapidsmpf/integrations/dask/core.py
```diff
@@ -344,6 +364,10 @@ def bootstrap_dask_cluster(
         memory pool raises OOM errors.
     enable_statistics
         Whether to track shuffler statistics.
+    options
+        Configuration options. This argument takes precedence: if not None, the
+        other arguments (beside `client`) are ignored!
+        TODO: remove the other arguments beside `client`.
```
rjzamora, line 370: Seem like this is just to provide temporary backward compatibility, correct?

rjzamora: If we can get the cudf component ready while this PR is still in review, it probably makes sense to just break the  `bootstrap_dask_cluster` API.

python/rapidsmpf/rapidsmpf/integrations/dask/core.py
```diff
@@ -233,8 +222,10 @@ def rmpf_worker_setup(
     """
     ctx = get_worker_context(dask_worker)
     with ctx.lock:
+        ctx.options = options
+
         # Print statistics at worker shutdown.
-        if enable_statistics:
+        if ctx.options.get_or_default("dask_statistics", default_value=False):
```
rjzamora, line 228: Just to check my understanding: It looks like we need to specify the default value whenever we try to extract a config value, and so the default for the same option can diverge in different parts of the code. There is no central place where user-defined options are validated - Is that correct?

rjzamora (approved):
Thanks @madsbk

I think the use of `Options` is a good change. I **am** concerned that the lack of centralized defaults/validation could lead to issues, but I also thing that can be addressed later.

python/rapidsmpf/rapidsmpf/integrations/dask/core.py
madsbk: Yes, the philosophy is that the component using the option knows how to parse it.

PR #320, opened 2025-05-29
Remove ucxx Python CMake workarounds

KyleFromNVIDIA:
We previously had to manually set `ucxx_DIR` and the include directories due to issues with the ucxx build. Those issues have been resolved as of https://github.com/rapidsai/ucxx/pull/429, so remove these workarounds.

Follow-up to https://github.com/rapidsai/rapidsmpf/pull/269.


gforsyth (approved):
Nice, thanks for tracking these down and fixing them, @KyleFromNVIDIA !

bdice (approved):

PR #324, opened 2025-05-29
[MINOR] Adding shuffle benchmark with no partitioning

nirandaperera:
This PR adds a shuffle benchmark without any hash partitioning operation. It simulates hash-partitioning by duplicating a chunk for each output partition.


cpp/benchmarks/bench_shuffle_no_partition.cpp
```diff
@@ -0,0 +1,387 @@
+/**
```
madsbk, line 1: Would it make sense to do the change in `benchmark_shuffle.cpp` as an option instead?

nirandaperera: Sure. Maybe, I'll add it as a cmd line option.

cpp/benchmarks/bench_shuffle.cpp
```diff
@@ -271,6 +266,127 @@ rapidsmpf::Duration run(
...
 }
 
+rapidsmpf::Duration run_hash_partitioning(
+    std::shared_ptr<rapidsmpf::Communicator>& comm,
+    std::shared_ptr<rapidsmpf::ProgressThread>& progress_thread,
+    ArgumentParser const& args,
+    rmm::cuda_stream_view stream,
+    rapidsmpf::BufferResource* br,
+    std::shared_ptr<rapidsmpf::Statistics>& statistics
+) {
```
madsbk, line 276: It would be good with a docstring here describing the partition run

cpp/benchmarks/bench_shuffle.cpp
```diff
@@ -271,6 +266,127 @@ rapidsmpf::Duration run(
...
+}
+
+rapidsmpf::Duration run_duplicate_input_tables(
+    std::shared_ptr<rapidsmpf::Communicator>& comm,
+    std::shared_ptr<rapidsmpf::ProgressThread>& progress_thread,
+    ArgumentParser const& args,
+    rmm::cuda_stream_view stream,
+    rapidsmpf::BufferResource* br,
+    std::shared_ptr<rapidsmpf::Statistics>& statistics
+) {
```
madsbk, line 331: And a docstring here describing the duplicate run

madsbk (approved):

PR #319, opened 2025-05-29
add wheel publishing

jameslamb:
Followup to #269

Adds wheel publishing to https://pypi.anaconda.org/rapidsai-wheels-nightly/simple/ on merges to `branch-25.06` + nightly runs.

PR #325, opened 2025-05-29
add wheel publishing

jameslamb:
Adds wheel publishing.

This was added on `branch-25.06` in #319, but those changes were lost in a forward-merge.

Note that I'm intentionally NOT adding the `rapids-upload-to-s3` changes from that PR... those are no longer necessary in 25.08, for the reasons described in https://github.com/rapidsai/build-planning/issues/181


gforsyth (approved):
:shipit:

PR #341, opened 2025-06-10
Add missing `__dealloc__` to `SpillManager`

pentschev:
Follow-up to #208 and #294, ensuring GIL is dropped when calling `SpillManager`'s C++ destructor from Python.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/buffer/spill_manager.pyx
+++ b/python/rapidsmpf/rapidsmpf/buffer/spill_manager.pyx
@@ -76,6 +76,17 @@ cdef class SpillManager:
     def __init__(self):
         raise TypeError("Please get a `SpillManager` from a buffer resource instance")
 
+    def __dealloc__(self):
+        """
+        Deallocate resource without holding the GIL.
+
+        This is important to ensure owned resources, like the underlying C++
+        `SpillManager` object is destroyed, ensuring any threads can be
+        joined without risk of deadlocks if both thread compete for the GIL.
+        """
+        with nogil:
+            self._handle.reset()
+
     @classmethod
     def _create(cls, BufferResource br):
         """Construct a SpillManager associated the specified buffer resource.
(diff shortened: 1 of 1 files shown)


madsbk (approved):
Thanks @pentschev

pentschev:
I’m actually not quite sure about this anymore [as it failed to build](https://github.com/rapidsai/rapidsmpf/actions/runs/15560297985/job/43811088933?pr=341#step:10:999), [`self._handle`](https://github.com/rapidsai/rapidsmpf/blob/9227919f8056cfdfa00ae1750407359139302ede/python/rapidsmpf/rapidsmpf/buffer/spill_manager.pyx#L96) is not a `shared_ptr` like in other cases, nor an object owned by the `SpillManager` Cython class, but rather a [raw pointer that is returned as reference](https://github.com/rapidsai/rapidsmpf/blob/9227919f8056cfdfa00ae1750407359139302ede/cpp/src/buffer/resource.cpp#L165-L167) from the [C++ `SpillManager` object owned by the C++ `BufferResource`](https://github.com/rapidsai/rapidsmpf/blob/9227919f8056cfdfa00ae1750407359139302ede/cpp/include/rapidsmpf/buffer/resource.hpp#L389).

While the problem is calling `SpillManager::~SpillManager`, and the frame that calls it comes from `resource.cpython-312-aarch64-linux-gnu.so`, I’m not sure who’s still owning the object that needs to be destroyed without the GIL. Maybe it is because of the reference and [my latest change](https://github.com/rapidsai/rapidsmpf/pull/341/commits/ecea0980c440eeb9ac69d6796c07be999c0219c2) may resolve the issue, I’m not entirely confident of that.

pentschev:
I'm not entirely confident this resolves the problem, but given we just had another instance in https://github.com/rapidsai/rapidsmpf/actions/runs/15590717940/attempts/1?pr=336 and at least I don't expect it to be harmful, I'll just merge this so we can see if it improves the situation.

Issue #337, opened 2025-06-06
Add / document extra dependencies


TomAugspurger:
https://github.com/rapidsai/cudf/pull/19101 is adding `rapidsmpf` to test the cudf-polars dask integration. Just adding `rapidsmpf` failed with:

```
  File "/__w/cudf/cudf/python/cudf_polars/tests/experimental/test_rapidsmpf.py", line 35, in test_join_rapidsmpf
    from rapidsmpf.integrations.dask import bootstrap_dask_cluster
  File "/pyenv/versions/3.10.18/lib/python3.10/site-packages/rapidsmpf/integrations/dask/__init__.py", line 7, in <module>
    from rapidsmpf.integrations.dask.core import (
  File "/pyenv/versions/3.10.18/lib/python3.10/site-packages/rapidsmpf/integrations/dask/core.py", line 13, in <module>
    import ucxx._lib.libucxx as ucx_api
ModuleNotFoundError: No module named 'ucxx'
```

We should probably add an extra for each integration that contains all the dependencies needed for that integration to function.

TomAugspurger:
> What is not a hard-dependency is `distributed-ucxx` which is something that would be required for the Dask integration alone

My understanding is that distributed-ucxx is *not* a hard requirement. I don't believe that [boostrap_dask_cluster](https://github.com/rapidsai/rapidsmpf/blob/9227919f8056cfdfa00ae1750407359139302ede/python/rapidsmpf/rapidsmpf/integrations/dask/core.py#L337) puts any special requirements on the `Comms` used for the cluster. So we're able to use `protocol="dask"` which AFAIU doesn't use distributed-ucxx (i.e. `protocol="ucxx"`).

So I believe that our imports in `rapidsmpf.integrations.dask` should reflect our requirements. I believe that `ucxx` is the only one not covered (aside from `distributed`, which is implicitly covered by `dask_cuda`). I'll push a PR shortly.

TomAugspurger:
https://github.com/rapidsai/rapidsmpf/pull/339 handles the PyPI side of things. It doesn't *not* handle the conda side. I'm not sure what the best practices are there, but is it possible to have something like a `rapidsmpf-dask` or `rapidsmpf-ray`? Or do we need `rapidsmpf` to mean "all dependencies" with a `rapidsmpf-core` that has just the required?

cc @btepera.

pentschev:
[...]

PR #342, opened 2025-06-10
fix(deps): ensure runtime deps come along with `librapidsmpf`

gforsyth:
xref #337


pentschev (approved):
This only applies for `librapidsmpf` wheels, is that right @gforsyth ? That looks fine to me.

KyleFromNVIDIA (approved):

PR #348, opened 2025-06-16
[MINOR] Host memory resource benchmarks

nirandaperera:
This PR adds host memory benchmarks, that measures the following for pinned vs new_delete rmm host memory resources, 
- allocations
- deallocations
- DeviceToHost copy
- HostToDeviceCopy

## Bench Results from PDX cluster 

[notebook](https://colab.research.google.com/drive/1T_Z3Kz6B7sPfwe8IfVdHzqllZrhJQhKx?usp=sharing)

| MB/s         |                  |           |                  |           |
|--------------|------------------|-----------|------------------|-----------|
| direction    | DeviceToHostCopy |           | HostToDeviceCopy |           |
| mem resource | new_delete       | pinned    | new_delete       | pinned    |
|         size |                  |           |                  |           |
|         1024 |           122.92 |    342.46 |           452.97 |    320.00 |
|       512000 |        10,607.24 | 39,848.52 |        20,731.64 | 39,536.86 |
|      1048576 |        12,284.28 | 44,202.90 |        22,535.32 | 44,606.30 |
|    524288000 |        14,865.78 | 51,572.13 |        11,344.72 | 52,449.99 |
|   1073741824 |        12,047.48 | 51,300.43 |        11,100.86 | 52,449.32 |

image
image

It seems using the pinned host memory, at least as a staging buffers for spilling is worthwhile.

### DIFF
--- a/cpp/benchmarks/bench_memory_resources.cpp
+++ b/cpp/benchmarks/bench_memory_resources.cpp
@@ -90,7 +90,7 @@ static void BM_DeviceToHostCopy(benchmark::State& state) {
     benchmark::DoNotOptimize(device_buffer);
     benchmark::DoNotOptimize(host_ptr);
 
-    cudaStreamSynchronize(stream);
+    RAPIDSMPF_CUDA_TRY(cudaStreamSynchronize(stream));
     for (auto _ : state) {
         for (size_t i = 0; i < kNumCopies; ++i) {
             cudaMemcpyAsync(
@@ -101,7 +101,7 @@ static void BM_DeviceToHostCopy(benchmark::State& state) {
                 stream
             );
         }
-        cudaStreamSynchronize(stream);
+        RAPIDSMPF_CUDA_TRY(cudaStreamSynchronize(stream));
     }
     // Cleanup
     host_mr->deallocate(host_ptr, transfer_size);
@@ -132,7 +132,7 @@ static void BM_HostToDeviceCopy(benchmark::State& state) {
     benchmark::DoNotOptimize(device_buffer);
     benchmark::DoNotOptimize(host_ptr);
 
-    cudaStreamSynchronize(stream);
+    RAPIDSMPF_CUDA_TRY(cudaStreamSynchronize(stream));
 
     for (auto _ : state) {
         for (size_t i = 0; i < kNumCopies; ++i) {
@@ -144,7 +144,7 @@ static void BM_HostToDeviceCopy(benchmark::State& state) {
                 stream
             );
         }
-        cudaStreamSynchronize(stream);
+        RAPIDSMPF_CUDA_TRY(cudaStreamSynchronize(stream));
     }
     // Cleanup
     host_mr->deallocate(host_ptr, transfer_size);


cpp/benchmarks/CMakeLists.txt
```diff
@@ -86,3 +86,28 @@ install(
...
+
+add_executable(bench_memory_resources "bench_memory_resources.cpp" ${UTILS_SOURCES})
+set_target_properties(
+  bench_memory_resources
+  PROPERTIES RUNTIME_OUTPUT_DIRECTORY "$<BUILD_INTERFACE:${RAPIDSMPF_BINARY_DIR}/benchmarks>"
+             CXX_STANDARD 17
+             CXX_STANDARD_REQUIRED ON
+             CXX_EXTENSIONS ON
+             CUDA_STANDARD 17
+             CUDA_STANDARD_REQUIRED ON
```
madsbk, line 98:
```suggestion
             CXX_STANDARD 20
             CXX_STANDARD_REQUIRED ON
             CXX_EXTENSIONS ON
             CUDA_STANDARD 20
             CUDA_STANDARD_REQUIRED ON
```

cpp/benchmarks/bench_memory_resources.cpp
```diff
@@ -0,0 +1,181 @@
...
+    const std::string resource_type = state.range(1) == 0 ? "new_delete" : "pinned";
+
+    auto host_mr = create_host_memory_resource(resource_type);
+    auto device_mr = std::make_unique<rmm::mr::cuda_memory_resource>();
+    rmm::cuda_stream_view stream = rmm::cuda_stream_default;
+
+    // Allocate device memory
+    auto device_buffer = rmm::device_buffer(transfer_size, stream, device_mr.get());
+    // Initialize device memory with some data
+    cudaMemset(device_buffer.data(), 0, transfer_size);
```
madsbk, line 82: use `RAPIDSMPF_CUDA_TRY`

cpp/benchmarks/bench_memory_resources.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    // Allocate host memory and copy from device
+    void* host_ptr = host_mr->allocate(transfer_size);
```
madsbk, line 85: Is the allocations/deallocations included in the timing?

nirandaperera: @madsbk I was thinking no. But I should have paused the timing here. Let me do that.

pentschev (commented):
Do we really want this benchmark in `rapidsmpf`? While it's a useful demonstration of pinned memory performance, in itself it has little or nothing to do with `rapidsmpf` and this may confuse users. I think this should be moved to a separate "playground" repository of its own or find another project that has a more direct connection with this benchmark (RMM is what first comes to mind).

nirandaperera:
> Do we really want this benchmark in `rapidsmpf`? While it's a useful demonstration of pinned memory performance, in itself it has little or nothing to do with `rapidsmpf` and this may confuse users. I think this should be moved to a separate "playground" repository of its own or find another project that has a more direct connection with this benchmark (RMM is what first comes to mind).

@pentschev good point. IMO lets move this to a playground repo under the rapids org. I am not sure if there is any at the moment. This is an evaluation of whether we should use pinned memory or not. I could also extend this to do some send/ resc operations using pinned buffers. So, let's keep this inside rapidsmpf for the moment. 

Maybe @quasiben can weigh in here. Do we have a playground repo where we could push stuff like these?

madsbk:
IMHO, I think is OK to have it here. But a dedicated "playground" repository could also be useful :)

PR #350, opened 2025-06-23
Deadlock detection

madsbk:
When `RAPIDSMPF_DEBUG` is defined, all lock guards have a timeout of 60 seconds.


pentschev (commented):
In general this looks fine, I'm still wondering how you plan to enable this in CI. I believe we don't want to run always in debug mode, and duplicating runs may be a little challenging given the separation of build and test jobs. One thing we can do is to enable it as a separate nightly workflow, but the amount of runs is greatly reduced versus runs in PRs.

PR #353, opened 2025-06-25
Allow building with AddressSanitizer

pentschev:
Make changes to CMake and `build.sh` to allow building RapidsMPF with AddressSanitizer both in C++ and Cython shared libraries.


python/rapidsmpf/CMakeLists.txt
```diff
@@ -19,6 +19,14 @@ project(
   LANGUAGES CXX CUDA
 )
 
+# Add global Address Sanitizer flags only if explicitly requested
+option(RAPIDSMPF_PYTHON_ASAN "Enable Address Sanitizer for Python build" OFF)
+if(RAPIDSMPF_PYTHON_ASAN)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=address")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -fsanitize=address")
+endif()
```
jakirkham, line 28: Think we would add these flags to the target we are building

Also this likely needs to moved to the end of the file for the other RAPIDS CMake logic to take effect

```suggestion
if(RAPIDSMPF_PYTHON_ASAN)
  set(RAPIDSMPF_ASAN_FLAGS -fsanitize=address)
  target_compile_options(
    TARGET PRIVATE "$<$<COMPILE_LANGUAGE:CXX>:${RAPIDSMPF_ASAN_FLAGS}>"
                             "$<$<COMPILE_LANGUAGE:CUDA>:${UDF_CUDA_FLAGS}>"
  )
  target_link_options(TARGET PRIVATE "LINKER:${RAPIDSMPF_ASAN_FLAGS}")
endif()
```

jakirkham (commented):
Thanks Peter! 🙏

Think we need a few changes on the Cython side. Have noted what those might look like below

Possibly this will need more tweaks. Though hopefully this gets us moving in the right direction

python/rapidsmpf/CMakeLists.txt
pentschev: Thanks @jakirkham , I've tried that locally but the result is exactly the same:

```
building rapidsmpf...
Processing /datasets/pentschev/src/rapidsmpf/python/rapidsmpf
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: rapidsmpf
  Building wheel for rapidsmpf (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Building wheel for rapidsmpf (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [3 lines of output]
      *** scikit-build-core 0.11.4 using CMake 3.31.6 (wheel)
      *** Configuring CMake...
      ==4051253==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for rapidsmpf
Failed to build rapidsmpf
ERROR: Failed to build installable wheels for some pyproject.toml based projects (rapidsmpf)
```

I attempted both to keep the position and to move it to the end of the file with no changes.

jakirkham: Do we need it in `CFLAGS` as well? If so, it would need to a change to account for that

python/rapidsmpf/CMakeLists.txt
```diff
@@ -19,6 +19,16 @@ project(
   LANGUAGES CXX CUDA
 )
 
+# Add global AddressSanitizer flags only if explicitly requested
+option(RAPIDSMPF_PYTHON_ASAN "Enable AddressSanitizer for Python build" OFF)
+add_library(maybe_asan INTERFACE)
+target_compile_options(
```
robertmaynard, line 25: You can simplify this by using `$<COMPILE_LANGUAGE:C,CXX>`  instead

pentschev: According to [this](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:COMPILE_LANGUAGE:languages), my understanding was correct.

PR #355, opened 2025-06-25
The C++ package is called rapidsmpf

wence-:


pentschev (approved):
LGTM, thanks Lawrence.

madsbk (approved):

wence-:
> Really, the CMake project is called `RAPIDSMPF`. I wish we could make the casing consistent. CMake does some normalization / lookup priority of various casings, so this should work -- but it's not ideal to have mismatched casing.
> 
> https://github.com/rapidsai/rapidsmpf/blob/branch-25.08/cpp/CMakeLists.txt#L19

I mean we can downcase the C++ project.

wence-:
> That could be a good idea. We've been terribly inconsistent across RAPIDS with casing of CMake project names, iirc.

Did that.

wence-:
> > That could be a good idea. We've been terribly inconsistent across RAPIDS with casing of CMake project names, iirc.
> 
> Did that.

Didn't do that, everything broke.

I instead used uppercase in the librapidsmpf cmakelists.

PR #358, opened 2025-06-26
PostBox with multiple mutexes 

nirandaperera:
This PR evaluates a new PostBox impl with multiple mutexes. The reason for this is, I am trying to concat chunks inside the postbox from a separate thread. Even now, there are several threads accessing the postbox (caller thread, postbox spilling, shuffle progress thread, etc).

The new impl splits the pigenholes map into `N` buckets, and a key is designated a bucket based on `key%N`. Each bucket has its own mutex. 

Benchmark results: 

| MB/s    |         | nBuckets (Throughput MB/s)    |       |       |       |
|---------|---------|-------------|-------|-------|-------|
| KeyType | chunks  | 0 (current) |     2 |     4 |     8 |
| PartID  |   10000 |       46.27 | 49.03 | 50.86 | 48.61 |
| PartID  |  100000 |       47.44 | 45.04 | 49.59 | 50.23 |
| PartID  | 1000000 |       46.15 | 47.71 | 48.02 | 48.88 |
| Rank    |   10000 |       48.08 | 51.77 | 47.76 | 53.10 |
| Rank    |  100000 |       47.16 | 49.63 | 49.35 | 49.70 |
| Rank    | 1000000 |       46.19 | 49.24 | 47.95 | 48.62 |

PostBox_PartID_

PostBox_Rank_

### DIFF
--- a/cpp/benchmarks/CMakeLists.txt
+++ b/cpp/benchmarks/CMakeLists.txt
@@ -111,3 +111,29 @@ install(
   DESTINATION bin/benchmarks/librapidsmpf
   EXCLUDE_FROM_ALL
 )
+
+add_executable(bench_postbox "postbox/bench_postbox.cpp" "postbox/postbox2.hpp" ${UTILS_SOURCES})
+set_target_properties(
+  bench_postbox
+  PROPERTIES RUNTIME_OUTPUT_DIRECTORY "$<BUILD_INTERFACE:${RAPIDSMPF_BINARY_DIR}/benchmarks>"
+             CXX_STANDARD 20
+             CXX_STANDARD_REQUIRED ON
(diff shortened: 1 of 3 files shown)


nirandaperera:
@madsbk I saw there was a comment in the `PostBox` class to use multiple mutexes. WDYT about this? TBH I was expecting a bit more performance improvement. It was a bit underwhelming to me.

wence-:
How many chunks do we think is a realistic max?

nirandaperera:
> How many chunks do we think is a realistic max?

@wence- IINM chunk ID can go upto 38bits. 
https://github.com/rapidsai/rapidsmpf/blob/015f0b864451a85bff634a48943522595ed9d0d5/cpp/src/shuffler/shuffler.cpp#L843

madsbk:

PR #361, opened 2025-06-27
Adding python bindings for insert_grouped

nirandaperera:
This adds python bindings for ~`Shuffler::insert_grouped`~ `Shuffler::concat_insert` methods


python/rapidsmpf/rapidsmpf/shuffler.pyx
```diff
@@ -122,7 +124,7 @@ cdef class Shuffler:
         """
         return self._comm
 
-    def insert_chunks(self, chunks):
+    def insert_chunks(self, chunks, bool grouped = False):
```
TomAugspurger, line 127: Do we have a good reason for having a single Python method (`insert_chunks`) with a boolean flag, rather than having two methods to match the C++ API (`insert`, `insert_grouped`)? If there aren't strong reasons I'd probably vote for consistency with the C++ API (although I'm noticing now we're already inconsistent: `insert_chunks` vs `insert`.)

python/rapidsmpf/rapidsmpf/shuffler.pyx
```diff
@@ -139,33 +144,47 @@ cdef class Shuffler:
...
         Notes
         -----
         Once a partition is marked as finished, it is considered complete and no
         further chunks will be accepted for that partition.
         """
+        cdef vector[uint32_t] _pids
+
+        if isinstance(pids, Iterable):
+            # Reserve space and populate vector
+            _pids.reserve(len(pids))
```
TomAugspurger, line 180: Iterables don't necessarily have a length 

```python
In [3]: xs = (x for x in range(3))

In [4]: isinstance(xs, collections.abc.Iterable)
Out[4]: True

In [5]: len(xs)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[5], line 1
----> 1 len(xs)

TypeError: object of type 'generator' has no len()
```

Maybe convert it to a Python list first, and then https://docs.cython.org/en/latest/src/userguide/wrapping_CPlusPlus.html#standard-library suggests that this should work:

```
cdef vector[uint32_t] _pids = pids_list
```

python/rapidsmpf/rapidsmpf/shuffler.pyx
nirandaperera: @TomAugspurger IMO, the two methods have different implementations in C++ side. It was more cleaner that way rather than branching out within a single method. But in python side, if we had two methods, both would have identical impl, because `insert` and `insert_grouped` have identical signatures. 
That's why I chose a flag in the python side.

python/rapidsmpf/rapidsmpf/shuffler.pyx
nirandaperera: @TomAugspurger good catch. I'll remove that `reserve`. I think that should be good enough. WDYT?

TomAugspurger: I'd do this if it works:

```python
cdef vector[uint32_t] _pids

if isinstance(pids, Iterable):
    pids_list = list(pids)
    _pids = pids
```

Otherwise, skipping the reserve is probably fine. I wouldn't expect this to take much time regardless.

nirandaperera: It seems `cdef vector[int] vect = range(1, 10, 2)`  is also [allowed](https://docs.cython.org/en/latest/src/userguide/wrapping_CPlusPlus.html#standard-library), 

So, then technically we should be able to do, `_pids = pids`, isnt it?

TomAugspurger: > So, then technically we should be able to do, _pids = pids, isnt it?

I'm not sure, but maybe! If it compiles then we're probably good to go (`range(...)` does have a length, so it's a bit different than the generator comprehension I showed above). I'm guessing / hoping that Cython does the C equivalent of `list(pids)` on whatever iterable is passed.

python/rapidsmpf/rapidsmpf/shuffler.pyx
TomAugspurger: Works for me. I just wanted to confirm we weren't violating a rule of the Python API being a strict wrapper around the C++ API, like pylibcudf does.

python/rapidsmpf/rapidsmpf/shuffler.pyx
```diff
@@ -131,6 +133,9 @@ cdef class Shuffler:
         chunks
             A map where keys are partition IDs (``int``) and values are packed
             data (``PackedData``).
+        grouped
+            If ``True``, the chunks are grouped by the destination rank of the
+            partition ID.
```
TomAugspurger, line 138: A bit of context around when a user might (or might not) want to do this. IIUC, it's most beneficial when there are many partitions relative to the number of nodes because fewer messages are required. Is there any downside to `grouped=True`?

TomAugspurger (commented):
One general question: is there any coupling between `insert_chunks(grouped=True)` and `insert_finished` with a list of partition IDs? Or can you use the grouped `insert_chunks` with integer `insert_finished` and vice-versa? (I'm guessing the answer is "there's no coupling; do whatever you want")

python/rapidsmpf/rapidsmpf/shuffler.pxd
```diff
@@ -30,7 +30,9 @@ cdef extern from "<rapidsmpf/shuffler/shuffler.hpp>" nogil:
         ) except +
         void shutdown() except +
         void insert(unordered_map[uint32_t, cpp_PackedData] chunks) except +
+        void insert_grouped(unordered_map[uint32_t, cpp_PackedData] chunks) except +
```
pentschev, line 33: Not sure if I'm missing something from previous discussion, but `insert_grouped` is not in the `pyi`/`pyx` files, is that intentional? Similarly, `insert_chunks` is in `pyi`/`pyx` but not here.

pentschev: Ok, that was in fact my mistake, for some reason I overlooked and thought `insert_grouped` was not even called but it is. Please disregard my comment.

python/rapidsmpf/rapidsmpf/shuffler.pyx
```diff
@@ -136,9 +139,16 @@ cdef class Shuffler:
         -----
         This method adds the given chunks to the shuffle, associating them with their
         respective partition IDs.
+
+        There are some considerations for using ``grouped=True``:
```
TomAugspurger, line 143: I suspect this portion of the docstring isn't needed?

May be worth adding a `See Also` section This would be roughly:

```
See Also
--------
Shuffler.concat_insert
```

and likewise for `Shuffler.concat_insert` to refer to `Shuffler.insert_chunks`.

python/rapidsmpf/rapidsmpf/shuffler.pyx
```diff
@@ -147,25 +157,66 @@ cdef class Shuffler:
...
+            data (``PackedData``).
+
+        Notes
+        -----
+        This method adds the given chunks to the shuffle, associating them with their
+        respective partition IDs.
+
+        There are some considerations for using this method:
+        - The chunks are grouped by the destination rank of the partition ID and
+        concatenated on device memory.
```
TomAugspurger, line 178: Sphinx might warn about this, I'm not sure:

```suggestion
        - The chunks are grouped by the destination rank of the partition ID and
          concatenated on device memory.
```

python/rapidsmpf/rapidsmpf/shuffler.pyx
```diff
@@ -147,25 +157,66 @@ cdef class Shuffler:
...
+        - Caller thread will perform the concatenation, and hence it will be blocked.
```
TomAugspurger, line 179: Will the caller thread "be blocked", or will *it* block while doing the concatenation (like the actual set of memcpy operations needed to do the concat)?

```suggestion
        - The caller thread will perform the concatenation, and hence it will be blocked.
```

nirandaperera: Caller thread will be doing the actual memory copies.

TomAugspurger:
FYI, https://downloads.rapids.ai/ci/rapidsmpf/pull-request/361/d8b8c92/docs/rapidsmpf/html/api/#rapidsmpf.shuffler.Shuffler.concat_insert has the rendered HTML (VPN required)

image


Might need a newline before starting the list?

nirandaperera:
@TomAugspurger thanks. I added it now.

TomAugspurger (approved):

nirandaperera:
@TomAugspurger https://github.com/rapidsai/rapidsmpf/actions/runs/15984361460/job/45085669282?pr=361 Yikes, it seems precommit doesnt agree with that change :slightly_frowning_face:

nirandaperera:
\merge

PR #365, opened 2025-06-30
Use transition states in PausableThreadLoop

### LINKED ISSUE #362: Test failure
Had a test failure on an unrelated code [here](https://github.com/rapidsai/rapidsmpf/actions/runs/15932588402/job/44946092752#step:10:53860)

```
10: [ RUN      ] ShuffleTestP/ShuffleInsertGroupedTest.InsertPackedData/total_nparts_100__nbytes_10
10: /tmp/conda-bld-output/bld/rattler-build_librapidsmpf/work/cpp/tests/test_shuffler.cpp:649: Failure
10: Expected equality of these values:
10:   outbound_chunks
10:     Which is: { (98, 2), (94, 2), (90, 2), (86, 2), (82, 2), (78, 2), (74, 2), (70, 2), (66, 2), (62, 2), (58, 2), (54, 2), (50, 2), (46, 2), (42, 2), (38, 2), (34, 2), (30, 2), (26, 2), (22, 2), (18, 2), (14, 2), (10, 2), (6, 2), (2, 2), (99, 2), (95, 2), (91, 2), (87, 2), (83, 2), (79, 2), (75, 2), ... }
10:   shuffler.outbound_chunk_counter_
10:     Which is: { (55, 2), (51, 2), (47, 2), (43, 2), (39, 2), (35, 2), (31, 2), (11, 2), (7, 2), (3, 2), (15, 2), (19, 2), (23, 2), (27, 2), (59, 2), (63, 2), (67, 2), (71, 2), (75, 2), (79, 2), (83, 2), (87, 2), (91, 2), (95, 2), (99, 2), (57, 2), (53, 2), (49, 2), (45, 2), (41, 2), (37, 2), (33, 2), ... }
10: 
10: /tmp/conda-bld-output/bld/rattler-build_librapidsmpf/work/cpp/tests/test_shuffler.cpp:652: Failure
10: Expected equality of these values:
10:   pids.size()
10:     Which is: 100
10:   n_data_messages
10:     Which is: 50
10: 
10: /tmp/conda-bld-output/bld/rattler-build_librapidsmpf/work/cpp/tests/test_shuffler.cpp:653: Failure
10: Expected equality of these values:
10:   num_bytes * pids.size()
10:     Which is: 1000
10:   n_total_data_size
10:     Which is: 500
10: 
10: /tmp/conda-bld-output/bld/rattler-build_librapidsmpf/work/cpp/tests/test_shuffler.cpp:654: Failure
10: Expected equality of these values:
10:   num_bytes * pids.size()
10:     Which is: 1000
10:   n_total_metadata_size
10:     Which is: 500
10: 
10: [       OK ] ShuffleTestP/ShuffleInsertGroupedTest.InsertPackedData/total_nparts_100__nbytes_10 (6 ms)
10: [ RUN      ] ShuffleTestP/ShuffleInsertGroupedTest.InsertPackedData/total_nparts_100__nbytes_1000
10: [  FAILED  ] ShuffleTestP/ShuffleInsertGroupedTest.InsertPackedData/total_nparts_100__nbytes_10, where GetParam() = (100, 10) (6 ms)
```

wence-:
Since pausing and stopping a thread does not interrupt the execution of the current iteration of its event loop, to correctly track state, we introduce intermediate stopping and pausing states.

This is somewhat delicate because we were previously making (I think incorrect) assumptions about the state of the thread loop after calling pause. Fix those with some judicious wakes.

- Closes #362

### DIFF
--- a/cpp/src/pausable_thread_loop.cpp
+++ b/cpp/src/pausable_thread_loop.cpp
@@ -12,8 +12,22 @@ PausableThreadLoop::PausableThreadLoop(std::function<void()> func, Duration slee
         while (true) {
             {
                 std::unique_lock<std::mutex> lock(mutex_);
-                cv_.wait(lock, [this]() { return !paused_ || !active_; });
-                if (!active_) {
+                cv_.wait(lock, [this]() { return state_ != State::Paused; });
+                // Note: state changes here must notify any waiters.
+                switch (state_) {
+                case State::Running:
+                    break;
+                case State::Pausing:
+                    state_ = State::Paused;
+                    lock.unlock();
+                    cv_.notify_one();
+                case State::Paused:
+                    continue;
+                case State::Stopping:
+                    state_ = State::Stopped;
+                    lock.unlock();
+                    cv_.notify_one();
+                case State::Stopped:
                     return;
                 }
             }
@@ -37,18 +51,30 @@ PausableThreadLoop::~PausableThreadLoop() {
 
 bool PausableThreadLoop::is_running() const noexcept {
     std::lock_guard<std::mutex> lock(mutex_);
-    return !paused_;
+    return state_ != State::Paused;
+}
+
+void PausableThreadLoop::pause_nb() {
+    {
+        std::lock_guard<std::mutex> lock(mutex_);
+        state_ = State::Pausing;
+    }
+    cv_.notify_one();
 }
 
 void PausableThreadLoop::pause() {
-    std::lock_guard<std::mutex> lock(mutex_);
-    paused_ = true;
+    pause_nb();
+    // And wait for the loop to flip to paused state. Behaviour is
+    // undefined if someone else called resume/stop while we're in
+    // this function.
+    std::unique_lock lock(mutex_);
+    cv_.wait(lock, [this]() { return state_ == State::Paused; });
 }
 
 void PausableThreadLoop::resume() {
     {
         std::lock_guard<std::mutex> lock(mutex_);
-        paused_ = false;
+        state_ = State::Running;
     }
     cv_.notify_one();
 }
@@ -56,8 +82,7 @@ void PausableThreadLoop::resume() {
 void PausableThreadLoop::stop() {
     {
         std::lock_guard<std::mutex> lock(mutex_);
-        active_ = false;
-        paused_ = false;  // Ensure it's not stuck in pause
+        state_ = State::Stopping;
     }
     cv_.notify_one();  // Wake up thread to exit
     if (thread_.joinable()) {
--- a/cpp/src/progress_thread.cpp
+++ b/cpp/src/progress_thread.cpp
@@ -102,7 +102,11 @@ void ProgressThread::remove_function(FunctionID function_id) {
     functions_.erase(function_id.function_index);
 
     if (functions_.empty()) {
-        thread_.pause();
+        // Pausing the thread is not atomic, and so `event_loop` may
+        // run, needing the lock, we must therefore drop the lock before
+        // pausing.
+        lock.unlock();
+        thread_.pause_nb();
     }
 }
 
@@ -113,6 +117,7 @@ void ProgressThread::pause() {
 
 void ProgressThread::resume() {
     thread_.resume();
+    cv_.notify_all();
 }
 
 bool ProgressThread::is_running() const {
--- a/cpp/include/rapidsmpf/pausable_thread_loop.hpp
+++ b/cpp/include/rapidsmpf/pausable_thread_loop.hpp
@@ -6,6 +6,7 @@
 #pragma once
 
 #include <condition_variable>
+#include <cstdint>
 #include <functional>
 #include <mutex>
 #include <thread>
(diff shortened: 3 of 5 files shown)


cpp/src/progress_thread.cpp
```diff
@@ -102,7 +102,11 @@ void ProgressThread::remove_function(FunctionID function_id) {
     functions_.erase(function_id.function_index);
 
     if (functions_.empty()) {
-        thread_.pause();
+        // Pausing the thread is not atomic, and so `event_loop` may
+        // run, needing the lock, we must therefore drop the lock before
+        // pausing.
+        lock.unlock();
+        thread_.pause_nb();
     }
```
wence-, line 110: I think this comment is true if we use the blocking `thread_.pause()` (which definitely waits for the iteration to complete with us holding the lock). With nonblocking pause I don't think we need to drop the lock. (This was what was saving us before).

cpp/src/progress_thread.cpp
```diff
@@ -113,6 +117,7 @@ void ProgressThread::pause() {
 
 void ProgressThread::resume() {
     thread_.resume();
+    cv_.notify_all();
```
wence-, line 120: We need to notify on state changes in both directions.

cpp/src/pausable_thread_loop.cpp
```diff
@@ -12,8 +12,22 @@ PausableThreadLoop::PausableThreadLoop(std::function<void()> func, Duration slee
...
+                case State::Pausing:
+                    state_ = State::Paused;
+                    lock.unlock();
+                    cv_.notify_one();
+                case State::Paused:
+                    continue;
+                case State::Stopping:
+                    state_ = State::Stopped;
+                    lock.unlock();
+                    cv_.notify_one();
```
wence-, line 29: When we're in the intermediate stages, when we then flip the state we must notify any waiter.

cpp/src/progress_thread.cpp
pentschev: Isn't this already achieved with https://github.com/rapidsai/rapidsmpf/pull/365/files#diff-07f80ad017d4088866533997c3a66643e01ef6bd52c0d53ef60dc69afa9d5a4dR79?

cpp/src/progress_thread.cpp
wence-: I put the comment in when I was using `pause` rather than `pause_nb`...

cpp/src/progress_thread.cpp
wence-: No? This one notifies waiters in the progressthread that its internal pausablethread has been updated in state to running.

pentschev (approved):
Thanks @wence- !

cpp/src/pausable_thread_loop.cpp
```diff
@@ -37,27 +51,40 @@ PausableThreadLoop::~PausableThreadLoop() {
...
+    // this function.
+    std::unique_lock lock(mutex_);
+    cv_.wait(lock, [this]() { return state_ == State::Paused; });
 }
 
 void PausableThreadLoop::resume() {
     {
         std::lock_guard<std::mutex> lock(mutex_);
-        paused_ = false;
+        state_ = State::Running;
```
nirandaperera, line 79: Should we check `Stopping` here? Ideally once we have started stopping, we should not be able to resume

cpp/src/pausable_thread_loop.cpp
```diff
@@ -37,27 +51,40 @@ PausableThreadLoop::~PausableThreadLoop() {
 
 bool PausableThreadLoop::is_running() const noexcept {
     std::lock_guard<std::mutex> lock(mutex_);
-    return !paused_;
+    return state_ != State::Paused;
```
nirandaperera, line 54:
```suggestion
    return state_ != State::Paused and state_ != Stopping;
```

nirandaperera (commented):
Just some minor comments on handling `Stopping` state. 
But +1 on the approach for sure. :innocent:

cpp/src/pausable_thread_loop.cpp
wence-: Stopping is still running (the thread loop might still be executing).

wence-:
> @wence- I've been prototyping `std::atomic::wait` instead of cv wait [here](https://github.com/rapidsai/rapidsmpf/pull/359), now that we are on c++20. WDYT about that?

Maybe, however, the memory ordering is more complicated. What are the advantages? The classes are smaller (because we don't have 40 bytes for the mutex and 8 for the condition_variable).

Disadvantages: the updates are no longer done in critical sections (maybe not an issue).

Is there a performance gain here?

TBH, I would prefer if we could move away from rolling our own concurrency solutions and find a good library that offers the building blocks we need.

cpp/src/pausable_thread_loop.cpp
```diff
@@ -37,27 +51,42 @@ PausableThreadLoop::~PausableThreadLoop() {
...
 }
 
 void PausableThreadLoop::stop() {
     {
         std::lock_guard<std::mutex> lock(mutex_);
-        active_ = false;
-        paused_ = false;  // Ensure it's not stuck in pause
+        state_ = State::Stopping;
     }
     cv_.notify_one();  // Wake up thread to exit
```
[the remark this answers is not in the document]
wence-: i.e. `stop()` induces the transition `running/paused -> stopping -> stopped` (of the thread loop). the thread calling stop wakes up the thread loop, then waits for the thread loop to `join()` (which is equivalent to the thread loop being notified, updating its state to `stopped` and then returning). If we in the thread loop change the state and notify before returning (and then here we wait for that notification) we just do a bit more waiting.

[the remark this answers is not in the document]
nirandaperera: I misread the wait predicate. The notification is still important. So, pls disregard that part

PR #369, opened 2025-07-02
[MINOR] Adding more nvtx info 

nirandaperera:
This PR adds more NVTX information to the codebase.

### DIFF
--- a/cpp/src/shuffler/shuffler.cpp
+++ b/cpp/src/shuffler/shuffler.cpp
@@ -356,7 +356,7 @@ class Shuffler::Progress {
         // Check if any data in transit is finished.
         {
             auto const t0_check_future_finish = Clock::now();
-            RAPIDSMPF_NVTX_SCOPED_RANGE("check_fut_finish", in_transit_futures_);
+            RAPIDSMPF_NVTX_SCOPED_RANGE("check_fut_finish", in_transit_futures_.size());
             if (!in_transit_futures_.empty()) {
                 std::vector<ChunkID> finished =
                     shuffler_.comm_->test_some(in_transit_futures_);
--- a/cpp/src/shuffler/shuffler.cpp
+++ b/cpp/src/shuffler/shuffler.cpp
@@ -207,6 +207,7 @@ class Shuffler::Progress {
         {
             auto const t0_metadata_recv = Clock::now();
             RAPIDSMPF_NVTX_SCOPED_RANGE("meta_recv", -1);
+            int i = 0;
             while (true) {
                 auto const [msg, src] = shuffler_.comm_->recv_any(metadata_tag);
                 if (msg) {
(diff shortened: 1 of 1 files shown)


cpp/src/shuffler/shuffler.cpp
```diff
@@ -170,195 +172,227 @@ class Shuffler::Progress {
...
-                    shuffler_.partition_owner(shuffler_.comm_, chunk.part_id(0))
-                        == shuffler_.comm_->rank(),
-                    "receiving chunk not owned by us"
-                );
-                incoming_chunks_.insert({src, std::move(chunk)});
-            } else {
-                break;
+        {
+            auto const t0_metadata_recv = Clock::now();
+            RAPIDSMPF_NVTX_SCOPED_RANGE("meta_recv", -1);
```
pentschev, line 209: The `-1` value is just to fill it because it's a required argument, right? Assuming there's a proper NVTX API for that, maybe a better solution is adding an alternative scoped range implementation that doesn't take a value instead, having a value here implies the user needs to read the code to figure out what that means.

cpp/src/shuffler/shuffler.cpp
```diff
@@ -170,195 +172,227 @@ class Shuffler::Progress {
...
-                    "shuffle-payload-send", chunk.concat_data_size()
-                );
-                fire_and_forget_.push_back(
-                    shuffler_.comm_->send(chunk.release_data_buffer(), src, gpu_data_tag)
-                );
-            } else {
-                break;
+        {
+            auto const t0_init_gpu_data_send = Clock::now();
+            RAPIDSMPF_NVTX_SCOPED_RANGE("init_gpu_send", -1);
```
pentschev, line 324: Similarly here, maybe a separate scoped range implementation without a value.

cpp/src/shuffler/shuffler.cpp
```diff
@@ -159,6 +160,7 @@ class Shuffler::Progress {
      * @return The progress state of the shuffler.
      */
     ProgressThread::ProgressState operator()() {
+        RAPIDSMPF_NVTX_SCOPED_RANGE("Shuffler.Progress", p_iters++);
```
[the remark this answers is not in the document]
nirandaperera: I was actually wanted to monitor the number of progress thread iterations. Mainly because, it gives an idea about how many times we call the `nanosleep` at the end of the iteration.

cpp/include/rapidsmpf/nvtx.hpp
```diff
@@ -88,21 +72,61 @@ struct rapidsmpf_domain {
...
+        RAPIDSMPF_NVTX_SCOPED_RANGE_IMPL_WITHOUT_VAL \
+    )                                                \
+    (__VA_ARGS__)
+
 /**
  * @brief Convenience macro for generating an NVTX scoped range in the `rapidsmpf` domain
  * to annotate a time duration.
  *
- * Takes two arguments (message, payload).
+ * Takes one or two arguments (message, [optional] payload).
```
pentschev, line 113: Isn't there a way to just document `@params`, and make some of them optional?

pentschev (approved):
Left one comment about docs, other LGTM. Thanks @nirandaperera !

PR #373, opened 2025-07-09
Remove need for wildcard receive in readyfordata ack

wence-:
When we send a message to a destination rank that we have a chunk for them, we know we're going to receive an ack, so pre-post the receive and explicitly manage those messages. This removes a wildcard tag match loop, and ensures the ready for data messages have matching receives available which will probably enable eager protocol optimisations in the underlying comms library.

### DIFF
--- a/cpp/include/rapidsmpf/communicator/communicator.hpp
+++ b/cpp/include/rapidsmpf/communicator/communicator.hpp
@@ -443,7 +443,7 @@ class Communicator {
      * futures are erased from the vector.
      * @return Completed futures.
      */
-    std::vector<std::unique_ptr<Future>> virtual test_some(
+    [[nodiscard]] virtual std::vector<std::unique_ptr<Future>> test_some(
         std::vector<std::unique_ptr<Future>>& future_vector
     ) = 0;
 
--- a/cpp/include/rapidsmpf/communicator/communicator.hpp
+++ b/cpp/include/rapidsmpf/communicator/communicator.hpp
@@ -439,11 +439,12 @@ class Communicator {
     /**
      * @brief Tests for completion of multiple futures.
      *
-     * @param future_vector Vector of Future objects.
-     * @return Indices of completed futures.
+     * @param[inout] future_vector Vector of Future objects. Completed
+     * futures are erased from the vector.
+     * @return Completed futures.
      */
-    std::vector<std::size_t> virtual test_some(
-        std::vector<std::unique_ptr<Future>> const& future_vector
+    std::vector<std::unique_ptr<Future>> virtual test_some(
+        std::vector<std::unique_ptr<Future>>& future_vector
     ) = 0;
 
     /**
--- a/cpp/src/communicator/mpi.cpp
+++ b/cpp/src/communicator/mpi.cpp
@@ -156,7 +156,10 @@ std::unique_ptr<Communicator::Future> MPI::recv(
 std::pair<std::unique_ptr<std::vector<uint8_t>>, Rank> MPI::recv_any(Tag tag) {
     int msg_available;
     MPI_Status probe_status;
-    RAPIDSMPF_MPI(MPI_Iprobe(MPI_ANY_SOURCE, tag, comm_, &msg_available, &probe_status));
+    MPI_Message matched_msg;
+    RAPIDSMPF_MPI(MPI_Improbe(
+        MPI_ANY_SOURCE, tag, comm_, &msg_available, &matched_msg, &probe_status
+    ));
     if (!msg_available) {
         return {nullptr, 0};
     }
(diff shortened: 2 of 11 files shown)


wence-:
cc @nirandaperera

nirandaperera:
> cc @nirandaperera

:sweat_smile: I thought I was working on it. LOL! Anyways, let me do a review.

cpp/include/rapidsmpf/communicator/communicator.hpp
```diff
@@ -439,11 +439,12 @@ class Communicator {
...
      *
-     * @param future_vector Vector of Future objects.
-     * @return Indices of completed futures.
+     * @param[inout] future_vector Vector of Future objects. Completed
+     * futures are erased from the vector.
+     * @return Completed futures.
      */
-    std::vector<std::size_t> virtual test_some(
-        std::vector<std::unique_ptr<Future>> const& future_vector
+    std::vector<std::unique_ptr<Future>> virtual test_some(
```
nirandaperera, line 446:
```suggestion
    virtual std::vector<std::unique_ptr<Future>> test_some(
```

cpp/src/communicator/mpi.cpp
```diff
@@ -151,7 +156,10 @@ std::unique_ptr<Communicator::Future> MPI::recv(
 std::pair<std::unique_ptr<std::vector<uint8_t>>, Rank> MPI::recv_any(Tag tag) {
     int msg_available;
     MPI_Status probe_status;
-    RAPIDSMPF_MPI(MPI_Iprobe(MPI_ANY_SOURCE, tag, comm_, &msg_available, &probe_status));
+    MPI_Message matched_msg;
+    RAPIDSMPF_MPI(MPI_Improbe(
+        MPI_ANY_SOURCE, tag, comm_, &msg_available, &matched_msg, &probe_status
+    ));
```
wence-, line 162: To do the same change in the UCXX communicator requires https://github.com/rapidsai/ucxx/issues/457

cpp/src/communicator/mpi.cpp
```diff
@@ -183,30 +185,43 @@ std::pair<std::unique_ptr<std::vector<uint8_t>>, Rank> MPI::recv_any(Tag tag) {
...
-    return std::vector<std::size_t>(completed.begin(), completed.end());
+    std::vector<std::unique_ptr<Communicator::Future>> completed;
+    completed.reserve(static_cast<std::size_t>(num_completed));
+    std::ranges::transform(
+        indices.begin(),
+        indices.begin() + num_completed,
+        std::back_inserter(completed),
+        [&](std::size_t i) {
+            std::unique_ptr<Communicator::Future> fut{nullptr};
+            std::swap(fut, future_vector[i]);
```
nirandaperera, line 219: Nit. Cant we do this?
```suggestion
            std::unique_ptr<Communicator::Future> fut{std::move(future_vector[i])};
```

nirandaperera (approved):
LGTM. Just some nitpicks.

nirandaperera:
> > > cc @nirandaperera
> > 
> > 
> > 😅 I thought I was working on it. LOL! Anyways, let me do a review.
> 
> You were busy in meetings :)

I started on it, but had a correction issue. Maybe I missed this, 
```c++
auto dst =
                        shuffler_.partition_owner(shuffler_.comm_, chunk.part_id(0));
```

wence-:
Oh yeah, that check for emptiness does beed to be added. Do you want to pick this up?

nirandaperera:
So, IINM what's happening is, we post receives for the ready ack immediately after outgoing chunks metadata sends have been posted. Say, we have two chunks (chunk id 0 and 1) to the same destination. Destination receives these metadata in order 0, 1. It puts the metadata in the `incoming_chunks_` multimap under source (I think order is still preserved). However, when iterating through `incoming_chunks_`, if chunk0's buffer is not ready, we skip and go to chunk1. Now, chunk1's ready ack send will be posted, before chunk0, and source will intepret it as chunk0's ack. It will then post sends for chunk0 data.

nirandaperera:
@pentschev I did some long running tests on this on DGX15. It seems to be stable enough. 

Let's ask @madsbk 's review on this as well.

madsbk:
> Now, chunk1's ready ack send will be posted, before chunk0, and source will interpret it as chunk0's ack

@nirandaperera, where is this happening in the code? I thought we used the ChunkID and not the message order to identify ACKs?
Like here: https://github.com/rapidsai/rapidsmpf/blob/11b8e394d407b5d910356c8b8e4a433beff5defd/cpp/src/shuffler/shuffler.cpp#L342

wence-:
> and source will intepret it as chunk0's ack. It will then post sends for chunk0 data.

I think (perhaps you also ended up agreeing) that this analysis is not correct.

What's going on:

- Send metadata `src` to `dst` for `k` chunks. We now expect to receive `k` `ready_ack`s
- On `dst` receive any metadata and start allocating buffers to receive into. Once a receive buffer is ready, post a receive for the data and send a message to `src` with the `ready_ack`. Crucially, this _contains_ chunk id we're expecting.

So, receives for the data are posted in (some) order corresponding to the order in which the data buffers become ready, we send the ready_ack messages back to `src` in the same order. For a given `src`/`dst` pair, the ready ack messages will be received in order on `src`. When we get the ack, we figure out which chunk it corresponds to (by decoding the message) and then send that data to `dst`. Since the ready ack messages don't overtake each other, we send data messages from `src` in the order `dst` is expecting them.

wence-:
I think the error we saw is that I wasn't checking `ready_ack_receives_` being empty in the finish condition for the shuffler event loop.

nirandaperera:
> I think the error we saw is that I wasn't checking `ready_ack_receives_` being empty in the finish condition for the shuffler event loop.

@wence- Maybe. My intuition was, both `ready_ack_receives` and `outgoing_chunks` map were consumed simultaneously. So, essentially `ready_ack_receives.size() == outgoing_chunks.size()`. I felt like testing one container should be sufficient.

nirandaperera:
@wence- I think we can merge this PR.

PR #375, opened 2025-07-14
Cleanup CUDA 11 entries and ensure custom Docker configs in wheels tests

pentschev:
Remove CUDA 11 GHA workflow filters and ensure wheels tests use the same custom Docker configurations required for UCX.


pentschev:
Thanks @bdice for the quick review!

PR #377, opened 2025-07-15
Adding an initial AllGather implementation

nirandaperera:
This PR adds the initial implementation of the AllGatherer operation. 

It piggybacks on the `Shuffler` to broadcast messages in each rank. This is done by _copying_ inserted data and assigning the rank as the partititon ID. An identity function is set as the `partition_owner` lambda for the Shuffler. 

Let's use this PR to iron out the Allgather API for the callers. 

There are several downsides of this initial impl. 
- Making copies during insertion (this hinders the scalability of Allgather operation)
    - To address this we need to add an API endpoint to send multiple ranks. Currently, when we send a buffer, the ownership of the buffer is transferred to the send request future. And the future only contains a single future. Once we fix this, we can have a separate progression path for allgather, where a single buffer will be send to all ranks, irrespective of PartitionID. 

- ~~There are no order guarantees for the data extracted. I believe this could be problematic for the callers. But I may be wrong.~~
After discussing with @wence- , I have added ordering guarantees as follows. 
  - `wait_and_extract`: No ordering guarantees.
  - `wait_and_extract_ordered`: Data is ordered by rank and insertion order. 
  ie. notating rank `i`'s `j`th partition by `y_i_j` , the results will be `[y_0_0, y_0_1, y_0_2, ... y_0_k, ..., y_n_0, ..., y_n_k]`. 
  It provides a `n_chunks_per_rank` vector, indicating `j` value in each rank. 

- The extraction can only be done after calling `insert_finished`. Therefore, data can not be consumed in a streaming fashion.


cpp/include/rapidsmpf/all_gatherer/all_gatherer.hpp
```diff
@@ -0,0 +1,99 @@
...
+namespace rapidsmpf::experimental::all_gatherer {
+
+/**
+ * @brief All-gather service for cuDF tables.
+ *
+ * The `AllGatherer` class provides an interface for performing an all-gather operation
+ * on cuDF tables, collecting data from all ranks and distributing the complete dataset
+ * to all ranks.
+ */
+class AllGatherer {
```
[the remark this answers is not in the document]
nirandaperera: @pentschev I prefer that too. The only reason I chose that was the `ShufflER` name.

cpp/include/rapidsmpf/all_gatherer/all_gatherer.hpp
```diff
@@ -0,0 +1,102 @@
+/**
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION & AFFILIATES.
```
madsbk, line 2:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.
```

cpp/include/rapidsmpf/all_gatherer/all_gatherer.hpp
```diff
@@ -0,0 +1,102 @@
...
+#include <rmm/cuda_stream_view.hpp>
+
+#include <rapidsmpf/buffer/packed_data.hpp>
+#include <rapidsmpf/buffer/resource.hpp>
+#include <rapidsmpf/communicator/communicator.hpp>
+#include <rapidsmpf/progress_thread.hpp>
+#include <rapidsmpf/shuffler/shuffler.hpp>
+#include <rapidsmpf/statistics.hpp>
+
+namespace rapidsmpf::experimental::all_gatherer {
```
madsbk, line 21: I don't think er need `experimental`, everything is experimental at this point :)

cpp/include/rapidsmpf/all_gatherer/all_gatherer.hpp
```diff
@@ -0,0 +1,99 @@
...
+    [[nodiscard]] bool finished() const;
+
+    /**
+     * @brief Wait for all ranks to finish and extract all gathered data.
+     *
+     * @param timeout Optional timeout (ms) to wait.
+     * @return A vector of packed data containing all gathered data from all ranks.
+     *
+     * @note There are no guarantees on the order of the data across ranks.
+     */
```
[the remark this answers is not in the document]
madsbk: +1

cpp/src/all_gatherer.cpp
```diff
@@ -0,0 +1,112 @@
+/**
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION & AFFILIATES.
```
madsbk, line 2:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.
```

nirandaperera:
I think this test failure is a valid case. I just reproduced this locally. Enabling `TRACE` logs show some processes are stuck on Barrier during Shutdown.

madsbk (approved):
Looks good to me

cpp/include/rapidsmpf/all_gather/all_gather.hpp
```diff
@@ -0,0 +1,104 @@
...
+#include <rapidsmpf/statistics.hpp>
+
+namespace rapidsmpf::all_gather {
+
+/**
+ * @brief All-gather service for cuDF tables.
+ *
+ * The `AllGather` class provides an interface for performing an all-gather operation
+ * on cuDF tables, collecting data from all ranks and distributing the complete dataset
+ * to all ranks.
```
wence-, line 28: This is not only for tables, right? The interface takes `PackedData`.

cpp/include/rapidsmpf/all_gather/all_gather.hpp
```diff
@@ -0,0 +1,104 @@
...
+     */
+    [[nodiscard]] bool finished() const;
+
+    /**
+     * @brief Wait for all ranks to finish and extract all gathered data.
+     *
+     * @param timeout Optional timeout (ms) to wait.
+     * @return A vector of packed data containing all gathered data from all ranks.
+     *
+     * @note There are no guarantees on the order of the data across ranks.
```
wence-, line 89: What does this mean? I really want the invariant that after `x = allgather(...)` then on every pair of ranks `i` and `j`, the value of `x` on rank `i` is equal to the value of `x` on rank `j`.

In fact, this is really a _necessity_ for usage by an (ordered) dataframe library.

Consider, for example, having some input (distributed/streamed) dataframe that is sorted on some column, it must be the case that after allgather, the concatenated dataframe is _still_ sorted on that column.

cpp/include/rapidsmpf/all_gather/all_gather.hpp
```diff
@@ -0,0 +1,104 @@
...
+     * @brief Shutdown the all-gather, blocking until all inflight communication is done.
+     */
+    void shutdown();
+
+    /**
+     * @brief Insert data into the all-gather.
+     *
+     * @param data The data to insert.
+     */
+    void insert(PackedData&& data);
```
wence-, line 68: We need to think about what the semantics of a streamed allgather are.

The classic MPI collective allgather provides that when we call `x = allgather(y)` then (annotating rank `i`s `y` data by `y_i`) we have `x = [y_0, y_1, ..., y_n]` on all ranks.

If we're doing streamed insertion then when we're doing

```
gatherer = AllGather(...);

while (...) {
    gatherer.insert(y_j); // each rank inserts j = 0..k partitions
}

result = gatherer.wait_and_extract();
```

What is the order of `result`, in this API it's a `std::vector<PackedData>`, notating rank `i`'s `j`th partition by `y_i_j` what order is the packed data in?

Is it `[y_0_0, y_0_1, y_0_2, ... y_0_k, ..., y_n_0, ..., y_n_k]` i.e. "as if" we had concatenated the local `y_j`s and then done an allgather)

Or is it `[y_0_0, y_1_0, ..., y_n_0, ..., y_0_k, ..., y_n_k]` i.e. "as if" we allgather partition at a time and then concatenate.

I think it must be the first one, because we don't necessarily have the same number of partitions on each rank. Furthermore I think the sensible mental model for streaming + distributed is that the distributed part of the parallelism is "outermost" when we care about ordering of data. i.e. to materialise at some point a streamed distributed dataframe then we first locally concatenate our streamed chunks and then concatenate globally.

wence- (changes requested):
I think we need to think about, and provide guarantees on, the ordering of the return values.

cpp/include/rapidsmpf/all_gather/all_gather.hpp
nirandaperera: @wence- I agree. This ordering is not guaranteed at the moment. In fact myself and @madsbk discussed this ordered tables use case on Tue. We need to attach some additional info like, origin, source chunk ID, etc when pulling data out of the ready postbox. However, we decided to punt it for the moment, because we are focusing on broadcast joins which does not depend on ordering guarantees.

cpp/include/rapidsmpf/all_gather/all_gather.hpp
nirandaperera: @wence- thank you for clearly expressing the idea with the notations. :innocent: 
The current impl, does not guarantee either of the orderings, because when we send out data, we use the `rank` as the `partitionID` for each copy. So, when receiving, every rank will have 1 partition with many chunks, and this is the reason why ordering is not guaranteed. 

>Is it [y_0_0, y_0_1, y_0_2, ... y_0_k, ..., y_n_0, ..., y_n_k] i.e. "as if" we had concatenated the local y_js and then done an allgather)
>
>Or is it [y_0_0, y_1_0, ..., y_n_0, ..., y_0_k, ..., y_n_k] i.e. "as if" we allgather partition at a time and then concatenate.

IIIUC, first is, `rank and then insertion` order, while second is, `insertion and then rank` order. 

I agree with the first ordering semantic. Or how about, returning 
`tuple</*data*/ vector<PackedData>, /*src*/vector<Rank>, /*inserted order*/vector<PartID>>` so that users may sort `data` in any order they want?

cpp/include/rapidsmpf/all_gather/all_gather.hpp
nirandaperera: @wence- Yes. I think the biggest blocker ATM for allgather is, `Communicator::send` only accepts a single rank and takes ownership of the send buffer. I'm currently working on adding a new send/ broadcast method for the `Communicator`, that accepts a vector/ hashset of destination ranks.

nirandaperera: So, I think we'll end up having a similar progress loop like the shuffler for allgather.

cpp/include/rapidsmpf/all_gather/all_gather.hpp
nirandaperera: @wence- I have added ordering in the `wait_and_extract_ordered` method now.

cpp/tests/test_all_gather.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+    void TearDown() override {
+        GlobalEnvironment->barrier();
+    }
+
+    rmm::cuda_stream_view stream;
+    rapidsmpf::Communicator* comm;
+    std::unique_ptr<rapidsmpf::BufferResource> br;
+    std::unique_ptr<AllGather> all_gather;
+    std::unique_ptr<rmm::mr::device_memory_resource> mr;
```
wence-, line 109: dtor order is the reverse of this. So by the time you come to destruct the allgather and the buffer resource, the memory resource will be gone. This is a problem. You need to make sure that the member order matches the initialisation order requirements

nirandaperera: Thanks @wence-. I fixed that. But its not only that. I've noticed sometimes, some ranks go into the `TearDown` and start a barrier, while other ranks are still inside the test case. I think let's keep an eye on that. If that comes back again, we'll investigate more. 
I feel like we should differentiate test runs using the OpID. It gives each test case its own communication space.

cpp/include/rapidsmpf/shuffler/chunk.hpp
```diff
@@ -62,6 +62,34 @@ class Chunk {
...
+                std::move(other.expected_num_chunks_);
+            const_cast<std::vector<uint32_t>&>(meta_offsets_) =
+                std::move(other.meta_offsets_);
+            const_cast<std::vector<uint64_t>&>(data_offsets_) =
+                std::move(other.data_offsets_);
+            metadata_ = std::move(other.metadata_);
+            data_ = std::move(other.data_);
+            return *this;
+        }
+        return *this;
```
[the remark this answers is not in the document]
nirandaperera: I'll remove const modifier. Dont see a significant downside by doing that.

cpp/src/allgather.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+    for (auto&& chunk : chunks) {
+        auto rank = shuffler::Shuffler::extract_rank(chunk.chunk_id());
+        n_chunks_per_rank[static_cast<size_t>(rank)]++;
+        packed_data.emplace_back(
+            chunk.release_metadata_buffer(), chunk.release_data_buffer()
+        );
+    }
+
+    return {std::move(packed_data), std::move(n_chunks_per_rank)};
```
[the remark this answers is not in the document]
nirandaperera: I feel the caller might need to know how many messages are coming from each `Rank`, especially because we are semantically doing a collection of `AllgatherV`s.

cpp/src/allgather.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+AllGather::~AllGather() {
+    shutdown();
+}
+
+void AllGather::shutdown() {
+    if (shuffler_) {
+        shuffler_->shutdown();
+        shuffler_.reset();
+    }
```
wence-, line 69: Surely just `shuffler_.reset()` since the dtor for the shuffler will run `shutdown()` for us.

cpp/src/allgather.cpp
```diff
@@ -0,0 +1,149 @@
...
+bool AllGather::finished() const {
+    return shuffler_->finished();
+}
+
+std::vector<PackedData> AllGather::wait_and_extract(
+    std::optional<std::chrono::milliseconds> timeout
+) {
+    RAPIDSMPF_EXPECTS(
+        insert_finished_.load(), "insertion has not finished yet.", std::runtime_error
+    );
```
wence-, line 113: This tracking locally whether I have finished inserting. But that is not equivalent to saying that everyone has finished inserting. So I think this flag should be removed because the error checking it provides is not correct.

wence-: IOW, marking locally that I have finished insertions is a necessary *but not sufficient* condition for knowing that it is safe to extract data.

cpp/src/allgather.cpp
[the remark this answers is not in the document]
wence-: In what scenario do we need to know this? I would much prefer that `wait_and_extract_ordered` has the same API surface as `wait_and_extract`.

cpp/tests/test_allgather.cpp
```diff
@@ -0,0 +1,289 @@
...
+    std::vector<uint8_t> copied_data(n_elements * sizeof(int));
+    RAPIDSMPF_CUDA_TRY(cudaMemcpyAsync(
+        copied_data.data(),
+        packed_data.data->data(),
+        n_elements * sizeof(int),
+        cudaMemcpyDefault,
+        stream
+    ));
+
+    RAPIDSMPF_CUDA_TRY(cudaStreamSynchronize(stream));
```
wence-, line 77: This is a `Buffer` right? use `br.move_to_host_vector` on the result instead.

cpp/tests/test_allgather.cpp
```diff
@@ -0,0 +1,289 @@
...
+constexpr auto gen_offset(int i, int r) {
+    return i * 10 + r;
+};
+
+TEST_P(AllGatherTest, basic_allgather) {
+    auto this_rank = comm->rank();
+
+    for (int i = 0; i < n_inserts; i++) {
+        auto packed_data =
+            generate_packed_data(n_elements, gen_offset(i, this_rank), stream, *br);
```
wence-, line 155: Can we use a different number of elements on each rank, to test the `allgather_v_`-ness of this.

wence- (changes requested):
Some smaller concerns. I think the API surface is now looking pretty good.

PR #380, opened 2025-07-16
Add single-worker shuffle integration and Dask-cuDF example

rjzamora:
This adds a new task-based shuffle wrapper: `rapidsmpf.integrations.rapidsmpf_shuffle_graph`, similar to `rapidsmpf.integrations.dask.rapidsmpf_shuffle_graph`, which is based on the single-process `Communicator`.

A few refactors are included:

- Adds `rapidsmpf.integrations.core` for classes that are not really "Dask specific".
  - Moves  `rapidsmpf.integrations.dask.DataFrameT` to `rapidsmpf.integrations.core.DataFrameT`
  - Moves  `rapidsmpf.integrations.dask.DaskWorkerContext` to `rapidsmpf.integrations.core.WorkerContext`
  - Moves  `rapidsmpf.integrations.dask.DaskIntegration` to `rapidsmpf.integrations.core.ShufflerIntegration
- Updates `rapidsmpf.examples.dask.dask_cudf_shuffle` to (optionally) use the new single `rapidsmpf_shuffle_graph`.
  - New `cluster_kind` argument allows the user to choose from `{'distributed', 'single', 'auto'}`.


madsbk (changes requested):
Good idea

python/rapidsmpf/rapidsmpf/examples/dask.py
```diff
@@ -184,6 +196,20 @@ def dask_cudf_shuffle(
...
         *sort_boundary_names,
     )
+    if cluster_kind == "single":
+        graph = single_rapidsmpf_shuffle_graph(*shuffle_graph_args)
+    else:
+        try:
+            from rapidsmpf.integrations.dask.shuffler import rapidsmpf_shuffle_graph
+
+            graph = rapidsmpf_shuffle_graph(*shuffle_graph_args)
+        except (ImportError, ValueError):
```
TomAugspurger, line 206: I've gone back and forth, but I *think* that I dislike the "auto" behavior here :)

I'm just not sure how many cases we have where we'd be fine with silently falling back to single-worker execution if you happen to mess up the install of rapidsmpf, or forget to start a Dask cluster, or if `rapidsmpf_shuffle_graph` happens to throw a `ValueError` for some new reason (*hopefully* it's just because you forgot to start a cluster, but with exceptions it's hard to know for sure). They just seem so drastically different that I'd want to know if something happened. I guess that argues for me specifying `cluster_kind="distributed"` but since it's an optional keyword I know I (and probably users) will forget to set it.

The upside of having a single entrypoint for shuffling dask stuff is appealing. But right now I'm just not sure it's worth the potential downside...

python/rapidsmpf/rapidsmpf/integrations/single.py
```diff
@@ -0,0 +1,495 @@
...
+    Retrieve the single `WorkerContext`.
+
+    If the worker context does not already exist on the worker, it
+    will be created and attached to `_single_rapidsmpf_worker_context`.
+
+    Returns
+    -------
+    The existing or newly initialized worker context.
+    """
+    with WorkerContext.lock:
```
TomAugspurger, line 54: Is this lock necessary for correctness, or is it just a consequence of trying to reuse some code? Put differently, do we actually have multiple threads active here?

python/rapidsmpf/rapidsmpf/integrations/single.py
```diff
@@ -0,0 +1,495 @@
...
+    with ctx.lock:
+        if ctx.comm is not None:
+            return  # Single worker already set up
+
+        ctx.options = options
+        ctx.comm = new_communicator(options)
+        ctx.comm.logger.trace("single communicator created.")
+
+        # Insert RMM resource adaptor on top of the current RMM resource stack.
+        mr = RmmResourceAdaptor(
```
TomAugspurger, line 84: IIUC, there's some duplicate code here with https://github.com/rapidsai/rapidsmpf/blob/branch-25.08/python/rapidsmpf/rapidsmpf/integrations/dask/core.py#L199C5-L199C22. Any chance at sharing some of this?

python/rapidsmpf/rapidsmpf/examples/dask.py
rjzamora: It could be worth considering that I don't actually want anyone **using** this function is practice (at least not yet).

With that said: Since it's a "DataFrame" API, it definitely feels "off" for the user to need to specify whether or not we are executing in "distributed mode". I don't think the `cluster_kind` argument really **need** to exist. If there is a distributed client available, we should run in distributed mode. Otherwise, we use "single".

The only reason the `cluster_kind` argument makes sense, is because some users may want to use "single" execution with `.compute(scheduler="synchronous")` when a distributed client happens to be active. Perhaps the argument could be `force_single: bool = False`?

TomAugspurger: Ah, I'd missed that this is in `rapidsmpf.examples`. In that case, maybe this doesn't matter either way.

python/rapidsmpf/rapidsmpf/integrations/core.py
```diff
@@ -0,0 +1,115 @@
...
+    br: BufferResource | None = None
+    progress_thread: ProgressThread | None = None
+    comm: Communicator | None = None
+    spill_collection: SpillCollection = field(default_factory=SpillCollection)
+    statistics: Statistics | None = None
+    shufflers: dict[int, Shuffler] = field(default_factory=dict)
+    options: Options = field(default_factory=Options)
+
+
+@runtime_checkable
```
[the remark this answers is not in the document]
rjzamora: We do check that the integration is a proper `ShufflerIntegration` in `rapidsmpf_shuffle_graph` (and we raise an error if it isn't). Does it make sense to remove this check (and the corresponding test)?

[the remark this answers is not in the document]
TomAugspurger: IMO it's fine to remove. Our type checker will catch things if we mess stuff up internally.

python/rapidsmpf/rapidsmpf/examples/dask.py
rjzamora: I added a note to the doc-string that this API is currently for demonstration and testing purposes, so I think we are good on this front for now.

python/rapidsmpf/rapidsmpf/integrations/single.py
rjzamora: It should be possible for multiple threads to call this function at the same time. I definitely saw issues in an earlier iteration of this branch without the lock. With that said, all tests seem to pass without the lock right now, so something may have changed.

rjzamora:
Thanks @TomAugspurger, @wence- and @madsbk for the reviews!

I did a lot of refactoring here to consolidate as much logic as possible in `rapidsmpf.integtations.core`. I think the PR is pretty close, but there is still something up with the docs build (I put this part off as long as possible because I'm terrible at fixing doc issues).

TomAugspurger (commented):
Thanks, looks nice. I'll take a look at the doc build later today.

python/rapidsmpf/rapidsmpf/integrations/single.py
TomAugspurger: Thinking through this again: I think we do need a lock here if we're hoping to support multi-threaded applications that might independently launch shuffles. We somehow need to protect access to this shared, global resource.

I can't remember, but I think my original question was motivated by whether this was needed by a single shuffle in isolation (which I hope it doesn't, but can't say for sure).

python/rapidsmpf/rapidsmpf/integrations/single.py
```diff
@@ -0,0 +1,219 @@
...
+class _SingleWorker:
+    """Mutable single-worker utility class."""
+
+    context: WorkerContext | None = None
+
+
+_single_rapidsmpf_worker: _SingleWorker = _SingleWorker()
+
+
+def get_single_worker_context(worker: _SingleWorker | None = None) -> WorkerContext:
```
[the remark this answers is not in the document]
rjzamora: >What do we want the public API of this integration to be? Do users of this need access to get_single_worker_context, or is single_rapidsmpf_shuffle_graph sufficient?

Good question - I'm thinking that `single_rapidsmpf_shuffle_graph` and `setup_single_worker` are all that need to be "public" here.

>I've started to look at the docs now and can handle that if you want.

Thanks :)

I'm actually still stuck on cudf-polars integration in a distinct cudf branch. I'll try to submit a draft PR soon, but something is still off :/

python/rapidsmpf/rapidsmpf/integrations/single.py
```diff
@@ -0,0 +1,222 @@
...
+
+def _get_occupied_ids_single() -> list[set[int]]:
+    ctx = _get_single_worker_context()
+    return [set(ctx.shufflers.keys())]
+
+
+def _single_worker_barrier(
+    shuffle_ids: tuple[int, ...],
+    partition_count: int,
+    dependencies: Sequence[None],
```
rjzamora, line 94:
```suggestion
    *dependencies: Sequence[None],
```
Oops.

TomAugspurger:
Hmm looks like we might have some unexpected commits here. I'll try to merge `branch-25.10`.

python/rapidsmpf/rapidsmpf/integrations/core.py
```diff
@@ -0,0 +1,469 @@
...
+            ctx.shufflers[shuffle_id] = Shuffler(
+                ctx.comm,
+                ctx.progress_thread,
+                op_id=shuffle_id,
+                total_num_partitions=partition_count,
+                stream=DEFAULT_STREAM,
+                br=ctx.br,
+                statistics=ctx.statistics,
+            )
+    return ctx.shufflers[shuffle_id]
```
wence-, line 229:
```suggestion
        return ctx.shufflers[shuffle_id]
```

This return needs to happen with the lock held I think.

python/rapidsmpf/rapidsmpf/integrations/single.py
```diff
@@ -0,0 +1,222 @@
...
+class _SingleWorker:
+    """Mutable single-worker utility class."""
+
+    context: WorkerContext | None = None
+
+
+_single_rapidsmpf_worker: _SingleWorker = _SingleWorker()
+
+
+def _get_single_worker_context(worker: _SingleWorker | None = None) -> WorkerContext:
```
wence-, line 36: *nit*: This function is private to this module, which is already called `single`, why do we need to repeat `Single` everywhere?

How about `_worker_context`?

python/rapidsmpf/rapidsmpf/integrations/single.py
```diff
@@ -0,0 +1,222 @@
...
+    The existing or newly initialized worker context.
+    """
+    with WorkerContext.lock:
+        worker = worker or _single_rapidsmpf_worker
+        if worker.context is None:
+            worker.context = WorkerContext()
+        return cast("WorkerContext", worker.context)
+
+
+def setup_single_worker(options: Options = Options()) -> None:
```
wence-, line 55: Similarly here, we're in the `single` module, do we really need `single` in the name of these functions?

wence- (commented):
Minor quibbles, but overall I think this looks good.

python/rapidsmpf/rapidsmpf/integrations/core.py
madsbk: Good catch, we should hold on to the lock. 

PS: the current version without the lock works in CPython because `dict[x]` is atomic.

wence- (approved):

TomAugspurger:
In addition to the shuffler teardown hanging, I'm maybe a bit surprised at the timing (i.e. at the end of the entire run, rather than a specific test or when the shuffle completes): I'll look into the lifetimes and ensure that we don't have any circular references or things that would be keeping these alive.

TomAugspurger:
Adapting that slightly to clear the finished shufflers:

```python

def clear_finished(dask_worker):
    ctx = get_worker_context(dask_worker)
    for shuffle_id in list(ctx.shufflers.keys()):
        if ctx.shufflers[shuffle_id].finished():
            del ctx.shufflers[shuffle_id]

...

def main():
    ...

    print("check")
    print(list(client.run(check).values()))

    print("clear finished")
    client.run(clear_finished)

    print("re-check")
    print(list(client.run(check).values()))

```

it prints out:

```
check
[(0, 0), (2, 2), (2, 2), (2, 2), (2, 2), (0, 0), (2, 2), (2, 2)]
clear finished
re-check
[(0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0)]
```

Is it possible that there's a race condition between our `extract_partition` that checks `shuffler.finished()` and cleans things up, and when `shuffler.finished()` is actually set, such that no Dask task running `extract_partition` actually sees a finished Shuffle?

wence-:
I can't reproduce the hang locally. But not sure what is going on...

wence-:
> The output at https://github.com/rapidsai/rapidsmpf/actions/runs/16731828173/job/47362118530?pr=380#step:10:752 shows that `tests/test_dask.py::test_many_shuffles` "leaks" shufflers, such that their cleanup runs upon process exit rather than when the shuffle finishes.
> 
> Locally, I see this thanks the snippet that tries to do too many concurrent shuffles:
> 
> https://github.com/rapidsai/rapidsmpf/blob/85a2b3bf60e05e71adaa609689b7bbbbf58d81f4/python/rapidsmpf/rapidsmpf/tests/test_dask.py#L201-L206
> 
> . If I comment out that section, we don't have any leaked `Shuffle` objects. That's perhaps not surprising: we create the `Shuffler` instances (in `rapidsmpf_shuffle_graph`) but never actually compute the task graph, and so never call `extract_partitions`, and so never attempt to call `Shuffler.finished` / drop the `Shuffler` and run its destructor.
> For now, I propose that:
> 
> 1. We manually clean up `Shuffle` instances upon completion of that test, under the theory that this will help isolate future hangs; this isn't a long-term solution for our users though.
> 2. We rerun CI a handful of times to see if it fails.
> 3. Merge this PR
> 
> I think there's still more to do here though, even if avoiding the leaked `Shuffler` instances in that test happens to fix things.

OK, so I guess the thing is here that `rapidsmpf_shuffle_graph` stages (i.e. creates) the shuffler objects during graph build, and then when we stage too many concurrent shuffles we throw and don't clean things up.

One way we could try and make this a little more robust might be to not bother staging the shuffle during graph build: `insert_partition` calls `get_shuffler` with the requested shuffle ID so we would collectively correctly create the shuffles if we just did that I think.

That would mean we don't make the shuffler until it's required.

It still doesn't solve the underlying problem that if we catch some exception in our compute we will leave some shuffle objects lying around on the workers, but at least if we fail during actual compute it's probably less recoverable.

I guess the general problem is that if you create a shuffle but then never complete it will leave a progress thread running and the Shuffler destructor will block.

I'm not sure there's a nice way to solve this problem. But we should think about it.

python/rapidsmpf/rapidsmpf/tests/test_dask.py
```diff
@@ -134,6 +181,16 @@ def test_boostrap_single_node_cluster_no_deadlock() -> None:
...
     pytest.importorskip("dask_cudf")
 
+    def clear_finished_shuffles(dask_worker: Worker) -> None:
+        # Avoid leaking Shuffler objects between tests.
+        ctx = get_worker_context(dask_worker)
+        for shuffle_id, shuffler in list(ctx.shufflers.items()):
+            if ctx.shufflers[shuffle_id].finished():
+                del ctx.shufflers[shuffle_id]
+            else:
+                shuffler.shutdown()
```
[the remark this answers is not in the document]
TomAugspurger: I don't have the output anymore, but I did confirm locally with the prints from https://github.com/rapidsai/rapidsmpf/pull/380/commits/e3ee664f48170bbfc3b426d5935d5970508a37ae the Dask worker didn't have any Shuffler instances in its `WorkerContext`. Likewise for the main process after the `single` test.

In this test we stage too many shuffles before we attempt to insert / extract any partitions. Perhaps under those conditions, `shuffler.shutdown()` won't block, since there's no outstanding operations to progress? I get a little lost around [here](https://github.com/rapidsai/rapidsmpf/blob/85a2b3bf60e05e71adaa609689b7bbbbf58d81f4/cpp/src/shuffler/shuffler.cpp#L516).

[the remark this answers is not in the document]
wence-: So I think this is fine.

wence- (approved):

python/rapidsmpf/rapidsmpf/integrations/core.py
```diff
@@ -0,0 +1,480 @@
...
+    """Shuffle-integration protocol."""
+
+    @staticmethod
+    def insert_partition(
+        df: DataFrameT,
+        partition_id: int,
+        partition_count: int,
+        shuffler: Shuffler,
+        options: Any,
+        get_worker_context: Callable[[], WorkerContext],
```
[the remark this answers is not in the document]
TomAugspurger: 337e5268feeeae4044605a461c6f90b6b89a98ee avoids this API change, by putting the cluster kind (distributed, single) into the `options` that's already passed through to insert / extract. The default behavior is to use `distributed` (backwards compatible). `single` clusters will need to pass that through otherwise things will error since we'll try to get the (dask) `WorkerContext`, which won't exist, and the task will error.

I'm not 100% happy with this design but I think it'll do for now.

bdice (approved):
Approving CI changes.

PR #393, opened 2025-07-25
Revert "Remove need for wildcard receive in readyfordata ack (#373)"

wence-:
This reverts commit a47eece9ed7da9a4aeb3dec91bab2a4d1fcd468e.


wence-:
Saw another test suite failure, so reverting until we can figure things out.

pentschev (approved):

PR #396, opened 2025-07-28
Disable scanning for C++ files

pentschev:
Building 25.10 wheels packages fail due to the use of GCC 14, https://github.com/rapidsai/kvikio/pull/749/ implements a fix for that case that will probably fix it here as well.


bdice:
Maybe a similar change is needed in the python CMake code? https://github.com/rapidsai/rapidsmpf/blob/branch-25.08/python/rapidsmpf/CMakeLists.txt

pentschev:
Thanks @vyasr !

PR #399, opened 2025-07-29
Reduce wildcard message matching

wence-:
A second go at #373. To ensure that we observe in-order completion of ucx requests we separate ack receives by rank and break when we observe a gap.


cpp/src/shuffler/shuffler.cpp
```diff
@@ -325,36 +336,36 @@ class Shuffler::Progress {
...
+                std::transform_reduce(
+                    ready_ack_receives_.begin(),
+                    ready_ack_receives_.end(),
+                    0,
+                    std::plus<>(),
+                    [](auto& kv) { return kv.second.size(); }
+                )
+            );
+            for (auto& [dst, futures] : ready_ack_receives_) {
+                auto finished = shuffler_.comm_->test_some(futures);
```
pentschev, line 350: Maybe we should add a comment explaining why we need to loop over destinations individually and not the whole `ready_ack_receives_` at once. This is the type of thing I can see ourselves optimizing away in the future.

wence-:
> I tested it on my end and it is running as expected without crashes. However, should we also add the larger tests that were causing the errors? Below is the patch you sent me with changes that caused the failures.
> 

Thanks for testing.
[...]


> Other than the suggestion above, LGTM. Thanks @wence- .

Applied.

PR #401, opened 2025-07-29
Fixing Send to self segfault in rank 0

nirandaperera:
This PR fixes the send-to-self segfault in rank0 and adds a tests to verify. 

It also adds a `TestEnvironmentType` enum to easily skip unrelated tests, especially Single communicator tests.

### DIFF
--- a/cpp/tests/test_communicator.cpp
+++ b/cpp/tests/test_communicator.cpp
@@ -3,7 +3,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#include <cudf_test/base_fixture.hpp>
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
 #include <rmm/mr/device/cuda_memory_resource.hpp>
 #include <rmm/mr/device/device_memory_resource.hpp>
 #include <rmm/resource_ref.hpp>
@@ -15,11 +17,14 @@
 #include "environment.hpp"
 #include "utils.hpp"
 
-class BaseCommunicatorTest : public cudf::test::BaseFixture {
+class BaseCommunicatorTest : public ::testing::Test {
   protected:
     void SetUp() override {
         comm = GlobalEnvironment->comm_.get();
-        br = std::make_unique<rapidsmpf::BufferResource>(mr());
+        mr = std::unique_ptr<rmm::mr::device_memory_resource>(
+            new rmm::mr::cuda_memory_resource{}
+        );
+        br = std::make_unique<rapidsmpf::BufferResource>(mr.get());
         stream = rmm::cuda_stream_default;
         GlobalEnvironment->barrier();
     }
@@ -31,6 +36,7 @@ class BaseCommunicatorTest : public cudf::test::BaseFixture {
     virtual rapidsmpf::MemoryType memory_type() = 0;
 
     rapidsmpf::Communicator* comm;
+    std::unique_ptr<rmm::mr::device_memory_resource> mr;
     rmm::cuda_stream_view stream;
     std::unique_ptr<rapidsmpf::BufferResource> br;
 };


cpp/src/communicator/ucxx.cpp
```diff
@@ -1007,8 +1014,9 @@ constexpr ::ucxx::TagMask UserTagMask{std::numeric_limits<uint32_t>::max()};
 std::shared_ptr<::ucxx::Endpoint> UCXX::get_endpoint(Rank rank) {
     Logger& log = logger();
     try {
+        auto ep = shared_resources_->get_endpoint(rank);
         log.trace("Endpoint for rank ", rank, " already available, returning to caller");
-        return shared_resources_->get_endpoint(rank);
+        return ep;
```
pentschev, line 1019: Nit: revert this change. The reasoning is the logging will be omitted if `shared_resources_->get_endpoint(rank)` raises an exception, plus I think the original was more straightforward.

cpp/tests/test_communicator.cpp
```diff
@@ -0,0 +1,136 @@
...
+    // send data to self (ignore the return future)
+    auto send_future = comm->send(std::move(send_buf), self_rank, tag);
+
+    // receive data from self
+    std::vector<std::unique_ptr<Communicator::Future>> recv_futures;
+    auto recv_buf = make_buffer(n_elems);
+    recv_futures.emplace_back(comm->recv(self_rank, tag, std::move(recv_buf)));
+
+    while (!recv_futures.empty()) {
+        auto finished = comm->test_some(recv_futures);
```
pentschev, line 122: I know when things are going down the expected path this is not necessary, I would still suggest waiting for the `send_future`, the reason this is useful is in case something goes wrong in a previous test and a receive message is waiting to be received, which is possible because we do not destroy/restart workers after each test, in that case the you may be receiving an "old" message which will be hard to debug. Additionally, would be best to check that the status of the requests are `UCS_OK`, e.g., `recv_future->getStatus() == UCS_OK` to ensure the request succeeded.

cpp/tests/test_communicator.cpp
```diff
@@ -0,0 +1,136 @@
...
+            std::this_thread::sleep_for(std::chrono::microseconds(10));
+            continue;
+        }
+        // there should only be one finished future
+        auto recv_buf = comm->get_gpu_data(std::move(finished[0]));
+        EXPECT_EQ(n_elems, recv_buf->size);
+        std::vector<uint8_t> recv_data(n_elems);
+        copy_from_buffer(*recv_buf, recv_data.data(), n_elems);
+        EXPECT_EQ(data, recv_data);
+        log_vec("recv_data:", recv_data);
```
pentschev, line 134: Is this necessary/useful somehow?

pentschev (approved):
Thanks @nirandaperera for debugging and fixing this, if I'm understanding this correctly the issue only occurred because there was no endpoint to self before, is that right? Overall LGTM, left a few suggestions.

cpp/tests/test_communicator.cpp
pentschev: Yes, we should check for errors, however, I suggested the "old way", instead of `getStatus()` we whould `recv_future->checkError()`, which will raise an exception if an error occurred.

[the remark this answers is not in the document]
wence-: `test_some` calls `progress_worker` so I think that's ok

wence-: Did that in #403

cpp/tests/test_communicator.cpp
```diff
@@ -0,0 +1,136 @@
...
+    auto copy_from_buffer(Buffer& buf, void* dst, size_t size) {
+        if (memory_type() == MemoryType::HOST) {
+            std::memcpy(dst, buf.data(), size);
+        } else {
+            RAPIDSMPF_CUDA_TRY(
+                cudaMemcpyAsync(dst, buf.data(), size, cudaMemcpyDefault, stream)
+            );
+            RAPIDSMPF_CUDA_TRY(cudaStreamSynchronize(stream));
+        }
+    }
```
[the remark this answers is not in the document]
nirandaperera: We dont have utils to copy to/from arbitrary pointers in out buffer/ resource APIs, IINM.

cpp/tests/test_communicator.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include <rapidsmpf/buffer/resource.hpp>
+#include <rapidsmpf/communicator/communicator.hpp>
+
+#include "environment.hpp"
+#include "utils.hpp"
+
+using namespace rapidsmpf;
+
+// Base class for all communicator tests which
+class BaseCommunicatorTest : public cudf::test::BaseFixture {
```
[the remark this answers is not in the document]
nirandaperera: @wence- just to get the default mem resource :innocent: L27

nirandaperera:
@wence- This PR is on hold for the moment. There seems to be some bug in the UCX endpoint setup. That's why the cpp tests are failing. The newly added tests pass if we just run them independently. But when running the test suite (`ucxx_tests`) with higher parallelisms (3 or 5), there is some interference. :confused: 

I have already discussed with @pentschev. We are currently looking into it.

cpp/tests/test_communicator.cpp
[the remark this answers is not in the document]
wence-: Buffer to host (or device) vector:

```
auto host_vec = br->move_to_host_vector(std::move(buf), ...);
auto device_buffer = br->move_to_device_buffer(std::move(obuf), ...);
```

Vector to Buffer of given type (copies):
```
auto buf = br->move(mem_type, br->move(std::make_unique<decltype(vec)>(vec.begin(), vec.end()), stream, reservation);
```

`device_buffer` to Buffer (copies)

```
auto buf = br->move(mem_type, br->move(std::make_unique<rmm::device_buffer>(d_buf, stream), stream, reservation);
```

If we think we should have better utilities for doing this, let's figure out the right API and provide general purpose utils rather than hiding them in a test class.

cpp/src/communicator/ucxx.cpp
pentschev: Ah yeah, you're right, my bad.

PR #404, opened 2025-07-30
`PackedData` is memory-type agnostic

madsbk:
The gpu data of `PackedData` is now a `Buffer`, which can both host or device memory.

This means that we don't have to _unspill_ gpu data before inserting it into the shuffler. But for now, `Shuffler::extract()` unspill the buffer so consumers like `unpack_and_concat()` doesn't have to provide a memory reservation.

The shuffler and the partitioning functions how takes a `BufferResource` instead of a memory resource.

### DIFF
--- a/cpp/include/rapidsmpf/buffer/packed_data.hpp
+++ b/cpp/include/rapidsmpf/buffer/packed_data.hpp
@@ -10,6 +10,8 @@
 
 #include <rmm/device_buffer.hpp>
 
+#include <rapidsmpf/buffer/buffer.hpp>
+#include <rapidsmpf/buffer/resource.hpp>
 #include <rapidsmpf/error.hpp>
 
 namespace rapidsmpf {
@@ -22,39 +24,42 @@ namespace rapidsmpf {
  */
 struct PackedData {
     std::unique_ptr<std::vector<std::uint8_t>> metadata;  ///< The metadata
-    std::unique_ptr<rmm::device_buffer> gpu_data;  ///< The gpu data
+    std::unique_ptr<Buffer> gpu_data;  ///< The gpu data
 
-    /**
-     * @brief Construct packed data from metadata and gpu data, taking ownership.
-     *
-     * @param meta The metadata
-     * @param data The gpu data
-     */
     PackedData(
-        std::unique_ptr<std::vector<std::uint8_t>>&& meta,
-        std::unique_ptr<rmm::device_buffer>&& data
+        std::unique_ptr<std::vector<std::uint8_t>> metadata,
+        std::unique_ptr<Buffer> gpu_data
     )
-        : metadata{std::move(meta)}, gpu_data{std::move(data)} {
-        RAPIDSMPF_EXPECTS(
-            metadata != nullptr || gpu_data != nullptr,
-            "Metadata or GPU data must be non-null"
-        );
+        : metadata{std::move(metadata)}, gpu_data{std::move(gpu_data)} {}
 
-        RAPIDSMPF_EXPECTS(
-            (metadata->size() > 0 || gpu_data->size() == 0),
-            "Empty Metadata and non-empty GPU data is not allowed"
-        );
+    PackedData(
+        std::unique_ptr<std::vector<std::uint8_t>> metadata,
+        std::unique_ptr<rmm::device_buffer> gpu_data,
+        BufferResource* br,
+        rmm::cuda_stream_view stream,
+        std::shared_ptr<Buffer::Event> event = nullptr
+    )
+        : PackedData(std::move(metadata), br->move(std::move(gpu_data), stream, event)) {}
+
+    PackedData(
+        std::unique_ptr<std::vector<std::uint8_t>> metadata,
+        std::unique_ptr<rmm::device_buffer> gpu_data,
+        BufferResource* br
+    )
+        : metadata{std::move(metadata)} {
+        auto stream = gpu_data->stream();
+        this->gpu_data = br->move(std::move(gpu_data), stream);
     }
 
-    /**
-     * @brief Construct an empty PackedData object.
-     *
-     * This constructor initializes both the metadata and GPU data to empty
-     * buffers.
-     */
-    PackedData()
-        : metadata{std::make_unique<std::vector<std::uint8_t>>()},
-          gpu_data{std::make_unique<rmm::device_buffer>()} {}
+    // /**
+    //  * @brief Construct an empty PackedData object.
+    //  *
+    //  * This constructor initializes both the metadata and GPU data to empty
+    //  * buffers.
+    //  */
+    // PackedData()
+    //     : metadata{std::make_unique<std::vector<std::uint8_t>>()},
+    //       gpu_data{std::make_unique<rmm::device_buffer>()} {}
 
     ~PackedData() = default;
 
@@ -76,7 +81,7 @@ struct PackedData {
      * @return True if the packed data is empty, false otherwise.
      */
     [[nodiscard]] bool empty() const {
-        return metadata->empty() && gpu_data->size() == 0;
+        return metadata->empty() && gpu_data->size == 0;
     }
 };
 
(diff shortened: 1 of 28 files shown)


cpp/include/rapidsmpf/buffer/packed_data.hpp
```diff
@@ -22,39 +24,39 @@ namespace rapidsmpf {
  */
 struct PackedData {
     std::unique_ptr<std::vector<std::uint8_t>> metadata;  ///< The metadata
-    std::unique_ptr<rmm::device_buffer> gpu_data;  ///< The gpu data
+    std::unique_ptr<Buffer> gpu_data;  ///< The gpu data
```
wence-, line 27: *nit*: should we just call this `data` now? It might never have been on gpu.

cpp/include/rapidsmpf/buffer/packed_data.hpp
```diff
@@ -22,39 +24,39 @@ namespace rapidsmpf {
...
+    /// @brief Default constructor
+    PackedData() = default;
 
     /**
-     * @brief Construct packed data from metadata and gpu data, taking ownership.
+     * @brief Construct from metadata and gpu data.
```
wence-, line 32:
```suggestion
     * @brief Construct from metadata and gpu data, taking ownership.
```
This still takes ownership of the provided objects.

cpp/include/rapidsmpf/shuffler/shuffler.hpp
```diff
@@ -234,9 +234,7 @@ class Shuffler {
      * @param stream The CUDA stream for BufferResource memory operations.
      * @param event The event to use for the new chunk.
```
wence-, line 235: These parameters don't exist any more.
```suggestion
```

cpp/include/rapidsmpf/buffer/packed_data.hpp
```diff
@@ -22,39 +24,39 @@ namespace rapidsmpf {
...
-        : metadata{std::make_unique<std::vector<std::uint8_t>>()},
-          gpu_data{std::make_unique<rmm::device_buffer>()} {}
+    PackedData(
+        std::unique_ptr<std::vector<std::uint8_t>> metadata,
+        std::unique_ptr<rmm::device_buffer> gpu_data,
+        rmm::cuda_stream_view stream,
+        BufferResource* br,
+        std::shared_ptr<Buffer::Event> event = nullptr
+    )
+        : PackedData(std::move(metadata), br->move(std::move(gpu_data), stream, event)) {}
```
wence-, line 59: Since the device buffer is on GPU, is the idea with this move that we _might_ spill to host (hence the stream/event)?

cpp/src/integrations/cudf/partition.cpp
```diff
@@ -78,37 +83,45 @@ std::unordered_map<shuffler::PartID, PackedData> partition_and_pack(
...
     RAPIDSMPF_NVTX_FUNC_RANGE();
     RAPIDSMPF_MEMORY_PROFILE(statistics);
     std::unordered_map<shuffler::PartID, PackedData> ret;
-    auto packed = cudf::contiguous_split(table, splits, stream, mr);
+    auto packed = cudf::contiguous_split(table, splits, stream, br->device_mr());
     for (shuffler::PartID i = 0; static_cast<std::size_t>(i) < packed.size(); i++) {
         auto pack = std::move(packed[i].data);
-        ret.emplace(i, PackedData(std::move(pack.metadata), std::move(pack.gpu_data)));
+        ret.emplace(
+            i, PackedData(std::move(pack.metadata), std::move(pack.gpu_data), stream, br)
```
wence-, line 115: Previously we would never spill at this point, now there is an opportunity to do so?

cpp/include/rapidsmpf/buffer/packed_data.hpp
```diff
@@ -74,9 +76,16 @@ struct PackedData {
...
+     * @throw std::invalid_argument if metadata and gpu_data is null and non-null.
      */
     [[nodiscard]] bool empty() const {
-        return metadata->empty() && gpu_data->size() == 0;
+        RAPIDSMPF_EXPECTS(
+            (!metadata) == (!gpu_data),
+            "the metadata and gpu_data pointers cannot be null and non-null",
+            std::invalid_argument
+        );
+        return metadata->empty() && gpu_data->size == 0;
```
wence-, line 88: *question*: Can a packed data ever have both `metadata` and `gpu_data` as nullptr?

cpp/tests/test_shuffler.cpp
```diff
@@ -20,6 +20,7 @@
 #include <rapidsmpf/utils.hpp>
 
 #include "environment.hpp"
+#include "rapidsmpf/buffer/buffer.hpp"
```
wence-, line 23:
```suggestion
#include <rapidsmpf/buffer/buffer.hpp>
```

And then move it up.

cpp/tests/test_chunk.cpp
```diff
@@ -40,15 +40,16 @@ namespace {
...
+    BufferResource* br
 ) {
     auto metadata_ptr =
         std::make_unique<std::vector<uint8_t>>(metadata.begin(), metadata.end());
     auto data_ptr = std::make_unique<rmm::device_buffer>(data.size(), stream);
     RAPIDSMPF_CUDA_TRY(
         cudaMemcpy(data_ptr->data(), data.data(), data.size(), cudaMemcpyHostToDevice)
     );
-    return PackedData{std::move(metadata_ptr), std::move(data_ptr)};
+    return PackedData{std::move(metadata_ptr), std::move(data_ptr), stream, br};
```
wence-, line 52: I _think_ I would prefer if the `PackedData` ctor were trivial and the conversion into `Buffer` (from `device_buffer`)  was explicit. i.e. here:

```
auto buf = br->move_from_device_buffer(data_ptr, stream);r

return PackedData{std::move(metadata), std::move(buf)};
```
WDYT?

cpp/src/buffer/resource.cpp
```diff
@@ -141,6 +141,17 @@ std::unique_ptr<rmm::device_buffer> BufferResource::move_to_device_buffer(
...
 }
 
+std::unique_ptr<rmm::device_buffer> BufferResource::move_to_device_buffer(
+    std::unique_ptr<Buffer> buffer
+) {
+    RAPIDSMPF_EXPECTS(
+        buffer->mem_type() == MemoryType::DEVICE,
+        "cannot move between memory types without a reservation",
+        std::invalid_argument
+    );
```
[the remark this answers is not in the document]
madsbk: Good point, removed

cpp/tests/test_chunk.cpp
madsbk: good idea, done

cpp/tests/test_shuffler.cpp
madsbk: done

cpp/include/rapidsmpf/buffer/packed_data.hpp
madsbk: I still like to use `gpu_data`. It might not be located on device, but it is still gpu data?

cpp/include/rapidsmpf/buffer/packed_data.hpp
```diff
@@ -22,39 +24,39 @@ namespace rapidsmpf {
...
-        : metadata{std::move(meta)}, gpu_data{std::move(data)} {
-        RAPIDSMPF_EXPECTS(
-            metadata != nullptr || gpu_data != nullptr,
-            "Metadata or GPU data must be non-null"
-        );
-
-        RAPIDSMPF_EXPECTS(
-            (metadata->size() > 0 || gpu_data->size() == 0),
-            "Empty Metadata and non-empty GPU data is not allowed"
-        );
```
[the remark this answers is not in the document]
madsbk: done

madsbk:
Thanks @wence- and @nirandaperera, it is ready for another review round

cpp/src/integrations/cudf/partition.cpp
wence-: Ah ok, it's only `br->move`s with a memory reservation that might spill?

madsbk: yes!

cpp/include/rapidsmpf/buffer/packed_data.hpp
madsbk: Good point, done

wence- (approved):

madsbk:
Thanks @wence-  @nirandaperera

PR #408, opened 2025-07-31
Remove `RegisterRank` control op from UCXX communicator

pentschev:
The `RegisterRank` control operation was initially added as an optimization, where a UCX endpoint handle (in the listener callback) would be associated to the rank before the UCXX endpoint was made available and properly registered. In practice that never occurs, `associate_endpoint_rank` only completes after `get_endpoint(rank, ucxx_endpoint)` is called on both processes that communicate with each other, and that would cause `rank_to_endpoint_`. Overwriting `rank_to_endpoint_` per se wouldn't be a problem, but there was also an error in the `RegisterRank` message where the remote rank would be sent to the remote endpoint (`rank`) instead of the local endpoint (`shared_resources_->rank()`).

The number of endpoints being created before and after changes contained here remain the same, in tests exactly `N^2` endpoints are created, for example running `4` ranks incurs in exactly `16` endpoints being created, therefore in practice this change is a no-op for existing use cases. The case where the existing implementation is problematic only happens when a rank communicates to self which was previously unused/unsupported.

Ultimately, this change also makes the implementation simpler and easier to understand.

### DIFF
--- a/cpp/src/communicator/ucxx.cpp
+++ b/cpp/src/communicator/ucxx.cpp
@@ -219,8 +219,14 @@ class SharedResources {
     void register_listener(std::shared_ptr<::ucxx::Listener> listener) {
         std::lock_guard<std::mutex> lock(listener_mutex_);
         auto worker = std::dynamic_pointer_cast<::ucxx::Worker>(listener->getParent());
-        rank_to_listener_address_[rank_] =
-            ListenerAddress{.address = worker->getAddress(), .rank = rank_};
+        RAPIDSMPF_EXPECTS(
+            rank_to_listener_address_
+                .emplace(
+                    rank_, ListenerAddress{.address = worker->getAddress(), .rank = rank_}
+                )
+                .second,
+            "listener for given rank already exists"
+        );
         listener_ = std::move(listener);
     }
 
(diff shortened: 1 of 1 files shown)


cpp/src/communicator/ucxx.cpp
```diff
@@ -241,26 +239,12 @@ class SharedResources {
...
-     *
-     * Associate a previously registered endpoint to a specific rank.
-     *
-     * @param rank The rank to register the endpoint for.
-     * @param endpoint_handle The handle of the endpoint to register.
-     */
-    void associate_endpoint_rank(Rank const rank, ucp_ep_h const endpoint_handle) {
-        std::lock_guard<std::mutex> lock(endpoints_mutex_);
-        rank_to_endpoint_[rank] = endpoints_[endpoint_handle];
+        endpoints_[endpoint->getHandle()] = endpoint;
```
nirandaperera, line 247: Shall we add some overwrite protection here? 
```suggestion
       RAPIDSMPF_EXPECTS(endpoints_.emplace(endpoint->getHandle(), std::move(endpoint)), "endpoint handle already exists");
```

nirandaperera: same goes to L234-235. I feel like it'll make things fail fast if some things behave unexpectedly.

nirandaperera (changes requested):

cpp/src/communicator/ucxx.cpp
pentschev: Done in https://github.com/rapidsai/rapidsmpf/pull/408/commits/f02fa1881904eaf293cb50370bbbf96826cf847a

nirandaperera (approved):
LGTM

PR #410, opened 2025-08-01
Remove `mem_type` from some ops in Buffer API

### LINKED ISSUE #280: Inconsistency with multiple buffer resources
There is an inconsistency when using multiple buffer resources in the same application. 

- Say we have two buffer resources, `r1` and `r2`. 
- We allocate and a buffer, `b1` using `r1` , and populate some data. 
- Say we want to make a copy of `b1` using an `r2` allocation. 
- We make a reservation on `r2`, and pass it to `r2.copy` method alongside `b1` buffer. 
- But now, we call `b1.copy` which does not accept a buffer resource arg. So, it allocates memory from its own  buffer resource, which is `ra`.
- So, we will be allocating on `ra` without reserving and not using the reservation on `rb`.

nirandaperera:
In the `BufferResource` utils such as `copy`, we accept a `mem_type` and a `reservation`. I think this is redundant and error prone. 
One could provide a conflicting mem_type and reservation, device/ host - We throw when this happens. But IMO this is unnecessary, because allocation is already made for the required mem_type. 
Also we still have the inconsistency reported in #280. 

So, this PR removes the redundant mem_type, and forces buffer utils to use reservations to make new allocations and removes buffer resource/ mem type args. This also this fixes #280

~~Also this PR makes buffer resource `move*` methods static, as `Buffer` no longer requires `br*`~~

### DIFF
--- a/cpp/src/buffer/resource.cpp
+++ b/cpp/src/buffer/resource.cpp
@@ -128,13 +128,13 @@ std::unique_ptr<rmm::device_buffer> BufferResource::move_to_device_buffer(
         "the memory type of MemoryReservation doesn't match",
         std::invalid_argument
     );
-    return move(std::move(buffer), stream, reservation)->device();
+    return move(std::move(buffer), stream, reservation)->release_device();
 }
 
 std::unique_ptr<rmm::device_buffer> BufferResource::move_to_device_buffer(
     std::unique_ptr<Buffer> buffer
 ) {
-    return std::move(buffer->device());
+    return buffer->release_device();
 }
 
 std::unique_ptr<std::vector<uint8_t>> BufferResource::move_to_host_vector(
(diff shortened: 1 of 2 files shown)


cpp/src/buffer/resource.cpp
```diff
@@ -136,9 +123,12 @@ std::unique_ptr<rmm::device_buffer> BufferResource::move_to_device_buffer(
...
     MemoryReservation& reservation
 ) {
-    return std::move(
-        move(MemoryType::DEVICE, std::move(buffer), stream, reservation)->device()
+    RAPIDSMPF_EXPECTS(
+        reservation.mem_type_ == MemoryType::DEVICE,
+        "the memory type of MemoryReservation doesn't match",
+        std::invalid_argument
     );
+    return std::move(move(std::move(buffer), stream, reservation)->device());
```
madsbk, line 131:
```suggestion
    return move(std::move(buffer), stream, reservation)->device();
```

cpp/src/buffer/resource.cpp
```diff
@@ -152,9 +142,12 @@ std::unique_ptr<std::vector<uint8_t>> BufferResource::move_to_host_vector(
...
     MemoryReservation& reservation
 ) {
-    return std::move(
-        move(MemoryType::HOST, std::move(buffer), stream, reservation)->host()
+    RAPIDSMPF_EXPECTS(
+        reservation.mem_type_ == MemoryType::HOST,
+        "the memory type of MemoryReservation doesn't match",
+        std::invalid_argument
     );
+    return std::move(move(std::move(buffer), stream, reservation)->host());
```
madsbk, line 150:
```suggestion
    return move(std::move(buffer), stream, reservation)->host();
```

madsbk (commented):
Overall looks good but I think it is best to keep the buffer resource methods non-static. In the future, I think the buffer resource might want to track the copy/move operations.

cpp/src/buffer/resource.cpp
nirandaperera: This move is required because  `..->device()` is a ref, but I agree its clunky. I added a separate `release_device()` method to make it bit more "less clunky" :D

PR #413, opened 2025-08-05
Shuffle.extract(): don't unspill extracted partitions

madsbk:
Instead of implicit unspilling, use `unspill_partitions()`.


madsbk:
Thanks @nirandaperera

Issue #421, opened 2025-08-08
Segmentation fault from `Statistics.get_stat` with invalid statistic name


TomAugspurger:
```python

from rapidsmpf.statistics import Statistics
stats = Statistics(enable=True)
stats.get_stat('segfault')
```

segfaults with

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at
Aborted (core dumped)
```

PR #423, opened 2025-08-08
Fixed exception handling in Statistics.get_stat

TomAugspurger:
This fixes the segfault during `Statistics.get_stat`. Previously, these methods were marked as `nogil`, but we need to catch the C++ `std::out_of_range` so now they are `except +`.

Now we'll raise a `KeyError` instead of segfaulting.

Also fixes an issue with the type on `cpp_get_statistic_value`, which should be a double rather than a size_t.

Closes https://github.com/rapidsai/rapidsmpf/issues/421

### DIFF
--- a/python/rapidsmpf/rapidsmpf/statistics.pyx
+++ b/python/rapidsmpf/rapidsmpf/statistics.pyx
@@ -23,14 +23,14 @@ cdef extern from *:
     ) {
         return stats.get_stat(name).count();
     }
-    std::size_t cpp_get_statistic_value(
+    double cpp_get_statistic_value(
         rapidsmpf::Statistics const& stats, std::string const& name
     ) {
         return stats.get_stat(name).value();
     }
     """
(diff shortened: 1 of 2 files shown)


python/rapidsmpf/rapidsmpf/statistics.pyx
```diff
@@ -23,14 +23,14 @@ cdef extern from *:
...
+    double cpp_get_statistic_value(
         rapidsmpf::Statistics const& stats, std::string const& name
     ) {
         return stats.get_stat(name).value();
     }
     """
-    size_t cpp_get_statistic_count(cpp_Statistics stats, string name) nogil
-    double cpp_get_statistic_value(cpp_Statistics stats, string name) nogil
+    size_t cpp_get_statistic_count(cpp_Statistics stats, string name) except +
+    double cpp_get_statistic_value(cpp_Statistics stats, string name) except +
```
wence-, line 33: We do still need to drop the gil because accessing the statistics acquires a lock, so this can now deadlock due to lock ordering.

PR #424, opened 2025-08-08
[MINOR] Add proper shuffler shutdown for c++ tests 

nirandaperera:
This PR adds proper shuffler shutdown order in shuffler tests.

### DIFF
--- a/cpp/tests/test_shuffler.cpp
+++ b/cpp/tests/test_shuffler.cpp
@@ -356,6 +356,8 @@ class ConcurrentShuffleTest
             stream,
             br.get()
         ));
+
+        shuffler.shutdown();
     }
 
     template <typename InsertFn, typename InsertFinishedFn>
@@ -365,17 +367,19 @@ class ConcurrentShuffleTest
 
         for (int t_id = 0; t_id < num_shufflers; t_id++) {
             // pass a copy of the insert_fn and insert_finished_fn to each thread
-            futures.push_back(std::async(
-                std::launch::async,
-                [this,
-                 t_id,
-                 insert_fn1 = insert_fn,
-                 insert_finished_fn1 = insert_finished_fn] {
-                    ASSERT_NO_FATAL_FAILURE(this->RunTest(
-                        t_id, std::move(insert_fn1), std::move(insert_finished_fn1)
-                    ));
-                }
-            ));
+            futures.push_back(
+                std::async(
+                    std::launch::async,
+                    [this,
+                     t_id,
+                     insert_fn1 = insert_fn,
+                     insert_finished_fn1 = insert_finished_fn] {
+                        ASSERT_NO_FATAL_FAILURE(this->RunTest(
+                            t_id, std::move(insert_fn1), std::move(insert_finished_fn1)
+                        ));
+                    }
+                )
+            );
         }
 
         for (auto& f : futures) {
@@ -475,9 +479,9 @@ class ShuffleInsertGroupedTest
 
         stream = cudf::get_default_stream();
 
-        progress_thread =
-            std::make_shared<rapidsmpf::ProgressThread>(GlobalEnvironment->comm_->logger()
-            );
+        progress_thread = std::make_shared<rapidsmpf::ProgressThread>(
+            GlobalEnvironment->comm_->logger()
+        );
 
         GlobalEnvironment->barrier();
     }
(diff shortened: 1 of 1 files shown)


cpp/tests/test_shuffler.cpp
```diff
@@ -364,19 +365,17 @@ class ConcurrentShuffleTest
...
+                std::launch::async,
+                [this,
+                 t_id,
+                 insert_fn1 = insert_fn,
+                 insert_finished_fn1 = insert_finished_fn] {
+                    ASSERT_NO_FATAL_FAILURE(this->RunTest(
+                        t_id, std::move(insert_fn1), std::move(insert_finished_fn1)
+                    ));
+                }
+            ));
```
pentschev, line 378: Are all these formatting changes coming from your IDE and not pre-commit? Are they necessary?

nirandaperera: sorry. No it was coming from the IDE. I think my clangformat in the cursor is different from the precommit params. I'll check that

cpp/tests/test_shuffler.cpp
```diff
@@ -173,8 +173,6 @@ void test_shuffler(
             sort_table(result), sort_table(expect_partitions[finished_partition])
         );
     }
-
-    shuffler.shutdown();
```
pentschev, line 177: IIRC, this was very much needed to ensure proper synchronization, maybe we'll need at least a `barrier()`, but let's see.

cpp/tests/test_shuffler.cpp
```diff
@@ -355,6 +356,8 @@ class ConcurrentShuffleTest
             stream,
             br.get()
         ));
+
+        shuffler.shutdown();
```
pentschev, line 360: Not sure I follow, why did `shutdown()` was removed from `test_shuffler` to be added here?

cpp/tests/test_shuffler.cpp
```diff
@@ -484,22 +487,13 @@ class ShuffleInsertGroupedTest
...
-        // function is marked as done. This is important to ensure that the test does
-        // not hang.
-        progress_thread->resume();
-
-        if (shuffler) {
-            shuffler->shutdown();
-        }
-
-        progress_thread->stop();
+        progress_thread = nullptr;
```
pentschev, line 490: I think the `progress_thread` will not stop if `shuffler` isn't stopped, will it?

cpp/tests/test_shuffler.cpp
nirandaperera: Shuffler is shutdown where it was created or if its a class member, during the `TearDown` method.

cpp/tests/test_shuffler.cpp
nirandaperera: Yes, shuffler is stopped at the end of the test run (inside the test case)

cpp/tests/test_shuffler.cpp
```diff
@@ -666,6 +661,9 @@ TEST_P(ShuffleInsertGroupedTest, InsertPackedDataNoHeadroom) {
     // resume progress thread - this will guarantee that shuffler progress function is
     // marked as done. This is important to ensure that the test does not hang.
     progress_thread->resume();
+
+    shuffler.reset();
+    br.reset();
```
wence-, line 666: This must be wrong, otherwise our RAII destructors are broken. What is this for?

nirandaperera: > This must be wrong, otherwise our RAII destructors are broken. 

@wence- Sorry, could you clarify this comment?

> What is this for?

Previously `br` and `shuffler` were class members, and their destruction happened during the test object destruction. Since we don't create the objects during `Setup`, I thought its best to move the creation and destruction inside the test impl.

wence- (changes requested):

cpp/tests/test_shuffler.cpp
nirandaperera: I understand. This was just to explicitly indicate that shuffler is shutdown. It's more verbose and easy to follow. 

But its fine. I'll remove this line.

cpp/tests/test_shuffler.cpp
nirandaperera: @madsbk I understand. I was trying to be more explicit. But that's fine, I'll remove these lines.

PR #426, opened 2025-08-11
fix `RAPIDSMPF_BUILD_EXAMPLES`

madsbk:
Fix the `RAPIDSMPF_BUILD_EXAMPLES` option, which we were ignoring. 
Also some minor clean up.


bdice:
For a bit of context on the above -- typically we don't run `clang-tidy` as a pre-commit hook because it usually requires building the project (or a pre-built project, with `compile_commands.json`, or similar).

What we do instead with RAPIDS projects is to run a separate CI job for clang-tidy. Here are the relevant files/configs for cudf. I believe we have a similar setup for cuml and maybe other repos too.

- https://github.com/rapidsai/cudf/blob/314dcbc72feda715830815a74a0ad3f91ae493e1/.github/workflows/pr.yaml#L145-L152
- https://github.com/rapidsai/cudf/blob/314dcbc72feda715830815a74a0ad3f91ae493e1/ci/cpp_linters.sh
- https://github.com/rapidsai/cudf/blob/314dcbc72feda715830815a74a0ad3f91ae493e1/cpp/CMakeLists.txt#L90-L92
- https://github.com/rapidsai/cudf/blob/314dcbc72feda715830815a74a0ad3f91ae493e1/cpp/.clang-tidy

PR #427, opened 2025-08-11
Prevent installing OpenMPI with broken CUDA

pentschev:
Current OpenMPI package in conda-forge has broken CUDA support, see https://github.com/conda-forge/openmpi-feedstock/issues/203. Prevent installing the current 5.0.8 version until that's fixed.


pentschev:
Thanks @KyleFromNVIDIA and @wence- .

PR #428, opened 2025-08-11
Drop gil when getting statistics

wence-:
Although these functions may throw, we must still drop the gil because they acquire a lock internally. If we don't then two threads attempting to access the statistics at the same time might deadlock.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/statistics.pyx
+++ b/python/rapidsmpf/rapidsmpf/statistics.pyx
@@ -108,8 +108,9 @@ cdef class Statistics:
         cdef size_t count
         cdef double value
         try:
-            count = cpp_get_statistic_count(deref(self._handle), name_)
-            value = cpp_get_statistic_value(deref(self._handle), name_)
+            with nogil:
+                count = cpp_get_statistic_count(deref(self._handle), name_)
+                value = cpp_get_statistic_value(deref(self._handle), name_)
         except IndexError:
             # The C++ implementation throws a std::out_of_range exception
             # which we / Cython translate to a KeyError.
--- a/python/rapidsmpf/rapidsmpf/statistics.pyx
+++ b/python/rapidsmpf/rapidsmpf/statistics.pyx
@@ -29,8 +29,8 @@ cdef extern from *:
         return stats.get_stat(name).value();
     }
     """
-    size_t cpp_get_statistic_count(cpp_Statistics stats, string name) nogil except +
-    double cpp_get_statistic_value(cpp_Statistics stats, string name) nogil except +
+    size_t cpp_get_statistic_count(cpp_Statistics stats, string name) except + nogil
+    double cpp_get_statistic_value(cpp_Statistics stats, string name) except + nogil
 
 cdef class Statistics:
     """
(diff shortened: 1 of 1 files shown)


wence-:
cc @TomAugspurger

TomAugspurger:
Does the [`with nogil`](https://github.com/rapidsai/rapidsmpf/pull/423/files#diff-9e899bc69b1c500d4a7be9a72b132ad7908266665c1de10d0817a93ca575557aL110) I removed here matter? Or is it just the C++ bit that being marked `nogil` avoids the deadlock?

pentschev:
> Does the [`with nogil`](https://github.com/rapidsai/rapidsmpf/pull/423/files#diff-9e899bc69b1c500d4a7be9a72b132ad7908266665c1de10d0817a93ca575557aL110) I removed here matter? Or is it just the C++ bit that being marked `nogil` avoids the deadlock?

It _could_ matter. I don't know if it's the case that those calls specifically need a mutex currently, but in theory nothing prevents them from holding a mutex (now or in the future) while some other code is already holding the mutex and attempting to acquire the GIL simultaneously, which could lead to a deadlock. In general, I'd argue it's better to make the code future-proof as well, it's unlikely that we'll remember to check if the GIL is indeed being dropped when we change C++-only code that will be modified to hold a mutex.

madsbk (changes requested):
AFAIK, declaring them `nogil` doesn't release the GIL!

We also need to wrap the calls in a `with nogil:`, which will still catch exceptions.

python/rapidsmpf/rapidsmpf/statistics.pyx
```diff
@@ -29,8 +29,8 @@ cdef extern from *:
         return stats.get_stat(name).value();
     }
     """
-    size_t cpp_get_statistic_count(cpp_Statistics stats, string name) except +
-    double cpp_get_statistic_value(cpp_Statistics stats, string name) except +
+    size_t cpp_get_statistic_count(cpp_Statistics stats, string name) nogil except +
+    double cpp_get_statistic_value(cpp_Statistics stats, string name) nogil except +
```
madsbk, line 33:
```suggestion
    size_t cpp_get_statistic_count(cpp_Statistics stats, string name) except+ nogil
    double cpp_get_statistic_value(cpp_Statistics stats, string name) except+ nogil
```

`nogil` goes last apparently :)
```
  warning: /home/mkristensen/repos/rapidsmpf/python/rapidsmpf/rapidsmpf/statistics.pyx:32:84: The keyword 'nogil' should appear at the end of the function signature line. Placing it before 'except' or 'noexcept' will be disallowed in a future version of Cython.
  warning: /home/mkristensen/repos/rapidsmpf/python/rapidsmpf/rapidsmpf/statistics.pyx:33:84: The keyword 'nogil' should appear at the end of the function signature line. Placing it before 'except' or 'noexcept' will be disallowed in a future version of Cython.
```

wence-:
> AFAIK, declaring them `nogil` doesn't release the GIL!
> 
> We also need to wrap the calls in a `with nogil:`, which will still catch exceptions.

Thanks yes, this is true. Will update

python/rapidsmpf/rapidsmpf/statistics.pyx
wence-: Done.

PR #429, opened 2025-08-11
Use build cluster in devcontainers

trxcllnt:
RAPIDS has deployed an autoscaling cloud build cluster that can be used to accelerate building large RAPIDS projects. This contributes to https://github.com/rapidsai/build-planning/issues/209.


.devcontainer/cuda12.9-conda/devcontainer.json
```diff
@@ -11,7 +11,9 @@
   "runArgs": [
     "--rm",
     "--name",
-    "${localEnv:USER:anon}-rapids-${localWorkspaceFolderBasename}-25.10-cuda12.9-conda"
+    "${localEnv:USER:anon}-rapids-${localWorkspaceFolderBasename}-25.10-cuda12.9-conda",
+    "--ulimit",
+    "nofile=500000"
```
pentschev, line 16: It might be worth replicating options we use in CI: https://github.com/rapidsai/rapidsmpf/blob/2de638653ebaf4882e9e1f4d9a1d5eaa571b80f4/.github/workflows/pr.yaml#L98, those are primarily important for UCX.

trxcllnt:
Looks like there's a link error with UCX v1.19:
```
/usr/bin/ld: /usr/lib/libucp.so: undefined reference to `uct_rkey_unpack_v2'
collect2: error: ld returned 1 exit status
```

Is this expected? Should I downgrade the CUDA 12.9 builds to use UCX v1.18 again?

pentschev:
> Looks like there's a link error with UCX v1.19:
> 
> ```
> /usr/bin/ld: /usr/lib/libucp.so: undefined reference to `uct_rkey_unpack_v2'
> collect2: error: ld returned 1 exit status
> ```
> 
> Is this expected? Should I downgrade the CUDA 12.9 builds to use UCX v1.18 again?

I wasn't aware of that, is it rebuilding or just relinking with v1.19?

pentschev:
> I opened [this PR](https://github.com/rapidsai/devcontainers/pull/559) to downgrade CUDA 12.9 images to UCX 1.18 if necessary, but before I do that I want to make sure it's the right thing for rapidsmpf.

No, UCX 1.19 should work, the fact that it is failing points to something going wrong. I just checked the conda-forge build of UCX 1.19 and everything seems correct, I don't get any linking errors and the symbols exist:

```
$ nm /opt/miniforge3/envs/ucx-1.19-test/lib/libucp.so | grep _v2
00000000000ad120 T ucp_wireup_ep_connect_to_ep_v2
00000000000b1c30 T ucp_wireup_iface_lat_distance_v2
                 U uct_ep_connect_to_ep_v2
                 U uct_iface_is_reachable_v2
                 U uct_md_mem_dereg_v2
                 U uct_md_mem_reg_v2
                 U uct_md_mkey_pack_v2
                 U uct_md_query_v2
                 U uct_rkey_unpack_v2
$ nm /opt/miniforge3/envs/ucx-1.19-test/lib/libuct.so | grep _v2
0000000000019000 T uct_ep_connect_to_ep_v2
00000000000186d0 T uct_iface_is_reachable_v2
0000000000015130 t uct_md_attr_v2_init
00000000000162d0 T uct_md_mem_dereg_v2
00000000000161a0 T uct_md_mem_reg_v2
00000000000159a0 T uct_md_mkey_pack_v2
0000000000015c00 T uct_md_query_v2
000000000001b300 t uct_mm_iface_is_reachable_v2
0000000000015a10 T uct_rkey_unpack_v2
0000000000020160 t uct_self_iface_is_reachable_v2
00000000000224c0 T uct_tcp_ep_connect_to_ep_v2
0000000000025370 t uct_tcp_iface_is_reachable_v2
```

It looks to me like the devcontainer build is not finding or finding the wrong `libuct.so`.

pentschev:
> You're right, from the build log it appears we're installing the UCX v1.19.0 debs from the `ucx-1.19.0-ubuntu24.04-mofed5-cuda12-x86_64.tar.bz2` [github release](https://github.com/openucx/ucx/releases/tag/v1.19.0) artifact. I'll check if libuct.so is getting overridden by another step somehow.

Is that intentional? Those are prebuilt binaries, and TBH I've never tested them, although I believe they should generally work if you have a matching OS.

trxcllnt:
I see the symbol in the library in the image:
```shell
$ docker run --rm -it rapidsai/devcontainers:25.10-cpp-cuda12.9-ucx1.19.0-openmpi5.0.7 sh -c 'nm -D $(realpath -m /lib/libuct.so) | grep uct_rkey_unpack_v2'
0000000000016a20 T uct_rkey_unpack_v2
```

So I'm not sure why it's failing to link here.

trxcllnt:
@pentschev I see the issue. The linker is looking for the symbol in libuc**p**.so, not libuc**t**.so:
```
/usr/bin/ld: /usr/lib/libucp.so: undefined reference to `uct_rkey_unpack_v2'
```

And indeed, that symbol is missing in `libucp.so`:
```shell
$ docker run --rm -it rapidsai/devcontainers:25.10-cpp-cuda12.9-ucx1.19.0-openmpi5.0.7 sh -c 'nm -D $(realpath -m /lib/libucp.so) | grep uct_rkey_unpack_v2'
                 U uct_rkey_unpack_v2
```

This is also true in the conda version:
```
$ nm /opt/miniforge3/envs/ucx-1.19-test/lib/libucp.so | grep _v2
...
                 U uct_rkey_unpack_v2
```

So is it possible that the `benchmarks/bench_shuffle` target is linked to the wrong lib, and conda just adds both to its implicit link libraries?

pentschev:
> I _think_ the issue is the optional `.symtab` is missing? It seems we have to pass `nm -D` to check the dynamic symbol table:
> 
> ```
> $ docker run --rm -it rapidsai/devcontainers:25.10-cpp-cuda12.9-ucx1.19.0-openmpi5.0.7 \
>   sh -c 'nm $(realpath -m /lib/libucp.so) | grep uct_rkey_unpack_v2'
> nm: /usr/lib/libucp.so.0.0.0: no symbols
> ```

I don't know how that implicates in the issue we're seeing, isn't `.symtab` just for debugging/inspection?

trxcllnt:
Yeah that was just a guess. I pinged @robertmaynard about it. Maybe we're missing a link flag that conda automatically sets?

pentschev:
> Yeah that was just a guess. I pinged @robertmaynard about it. Maybe we're missing a link flag that conda automatically sets?

I think not, AFAIU, as long as the symbol is defined and the library that defines it is dynamically linked it should be found. The fact this is failing is very strange given we have confirmed all those conditions are indeed being satisfied. Let's see what Robert can tell us about this.

trxcllnt:
So `readelf -a /lib/libucp.so` lists the libraries as a `NEEDED` entry, but doesn't have any RPATH entries. Robert mentioned this sounds like `--no-allow-shlib-undefined` biting us, because it does a recursive walk over all the dependencies checking for missing symbols at link time. For some reason `ldd` is fine with just the NEEDED entries, but `ld` needs the RPATH entries.

Adding `LINK_FLAGS "-Wl,--allow-shlib-undefined"` to the benchmark, test, and example targets fixes the link issues, so I'll push that up now.

PR #435, opened 2025-08-13
Set default memory type to pool

pentschev:
Clarify warning when running benchmarks with `-m cuda`, this will cause a "memory leak" when UCX is running with CUDA IPC. This is a known limitation and there's no ETA on being addressed. This happens because UCX needs to map memory allocations with `cuIpcOpenMemHandle`, but it does not support `cuIpcCloseMemHandle`, and thus causing allocations that have been used to transfer over CUDA IPC to never release memory, even after `cudaFree`/`cuMemFree` is called.

Due to the limitation described above, it's probably a better and more sane to make `"pool"` allocator the default, which doesn't require remapping given allocations are never freed until the application terminates.

### DIFF
--- a/ci/run_cpp_benchmark_smoketests.sh
+++ b/ci/run_cpp_benchmark_smoketests.sh
@@ -15,5 +15,3 @@ export OMPI_MCA_opal_cuda_support=1  # enable CUDA support in OpenMPI
 # Ensure that benchmarks are runnable
 mpirun --map-by node --bind-to none -np 3 ./bench_shuffle -m cuda
 mpirun --map-by node --bind-to none -np 3 ./bench_comm -m cuda
-mpirun --map-by node --bind-to none -np 3 ./bench_shuffle -m async
-mpirun --map-by node --bind-to none -np 3 ./bench_comm -m async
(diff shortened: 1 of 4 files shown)


nirandaperera (approved):
LGTM.

pentschev:
Unfortunately I realized this won't work the way tests are currently written. In CI we have a single-GPU and [we then oversubscribe that GPU to run 3 processes](https://github.com/rapidsai/rapidsmpf/blob/26ee2f615c631176749376487bdb031e9deeed49/ci/run_cpp_benchmark_smoketests.sh#L16-L17), this sometimes works, but in general it isn't safe. By default [pool pre-allocates 80% of the GPU memory](https://github.com/rapidsai/rapidsmpf/blob/26ee2f615c631176749376487bdb031e9deeed49/cpp/benchmarks/utils/rmm_stack.hpp#L41), even if not used (like in the smoke tests).

The alternatives I see are:
1. Reduce the minimum pool to something much smaller (e.g., 10%)
  i. Pros: allows to oversubscribe multiple processes on a single-GPU, provided the data is small enough
  ii. Cons: may cause suboptimal performance in non-oversubscribed scenarios, since multiple buffers may be required to form the pool, additionally it may cause more fragmentation or make it impossible to test _very large_ buffer sizes
2. Switch the default to `"async"` allocator
  i. Pros: closer to most (but not necessarily all) of the use cases we're expecting to see, since that reduces fragmentation
  ii. Cons: may present slower performance with the new default allocator, given the need to use staging buffers for transports like CUDA IPC
3. Test with a different allocator than the default
  i. Pros: no need to change the current way we test
  ii. Cons: we won't test the default anymore

WDYT @madsbk @nirandaperera ? Do you have any other ideas or suggestions?

madsbk:
I vote on option **3.**

PR #438, opened 2025-08-15
Properly track shuffle device memory allocations

nirandaperera:
Currently when performing shuffle partitioning, splitting, or during data generation, we pass simply pass the device memory resource to cudf utils. Then we loose track of those allocations in the memory resource counters. This causes OOM issues in memory constrained environments (ex: running bench_shuffle with limited memory). 

This PR adds memory reservations prior to calling cudf utils that perform deep copies. 

NOTE: 
I'm suspecting this change might cause some performance degradation, but it should improve stability.

### DIFF
--- a/cpp/include/rapidsmpf/buffer/resource.hpp
+++ b/cpp/include/rapidsmpf/buffer/resource.hpp
@@ -540,7 +540,7 @@ auto with_memory_reservation(MemoryReservation&& reservation, auto&& f) {
         return std::forward<F>(f)();
     } else {
         static_assert(
-            [] { return false; }(),
+            std::integral_constant<F, false>(),
             "f must be callable with no args or MemoryReservation&"
         );
     }
--- a/cpp/include/rapidsmpf/buffer/resource.hpp
+++ b/cpp/include/rapidsmpf/buffer/resource.hpp
@@ -487,7 +487,8 @@ MemoryReservation reserve_or_fail(
  * memory is freed to satisfy the reservation; otherwise, allows overbooking even
  * if spilling was insufficient.
  * @return The memory reservation.
- * @throw std::overflow_error if the buffer resource cannot reserve enough memory
+ * @throw std::overflow_error if allow_overbooking is false and the buffer resource
+ * cannot reserve and spill enough memory.
  */
 MemoryReservation reserve_device_memory_and_spill(
     BufferResource* br, size_t size, bool allow_overbooking
@@ -511,7 +512,8 @@ auto with_memory_reservation(MemoryReservation&& reservation, auto&& f) {
         return std::forward<F>(f)();
     } else {
         static_assert(
-            [] { return false; }(), "f must be callable with either 0 or 1 argument"
+            [] { return false; }(),
+            "f must be callable with no args or MemoryReservation&"
         );
     }
 }
--- a/cpp/include/rapidsmpf/buffer/resource.hpp
+++ b/cpp/include/rapidsmpf/buffer/resource.hpp
@@ -38,6 +38,11 @@ class MemoryReservation {
     ~MemoryReservation() noexcept;
 
     /**
+     * @brief Clear the remaining size of the reservation.
+     */
+    void clear() noexcept;
+
+    /**
      * @brief Move constructor for MemoryReservation.
      *
      * @param o The memory reservation to move from.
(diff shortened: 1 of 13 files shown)


cpp/include/rapidsmpf/buffer/resource.hpp
```diff
@@ -478,4 +478,44 @@ MemoryReservation reserve_or_fail(
...
+        return std::forward<F>(f)(reservation);
+    } else if constexpr (std::invocable<F>) {
+        return std::forward<F>(f)();
+    } else {
+        static_assert(
+            [] { return false; }(),
+            "f must be callable with no args or MemoryReservation&"
+        );
+    }
+}
```
madsbk, line 519: What about `with_memory_reservation` also makes the reservation?

Then we can add a `MemoryReservation.clear()` that clears the remaining reservation (like in `~MemoryReservation()`) and use that in the general case?

cpp/src/integrations/cudf/partition.cpp
```diff
@@ -20,6 +21,61 @@
...
+namespace {
+
+
+/**
+ * @brief Calculate the memory usage of a column.
+ *
+ * @param col The column to calculate the memory usage of.
+ * @return The memory usage of the column.
+ */
+size_t column_memory_usage(cudf::column_view const& col) {
```
madsbk, line 34: What about submitting a small PR that introduces an util function `approx_device_memory_usage(x)` and implements overloads for common data structures like `cudf::table_view`, `cudf::coulmn_view` and vectors/ranges of them?

cpp/include/rapidsmpf/buffer/resource.hpp
```diff
@@ -478,4 +478,44 @@ MemoryReservation reserve_or_fail(
...
+ *
+ * @param reservation moved memory reservation.
+ * @param f The function to execute.
+ *
+ * @return The result of the function.
+ */
+auto with_memory_reservation(MemoryReservation&& reservation, auto&& f) {
+    using F = std::decay_t<decltype(f)>;
+
+    if constexpr (std::invocable<F, MemoryReservation&>) {
```
madsbk, line 509: Do we need to support this signature?

madsbk (commented):
Overall looks great

cpp/include/rapidsmpf/buffer/resource.hpp
nirandaperera: I was thinking we could use `with_memory_reservation` for any reservation. So, this signature would be more general purpose IMO.

cpp/include/rapidsmpf/buffer/resource.hpp
```diff
@@ -478,4 +518,32 @@ MemoryReservation reserve_or_fail(
...
+        return std::forward<F>(f)(reservation);
+    } else if constexpr (std::invocable<F>) {
+        return std::forward<F>(f)();
+    } else {
+        static_assert(
+            [] { return false; }(),
```
wence-, line 543:
```suggestion
            std::integral_constant<F, false>(),
```

Your example happens to work today but a sufficiently smart compiler might decide to compile-time evaluate the lambda and then you run afoul of the issue that `static_assert(false)` is not allowed. See also https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2593r0.html

cpp/include/rapidsmpf/integrations/cudf/partition.hpp
```diff
@@ -122,9 +128,13 @@ partition_and_split(
  * @param stream CUDA stream used for device memory operations and kernel launches.
  * @param br Buffer resource for memory allocations.
  * @param statistics The statistics instance to use (disabled by default).
- *
+ * @param allow_overbooking If true, allow overbooking (false by default)
  * @return The unpacked and concatenated result.
  *
+ * @throw std::overflow_error if the buffer resource cannot reserve enough memory
+ * to concatenate all partitions. std::logic_error if the parititons are not in device
+ * memory.
```
wence-, line 136:
```suggestion
 * @throw std::overflow_error if the buffer resource cannot reserve enough memory
 * to concatenate all partitions. 
 * @std::logic_error if the partitions are not in device memory.
```

cpp/src/integrations/cudf/utils.cpp
```diff
@@ -37,6 +45,43 @@ struct str_cudf_column_scalar_fn {
...
+        return bitmask_size(col);
+    }
+
+    template <typename T>
+    size_t operator()(cudf::column_view const& col, rmm::cuda_stream_view) {
+        RAPIDSMPF_FAIL("not implemented for type: " + cudf::type_to_name(col.type()));
+    }
+
+    static size_t bitmask_size(cudf::column_view const& col) {
+        return col.has_nulls() ? cudf::bitmask_allocation_size_bytes(col.size()) : 0;
```
wence-, line 81:
```suggestion
        return col.nullable() ? cudf::bitmask_allocation_size_bytes(col.size()) : 0;
```

cpp/src/integrations/cudf/utils.cpp
```diff
@@ -37,6 +45,43 @@ struct str_cudf_column_scalar_fn {
...
+        return static_cast<size_t>(col.size()) * cudf::size_of(col.type())
+               + bitmask_size(col);
+    }
+
+    // string type specialization
+    template <typename T>
+        requires(std::is_same_v<T, cudf::string_view>)
+    size_t operator()(cudf::column_view const& col, rmm::cuda_stream_view stream) {
+        cudf::strings_column_view sv(col);
+        return static_cast<size_t>(sv.chars_size(stream)) + bitmask_size(col);
```
wence-, line 62: Note that this launches a blocking kernel on `stream`...

cpp/src/integrations/cudf/utils.cpp
```diff
@@ -74,4 +119,42 @@ std::string str(
     return ss.str();
 }
 
+/**
+ * @brief Calculate the memory usage of a column.
+ *
+ * @param col The column to calculate the memory usage of.
+ * @return The memory usage of the column.
+ */
```
wence-, line 127:
```suggestion
```

These are documented (more completely) in the header file.

cpp/src/integrations/cudf/utils.cpp
```diff
@@ -74,4 +119,42 @@ std::string str(
...
+        }
+    );
+}
+
+/**
+ * @brief Calculate the memory usage of a table.
+ *
+ * @param tbl The table to calculate the memory usage of.
+ * @return The memory usage of the table.
+ */
```
wence-, line 147:
```suggestion
```

wence- (approved):
Tiny suggestions

cpp/src/integrations/cudf/utils.cpp
nirandaperera: I felt like its inevitable, because we dont have access to the rmm buffer sizes with the column view. Isnt it?

cpp/src/integrations/cudf/utils.cpp
nirandaperera: Ah! TIL. I was looking at [`column::nullable` and it didn't really give me that impression](https://docs.rapids.ai/api/libcudf/stable/classcudf_1_1column#abb4380659f07daa47db1892b7d02a978). 

Anyways, fixed it here #450

PR #440, opened 2025-08-15
Added nvtx annotation for shuffle barriers

TomAugspurger:
While looking at an nsys profile I noticed a gap between an insert_partition and extract_partition. I want to confirm where the time is being spent, so I've added an nvtx annotation here.


TomAugspurger:
This didn't end up being useful. Closing.

PR #441, opened 2025-08-18
CudaEvent helper class

madsbk:
Replace `Buffer::Event` with a more general class.


cpp/src/cuda_event.cpp
```diff
@@ -0,0 +1,77 @@
...
+    rmm::cuda_stream_view stream, unsigned flags
+) {
+    auto ret = std::make_shared<CudaEvent>(flags);
+    ret->record(stream);
+    return ret;
+}
+
+CudaEvent::~CudaEvent() noexcept {
+    if (event_ != nullptr) {
+        cudaEventDestroy(event_);
```
wence-, line 24: Just to note in passing, we're not allowed to make `static`-lifetime `CudaEvent`s

PR #448, opened 2025-08-19
Python API for gathering statistics

### LINKED ISSUE #422: Python API for structured Statistics
I'd like to collect `Statistics` for ad-hoc analysis, after a shuffle has completed.

At the moment, we have `Statistics.report()` which provides a nice human readable summary. I'd like to get similar information, but in a structured format so I can serialize it to JSON.

And my ultimate goal is to collect the statistics from all the nodes involved in a shuffle (via Dask in this case). Perhaps it's a separate issue, but a higher-level API that these statistics from all the workers and aggregates them might be in scope for rapidsmpf.

TomAugspurger:
This adds a python API for gathering statistics.

```python
import pprint
import dask.distributed
from dask_cuda import LocalCUDACluster

from rapidsmpf.config import Options
from rapidsmpf.examples.dask import dask_cudf_shuffle
from rapidsmpf.integrations.dask.shuffler import gather_shuffle_statistics


def main():
    config_options = Options({"dask_statistics": "true"})

    # RapidsMPF is compatible with `dask_cuda` workers.
    # Use an rmm pool for optimal performance.
    cluster =  LocalCUDACluster(rmm_pool_size=0.8)
    client = dask.distributed.Client(cluster)

    df = dask.datasets.timeseries().reset_index(drop=True).to_backend("cudf")
    shuffled = dask_cudf_shuffle(df, on=["name"], config_options=config_options)

    # collect the results in memory.
    shuffled.compute()

    stats = gather_shuffle_statistics(client)
    pprint.pprint(stats)


if __name__ == "__main__":
    main()

```

which prints:

```
{'event-loop-check-future-finish': {'count': 414255,
                                    'value': 0.21929259299999673},
 'event-loop-init-gpu-data-send': {'count': 414255, 'value': 1.084213984999997},
 'event-loop-metadata-recv': {'count': 414255, 'value': 18.802375738000066},
 'event-loop-metadata-send': {'count': 414255, 'value': 5.232919412000041},
 'event-loop-post-incoming-chunk-recv': {'count': 414255,
                                         'value': 12.169733180998538},
 'event-loop-total': {'count': 828510, 'value': 76.28069178999999},
 'shuffle-payload-recv': {'count': 569, 'value': 86509952.0},
 'shuffle-payload-send': {'count': 569, 'value': 86509952.0},
 'spill-bytes-host-to-device': {'count': 30, 'value': 0.0},
 'spill-time-host-to-device': {'count': 30, 'value': 0.000358106}}
```

Closes #422

### DIFF
--- a/python/rapidsmpf/rapidsmpf/integrations/core.py
+++ b/python/rapidsmpf/rapidsmpf/integrations/core.py
@@ -28,6 +28,7 @@ from rapidsmpf.statistics import Statistics
 
 if TYPE_CHECKING:
     from collections.abc import Callable, Sequence
+    from numbers import Number
 
     from rapidsmpf.communicator.communicator import Communicator
 
@@ -82,6 +83,20 @@ def get_new_shuffle_id(get_occupied_ids: Callable[[], Sequence[set[int]]]) -> in
         return _shuffle_id_vacancy.pop()
 
 
+_STATISTICS_KEYS = [
+    "event-loop-check-future-finish",
+    "event-loop-init-gpu-data-send",
+    "event-loop-metadata-recv",
+    "event-loop-metadata-send",
+    "event-loop-post-incoming-chunk-recv",
+    "event-loop-total",
+    "shuffle-payload-recv",
+    "shuffle-payload-send",
+    "spill-bytes-host-to-device",
+    "spill-time-host-to-device",
+]
+
+
 @dataclass
 class WorkerContext:
     """
@@ -116,6 +131,27 @@ class WorkerContext:
     spill_collection: SpillCollection = field(default_factory=SpillCollection)
     shufflers: dict[int, Shuffler] = field(default_factory=dict)
     options: Options = field(default_factory=Options)
+
+    def get_statistics(self) -> dict[str, dict[str, Number]]:
+        """
+        Get the statistics from the worker context.
+
+        Returns
+        -------
+        statistics
+            A dictionary of statistics. The keys are the names of the statistics.
+            The values are dictionaries with two keys:
+
+            - "count" is the number of times the statistic was recorded.
+            - "value" is the value of the statistic.
+
+        Notes
+        -----
+        Statistics are global across all shuffles. To measure statistics for any
+        given shuffle, gather statistics before and after the shuffle and compute
+        the difference.
+        """
+        return {stat: self.statistics.get_stat(stat) for stat in _STATISTICS_KEYS}
 
 
 class ShufflerIntegration(Protocol[DataFrameT]):
--- a/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
+++ b/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+from collections import defaultdict
 from functools import partial
 from typing import TYPE_CHECKING, Any
 
@@ -11,6 +12,7 @@ from distributed import get_worker
 
 from rapidsmpf.config import Options
 from rapidsmpf.integrations.core import (
+    _STATISTICS_KEYS,
     extract_partition,
     get_new_shuffle_id,
     get_shuffler,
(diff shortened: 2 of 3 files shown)


python/rapidsmpf/rapidsmpf/integrations/core.py
```diff
@@ -82,6 +83,20 @@ def get_new_shuffle_id(get_occupied_ids: Callable[[], Sequence[set[int]]]) -> in
         return _shuffle_id_vacancy.pop()
 
 
+_STATISTICS_KEYS = [
```
TomAugspurger, line 86: I'd like to avoid repeating statistic names here. Do we have a list of valid statistic names in C++ somewhere (looking at https://github.com/rapidsai/rapidsmpf/blob/d5aefa1d43f987115d7722b979a93414bcb3527f/cpp/src/shuffler/shuffler.cpp#L412, I think the answer is "no")?

TomAugspurger (commented):
/ok to test 99aef0f

python/rapidsmpf/rapidsmpf/integrations/core.py
pentschev: Yeah, I think the answer is no, they are added on spot when needed, so this is likely to change over time as well.

TomAugspurger: What do people think about using an enum (in C++) and updating `Statistics.add_stat` to take the enum member rather than a string?

For consumers of these statistics, it's helpful to know what's allowed. IMO that's worth a bit more friction on the producer side (adding a new member to the enum).

pentschev: I think a big problem with that (not immediately, but in the future) is that we'll break API and ABI every time we want to modify anything in the statistics, furthermore, we would have to register everything in a central place a priori, which is probably not something we really want to do.

@madsbk WDYT?

TomAugspurger: That will avoid the current repetition, but won't help with knowing what the valid keys are.

pentschev: > won't help with knowing what the valid keys are.

But why is it important to know that before it runs?

pentschev:
> The last thing I'd like to raise here: Currently collecting statistics is coupled to printing them upon shutdown
> Now that we have another way of getting statistics, I'd like to decouple those. I'll take the current value to mean whether or not to collect statistics, and add a new configuration option for whether to print them on shutdown.

Sounds good to me, I think the new configuration should be enabled by default, but +1 to that.

python/rapidsmpf/rapidsmpf/integrations/core.py
madsbk: I agree with Peter, being able to add new statistics on the spot can be very useful!

python/rapidsmpf/rapidsmpf/integrations/core.py
```diff
@@ -432,14 +457,18 @@ def rmpf_worker_setup(
     # Print statistics at worker shutdown.
     if options.get_or_default(f"{option_prefix}statistics", default_value=False):
         statistics = Statistics(enable=True, mr=mr)
+    else:
+        statistics = Statistics(enable=False)
+
+    if options.get_or_default(f"{option_prefix}print_statistics", default_value=False):
+        if not statistics.enabled:
+            raise ValueError("Cannot print statistics if statistics are not enabled.")
```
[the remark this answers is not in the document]
TomAugspurger: I worry a bit about the `{"dask_print_statistics": True}`  silently not doing anything when combined with the `{"dask_statistics": False}` setting.

That might be OK, since the default is to not collect statistics. Users looking to enable statistics would hopefully see `dask_statistics` in addition to `dask_print_statistics`.

If that's the preference then I'm happy to make the change.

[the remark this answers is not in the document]
TomAugspurger: Changed this to print by default (when stats are enabled). This should mean no change from the current behavior, but adds the ability to silence the printing when you're analyzing the stats separately.

I didn't look at all really, but I wasn't sure how to get a `bool | str` to work with the `Options` object (which seems to require specific types for fields).

Issue #449, opened 2025-08-19
Disable overbooking in cudf partition utils by default


nirandaperera:
To preserve the behavior prior to #438 , overbooking is enabled by default (438 disabled it by default). 

Disabling would be the safer option. So, revert it once the nightly TPCH run OOM issue is resolved.

PR #450, opened 2025-08-19
[MINOR] Enable overbooking by default in cudf utils 

nirandaperera:
A temporary measure to preserve behavior prior to #438. But this needs to be reverted in #449


pentschev (approved):
LGTM.

PR #451, opened 2025-08-20
Fix linker flags in new streaming benchmarks CMakeLists

pentschev:
https://github.com/rapidsai/rapidsmpf/pull/429 added linker flags to allow linking to pass in devcontainers, unfortunately the new streaming engine ended up being merged just after that without the same changes, which are now included in this PR.


pentschev:
Thanks Mads!

PR #452, opened 2025-08-20
API for clearing Statistics

TomAugspurger:
This adds an API for clearing / resetting the statistics captured by a Statistics object. This will aid in analyzing the performance of individual shuffles, since users can reset statistics between shuffles (or in the case of cudf-polars, queries).

### DIFF
--- a/cpp/include/rapidsmpf/statistics.hpp
+++ b/cpp/include/rapidsmpf/statistics.hpp
@@ -234,6 +234,8 @@ class Statistics {
 
     /**
      * @brief Clears all statistics.
+     *
+     * Memory profiling records are not cleared.
      */
     void clear();
 
--- a/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
+++ b/python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
@@ -323,6 +323,8 @@ def clear_shuffle_statistics(client: Client) -> None:
     """
     Clear all statistics for all workers.
 
+    Memory profiling records are not cleared.
+
     Parameters
     ----------
     client
--- a/python/rapidsmpf/rapidsmpf/statistics.pyx
+++ b/python/rapidsmpf/rapidsmpf/statistics.pyx
@@ -238,6 +238,8 @@ cdef class Statistics:
     def clear(self) -> None:
         """
         Clears all statistics.
+
+        Memory profiling records are not cleared.
         """
         with nogil:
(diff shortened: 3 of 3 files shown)


cpp/src/statistics.cpp
```diff
@@ -81,6 +81,12 @@ std::vector<std::string> Statistics::list_stat_names() const {
     return std::vector<std::string>{ks.begin(), ks.end()};
 }
 
+void Statistics::clear() {
+    std::lock_guard<std::mutex> lock(mutex_);
+    stats_.clear();
+    memory_records_.clear();
+}
```
madsbk, line 88: What happens if we are in the middle of a `MemoryRecorder`?
I think we need to throw an error if `memory_records_` isn't empty otherwise `Statistics::MemoryRecorder::~MemoryRecorder()` will fail.

TomAugspurger: I've done statistics only for now.

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -232,6 +232,11 @@ class Statistics {
      */
     std::vector<std::string> list_stat_names() const;
 
+    /**
+     * @brief Clears all statistics.
```
madsbk, line 236:
```suggestion
     * @brief Clears all statistics.
     * 
     * Memory profiling records remain unchanged.
```

python/rapidsmpf/rapidsmpf/integrations/dask/shuffler.py
```diff
@@ -312,6 +312,25 @@ def _gather_worker_shuffle_statistics(
...
+    dask_worker: Worker,
+) -> None:
+    context = get_worker_context(dask_worker)
+    context.statistics.clear()
+
+
+def clear_shuffle_statistics(client: Client) -> None:
+    """
+    Clear all statistics for all workers.
+
```
madsbk, line 325:
```suggestion
    Clear all statistics for all workers.

    Memory profiling records remain unchanged.
```

madsbk (approved):
Looks good

cpp/src/statistics.cpp
madsbk: > I'm not familiar with `MemoryRecorder`. My hope was that the lock we acquire here in `clear` would prevent us from modifying things when we aren't supposed to.

When memory profiling is enabled, a `MemoryRecorder` register a record on creation and tally it up on destruction.

PR #453, opened 2025-08-21
Disable shallow clone to be able to checkout the proper commit

pentschev:
Cloning a git repository in shallow mode (i.e., `--depth 1`) does not allow checking out arbitrary commits since git will not pull the entire history and thus only the head commit exists, or proper tags, since tags are pulled even in shallow clones.


pentschev:
Thanks Mads!

PR #455, opened 2025-08-21
Refactor dependency specifications to align with RAPIDS standards

bdice:
Follow-up from an issue discussed with @pentschev.

This refactors dependency specifications in rapidsmpf to better align with the way we define things in other RAPIDS repos.

This now uses the `depends_on_...` format for getting the librapidsmpf/rapidsmpf packages into the test environments used in CI scripts.

I also refactored `test_python` and some other dependency lists to separate it in the same way we handle other packages.


jameslamb (approved):
Support this! I'm glad it also removed some duplicate dependencies between the hard runtime dependencies and the `[test]` extra.

pentschev (approved):
Only giving a "support approval", I'm not quite certain of the changes but I trust James' review. It seems there are few a docs failures attempting to import `rapidsmpf` Python modules @bdice , I think those are legit from changes here.

PR #464, opened 2025-08-26
BaseChunk: common base for streaming chunks 

madsbk:
Introduce a common base for streaming chunks and fixes a bug in the shuffler where the stream argument wasn't set.

This also simplifies the Python bindings a lot.

### DIFF
--- a/cpp/include/rapidsmpf/streaming/core/base_chunk.hpp
+++ b/cpp/include/rapidsmpf/streaming/core/base_chunk.hpp
@@ -29,7 +29,7 @@ class BaseChunk {
     constexpr BaseChunk(
         std::uint64_t sequence_number, rmm::cuda_stream_view stream
     ) noexcept
-        : _sequence_number{sequence_number}, _stream{stream} {}
+        : sequence_number_{sequence_number}, stream_{stream} {}
 
     /**
      * @brief Virtual destructor.
@@ -56,7 +56,7 @@ class BaseChunk {
      * @return The sequence number.
      */
     [[nodiscard]] constexpr std::uint64_t sequence_number() const noexcept {
-        return _sequence_number;
+        return sequence_number_;
     }
 
     /**
@@ -65,12 +65,12 @@ class BaseChunk {
      * @return The associated rmm::cuda_stream_view.
      */
     [[nodiscard]] rmm::cuda_stream_view stream() const noexcept {
-        return _stream;
+        return stream_;
     }
 
   private:
-    std::uint64_t _sequence_number;
-    rmm::cuda_stream_view _stream;
+    std::uint64_t sequence_number_;
+    rmm::cuda_stream_view stream_;
 };
 
 }  // namespace rapidsmpf::streaming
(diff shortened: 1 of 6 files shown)


madsbk:
> @madsbk now that a `Chunk` is polymorphic, should we use shared/unique ptrs?

Where?

cpp/include/rapidsmpf/streaming/cudf/partition.hpp
```diff
@@ -24,56 +25,107 @@ namespace rapidsmpf::streaming {
...
-struct PartitionVectorChunk {
+class PartitionVectorChunk : public BaseChunk {
+  public:
     /**
-     * @brief Sequence number used to preserve chunk ordering.
+     * @brief Chunk of packed partitions stored as a vector.
+     *
+     * @param sequence_number Ordering identifier for the chunk.
+     * @param stream CUDA stream associated with buffers referenced by the packed data.
+     * @param data The packed-data vector to take ownership of (moved).
```
wence-, line 90: *nit*: reorder parameters to match argument order.

cpp/include/rapidsmpf/streaming/cudf/partition.hpp
```diff
@@ -24,56 +25,107 @@ namespace rapidsmpf::streaming {
...
-struct PartitionMapChunk {
+class PartitionMapChunk : public BaseChunk {
+  public:
     /**
-     * @brief Sequence number used to preserve chunk ordering.
+     * @brief Construct a chunk with a partition map.
+     *
+     * @param sequence_number Ordering identifier for the chunk.
+     * @param stream CUDA stream associated with buffers referenced by the packed data.
+     * @param data The partition → packed-data map to take ownership of (moved).
```
wence-, line 40: *nit*: ordering doesn't match ctor argument order.

cpp/include/rapidsmpf/streaming/core/base_chunk.hpp
```diff
@@ -0,0 +1,76 @@
...
+     */
+    constexpr BaseChunk(
+        std::uint64_t sequence_number, rmm::cuda_stream_view stream
+    ) noexcept
+        : sequence_number_{sequence_number}, stream_{stream} {}
+
+    /**
+     * @brief Virtual destructor.
+     */
+    virtual ~BaseChunk() = default;
```
wence-, line 37: *question*: Does this class need to be virtual, or could it "just" be a mixin instead?

It seems like we never need to be able to store polymorphic child classes of `BaseChunk` in the same container...

cpp/include/rapidsmpf/streaming/core/base_chunk.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+#include <rmm/cuda_stream_view.hpp>
+
+namespace rapidsmpf::streaming {
+
+/**
+ * @brief Common base for streaming chunks.
+ *
+ * Holds the per-chunk sequence number (for ordering) and the CUDA stream
+ * on which the chunk's resources were created.
```
wence-, line 18: *minor*: Not all chunks will need a stream (if we're communicating non-gpu data through the channel). I suppose we can just use the default stream in that case.

PR #470, opened 2025-08-27
Avoid setting memory resource in `rmpf_worker_setup`

TomAugspurger:
This updates `rmpf_worker_setup` to avoid setting the (global) default device memory resource. rapidsmpf *shouldn't* need it set, and setting it can interfere with applications using RMM and rapidsmpf.


TomAugspurger:
It sounds like we can't remove this until https://github.com/rapidsai/cudf/issues/14229 is resolved. Closing for now.

PR #473, opened 2025-08-28
[Streaming] Type-erased channels

madsbk:
Redesigning the streaming network to use type-erased channels rather than templates. This will simplify future Python bindings and the runtime-generation of streaming networks.


cpp/include/rapidsmpf/streaming/core/channel.hpp
```diff
@@ -6,76 +6,157 @@
...
+
+/**
+ * @brief A coroutine-based channel for sending and receiving messages asynchronously.
+ */
+class Channel {
+  public:
+    /**
+     * @brief Asynchronously send a message into the channel.
+     *
+     * Suspends if the channel is empty.
```
wence-, line 130:
```suggestion
     * Suspends if the channel is full.
```
?

cpp/include/rapidsmpf/streaming/core/channel.hpp
```diff
@@ -107,30 +188,9 @@ class Channel {
     }
 
   private:
-    coro::ring_buffer<T, Capacity> rb_;
+    coro::ring_buffer<Message, 1> rb_;
```
wence-, line 191: We should do some experimenting around whether it makes sense to be able to control the buffering capacity of the channel, but for now I think hard-coding to 1 is fine.

cpp/include/rapidsmpf/streaming/cudf/partition.hpp
```diff
@@ -85,8 +84,8 @@ namespace node {
  * partitions, and sends them to an output channel.
  *
  * @param ctx The context to use.
- * @param ch_in Input channel providing tables to partition.
- * @param ch_out Output channel to which packed partitioned table are sent.
+ * @param ch_in Input channel providing TableChunk to partition.
```
wence-, line 87:
```suggestion
 * @param ch_in Input channel providing `TableChunk`s to partition.
```

cpp/include/rapidsmpf/streaming/cudf/partition.hpp
```diff
@@ -85,8 +84,8 @@ namespace node {
...
+ * @param ch_out Output channel to which PartitionMapChunk are sent.
```
wence-, line 88:
```suggestion
 * @param ch_out Output channel to which `PartitionMapChunk`s are sent.
```

cpp/src/streaming/cudf/partition.cpp
```diff
@@ -24,14 +25,15 @@ Node partition_and_pack(
 
     co_await ctx->executor()->schedule();
     while (true) {
-        auto table = co_await ch_in->receive_or(nullptr);
-        if (table == nullptr) {
+        auto msg = co_await ch_in->receive();
+        if (msg.empty()) {
             break;
         }
+        auto table = msg.release<TableChunk>();
```
wence-, line 32: Could we have `release` return a `nullptr` for empty messages then we could have:

```
auto table = co_await ch_in->receive().release<TableChunk>();
```

Or even, (maybe not worth it) template `receive` so we would do:
```
auto table = co_await ch_in->receive_or<TableChunk>(nullptr);
```

WDYT?

cpp/include/rapidsmpf/streaming/core/channel.hpp
```diff
@@ -6,76 +6,157 @@
...
-        auto msg = co_await rb_.consume();
-        if (msg.has_value()) {
-            co_return std::make_optional(std::move(*msg));
-        } else {
-            co_return std::nullopt;
-        }
+    template <typename T>
+    T release() {
+        auto ret = get_ptr<T>();
+        reset();
```
wence-, line 85: IIUC, if some other message also has a reference to this payload they continue to do so?

cpp/src/streaming/cudf/partition.cpp
```diff
@@ -24,14 +25,15 @@ Node partition_and_pack(
...
         auto reservation = ctx->br()->reserve_and_spill(
-            MemoryType::DEVICE, table->make_available_cost(), false
+            MemoryType::DEVICE, table.make_available_cost(), false
         );
-        auto tbl = table->make_available(reservation, table->stream());
+        auto tbl = table.make_available(reservation, table.stream());
```
wence-, line 36: *question*: Since the data in a message is now stored in a shared pointer, does this now potentially make a _copy_ of the `table` (since someone else might also have the same message)?

wence- (commented):
Nothing major, just thinking about the semantics a bit

cpp/src/streaming/cudf/partition.cpp
wence-: ah, true

cpp/include/rapidsmpf/streaming/core/channel.hpp
wence-: Thanks.

wence- (approved):

madsbk:
Thanks @pentschev, do you have anything else?

Issue #486, opened 2025-09-03
Occasional UCX error


ayushdg:
I'm running a RapidsMPF shuffle with Ray and occasionally (twice in 15 runs) I see UCX errors during the shuffle. This was with rapidsmpf 25.06. Will revisit with 25.08:

```
2025-09-03 10:49:24.917 | INFO     | nemo_curator.backends.experimental.ray_actor_pool.executor:_create_rapidsmpf_actors:219 -     UCXX setup complete
(ShuffleStageAdapter pid=1323505, ip=10.52.48.39) [2025-09-03 10:50:30,967 E 1323505 1324986] logging.cc:118: Unhandled exception: N4ucxx13TimedOutErrorE. what(): Operation timed out
(ShuffleStageAdapter pid=1323505, ip=10.52.48.39) [2025-09-03 10:50:30,974 E 1323505 1324986] logging.cc:125: Stack trace: 
(ShuffleStageAdapter pid=1323505, ip=10.52.48.39)  /opt/venv/lib/python3.12/site-packages/ray/_raylet.so(+0x152da9a) [0x7ffff6eefa9a] ray::operator<<()
(ShuffleStageAdapter pid=1323505, ip=10.52.48.39) /opt/venv/lib/python3.12/site-packages/ray/_raylet.so(+0x15309a2) [0x7ffff6ef29a2] ray::TerminateHandler()
(ShuffleStageAdapter pid=1323505, ip=10.52.48.39) /usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0xbb0da) [0x7ffff57ff0da]
(ShuffleStageAdapter pid=1323505, ip=10.52.48.39) /usr/lib/x86_64-linux-gnu/libstdc++.so.6(_ZSt10unexpectedv+0) [0x7ffff57e9a55] std::unexpected()
(ShuffleStageAdapter pid=1323505, ip=10.52.48.39) /usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0xbb391) [0x7ffff57ff391]
(ShuffleStageAdapter pid=1323505, ip=10.52.48.39) /opt/venv/lib/python3.12/site-packages/libucxx/lib64/libucxx.so(+0x42158) [0x7fcf0c153158]
[...]

PR #487, opened 2025-09-03
Make UCXX progress mode configurable

pentschev:
https://github.com/rapidsai/rapidsmpf/pull/142 was automatically closed because it was a fork from before the repository was renamed, thus it cannot be reopened. This change should allow circumventing the error from https://github.com/rapidsai/rapidsmpf/issues/486, but performance profile may be affected.


cpp/src/communicator/ucxx.cpp
```diff
@@ -415,13 +423,18 @@ class SharedResources {
...
             worker_->progress();
-            // TODO: Support blocking progress mode
         }
 
         clear_completed_futures();
     }
 
+    void maybe_progress_worker() {
+        if (++progress_count % 100)
+            progress_worker();
```
madsbk, line 435: Why the `100` ?
```suggestion
        if (++progress_count % 100) { progress_worker(); }
```

cpp/src/communicator/ucxx.cpp
```diff
@@ -816,14 +829,41 @@ InitializedRank::InitializedRank(
 std::unique_ptr<rapidsmpf::ucxx::InitializedRank> init(
     std::shared_ptr<::ucxx::Worker> worker,
     Rank nranks,
-    std::optional<RemoteAddress> remote_address
+    std::optional<RemoteAddress> remote_address,
+    config::Options options
 ) {
-    auto create_worker = []() {
+    auto progress_mode = options.get<ProgressMode>("ucxx_progress_mode", [](auto s) {
```
madsbk, line 835:
```suggestion
    auto progress_mode = options.get<ProgressMode>("ucxx_progress_mode", [](auto const &s) {
```

cpp/src/communicator/ucxx.cpp
```diff
@@ -816,14 +829,25 @@ InitializedRank::InitializedRank(
...
+        RAPIDSMPF_EXPECTS(
+            progress_mode != ProgressMode::Blocking,
+            "Blocking progress mode not implemented yet."
+        );
+
+        if (progress_mode == ProgressMode::ThreadBlocking
+            || progress_mode == ProgressMode::ThreadPolling)
+        {
+            worker->setProgressThreadStartCallback(create_cuda_context_callback, nullptr);
+            worker->startProgressThread(progress_mode == ProgressMode::ThreadPolling);
```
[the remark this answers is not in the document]
madsbk: Could this also be a `option` ?

python/rapidsmpf/rapidsmpf/tests/conftest.py
```diff
@@ -22,9 +22,34 @@
...
+    parser.addoption(
+        "--disable-mpi",
+        action="store_true",
+        default=False,
+        help="Disable MPI tests",
+    )
+
+
+@pytest.fixture(scope="session")
+def _mpi_disabled(pytestconfig: pytest.Config) -> Any:
```
madsbk, line 36: I guess `pytestconfig.getoption("--disable-mpi")` return a boolean?

```suggestion
def _mpi_disabled(pytestconfig: pytest.Config) -> bool:
```

python/rapidsmpf/rapidsmpf/tests/conftest.py
```diff
@@ -22,9 +22,34 @@
...
-    Return the `mpi4py.MPI` module if MPI support is available or pytest.skip.
+    Return the `mpi4py.MPI` module or pytest.skip.
+
+    Return the `mpi4py.MPI` module if MPI support is available, skip tests if
+    `mpi4py.MPI` is unavailable or MPI tests are disabled via the --disable-mpi
+    command line option.
+
+    Parameters
+    ----------
+    mpi_disabled : bool, optional
```
madsbk, line 51:
```suggestion
    mpi_disabled

File: python/rapidsmpf/rapidsmpf/tests/conftest.py [hunk 6]
```diff
@@ -33,18 +58,21 @@ def _get_mpi_module_or_skip() -> MPI:
...
+        pytest.skip("MPI tests disabled")
+
     from mpi4py import MPI
 
     return MPI
 
 
 @pytest.fixture(scope="session")
-def _mpi_comm() -> Communicator:
+def _mpi_comm(_mpi_disabled: bool) -> Communicator:  # noqa: FBT001
```
madsbk, line 75:
```suggestion
def _mpi_comm(*, _mpi_disabled: bool) -> Communicator:
```

pentschev: Hmm, I thought this would be a positional argument that's why I didn't do it this way. Now I realized you're right, this could not be a positional argument and must be a kwarg, so this in fact works.

cpp/src/communicator/ucxx.cpp
pentschev: No specific reason, I just borrowed the `100` from [OpenMPI's default](https://github.com/open-mpi/ompi/blob/7ad7adad676773fc61203e9a536d17b2ebdfa9c8/opal/mca/common/ucx/common_ucx.c#L42).

cpp/src/communicator/ucxx.cpp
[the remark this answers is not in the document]
pentschev: Unless I'm misunderstanding your question, `progress_mode` is already read from an `option` just above that in line 835. My comment is that we previously implied `ThreadPolling`, but we're switching to `ThreadBlocking` as the default (which was my initial intention) in line 837.

python/rapidsmpf/rapidsmpf/tests/conftest.py
pentschev: Unfortunately this alone fails mypy:

```
python/rapidsmpf/rapidsmpf/tests/conftest.py:40: error: Returning Any from function declared to return "bool"  [no-any-return]
Found 1 error in 1 file (checked 87 source files)
```

I agree with you this should return a `bool`, but I think because this is a `pytestconfig` it's not properly evaluated during mypy and possibly `None` is seen as a return value. What we can do is cast this explicitly to `bool` which is valid so I'll do that too.

python/rapidsmpf/rapidsmpf/tests/conftest.py
```diff
@@ -22,9 +22,34 @@
...
+        action="store_true",
+        default=False,
+        help="Disable MPI tests",
+    )
+
+
+@pytest.fixture(scope="session")
+def _mpi_disabled(pytestconfig: pytest.Config) -> Any:
+    """Check if MPI tests are disabled via command line argument."""
+    return pytestconfig.getoption("--disable-mpi")
```
pentschev, line 38:
```suggestion
    return bool(pytestconfig.getoption("--disable-mpi"))
```

cpp/src/communicator/ucxx.cpp
madsbk: Add that in a comment

madsbk (approved):
Looks good @pentschev, thanks

cpp/src/communicator/ucxx.cpp
pentschev: Added in https://github.com/rapidsai/rapidsmpf/pull/487/commits/783d2d10d73a1f07429b5eb43171ea01ab9aa493

pentschev:
Thanks Mads!

ci/test_python.sh
```diff
@@ -37,8 +37,11 @@ nvidia-smi
 # Support invoking test_python.sh outside the script directory
 cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../
 
-rapids-logger "pytest rapidsmpf"
+rapids-logger "Pytest RapidsMPF (MPI+UCXX)"
 ./ci/run_pytests.sh && EXITCODE=$? || EXITCODE=$?;
```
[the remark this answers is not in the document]
pentschev: This is not really being changed here, I'm not sure why it was written like that. I'll open another PR to make this change.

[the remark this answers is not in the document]
pentschev: Opened https://github.com/rapidsai/rapidsmpf/pull/542 to tackle this and making things a bit more consistent.

PR #502, opened 2025-09-09
Fix test workflow

pentschev:
The test workflow was modified in #498 to add a custom job, however, the syntax was incorrect and broke the workflow, causing errors such as:

```
Invalid workflow file: .github/workflows/test.yaml
(Line: 38, Col: 12): Job 'conda-cpp-memcheck' depends on unknown job 'checks'.
```

This should fix the broken syntax and match the style of the file.


pentschev:
Thanks James!

PR #505, opened 2025-09-10
Add `known_first_party` configuration for isort

pentschev:
This is necessary to ensure isort doesn't break style if the clone of the repository has a different name than `rapidsmpf`.

To reproduce the original issue:

```
git clone https://github.com/rapidsai/rapidsmpf rapidsmpf-clone  # The directory name must be something other than rapidsmpf
cd rapidsmpf-clone
git checkout 9b36b33  # Current branch-25.10 commit, where this is known to fail
pre-commit run isort -a
```

That will make style changes to Cython files that are already in the expected format.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -72,7 +72,7 @@ repos:
     hooks:
       - id: isort
         name: isort (cython)
-        args: ["--config-root=python/", "--resolve-all-configs"]
+        args: ["--resolve-all-configs"]
         files: python/.*
         types_or: [cython]
   - repo: https://github.com/codespell-project/codespell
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -72,7 +72,7 @@ repos:
     hooks:
       - id: isort
         name: isort (cython)
-        args: ["--config-root=python/", "--resolve-all-configs"]
+        args: ["--resolve-all-configs"]
```
pentschev, line 75: Without this isort will look for a configuration file under `python/`, and we want it to use `pyproject.toml` in the root of the repository.

bdice (approved):

Issue #493, opened 2025-09-05
Tracking stream-ordered work in streaming execution


wence-:
In the streaming prototypes we are working on, all work is run on streams. We don't use PTDS because when a coroutine is suspended on thread A it might be resumed on thread B.

To facilitate this, the tables that we send around are decorated (as `TableChunk`s) with a sequence number (for order-preservation) and the stream the table was created on.

If we _only_ have a `Buffer`, we've lost the stream, so we need to keep track of that somewhere else.

Separately, for the purpose of shuffling and interacting with the parts of the world that are not stream ordered `Buffer`s optional hold a `cudaEvent` that can be queried to know if the `Buffer` is safe to hand to MPI/UCX.

I _think_ we want to move to a model where all `Buffer`s hold a `stream` (so we know which stream they are valid on), and we have a new type when interacting with non-stream-ordered libraries (`BufferWithEvent` or something?) that holds the event as well. [I know this is going back on my previous position that we shouldn't have this split].

Then the shuffler doesn't need to accept a stream (but could use a stream pool) because the inputs that it takes advertise the stream they are valid on, and the outputs it produces similarly do.

Rationale for this is to make it easy to compose pieces together without getting stream ordering wrong: if you pass Buffers on a stream in, the library makes sure that it is safe.

PR #507, opened 2025-09-12
Refactor and cleanup of `BufferResource`

madsbk:
Removing some of the buffer resource API that doesn't take a CUDA stream like `BufferResource::move_to_device_buffer()` and `BufferResource::allocate_empty_host_buffer()`.

Also adding some convenience functions and general code clean up. 

This is part of the https://github.com/rapidsai/rapidsmpf/issues/493 work, where `Buffer` will have an associated CUDA stream.

### DIFF
--- a/cpp/src/integrations/cudf/partition.cpp
+++ b/cpp/src/integrations/cudf/partition.cpp
@@ -216,7 +216,7 @@ std::vector<PackedData> spill_partitions(
     BufferResource* br,
     std::shared_ptr<Statistics> statistics
 ) {
-    auto const elapsed = Clock::now();
+    auto const start_time = Clock::now();
     // Sum the total size of all packed data in device memory.
     std::size_t device_size{0};
     for (auto& [_, data] : partitions) {
@@ -236,7 +236,7 @@ std::vector<PackedData> spill_partitions(
                 );
             }
             statistics->add_duration_stat(
-                "spill-time-device-to-host", Clock::now() - elapsed
+                "spill-time-device-to-host", Clock::now() - start_time
             );
             statistics->add_bytes_stat("spill-bytes-device-to-host", device_size);
             return ret;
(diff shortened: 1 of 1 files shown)


cpp/src/integrations/cudf/partition.cpp
```diff
@@ -161,28 +163,51 @@ std::unique_ptr<cudf::table> unpack_and_concat(
...
-    std::vector<cudf::table_view> unpacked;
-    std::vector<cudf::packed_columns> references;
-    unpacked.reserve(partitions.size());
-    references.reserve(partitions.size());
-    size_t concat_size = 0;
+
+    // Let's find the total size of the partitions and how much of the packed data we
+    // need to move to device memory (unspill).
+    size_t total_size = 0;
+    size_t non_device_size = 0;
```
pentschev, line 170: Any reason not to just call it `host_size`?

cpp/src/integrations/cudf/partition.cpp
```diff
@@ -238,25 +260,24 @@ std::vector<PackedData> unspill_partitions(
...
 
-    return with_memory_reservation(std::move(reservation), [&] {
-        // Unspill each partition.
-        std::vector<PackedData> ret;
-        ret.reserve(partitions.size());
-        for (auto& [metadata, data] : partitions) {
-            ret.emplace_back(
-                std::move(metadata), br->move(std::move(data), stream, reservation)
+            statistics->add_duration_stat(
+                "spill-time-host-to-device", Clock::now() - elapsed
```
pentschev, line 276: `elapsed` is a bad name, it should be `start_time` or something similar. Elapsed would be `Clock::now() - start_time`, but this is just using how it was already named before this PR, so this is fine.

pentschev (approved):
Left a question and comment, but otherwise LGTM.

madsbk:
thanks @pentschev

PR #511, opened 2025-09-15
Adding a streaming shuffler wrapper (`ShufflerAsync`)

nirandaperera:
This PR adds a `ShufflerAsync` class that wraps a `Shuffler` object and exposes the following async methods. 

```c++

    /**
     * @brief Asynchronously extracts all data for a specific partition.
     *
     * This coroutine will suspend until the specified partition is ready for extraction
     * (i.e., insert_finished has been called for this partition and all data has been
     * shuffled).
     *
     * @param pid The partition ID to extract data for.
     * @return A coroutine task that yields a vector of PackedData chunks for the
     * partition.
     *
     * @throws std::runtime_error if the partition ID is not found.
     *
     * @note Users should be careful when using `extract_async` and `extract_any_async`
     * together, because a pid intended for `extract_async` may be extracted by
     * `extract_any_async`.
     */
    coro::task<std::vector<PackedData>> extract_async(shuffler::PartID pid);

    /**
     * @brief Result type for extract_any_async operations.
     *
     * Contains the partition ID and associated data chunks from an extract operation.
     * Can represent either a valid result or an invalid/empty result.
     *
     * An invalid result is returned when no more partitions are available for extraction.
     */
    struct ExtractResult {
        shuffler::PartID pid;  ///< The partition ID that was extracted
        std::vector<PackedData> chunks;  ///< The data chunks for this partition

        /// @brief A sentinel value for an invalid partition ID.
        static constexpr auto InvalidPID = std::numeric_limits<shuffler::PartID>::max();

        /**
         * @brief Checks if this result represents valid extracted data.
         *
         * @return true if this result contains valid data, false otherwise.
         */
        constexpr bool is_valid() const {
            return pid != InvalidPID;
        }

        /**
         * @brief Creates an invalid/empty result.
         *
         * @return An ExtractResult with an invalid partition ID and empty chunks.
         */
        static constexpr ExtractResult invalid() {
            return {InvalidPID, {}};
        }
    };

    /**
     * @brief Asynchronously extracts data for any ready partition.
     *
     * This coroutine will suspend until at least one partition is ready for extraction,
     * then extract and return the data for one such partition. If no partitions become
     * ready and the shuffle is finished, returns an invalid result.
     *
     * @return A coroutine task that yields an ExtractResult containing the partition ID
     *         and data chunks, or an invalid result if no more partitions are available.
     *
     * @note Users should be careful when using `extract_async` and `extract_any_async`
     * together, because a pid intended for `extract_async` may be extracted by
     * `extract_any_async`.
     */
    coro::task<ExtractResult> extract_any_async();
```


cpp/include/rapidsmpf/streaming/cudf/shuffler.hpp
```diff
@@ -46,4 +211,39 @@ Node shuffler(
...
+ * It extracts finished partititons from the shuffler and inserts into the output channel.
+ *
+ * @param shuffler The shuffler to extract data from.
+ * @param ch_out The output channel receiving the extracted data.
+ *
+ * @return A streaming node that completes when the extraction has finished.
+ */
+Node shuffler_async_extract(
+    std::shared_ptr<ShufflerAsync> shuffler, std::shared_ptr<Channel> ch_out
+);
```
madsbk, line 245: Instead of introducing new nodes, I think it would be best to modify `Node shuffler()` to use the new async API.

PR #512, opened 2025-09-16
Buffer copy cleanup 

madsbk:
Replace `Buffer::copy()`, `Buffer::copy_to()`,  and `Buffer::copy_slice()` with a single function: `buffer_copy()`. 

This is part of the https://github.com/rapidsai/rapidsmpf/issues/493 work, where `Buffer` will have an associated CUDA stream.

### DIFF
--- a/cpp/tests/test_buffer_resource.cpp
+++ b/cpp/tests/test_buffer_resource.cpp
@@ -382,7 +382,7 @@ class BufferResourceCopySliceTest
             *source,
             length,
             0,  // dst_offset
-            offset,  // src_offset
+            std::ptrdiff_t(offset),  // src_offset
             stream,
             false
         );
@@ -607,7 +607,7 @@ class BufferResourceDifferentResourcesTest : public ::testing::Test {
 };
 
 TEST_F(BufferResourceDifferentResourcesTest, CopySlice) {
-    constexpr std::size_t slice_offset = 128;
+    constexpr std::ptrdiff_t slice_offset = 128;
     constexpr std::size_t slice_length = 512;
 
     auto buf1 = create_source_buffer();


cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -303,4 +258,31 @@ class Buffer {
...
+ * @throws std::invalid_argument If out of bounds.
+ */
+void buffer_copy(
+    Buffer& dst,
+    Buffer& src,
+    std::size_t size,
+    std::ptrdiff_t dst_offset,
+    std::ptrdiff_t src_offset,
+    rmm::cuda_stream_view stream,
+    bool attach_cuda_event
```
madsbk, line 285: The `stream` and `attach_cuda_event` argument will be removed once the https://github.com/rapidsai/rapidsmpf/issues/493 work is done.

cpp/src/buffer/buffer.cpp
```diff
@@ -61,175 +61,69 @@ std::byte const* Buffer::data() const {
...
-        storage_
-    );
-}
+    // Make sure we wait on any buffer event.
+    if (auto e = dst.get_event()) {
+        e->stream_wait(stream);
+    }
+    if (auto e = src.get_event()) {
+        e->stream_wait(stream);
+    }
```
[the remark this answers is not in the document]
nirandaperera: I think we just need to attach an event after calling memcpy if both buffers are not HOST, isnt it?

cpp/src/buffer/buffer.cpp
```diff
@@ -61,175 +61,69 @@ std::byte const* Buffer::data() const {
...
+        dst.data() + dst_offset, src.data() + src_offset, size, cudaMemcpyDefault, stream
+    ));
 
-void Buffer::wait_for_ready() const {
-    if (event_) {
-        event_->host_wait();
+    // Override the event to track the async copy.
+    if (attach_cuda_event) {
+        dst.override_event(CudaEvent::make_shared_record(stream));
     }
```
pentschev, line 127: Do we not need to override the event to the `src` as well? I _think_ it's unsafe not to, for example, if `cudaFreeAsync(src, src_stream)`, where `src_stream` is a different stream than the `stream` passed to `buffer_copy`, this will be potentially UB.

cpp/src/shuffler/chunk.cpp
```diff
@@ -356,8 +363,14 @@ Chunk Chunk::concat(
...
-            std::ignore = chunk.data_->copy_to(
-                *concat_data, std::ptrdiff_t(curr_data_offset), stream, false
+            buffer_copy(
+                *concat_data,
+                *chunk.data_,
+                chunk.data_->size,
+                std::ptrdiff_t(curr_data_offset),  // dst_offset
+                0,  // src_offset
+                stream,
+                false
```
pentschev, line 373: Just to be sure, this is `false` here because in line 394 below (new line number) the event is overridden, is that correct?

cpp/tests/test_buffer_resource.cpp
```diff
@@ -486,10 +465,17 @@ class BufferResourceCopyToTest : public BaseBufferResourceCopyTest,
         std::size_t const dest_offset
     ) {
         auto length = source->size;
-        auto bytes_written = source->copy_to(*dest, dest_offset, stream);
+        buffer_copy(
+            *dest,
+            *source,
+            source->size,
+            std::ptrdiff_t(dest_offset),  // dst_offset
```
pentschev, line 472: Why converting to `std::ptrdiff_t` explicitly here, but not for example in line 385 above? Just oversight or is there a different purpose?

cpp/src/buffer/buffer.cpp
madsbk: good catch

pentschev (approved):
LGTM, thanks Mads.

PR #514, opened 2025-09-17
Adding a streaming shuffler wrapper (`ShufflerAsync`) - Take 2

nirandaperera:
This PR adds a `ShufflerAsync` class that wraps a `Shuffler` object and exposes the following async methods. 

```c++

    /**
     * @brief Asynchronously extracts all data for a specific partition.
     *
     * This coroutine will suspend until the specified partition is ready for extraction
     * (i.e., insert_finished has been called for this partition and all data has been
     * shuffled).
     *
     * @param pid The partition ID to extract data for.
     * @return A coroutine task that yields a vector of PackedData chunks for the
     * partition.
     *
     * @throws std::runtime_error if the partition ID is not found.
     *
     * @note Users should be careful when using `extract_async` and `extract_any_async`
     * together, because a pid intended for `extract_async` may be extracted by
     * `extract_any_async`.
     */
    coro::task<std::vector<PackedData>> extract_async(shuffler::PartID pid);

    /**
     * @brief Result type for extract_any_async operations.
     *
     * Contains the partition ID and associated data chunks from an extract operation.
     * Can represent either a valid result or an invalid/empty result.
     *
     * An invalid result is returned when no more partitions are available for extraction.
     */
    struct ExtractResult {
        shuffler::PartID pid;  ///< The partition ID that was extracted
        std::vector<PackedData> chunks;  ///< The data chunks for this partition

        /// @brief A sentinel value for an invalid partition ID.
        static constexpr auto InvalidPID = std::numeric_limits<shuffler::PartID>::max();

        /**
         * @brief Checks if this result represents valid extracted data.
         *
         * @return true if this result contains valid data, false otherwise.
         */
        constexpr bool is_valid() const {
            return pid != InvalidPID;
        }

        /**
         * @brief Creates an invalid/empty result.
         *
         * @return An ExtractResult with an invalid partition ID and empty chunks.
         */
        static constexpr ExtractResult invalid() {
            return {InvalidPID, {}};
        }
    };

    /**
     * @brief Asynchronously extracts data for any ready partition.
     *
     * This coroutine will suspend until at least one partition is ready for extraction,
     * then extract and return the data for one such partition. If no partitions become
     * ready and the shuffle is finished, returns an invalid result.
     *
     * @return A coroutine task that yields an ExtractResult containing the partition ID
     *         and data chunks, or an invalid result if no more partitions are available.
     *
     * @note Users should be careful when using `extract_async` and `extract_any_async`
     * together, because a pid intended for `extract_async` may be extracted by
     * `extract_any_async`.
     */
    coro::task<ExtractResult> extract_any_async();
```

### DIFF
--- a/cpp/include/rapidsmpf/streaming/cudf/shuffler.hpp
+++ b/cpp/include/rapidsmpf/streaming/cudf/shuffler.hpp
@@ -117,51 +117,24 @@ class ShufflerAsync {
      * @brief Result type for extract_any_async operations.
      *
      * Contains the partition ID and associated data chunks from an extract operation.
-     * Can represent either a valid result or an invalid/empty result.
-     *
-     * An invalid result is returned when no more partitions are available for extraction.
      */
-    struct ExtractResult {
-        shuffler::PartID pid;  ///< The partition ID that was extracted
-        std::vector<PackedData> chunks;  ///< The data chunks for this partition
-
-        /// @brief A sentinel value for an invalid partition ID.
-        static constexpr auto InvalidPID = std::numeric_limits<shuffler::PartID>::max();
-
-        /**
-         * @brief Checks if this result represents valid extracted data.
-         *
-         * @return true if this result contains valid data, false otherwise.
-         */
-        [[nodiscard]] constexpr bool is_valid() const {
-            return pid != InvalidPID;
-        }
-
-        /**
-         * @brief Creates an invalid/empty result.
-         *
-         * @return An ExtractResult with an invalid partition ID and empty chunks.
-         */
-        static constexpr ExtractResult invalid() {
-            return {.pid = InvalidPID, .chunks = {}};
-        }
-    };
+    using ExtractResult = std::pair<shuffler::PartID, std::vector<PackedData>>;
 
     /**
      * @brief Asynchronously extracts data for any ready partition.
      *
      * This coroutine will suspend until at least one partition is ready for extraction,
      * then extract and return the data for one such partition. If no partitions become
-     * ready and the shuffle is finished, returns an invalid result.
+     * ready and the shuffle is finished, returns a nullopt.
      *
-     * @return ExtractResult containing the partition ID and data chunks, or an invalid
-     * result if no more partitions are available.
+     * @return ExtractResult containing the partition ID and data chunks, or a nullopt
+     * if no more partitions are available.
      *
      * @warning Users should be careful when using `extract_async` and `extract_any_async`
      * together, because a pid intended for `extract_async` may be extracted by
      * `extract_any_async`.
      */
-    coro::task<ExtractResult> extract_any_async();
+    coro::task<std::optional<ExtractResult>> extract_any_async();
 
   private:
     coro::mutex mtx_{};
--- a/cpp/src/streaming/cudf/shuffler.cpp
+++ b/cpp/src/streaming/cudf/shuffler.cpp
@@ -18,11 +18,8 @@ namespace rapidsmpf::streaming {
 namespace {
 
 /**
- * @brief Caller side corouting that inserts a partition ID into a set and notifies all
+ * @brief Caller side coroutine that inserts a partition ID into a set and notifies all
  * waiting tasks.
- *
- * Wrapping mtx lock and cv notify in a coroutine task to avoid calling coro::sync_wait
- * multiple times.
  *
  * @param mtx The mutex to use for synchronization.
  * @param cv The condition variable to use for notification.
@@ -37,7 +34,8 @@ coro::task<void> insert_and_notify(
     shuffler::PartID pid
 ) {
     // Note: this coroutine is not needed to be scheduled, because it is called from the
-    // progress thread.
+    // progress thread. Wrapping mtx lock and cv notify in a coroutine task will avoid
+    // calling coro::sync_wait multiple times.
     {
         auto lock = co_await mtx.scoped_lock();
         set.insert(pid);
@@ -63,7 +61,7 @@ ShufflerAsync::ShufflerAsync(
           stream,
           ctx_->br(),
           [this](shuffler::PartID pid) -> void {
-              ctx_->comm()->logger().trace("inserting finished partition ", pid);
+              ctx_->comm()->logger().trace("notifying waiters that ", pid, " is ready");
               coro::sync_wait(insert_and_notify(mtx_, cv_, ready_pids_, pid));
           },
           ctx_->statistics(),
@@ -120,7 +118,8 @@ coro::task<std::vector<PackedData>> ShufflerAsync::extract_async(shuffler::PartI
     co_return std::move(chunks);
 }
 
-coro::task<ShufflerAsync::ExtractResult> ShufflerAsync::extract_any_async() {
+coro::task<std::optional<ShufflerAsync::ExtractResult>>
+ShufflerAsync::extract_any_async() {
     // wait until at least one partition is ready for extraction
     auto lock = co_await mtx_.scoped_lock();
     co_await cv_.wait(lock, [this]() {
@@ -132,7 +131,7 @@ coro::task<ShufflerAsync::ExtractResult> ShufflerAsync::extract_any_async() {
     if (ready_pids_.empty()) {
         lock.unlock();
         ctx_->comm()->logger().trace("no partitions to extract");
-        co_return ExtractResult::invalid();
+        co_return std::nullopt;
     }
 
     auto pid = ready_pids_.extract(ready_pids_.begin()).value();
@@ -146,7 +145,7 @@ coro::task<ShufflerAsync::ExtractResult> ShufflerAsync::extract_any_async() {
         co_await cv_.notify_all();
     }
 
-    co_return {pid, std::move(chunks)};
+    co_return std::make_pair(pid, std::move(chunks));
 }
 
 namespace node {
@@ -192,16 +191,16 @@ Node shuffler(
     std::iota(finished.begin(), finished.end(), 0);
     shuffler_async.insert_finished(std::move(finished));
 
-    while (!shuffler_async.finished()) {
-        auto result = co_await shuffler_async.extract_any_async();
-        if (!result.is_valid()) {
-            break;
-        }
-
+    for (shuffler::PartID i = 0; i < shuffler_async.total_num_partitions(); i++) {
+        auto finished = co_await shuffler_async.extract_any_async();
+        RAPIDSMPF_EXPECTS(finished.has_value(), "Invalid result received");
         co_await ch_out->send(
-            std::make_unique<PartitionVectorChunk>(result.pid, std::move(result.chunks))
+            std::make_unique<PartitionVectorChunk>(
+                finished->first, std::move(finished->second)
+            )
         );
     }
+    RAPIDSMPF_EXPECTS(shuffler_async.finished(), "Shuffler not finished");
 
     co_await ch_out->drain(ctx->executor());
 }
--- a/cpp/tests/streaming/test_shuffler.cpp
+++ b/cpp/tests/streaming/test_shuffler.cpp
@@ -322,13 +322,14 @@ TEST_P(ShufflerAsyncTest, multi_consumer_extract) {
 
         while (!shuffler->finished()) {
             auto result = co_await shuffler->extract_any_async();
-            if (!result.is_valid()) {
+            if (!result.has_value()) {
                 break;
             }
 
             auto lock = co_await mtx.scoped_lock();
-            n_chunks_received += result.chunks.size();
-            finished_pids.push_back(result.pid);
+            auto& [pid, chunks] = *result;
+            n_chunks_received += chunks.size();
+            finished_pids.push_back(pid);
         }
         ctx->comm()->logger().debug(tid, " extract task finished");
     };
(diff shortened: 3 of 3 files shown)


cpp/include/rapidsmpf/shuffler/shuffler.hpp
```diff
@@ -143,6 +143,9 @@ class Shuffler {
 
     ~Shuffler();
 
+    Shuffler(const Shuffler&) = delete;
+    Shuffler& operator=(const Shuffler&) = delete;
```
madsbk, line 147:
```suggestion
    Shuffler(Shuffler const&) = delete;
    Shuffler& operator=(Shuffler const&) = delete;
```

cpp/include/rapidsmpf/streaming/cudf/shuffler.hpp
```diff
@@ -5,36 +5,202 @@
...
+        rmm::cuda_stream_view stream,
+        OpID op_id,
+        shuffler::PartID total_num_partitions,
+        shuffler::Shuffler::PartitionOwner partition_owner =
+            shuffler::Shuffler::round_robin
+    );
+
+    // Prevent copying
+    ShufflerAsync(const ShufflerAsync&) = delete;
+    ShufflerAsync& operator=(const ShufflerAsync&) = delete;
```
madsbk, line 58:
```suggestion
    ShufflerAsync(ShufflerAsync const&) = delete;
    ShufflerAsync& operator=(ShufflerAsync const&) = delete;
```

cpp/include/rapidsmpf/streaming/cudf/shuffler.hpp
```diff
@@ -5,36 +5,202 @@
...
+
+    ~ShufflerAsync() = default;
+
+    /**
+     * @brief Gets the streaming context associated with this shuffler.
+     *
+     * @return A reference to the shared context object.
+     */
+    constexpr auto& ctx() const {
```
madsbk, line 67: No `auto` in the public API
```suggestion
    constexpr std::shared_ptr<Context>& ctx() const {
```

cpp/include/rapidsmpf/streaming/cudf/shuffler.hpp
```diff
@@ -5,36 +5,202 @@
...
+    constexpr shuffler::PartID total_num_partitions() const {
+        return shuffler_.total_num_partitions;
+    }
 
+    /**
+     * @brief Gets the partition owner function used by this shuffler.
+     *
+     * @return A const reference to the function that maps partition IDs to owning ranks.
+     */
+    constexpr auto const& partition_owner() const {
```
madsbk, line 93:
```suggestion
    constexpr shuffler::Shuffler::PartitionOwner const& partition_owner() const {
```

madsbk (changes requested):
Overall looks great

cpp/tests/streaming/base_streaming_fixture.hpp
```diff
@@ -19,9 +19,19 @@
 class BaseStreamingFixture : public ::testing::Test {
   protected:
     void SetUp() override {
+        SetUp(1);  // default number of streaming threads
+    }
+
+    void SetUp(int num_streaming_threads) {
```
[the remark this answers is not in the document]
nirandaperera: That could conflict with the gtest test class `virtual SetUp()` method I think.

cpp/src/streaming/cudf/shuffler.cpp
```diff
@@ -11,36 +11,134 @@
...
+    auto lock = co_await mtx_.scoped_lock();
+    co_await cv_.wait(lock, [this]() {
+        return shuffler_.finished() || !ready_pids_.empty();
+    });
+
+    // no partitions to extract or shuffle is already finished. Gracefully return an
+    // invalid result.
+    if (ready_pids_.empty()) {
+        lock.unlock();
+        ctx_->comm()->logger().warn("no partitions to extract");
```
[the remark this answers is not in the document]
madsbk: Fair enough, but then we shouldn't warn.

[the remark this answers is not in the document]
nirandaperera: made it a trace

cpp/tests/streaming/test_shuffler.cpp
```diff
@@ -226,41 +197,261 @@ std::pair<Node, Node> shuffler_nb(
     // extract task: extract the packed chunks from the shuffler and send them to the
     // output channel
     auto extract_task = [](auto shuffler_ctx, auto ctx, auto ch_out) -> Node {
-        ShutdownAtExit c{ch_out};
+        ShutdownAtExit c{
+            ch_out
+        };  // TODO: could this be problematic with multiple consumers?
```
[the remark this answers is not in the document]
nirandaperera: I have fixed this in https://github.com/rapidsai/rapidsmpf/pull/516 with a `latch`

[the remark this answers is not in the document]
nirandaperera: This was not a problem with `single` communicator for some reason.

cpp/include/rapidsmpf/streaming/cudf/shuffler.hpp
```diff
@@ -5,36 +5,193 @@
...
+     * This coroutine will suspend until at least one partition is ready for extraction,
+     * then extract and return the data for one such partition. If no partitions become
+     * ready and the shuffle is finished, returns an invalid result.
+     *
+     * @return A coroutine task that yields an ExtractResult containing the partition ID
+     *         and data chunks, or an invalid result if no more partitions are available.
+     *
+     * @warning Users should be careful when using `extract_async` and `extract_any_async`
+     * together, because a pid intended for `extract_async` may be extracted by
+     * `extract_any_async`.
```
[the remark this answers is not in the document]
nirandaperera: @madsbk Hmm. It is still an issue TBH. For an example, say we had pids `0, 1`. And if `extract_any` and `extract(0)` was called in order, there's still no guarantee that `extract(0)` will receive it. Now, this will throw though. 
If we were to guarantee `extract(0)`, we will have to request a vector of `extract` pids to the shuffler constructor. Then those can be omitted by `extract_any`.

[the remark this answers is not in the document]
madsbk: Totally fine! but please add a description of the precise semantics of using extract_any and extract together. What guarantees do we provide, and what do we not provide?

cpp/include/rapidsmpf/streaming/cudf/utils.hpp
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Records @p event on @p secondary and inserts a wait for that event on @p primary.
+ * This is fully asynchronous with respect to the host thread; no host-side blocking.
+ *
+ * @param primary The stream that must not run ahead.
+ * @param secondary The stream whose already-enqueued work must complete first.
+ * @param event The CUDA event to use for synchronization. The same event may be reused
+ * across multiple calls; the caller does not need to provide an unique event each time.
+ */
+inline void sync_streams(
```
madsbk, line 28: Let's use the new [cuda_stream_join()](https://github.com/rapidsai/rapidsmpf/blob/f93215098f894f1914b380d02dbfc4b4f3708c02/cpp/include/rapidsmpf/cuda_stream.hpp#L33)

cpp/include/rapidsmpf/streaming/cudf/shuffler.hpp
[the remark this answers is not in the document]
nirandaperera: done

cpp/include/rapidsmpf/streaming/cudf/shuffler.hpp
```diff
@@ -5,36 +5,195 @@
...
+
+        /// @brief A sentinel value for an invalid partition ID.
+        static constexpr auto InvalidPID = std::numeric_limits<shuffler::PartID>::max();
+
+        /**
+         * @brief Checks if this result represents valid extracted data.
+         *
+         * @return true if this result contains valid data, false otherwise.
+         */
+        [[nodiscard]] constexpr bool is_valid() const {
```
wence-, line 136: noexcept? or is that implied by constexpr, I forget as always

cpp/src/streaming/cudf/shuffler.cpp
```diff
@@ -10,37 +10,147 @@
...
 
-namespace rapidsmpf::streaming::node {
+namespace rapidsmpf::streaming {
 
 namespace {
 
 /**
- * @brief Make @p primary wait until all work currently enqueued on @p secondary
- * completes.
+ * @brief Caller side corouting that inserts a partition ID into a set and notifies all
```
wence-, line 21: coroutine

cpp/src/streaming/cudf/shuffler.cpp
```diff
@@ -10,37 +10,147 @@
...
+ * waiting tasks.
  *
- * Records @p event on @p secondary and inserts a wait for that event on @p primary.
- * This is fully asynchronous with respect to the host thread; no host-side blocking.
+ * Wrapping mtx lock and cv notify in a coroutine task to avoid calling coro::sync_wait
+ * multiple times.
```
wence-, line 25: This feels like a comment on the implementation, rather than documentation on how to use the function.

cpp/src/streaming/cudf/shuffler.cpp
```diff
@@ -10,37 +10,147 @@
...
+        return shuffler_.finished() || !ready_pids_.empty();
+    });
+
+    // no partitions to extract or shuffle is already finished. Gracefully return an
+    // invalid result.
+    if (ready_pids_.empty()) {
+        lock.unlock();
+        ctx_->comm()->logger().trace("no partitions to extract");
+        co_return ExtractResult::invalid();
+    }
```
wence-, line 136: question: We're using a sentinel to indicate an invalid extract result. Would it be more ergonomic to use either `coro::expected` (or `std::expected`) or `std::optional`? Then we don't have to have a "different" type of optional thing as a return value.

cpp/src/streaming/cudf/shuffler.cpp
```diff
@@ -10,37 +10,147 @@
...
+    // before the pid was inserted into ready_pids_)
+    RAPIDSMPF_EXPECTS(
+        !extracted_pids_.contains(pid) && ready_pids_.erase(pid) > 0,
+        "partition already extracted or not found: " + std::to_string(pid),
+        std::out_of_range
+    );
+    extracted_pids_.emplace(pid);
+    lock.unlock();  // no longer need the lock
+
+    auto chunks = shuffler_.extract(pid);
```
wence-, line 113: This might fail (because `extract_any_async` can sneak in and steal the PID from us).

What about if we just drop the lock _after_ `shuffler_.extract(pid)` (and similarly in `extract_any_async`)? Then it is safe for `extract` and `extract_any_async` to run simultaneously I think.

cpp/src/streaming/cudf/shuffler.cpp
```diff
@@ -74,29 +176,35 @@ Node shuffler(
...
 
-    while (!shuffler.finished()) {
-        auto finished_partition = shuffler.wait_any();
-        auto packed_chunks = shuffler.extract(finished_partition);
         co_await ch_out->send(
-            std::make_unique<PartitionVectorChunk>(
-                sequence_number, std::move(packed_chunks), stream
-            )
+            std::make_unique<PartitionVectorChunk>(result.pid, std::move(result.chunks))
         );
```
wence-, line 203: We know how many tasks we must suspend now, I think. So could we change this to:

```
for (int i = 0; i < num_partitions; i++) {
    auto finished = co_await shuffler_async.extract_any_async();
    RAPIDSMPF_EXPECTS(finished.is_valid(), "Something went wrong");
    co_await ch_out->send(std::make_unique<PartitionVectorChunk>(result.pid, std::move(result.chunks)));
}
```

cpp/tests/streaming/test_shuffler.cpp
```diff
@@ -226,41 +201,263 @@ std::pair<Node, Node> shuffler_nb(
...
+TEST_P(StreamingShuffler, callbacks_1_consumer) {
+    EXPECT_NO_FATAL_FAILURE(run_test([&](auto ch_in, auto ch_out) -> Node {
+        return shuffler_nb(
+            ctx, stream, std::move(ch_in), std::move(ch_out), op_id, num_partitions, 1
+        );
+    }));
+}
+
+TEST_P(StreamingShuffler, callbacks_2_consumer) {
+    GTEST_SKIP() << "unreliable test";  // TODO: fix this
```
wence-, line 256: What is unreliable about this?

cpp/src/streaming/cudf/shuffler.cpp
```diff
@@ -10,37 +10,147 @@
...
+    : ctx_(std::move(ctx)),
+      shuffler_(
+          ctx_->comm(),
+          ctx_->progress_thread(),
+          op_id,
+          total_num_partitions,
+          stream,
+          ctx_->br(),
+          [this](shuffler::PartID pid) -> void {
+              ctx_->comm()->logger().trace("inserting finished partition ", pid);
```
wence-, line 66: *nit*: This message is slightly misleading, since we might expect that to mean that `insert_finished` has been called with this pid.

How about something like `"notifying waiters that ", pid, " is ready"`?

wence- (commented):
Only small suggestions from me.

cpp/tests/streaming/test_shuffler.cpp
nirandaperera: Multiple consumers were unreliable when I pin the tests to a single core. It was deadlocking intermittently. 
This will be fixed in https://github.com/rapidsai/rapidsmpf/pull/516

cpp/src/streaming/cudf/shuffler.cpp
nirandaperera: I moved this to a `std::optional` return.

PR #516, opened 2025-09-18
Enabling distributed streaming tests 

nirandaperera:
This PR enables distributed streaming tests. 

Depends on #514

### DIFF
--- a/cpp/src/streaming/cudf/shuffler.cpp
+++ b/cpp/src/streaming/cudf/shuffler.cpp
@@ -32,7 +32,7 @@ coro::task<void> insert_and_notify(
     std::unordered_set<shuffler::PartID>& set,
     shuffler::PartID pid
 ) {
-    // Note: this coroutine is not needed to be scheduled, because it is offloaded to the
+    // Note: this coroutine does not need to be scheduled, because it is offloaded to the
     // thread pool using `spawn`.
     {
         auto lock = co_await mtx.scoped_lock();


cpp/tests/CMakeLists.txt
```diff
@@ -131,16 +143,11 @@ if(RAPIDSMPF_HAVE_MPI)
     disable_sign_conversion_warning(ucxx_tests)
 
     rapidsmpf_mpirun_test_add(TEST_TARGET ucxx_tests)
-  endif()
-endif()
+  endif(RAPIDSMPF_HAVE_UCXX)
+endif(RAPIDSMPF_HAVE_MPI)
```
KyleFromNVIDIA, line 147: Please put this back the way it was. CMake has not required putting the condition in the `endif()` for well over a decade, and it is generally discouraged to do so.

```suggestion
  endif()
endif()
```

KyleFromNVIDIA (changes requested):

KyleFromNVIDIA (approved):

cpp/tests/streaming/test_shuffler.cpp
```diff
@@ -195,48 +223,70 @@ Node shuffler_nb(
         std::vector<rapidsmpf::shuffler::PartID> finished(total_num_partitions);
         std::iota(finished.begin(), finished.end(), 0);
         shuffler_ctx->shuffler->insert_finished(std::move(finished));
+
```
madsbk, line 226:
```suggestion
```

cpp/tests/streaming/test_shuffler.cpp
```diff
@@ -195,48 +223,70 @@ Node shuffler_nb(
...
-                std::make_unique<PartitionVectorChunk>(
-                    *expected, std::move(packed_chunks)
-                )
+                std::make_unique<PartitionVectorChunk>(pid, std::move(packed_chunks))
             );
 
             if (shuffler_ctx->shuffler->finished()) {
                 // if the shuffler is finished, shutdown & drain the ready_pids queue
                 co_await shuffler_ctx->ready_pids.shutdown_drain(ctx->executor());
+                // co_await shuffler_ctx->ready_pids.shutdown();
```
madsbk, line 261:
```suggestion
```

madsbk (approved):
Looks good

cpp/tests/streaming/test_shuffler.cpp
```diff
@@ -195,48 +223,70 @@ Node shuffler_nb(
...
+            // order.
+            auto result = shuffler_ctx->ready_pids.try_pop();
+            rapidsmpf::shuffler::PartID pid;
+            if (result.has_value()) {
+                pid = *result;  // partition id is found
+            } else if (result.error() == coro::queue_consume_result::stopped) {
+                break;  // queue is shutdown, so exit the loop
+            } else {  // queue is empty or busy, so yield and continue
+                co_await ctx->executor()->yield();
+                continue;
```
wence-, line 249: Can you explain how the deadlock may occur? It seems like `pop` either returns a value (if someone has `push`ed), a sentinel that the queue is shutdown, or suspends (if nothing is available). So I don't see where the deadlock occurs?

cpp/tests/streaming/test_shuffler.cpp
```diff
@@ -195,48 +223,70 @@ Node shuffler_nb(
...
             co_await ch_out->send(
-                std::make_unique<PartitionVectorChunk>(
-                    *expected, std::move(packed_chunks)
-                )
+                std::make_unique<PartitionVectorChunk>(pid, std::move(packed_chunks))
             );
 
             if (shuffler_ctx->shuffler->finished()) {
                 // if the shuffler is finished, shutdown & drain the ready_pids queue
                 co_await shuffler_ctx->ready_pids.shutdown_drain(ctx->executor());
```
wence-, line 260: The person _producing_ values into `ready_pids` should shut down the queue I think.

cpp/tests/streaming/test_shuffler.cpp
```diff
@@ -195,48 +223,70 @@ Node shuffler_nb(
...
+                // co_await shuffler_ctx->ready_pids.shutdown();
+                co_await ctx->executor()->yield();
```
wence-, line 262: IIUC you have these various yield points in and they are required for _correctness_. I think that speaks to underlying issues in the control flow. I think suspending the coroutine by inserting a yield point should never be necessary for correctness, only performance (in terms of wanting other tasks to run preferentially).

wence- (changes requested):
I am concerned about the various comments about deadlocks, and the use of yield and/or manual control flow to work around them. Can you please explain what's going on?

cpp/src/streaming/cudf/shuffler.cpp
```diff
@@ -191,16 +199,17 @@ Node shuffler(
...
-        RAPIDSMPF_EXPECTS(finished.has_value(), "Invalid result received");
+        if (!finished.has_value()) {
+            break;
+        }
         co_await ch_out->send(
             std::make_unique<PartitionVectorChunk>(
                 finished->first, std::move(finished->second)
             )
         );
     }
```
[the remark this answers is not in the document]
nirandaperera: currently we dont hold the local pids as a class member in the shuffler. We have a static util to create a vector of local pids. Maybe its something we should do?

cpp/tests/streaming/base_streaming_fixture.hpp
```diff
@@ -16,26 +16,37 @@
...
+    void SetUpWithThreads(int num_streaming_threads) {
+        // create a new options object, since we can not modify values in the global
+        // options object
+        rapidsmpf::config::Options options(GlobalEnvironment->options().get_strings());
         RAPIDSMPF_EXPECTS(
             options.insert_if_absent(
                 "num_streaming_threads", std::to_string(num_streaming_threads)
             ),
             "num_streaming_threads already set"
         );
```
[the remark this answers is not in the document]
wence-: Well just overwrite whatever is already set then, no?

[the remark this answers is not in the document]
nirandaperera: I asked @madsbk about overwriting values in `Options`. His comment was, "yeah, updating values will complicate things a lot " :sweat_smile: 
That's why I took the deep-copy approach

cpp/src/streaming/cudf/shuffler.cpp
```diff
@@ -33,13 +32,15 @@ coro::task<void> insert_and_notify(
     std::unordered_set<shuffler::PartID>& set,
     shuffler::PartID pid
 ) {
-    // Note: this coroutine is not needed to be scheduled, because it is called from the
-    // progress thread. Wrapping mtx lock and cv notify in a coroutine task will avoid
-    // calling coro::sync_wait multiple times.
+    // Note: this coroutine is not needed to be scheduled, because it is offloaded to the
+    // thread pool using `spawn`.
```
wence-, line 36:
```suggestion
    // Note: this coroutine does not need to be scheduled, because it is offloaded to the
    // thread pool using `spawn`.
```

cpp/tests/streaming/test_shuffler.cpp
wence-: Ah ok, that's too complicated.

cpp/tests/streaming/test_shuffler.cpp
```diff
@@ -200,43 +241,54 @@ Node shuffler_nb(
...
+    // shutdown task: shutdown the shuffler after all extract tasks have finished
+    auto shutdown_task =
+        [](auto shuffler_ctx, auto ctx, auto ch_out, auto& latch) -> Node {
+        ShutdownAtExit c{ch_out};
+        co_await ctx->executor()->schedule();
+
+        co_await latch;  // wait for all extract tasks to finish before clean up
         co_await ch_out->drain(ctx->executor());
+
+        shuffler_ctx->shuffler->shutdown();
```
wence-, line 278: This doesn't block because at this point we've extracted everything? I suppose it might block if we are still sending data to other people?

PR #519, opened 2025-09-22
A `Buffer` now tracts `last_write_event`

madsbk:
Replace manual attach/wait on buffer events with implicitly updated `Buffer::last_write_event`.


cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -136,31 +186,35 @@ class Buffer {
...
+     * @return `true` if the last recorded write event has completed; `false` otherwise.
      *
-     * @throws rapidsmpf::cuda_error if event wait fails (if set).
+     * @code{.cpp}
+     * // Example: send the buffer via MPI (non-stream-ordered).
+     * if (buffer.is_latest_write_done()) {
+     *   MPI_Isend(buffer.data(), buffer.size(), MPI_BYTE, dst, tag, comm, &req);
+     * } else {
+     *   // Ensure completion before handing to MPI.
+     *   cudaStreamSynchronize(buffer.stream());
```
wence-, line 210: To facilitate this usage, and to avoid oversynchronisation, we should introduce `buffer.wait_for_writes()` that does `cudaEventSynchronize` on the `latest_write_event_`

cpp/src/buffer/buffer.cpp
```diff
@@ -98,19 +74,18 @@ void buffer_copy(
...
+    dst.write_access(dst.stream(), [&](std::byte* dst_data) {
+        RAPIDSMPF_CUDA_TRY(cudaMemcpyAsync(
+            dst_data + dst_offset,
+            src.data() + src_offset,
+            size,
+            cudaMemcpyDefault,
+            dst.stream()
+        ));
+    });
+    cuda_stream_join(std::array{src.stream()}, std::array{dst.stream()});
```
wence-, line 89: We could potentially keep a (per-context and) per-thread cache of events around to use inside `cuda_stream_join`. I think because we never query the context of the stream, and are just using the runtime api, it only needs to be per-thread.

cpp/src/communicator/mpi.cpp
```diff
@@ -122,34 +124,31 @@ std::unique_ptr<Communicator::Future> MPI::send(
...
     );
     MPI_Request req;
-    RAPIDSMPF_MPI(MPI_Irecv(
-        recv_buffer->data(), recv_buffer->size, MPI_UINT8_T, rank, tag, comm_, &req
-    ));
+    recv_buffer->write_access(rmm::cuda_stream_default, [&](std::byte* recv_buffer_data) {
+        RAPIDSMPF_MPI(MPI_Irecv(
+            recv_buffer_data, recv_buffer->size, MPI_UINT8_T, rank, tag, comm_, &req
+        ));
+    });
```
wence-, line 151: *issue*: This oversynchronises with any work that happens to be queued on the default stream. You need this because you need access to the data pointer of the `Buffer` and this is now the only way to do that.

But after making this call, the `is_latest_write_done` method on `recv_buffer` is a lie. Because the write is not complete until the request is complete (rather than the work that happened to be recorded on the default stream when we made this call).

cpp/src/communicator/ucxx.cpp
```diff
@@ -1091,28 +1091,27 @@ std::unique_ptr<Communicator::Future> UCXX::send(
...
+    RAPIDSMPF_EXPECTS(recv_buffer->is_latest_write_done(), "msg must be ready");
+    auto req = recv_buffer->write_access(
+        recv_buffer->stream(), [&](std::byte* recv_buffer_data) {
+            return get_endpoint(rank)->tagRecv(
+                recv_buffer_data,
+                recv_buffer->size,
+                tag_with_rank(rank, tag),
+                ::ucxx::TagMaskFull
+            );
+        }
```
wence-, line 1114: *issue*: This has the same kind of issue as the same thing in `MPI::recv`.

Also, this also is indicative that `is_latest_write_done` is a bit of a misnomer. In this case what it means is "is the allocation done" (and I suppose, if there are writes to this buffer, are they flushed?)

cpp/src/cuda_event.cpp
```diff
@@ -37,37 +36,35 @@ CudaEvent& CudaEvent::operator=(CudaEvent&& other) {
...
-[[nodiscard]] bool CudaEvent::CudaEvent::is_ready() {
-    if (!done_.load(std::memory_order_acquire)) {
-        auto result = cudaEventQuery(event_);
-        done_.store(result == cudaSuccess, std::memory_order_release);
-        if (result != cudaSuccess && result != cudaErrorNotReady) {
-            RAPIDSMPF_CUDA_TRY(result);
-        }
-        return result == cudaSuccess;
+[[nodiscard]] bool CudaEvent::CudaEvent::is_ready() const {
+    std::lock_guard<std::mutex> lock(done_mutex_);
```
wence-, line 51: I wonder if we should just stop caching the result and always defer to `cudaEventQuery`.

cpp/tests/test_buffer_resource.cpp
```diff
@@ -384,10 +263,9 @@ class BufferResourceCopySliceTest
             0,  // dst_offset
             std::ptrdiff_t(offset)  // src_offset
         );
-
         EXPECT_EQ(slice->mem_type(), dest_type);
-        slice->wait_for_ready();
-        EXPECT_TRUE(slice->is_ready());
+        slice->stream().synchronize();
```
wence-, line 267: *suggestion*: again, we should expose the event so we do `slice->event.host_wait()` here.

cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -91,20 +97,64 @@ class Buffer {
...
-     * @brief Access the underlying memory buffer (host or device memory).
+     * @brief Provides write access to the buffer.
      *
-     * @return A const pointer to the underlying host or device memory.
+     * Calls @p f with a pointer to the buffer's memory. The callable must be invocable
+     * as `R(std::byte*)`; its return value (if any) is returned by this function.
      *
-     * @throws std::logic_error if the buffer does not manage any memory.
+     * The provided @p stream is the stream associated with this access. Any work enqueued
+     * on the buffer memory must use @p stream or synchronize with it before @p f returns.
```
wence-, line 113: Surely synchronize with it before f is _called_?

wence- (changes requested):

cpp/src/buffer/buffer.cpp
```diff
@@ -37,22 +36,47 @@ Buffer::Buffer(std::unique_ptr<rmm::device_buffer> device_buffer)
...
+    }
+}
+
+Buffer::DeviceStorageT const& Buffer::device() const {
+    RAPIDSMPF_EXPECTS(!is_locked(), "the buffer is locked");
+    if (const auto* ref = std::get_if<DeviceStorageT>(&storage_)) {
+        return *ref;
+    } else {
+        RAPIDSMPF_FAIL("Buffer is not device memory");
+    }
```
madsbk, line 75: Unchanged, just moved from `.hpp`

cpp/src/buffer/buffer.cpp
madsbk: But don't we get to the whole _CUDA static object problem_?

cpp/src/communicator/mpi.cpp
madsbk: Introducing:
```c++
    /**
     * @brief Acquire non-stream-ordered exclusive access to the buffer's memory.
     *
     * Alternative to `write_access()`. Acquires an internal exclusive lock so that
     * **any other access through the Buffer API** (including `write_access()`) will
     * fail with `std::logic_error` while the lock is held. The lock remains held
     * until `unlock()` is called.
     *
     * Use this when integrating with non-stream-aware consumer APIs that require a
     * raw pointer and cannot be expressed as work on a CUDA stream (e.g., MPI,
     * blocking host I/O).
     *
     * @note Prefer `write_access(stream, ...)` if you can express the operation as a
     * single callable on a stream, even if that requires manually synchronizing the
     * stream before the callable returns.
     *
     * @return Pointer to the underlying storage.
     *
     * @throws std::logic_error If the buffer is already locked.
     * @throws std::logic_error If `is_latest_write_done() != true`.
     *
     * @see write_access(), is_locked(), unlock()
     */
    std::byte* exclusive_data_access();
```

cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -80,31 +81,113 @@ class Buffer {
...
+        -> std::invoke_result_t<F, std::byte*> {
+        static_assert(
+            std::is_invocable_v<std::remove_reference_t<F>, std::byte*>,
+            "write_access() expects a callable with signature: R(std::byte*)"
+        );
+        using R = std::invoke_result_t<F, std::byte*>;
+
+        // After `f()` completes, an event is recorded on `stream`; this becomes the new
+        // latest-write event.
+        if constexpr (std::is_void_v<R>) {
```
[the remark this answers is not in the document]
madsbk: `data()` will throw if the buffer is locked.

> Or may be lock it and unlock at the end?

The lock is not for concurrency; it only exists to catch accidental access.

cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -80,31 +81,113 @@ class Buffer {
...
-     * @brief Access the underlying memory buffer (host or device memory).
+     * @brief Check whether the buffer is currently exclusively locked.
      *
-     * @return A const pointer to the underlying host or device memory.
+     * @return `true` if `exclusive_data_access()` has acquired the lock and `unlock()`
+     * has not yet been called; `false` otherwise.
+     */
+    [[nodiscard]] bool is_locked() const {
+        return lock_.load(std::memory_order_acquire);
+    }
```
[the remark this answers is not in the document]
madsbk: It is now a private method `throw_if_locked()`:

```c++
    /**
     * @brief Throws if the buffer is currently locked by `exclusive_data_access()`.
     *
     * @throws std::logic_error If the buffer is locked.
     */
    void throw_if_locked() const;
```

cpp/src/communicator/mpi.cpp
```diff
@@ -122,33 +124,34 @@ std::unique_ptr<Communicator::Future> MPI::send(
...
-    }
+    RAPIDSMPF_EXPECTS(recv_buffer->is_latest_write_done(), "msg must be ready");
     RAPIDSMPF_EXPECTS(
         recv_buffer->size <= std::numeric_limits<int>::max(),
         "recv buffer size exceeds MPI max count"
     );
     MPI_Request req;
     RAPIDSMPF_MPI(MPI_Irecv(
-        recv_buffer->data(), recv_buffer->size, MPI_UINT8_T, rank, tag, comm_, &req
+        recv_buffer->exclusive_data_access(),
```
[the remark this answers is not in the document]
madsbk: True, but `exclusive_data_access()` is the only why to get raw write access to the buffer now!

cpp/src/communicator/mpi.cpp
wence-: Yes, I think this looks like a good idea.

cpp/src/buffer/buffer.cpp
wence-: A few options:

1. Make them per-`BufferResource` object which is not static
2. Make them static but just don't destruct them

But I suspect this is a microopt we should revisit if we see things in the profiles.

cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -80,31 +81,105 @@ class Buffer {
...
+
+    /**
+     * @brief Acquire non-stream-ordered exclusive access to the buffer's memory.
+     *
+     * Alternative to `write_access()`. Acquires an internal exclusive lock so that
+     * **any other access through the Buffer API** (including `write_access()`) will
+     * fail with `std::logic_error` while the lock is held. The lock remains held
+     * until `unlock()` is called. This lock is not a concurrency mechanism; it only
+     prevents accedenalt access to the Buffer while  it simultaneous access through the
+     Buffer API.
```
wence-, line 156:
```suggestion
     * until `unlock()` is called. This lock is not a concurrency mechanism; it only
     * prevents accidental access to the Buffer through the rest of the Buffer API while locked.
```

Or something?

cpp/include/rapidsmpf/buffer/buffer.hpp
```diff
@@ -80,31 +81,105 @@ class Buffer {
...
+     *
+     * Use this when integrating with non-stream-aware consumer APIs that require a
+     * raw pointer and cannot be expressed as work on a CUDA stream (e.g., MPI,
+     * blocking host I/O).
+     *
+
+     *
```
wence-, line 163:
```suggestion
```

cpp/include/rapidsmpf/buffer/buffer.hpp
wence-: Ah, after discussion, no. However, I think we need to write down more carefully what the stream-ordered semantics of this `f` are required to be.

Suppose I have a buffer `buf`, with associated stream `streamA`. I have a stream I am going to do work on `streamB`, and some other streams `streamC`, ..., `streamX`.

Now I call:

```
buf.write_access(streamB, [&](std::byte *ptr) { do_something_that_modifies_ptr_on(streamB); });
```

What is my job? I think it is the following:

1. ensure that operations enqueued on `streamB` are ordered after any outstanding work on `streamA`
2. ensure that, after I enqueue the new operation on `streamB`, that `streamA` waits for that work.

(1) is so that I don't have a race modifying `buf`, (2) is so that if someone subsequently enqueues work on `buf` on its stream (`streamA`) they don't race with my modification.

So I think this means minimally my function must be:

```
buf.write_access(streamB, [&](std::byte *ptr) {
  cudaEventRecord(event, streamA);
  cudaStreamWaitEvent(streamB, event);
  do_work_on(streamB, ptr);
  cudaEventRecord(event, streamB);
  cudaStreamWaitEvent(streamA, event);
}
```

`write_access` then additionally records the work on `streamB` as the `latest_write_event`.

But IIUC this pattern will _always_ be needed. If all the operations are on `streamA` when we don't need the waits and the "pre" record, so we do:

```
buf.write_access(streamA, [&](std::byte *ptr) {
  do_work_on(streamA, ptr);
  cudaEventRecord(event, streamA); // this is equivalent to latest_write_event.record(streamA);
}
```

Could we therefore make this more foolproof to use by always doing these synchronisations. That way the promise for the user is easier to understand. "Inside your function, it is always safe to access `ptr` on the provided stream, after the function returns it is safe to access `buf` on its stream".

Something like:

```
auto write_access(rmm::cuda_stream_view stream, F&& f)
{
   auto const is_buf_stream = stream == stream();
   if (!is_buf_stream) {
      latest_write_event_.record(stream());
      latest_write_event_.stream_wait(stream);
   }
   std::invoke(f, ...);
   latest_write_event_.record(stream);
   if (!is_buf_stream) {
       latest_write_event_.stream_wait(stream());
   }
}
```

WDYT?
}

cpp/src/buffer/buffer.cpp
```diff
@@ -98,19 +153,19 @@ void buffer_copy(
...
+    dst.write_access(dst.stream(), [&](std::byte* dst_data) {
+        RAPIDSMPF_CUDA_TRY(cudaMemcpyAsync(
+            dst_data + dst_offset,
+            src.data() + src_offset,
+            size,
+            cudaMemcpyDefault,
+            dst.stream()
+        ));
+    });
+    cuda_stream_join(std::views::single(src.stream()), std::views::single(dst.stream()));
```
wence-, line 168: Question: to my mind it is clearer what is going on if the `cuda_stream_join`s are _inside_ the function. Is that still correct?

i.e.

```
dst.write_access(dst.stream(), [&](std::byte *dst_data) {
   cuda_stream_join(dst, src);
   cudamemcpyasync(dst, src);
   cuda_stream_join(src, dst);
});
```

cpp/src/buffer/resource.cpp
```diff
@@ -155,18 +156,12 @@ std::unique_ptr<Buffer> BufferResource::allocate(
...
-    return std::unique_ptr<Buffer>(new Buffer(std::move(data), stream));
-}
-
 std::unique_ptr<Buffer> BufferResource::move(
     std::unique_ptr<rmm::device_buffer> data, rmm::cuda_stream_view stream
 ) {
     auto upstream = data->stream();
     if (upstream.value() != stream.value()) {
-        cuda_stream_join(std::array{stream}, std::array{upstream});
+        cuda_stream_join(std::views::single(stream), std::views::single(upstream));
```
wence-, line 164: *question*: this is so common, should we provide an overload for `cuda_stream_join`?

cpp/src/communicator/ucxx.cpp
```diff
@@ -1091,25 +1091,20 @@ std::unique_ptr<Communicator::Future> UCXX::send(
 std::unique_ptr<Communicator::Future> UCXX::send(
     std::unique_ptr<Buffer> msg, Rank rank, Tag tag
 ) {
-    if (!msg->is_ready()) {
-        logger().warn("msg is not ready. This is irrecoverable, terminating.");
-        std::terminate();
-    }
+    RAPIDSMPF_EXPECTS(msg->is_latest_write_done(), "msg must be ready");
+    auto data = const_cast<std::byte*>(msg->data());
```
wence-, line 1095: Why not `exclusive_data_access`?

cpp/src/communicator/mpi.cpp
```diff
@@ -122,33 +124,34 @@ std::unique_ptr<Communicator::Future> MPI::send(
...
-    }
+    RAPIDSMPF_EXPECTS(msg->is_latest_write_done(), "msg must be ready");
     RAPIDSMPF_EXPECTS(
         msg->size <= std::numeric_limits<int>::max(),
         "send buffer size exceeds MPI max count"
     );
     MPI_Request req;
-    RAPIDSMPF_MPI(MPI_Isend(msg->data(), msg->size, MPI_UINT8_T, rank, tag, comm_, &req));
+    auto data = const_cast<std::byte*>(msg->data());
+    RAPIDSMPF_MPI(MPI_Isend(data, msg->size, MPI_UINT8_T, rank, tag, comm_, &req));
```
wence-, line 134: Why not `exclusive_data_access`?

wence- (commented):
I think my main open query is about docs for `write_access`. I am still a bit unsure if we are clear enough about what the user is expected to do and what we do for them. Additionally, whether or not we can make usage a bit more foolproof.

cpp/src/communicator/mpi.cpp
madsbk: but we don't need the `const_cast`, that was some old code. We now have:
```c++
RAPIDSMPF_MPI(MPI_Isend(msg->data(), msg->size, MPI_UINT8_T, rank, tag, comm_, &req));
```

cpp/include/rapidsmpf/buffer/buffer.hpp
madsbk: We now have:
```c++

    /**
     * @brief Provides stream-ordered write access to the buffer.
     *
     * Calls @p f with a pointer to the buffer's memory and the buffer's stream
     * (i.e., `this->stream()`).
     *
     * The callable must be invocable as:
     *   - `R(std::byte*, rmm::cuda_stream_view)`.
     *
     * All work performed by @p f must be stream-ordered on the buffer's stream.
     * Enqueuing work on any other stream without synchronizing with the buffer's
     * stream before and after the call is undefined behavior. In other words,
     * @p f must behave as a single stream-ordered operation, similar to issuing one
     * `cudaMemcpyAsync` on the buffer's stream. For non-stream-aware integrations,
     * use `exclusive_data_access()`.
     *
     * After @p f returns, an event is recorded on the buffer's stream, establishing
     * the new "latest write" for this buffer.
     *
     * @warning The pointer is valid only for the duration of the call. Using it
     * outside of @p f is undefined behavior.
     *
     * @tparam F Callable type.
     * @param f Callable that accepts `(std::byte*, rmm::cuda_stream_view)`.
     * @return Whatever @p f returns (`void` if none).
     *
     * @throws std::logic_error If the buffer is locked.
     *
     * @code{.cpp}
     * // Snippet: copy data from `src_ptr` into `buffer` on the buffer's stream.
     * buffer.write_access([&](std::byte* buffer_ptr, rmm::cuda_stream_view stream) {
     *   assert(buffer.stream().value() = stream.value());
     *   RAPIDSMPF_CUDA_TRY_ALLOC(cudaMemcpyAsync(
     *       buffer_ptr,
     *       src_ptr,
     *       num_bytes,
     *       cudaMemcpyDefault,
     *       stream
     *   ));
     * });
     * @endcode
     */
```

PR #528, opened 2025-09-24
Reuse existing Memory Resource in `rmpf_worker_setup` when possible

### LINKED ISSUE #523: Support user-provided Memory Resources
In https://github.com/rapidsai/cudf/pull/19895, I hit an issue with how rapidsmpf [creates and sets](https://github.com/rapidsai/rapidsmpf/blob/34d160647aed684c3c5f64da304b386e7938b058/python/rapidsmpf/rapidsmpf/integrations/core.py#L444-L455) new memory resources as part of the worker setup for the dask and single shufflers. In that PR, cudf-polars wants all allocations to be made on a [StatisticsResourceAdaptor](https://docs.rapids.ai/api/rmm/stable/python_api/#rmm.mr.StatisticsResourceAdaptor) so it can track memory usage. This runs into an error as part of a cudf-polars shuffle using rapidsmpf: rapidsmpf changes the memory resource out from under cudf-polars and an error is raised.

Concretely, I'd like to change the worker setup function to check what memory resource it gets from `rmm.mr.get_current_device_resource()`. As long as that Memory Resource satisfies rapidsmpf's needs (IIUC, that it's a `RmmResourceAdaptor`) then rapidsmpf should use it rather than creating a new one.

Alternatively, users could (optionally) pass in a concrete memory resource and have rapidsmpf use that, if it's valid, if we don't want to rely on passing around the MR via the global device resource. Either way, rapidsmpf will still set the memory resource as RMM's current device resource, so maybe getting it from the global isn't so bad.

There's one complication around handling RMM statistics. A couple spots in RMM specifically want a `RmmResourceAdaptor` rather than an arbitrary RMM memory resource. We'll need to either update those spots to handle `RmmStatisticsAdaptor(RmmResourceAdaptor(...))`, or extract the `RmmResourceAdaptor` from the `RmmStatisticsAdaptor` before handing them in.

TomAugspurger:
This updates `rmpf_worker_setup` to reuse the current device resource (i.e. not layer a new resource on top) when:

1. It's already an `RmmResourceAdaptor`
2. It's a `rmm.mr.StatisticsResourceAdaptor` whose upstream MR is an `RmmResourceAdaptor`

I think that supporting 1 should be safe. I'm hoping that supporting number 2 is also safe, but there are a couple spots (our `Statistics` and `LimitAvailableMemory`) that really do want a `RmmResourceAdaptor`. For now I've extracted the `RmmResourceAdaptor` from the `StatisticsResourceAdaptor`. I'm guessing that any allocations made directly against that `RmmResourceAdaptor` will throw off the RMM statistics.

Closes https://github.com/rapidsai/rapidsmpf/issues/523


python/rapidsmpf/rapidsmpf/integrations/core.py
```diff
@@ -440,23 +440,35 @@ def rmpf_worker_setup(
...
-            # Use a managed memory resource if OOM protection is enabled.
-            rmm.mr.ManagedMemoryResource()
-            if options.get_or_default(
-                f"{option_prefix}oom_protection", default_value=False
-            )
-            else None
-        ),
-    )
+    # Ensure that an RMM resource adaptor (or a StatisticsResourceAdaptor wrapping one)
+    # is on top of the current RMM resource stack.
```
madsbk, line 444: `RmmResourceAdaptor` doesn't have to be on top, as long as it is in the stack.

madsbk (commented):
I think your current approach, retrieving the `RmmResourceAdaptor` from the stack, is good!

But consider implementing [`get_rmm_memory_resource_stack()`](https://github.com/rapidsai/cudf/blob/ffa7ea2d298cbc85fa4c3dade3921fec07721e77/python/cudf/cudf/core/buffer/spill_manager.py#L40) to retrieve the current `RmmResourceAdaptor` and if none exists, create (and push) a new adaptor on the top of the stack.

madsbk (changes requested):
Overall looks good

python/rapidsmpf/rapidsmpf/integrations/core.py
```diff
@@ -440,23 +466,30 @@ def rmpf_worker_setup(
...
-            if options.get_or_default(
-                f"{option_prefix}oom_protection", default_value=False
-            )
-            else None
-        ),
-    )
-    rmm.mr.set_current_device_resource(mr)
+    # Ensure that an RMM resource adaptor is present in the current RMM resource stack.
+    mr = rmm.mr.get_current_device_resource()
+
```
madsbk, line 471:
```suggestion
```

python/rapidsmpf/rapidsmpf/tests/test_dask.py
```diff
@@ -117,14 +119,23 @@ def test_dask_cudf_integration(
...
+    sort: bool,
     cluster_kind: Literal["distributed", "single", "auto"],
+    preconfigure_mr: bool,
 ) -> None:
     # Test single-worker cuDF integration with Dask-cuDF
     pytest.importorskip("dask_cudf")
 
+    if preconfigure_mr:
+        mr = rapidsmpf.rmm_resource_adaptor.RmmResourceAdaptor(
+            rmm.mr.CudaMemoryResource()
```
TomAugspurger, line 135: This should also use the `device_mr` fixture

TomAugspurger:
Hmm, I guess a test is hanging: https://github.com/rapidsai/rapidsmpf/actions/runs/17982528193/job/51154821364?pr=528#step:10:2118... Trying to reproduce locally.

PR #529, opened 2025-09-24
Update `RAPIDS_BRANCH`, codify changes in `update-version.sh`

KyleFromNVIDIA:
This PR updates `RAPIDS_BRANCH` and modifies `update-version.sh` to make future changes automatically.


bdice (changes requested):
It seems like this repo isn't using `RAPIDS_BRANCH` right now, but it should. We might need to update `cmake/rapids_config.cmake` and `cmake/RAPIDS.cmake` to align with what we have in other repos.

PR #549, opened 2025-10-01
Adding pinned host buffer impl 

nirandaperera:
This PR adds `cccl::cudax` to the build along side a `PinnedHostBuffer` impl similar to `rmm::device_buffer`.

### DIFF
--- a/cpp/tests/CMakeLists.txt
+++ b/cpp/tests/CMakeLists.txt
@@ -175,9 +175,9 @@ target_link_libraries(
           GTest::gmock
           GTest::gtest
           $<TARGET_NAME_IF_EXISTS:conda_env>
+          $<$<BOOL:${RAPIDSMPF_HAVE_NUMA}>:numa>
           maybe_asan
           test_sources
-          $<$<BOOL:${RAPIDSMPF_HAVE_NUMA}>:numa>
 )
 disable_sign_conversion_warning(single_tests)
 add_test(NAME single_tests COMMAND "gtests/single_tests")


cpp/include/rapidsmpf/buffer/pinned_memory_resource.hpp
```diff
@@ -0,0 +1,236 @@
...
+     */
+    constexpr void set_stream(cuda::stream_ref stream) noexcept {
+        stream_ref_ = stream;
+    }
+
+    void* data_ = nullptr;  ///< Pointer to the allocated buffer data.
+    size_t size_;  ///< Size of the buffer in bytes.
+    cuda::stream_ref stream_ref_;  ///< CUDA stream used for memory operations.
+    PinnedMemoryResource*
+        p_resource_;  ///< Pointer to the memory resource used for allocation.
```
madsbk, line 233:
```suggestion
        mr_;  ///< Pointer to the memory resource used for allocation.
```

cpp/include/rapidsmpf/buffer/pinned_memory_resource.hpp
```diff
@@ -0,0 +1,236 @@
...
+    [[nodiscard]] constexpr void* data() const noexcept {
+        return data_;
+    }
+
+    /**
+     * @brief Gets a pointer to the buffer data.
+     *
+     * @return A pointer to the buffer data, or nullptr if not allocated.
+     */
+    constexpr void* data() noexcept {
```
madsbk, line 198:
```suggestion
    constexpr std::byte* data() noexcept {
```

cpp/include/rapidsmpf/buffer/pinned_memory_resource.hpp
```diff
@@ -0,0 +1,236 @@
...
+     * @param data Pointer to the source data to copy.
+     * @param size The size of the data to copy in bytes.
+     * @param stream The CUDA stream to use for memory operations.
+     * @param p_resource Pointer to the memory resource to use for allocation and
+     * deallocation.
+     */
+    PinnedHostBuffer(
+        void const* data,
+        size_t size,
+        cuda::stream_ref stream,
```
madsbk, line 164: Let's be consistent, either `rmm::cuda_stream_view` or `cuda::stream_ref` everywhere 
```suggestion
        rmm::cuda_stream_view stream,
```

cpp/include/rapidsmpf/buffer/pinned_memory_resource.hpp
```diff
@@ -0,0 +1,236 @@
...
+    /**
+     * @brief Sets the CUDA stream for this buffer.
+     *
+     * @param stream The new CUDA stream to use for memory operations.
+     */
+    constexpr void set_stream(cuda::stream_ref stream) noexcept {
+        stream_ref_ = stream;
+    }
+
+    void* data_ = nullptr;  ///< Pointer to the allocated buffer data.
```
madsbk, line 229:
```suggestion
    std::byte* data_ = nullptr;  ///< Pointer to the allocated buffer data.
```

cpp/include/rapidsmpf/buffer/pinned_memory_resource.hpp
```diff
@@ -0,0 +1,236 @@
...
+        cuda::stream_ref stream,
+        PinnedMemoryResource* p_resource
+    );
+
+    /**
+     * @brief Destroys the pinned host buffer and waits for the associated stream to
+     * complete.
+     */
+    ~PinnedHostBuffer() noexcept;
+
```
madsbk, line 173: Since the dtor deallocate the buffer, we need to implement the move operators and set `data_ = nullptr`.

cpp/src/buffer/pinned_memory_resource.cu
```diff
@@ -0,0 +1,119 @@
...
+      impl_(
+          std::make_unique<PinnedMemoryPoolImpl>(
+              numa_id, get_memory_pool_properties(properties_)
+          )
+      ) {}
+
+PinnedMemoryPool::~PinnedMemoryPool() = default;
+
+// PinnedMemoryResource implementation
+struct PinnedMemoryResource::PinnedMemoryResourceImpl {
```
madsbk, line 49: why the `PinnedMemoryResourceImpl` ?

cpp/include/rapidsmpf/buffer/pinned_memory_resource.hpp
nirandaperera: Good point. Let me fix that.

cpp/include/rapidsmpf/buffer/pinned_memory_resource.hpp
```diff
@@ -0,0 +1,236 @@
...
+};
+
+/**
+ * @brief A buffer that manages pinned host memory.
+ *
+ * This class provides a convenient interface for managing pinned host memory
+ * buffers. It automatically handles allocation and deallocation through a
+ * PinnedMemoryResource and supports both empty buffer creation and buffer
+ * creation with data copying.
+ */
```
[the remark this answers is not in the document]
nirandaperera: IINM `buf.data()` will be immediately available, but we should not use it until the allocation completes. 
https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY__POOLS.html#group__CUDART__MEMORY__POOLS_1g871003f518e27ec92f7b331307fa32d4

But it says, 
>A pointer to the allocated memory is returned immediately in *dptr. The allocation must not be accessed until the the allocation operation completes.

cpp/include/rapidsmpf/buffer/pinned_memory_resource.hpp
nirandaperera: I'm more inclined to use `cuda::stream_ref ` because we are not using any rmm dependencies in this class. rmm stream view is implicitly convertible to cuda stream ref. So, it won't be a big issue.

wence-: Is stream_ref implicitly convertible to cuda_stream_view? If not that is painful with the rest of rapidsmpf

cpp/include/rapidsmpf/buffer/pinned_memory_resource.hpp
nirandaperera: this is now fixed

cpp/src/buffer/pinned_memory_resource.cu
bdice: This is very similar to what we’ve proposed as a way to make RMM ABI-stable.

cmake/thirdparty/get_cccl_cudax.cmake
```diff
@@ -0,0 +1,36 @@
...
+
+# This function finds CCCL with CUDAX.
+function(find_and_configure_cccl_cudax)
+  # TODO: Replace this with rapids_cpm_cccl(... ENABLE_UNSTABLE) once
+  # #https://github.com/rapidsai/rapids-cmake/pull/919 is merged.
+
+  message(STATUS "Finding CCCL with CUDAX")
+
+  set(CCCL_VERSION 3.1.0)
+  set(CCCL_TAG v3.1.0-rc8)
```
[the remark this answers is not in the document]
nirandaperera: I am going to replace this with `rapids_cpm_cccl` once https://github.com/rapidsai/rapids-cmake/pull/919 merges

cpp/include/rapidsmpf/buffer/pinned_memory_resource.hpp
```diff
@@ -0,0 +1,346 @@
...
+            PinnedHostBuffer ret(src.size(), stream, std::move(mr));
+            // synchronize the downstream stream with upstream stream
+            cuda_stream_join(stream, src.stream());
+            RAPIDSMPF_CUDA_TRY(cudaMemcpyAsync(
+                ret.data(), src.data(), src.size(), cudaMemcpyDefault, stream
+            ));
+            return ret;
+        }
+        return PinnedHostBuffer(0, stream, std::move(mr));
+    }
```
madsbk, line 251: I suggest you drop the `stream` argument and just let the returned `PinnedHostBuffer` adapt the source's stream. 
Then we don't need `stream_synchronized_copy` but instead a device buffer overload:
```c++
    PinnedHostBuffer(
        rmm::device_buffer const& other,
        std::shared_ptr<PinnedMemoryResource> mr
    );
```

cpp/include/rapidsmpf/buffer/pinned_memory_resource.hpp
```diff
@@ -0,0 +1,346 @@
...
+     * @throws rapidsmpf::cuda_error if the stream synchronization fails.
+     */
+    void synchronize();
+
+  private:
+    std::byte* data_ = nullptr;  ///< Pointer to the allocated buffer data.
+    size_t size_;  ///< Size of the buffer in bytes.
+    rmm::cuda_stream_view stream_;  ///< CUDA stream used for memory operations.
+    std::shared_ptr<PinnedMemoryResource>
+        mr_;  ///< Shared pointer to the memory resource used for allocation/deallocation.
```
madsbk, line 343:
```suggestion
    std::shared_ptr<PinnedMemoryResource> mr_;  ///< Memory resource used for allocation/deallocation.
```

cpp/include/rapidsmpf/buffer/pinned_memory_resource.hpp
```diff
@@ -0,0 +1,406 @@
...
+    // only be used by the impl in .cu file.
+    struct PinnedMemoryPoolImpl;
+    std::unique_ptr<PinnedMemoryPoolImpl> impl_;
+};
+
+/**
+ * @brief A memory resource that allocates/deallocates pinned host memory from a pinned
+ * host memory pool. This internally uses
+ * `cuda::experimental::pinned_memory_resource`. See for more details:
+ * https://nvidia.github.io/cccl/cudax/api/classcuda_1_1experimental_1_1pinned__memory__resource.html
```
pentschev:
```suggestion
 * `cuda::experimental::pinned_memory_resource`.
 *
 * @sa https://nvidia.github.io/cccl/cudax/api/classcuda_1_1experimental_1_1pinned__memory__resource.html
```

cpp/include/rapidsmpf/buffer/pinned_memory_resource.hpp
```diff
@@ -0,0 +1,406 @@
...
+class PinnedMemoryResource;  // forward declaration
+
+/**
+ * @brief Properties for configuring a pinned memory pool. It is aimed to mimic
+ * `cuda::experimental::memory_pool_properties`. See for more details:
+ * https://nvidia.github.io/cccl/cudax/api/structcuda_1_1experimental_1_1memory__pool__properties.html
+ *
+ * Currently, this is a placeholder and does not have any effect. It was observed that
+ * priming async pools have little effect for performance. See for more details:
+ * https://github.com/rapidsai/rmm/issues/1931
```
pentschev:
```suggestion
 * priming async pools have little effect for performance.
 *
 * @sa https://github.com/rapidsai/rmm/issues/1931
```

cpp/include/rapidsmpf/buffer/pinned_memory_resource.hpp
```diff
@@ -0,0 +1,406 @@
...
+     * @brief Deallocates pinned memory asynchronously with alignment.
+     *
+     * @param stream The CUDA stream to use for the deallocation operation.
+     * @param ptr A pointer to the memory to deallocate.
+     * @param bytes The size of the memory to deallocate.
+     * @param alignment The alignment that was used for allocation.
+     */
+    void deallocate(
+        rmm::cuda_stream_view stream, void* ptr, size_t bytes, size_t alignment
+    ) noexcept;
```
pentschev: How do you even use alignment for deallocation? Isn't the pointer all you need/should use?

nirandaperera:
> I still think we don't necessarily need quite so much stubbed out code raising errors but hopefully all this will go away soon when we can use the cccl things directly

@wence- I agree. Created an issue for this https://github.com/rapidsai/rapidsmpf/issues/589

cpp/src/utils.cpp
```diff
@@ -74,19 +75,15 @@ bool parse_string(std::string const& value) {
...
     static const int numa_node_id = [] {
-        RAPIDSMPF_EXPECTS(nvmlInit() == NVML_SUCCESS, "failed to initialize NVML");
-        nvmlDevice_t device;
         RAPIDSMPF_EXPECTS(
-            nvmlDeviceGetHandleByIndex(0, &device) == NVML_SUCCESS,
-            "failed to get device handle"
+            numa_available() >= 0, "NUMA is not available", std::runtime_error
         );
-        uint32_t numa_node_id = 0;
+        int cpu = sched_getcpu();
```
pentschev: Ok, I think this works but _only if the CPU affinity has been properly set when launching the process_. Could you document that?

cpp/tests/test_utils.cpp
```diff
@@ -47,3 +47,7 @@ TEST(ParseStringTest, ParsesBooleansWithCaseAndWhitespace) {
 TEST(ParseStringTest, ThrowsOnInvalidBoolean) {
     EXPECT_THROW(parse_string<bool>("not_a_bool"), std::invalid_argument);
 }
+
+TEST(GetCurrentNumaNodeIdTest, ReturnsValidNumaNodeId) {
+    EXPECT_EQ(0, get_current_numa_node_id());  // default NUMA node ID is 0
```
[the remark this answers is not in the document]
nirandaperera: That's right. But if we were to test check it, we'll have to add the `get_current_numa_node_id` impl again on the test body, isnt it?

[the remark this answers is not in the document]
pentschev: I think we should just relax the condition to `>= 0`, and if possible, check the number of NUMA nodes of the system to check `0 < get_current_numa_node_id() < total_numa_nodes`.

cpp/CMakeLists.txt
```diff
@@ -72,6 +74,7 @@ set(RAPIDSMPF_HAVE_CUPTI ${BUILD_CUPTI_SUPPORT})
 set(RAPIDSMPF_BUILD_TESTS ${BUILD_TESTS})
 set(RAPIDSMPF_BUILD_BENCHMARKS ${BUILD_BENCHMARKS})
 set(RAPIDSMPF_BUILD_EXAMPLES ${BUILD_EXAMPLES})
+set(RAPIDSMPF_HAVE_NUMA ${BUILD_NUMA_SUPPORT})
```
pentschev: Ditto

cpp/include/rapidsmpf/utils.hpp
```diff
@@ -377,4 +380,23 @@ struct overloaded : Ts... {
...
+/**
+ * @brief Gets the NUMA node ID of the current CPU process.
+ *
+ * @note This function is only available if built with NUMA support. (See
+ * `RAPIDSMPF_NUMA_SUPPORT` CMake option.)
+ *
+ * @return The NUMA node ID of the current CPU process.
+ *
+ * @throws std::runtime_error If NUMA is not available or if the NUMA node ID cannot be
+ * retrieved.
```
pentschev:
```suggestion
 * @throws std::runtime_error If built with NUMA support but libnuma is not available
 * at runtime or if the NUMA node ID cannot be retrieved.
```

KyleFromNVIDIA (changes requested):

PR #557, opened 2025-10-02
Refactor `ShufflerAsync`

madsbk:
Clean up ShufflerAsync's extract methods.

### DIFF
--- a/cpp/src/streaming/cudf/shuffler.cpp
+++ b/cpp/src/streaming/cudf/shuffler.cpp
@@ -93,7 +93,9 @@ bool ShufflerAsync::all_extracted_unsafe() const {
     return extracted_pids_.size() == shuffler_.local_partitions().size();
 }
 
-coro::task<std::vector<PackedData>> ShufflerAsync::extract_async(shuffler::PartID pid) {
+coro::task<std::optional<std::vector<PackedData>>> ShufflerAsync::extract_async(
+    shuffler::PartID pid
+) {
     // Wait until the partition is finished
     auto lock = co_await mtx_.scoped_lock();
 
@@ -108,11 +110,9 @@ coro::task<std::vector<PackedData>> ShufflerAsync::extract_async(shuffler::PartI
     });
 
     // Did we wake up because all partitions have been extracted?.
-    RAPIDSMPF_EXPECTS(
-        !all_extracted_unsafe(),
-        "all partition have been extracted already",
-        std::out_of_range
-    );
+    if (all_extracted_unsafe()) {
+        co_return std::nullopt;
+    }
 
     // pid has now been extracted and isn't ready anymore.
     RAPIDSMPF_EXPECTS(ready_pids_.erase(pid) > 0, "something went wrong");
--- a/cpp/src/streaming/cudf/shuffler.cpp
+++ b/cpp/src/streaming/cudf/shuffler.cpp
@@ -104,33 +104,20 @@ coro::task<std::vector<PackedData>> ShufflerAsync::extract_async(shuffler::PartI
     );
 
     co_await cv_.wait(lock, [this, pid]() {
-        // Note: purposefully not checking for extracted_pids_.contains(pid) here,
-        // because it would require notifying the cv every time a partition is extracted.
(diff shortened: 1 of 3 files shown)


cpp/src/streaming/cudf/shuffler.cpp
```diff
@@ -104,33 +102,18 @@ coro::task<std::vector<PackedData>> ShufflerAsync::extract_async(shuffler::PartI
...
+    // Did we wake up because all partitions have been extracted?.
+    if (all_extracted_unsafe()) {
+        co_return std::nullopt;
     }
 
-    co_return std::move(chunks);
+    // pid has now been extracted and isn't ready anymore.
+    RAPIDSMPF_EXPECTS(ready_pids_.erase(pid) > 0, "something went wrong");
+    RAPIDSMPF_EXPECTS(extracted_pids_.emplace(pid).second, "something went wrong");
+    co_return shuffler_.extract(pid);
```
nirandaperera, line 116: I think we need to notify `cv_`, if this is the last PID. Otherwise we'd have to rely on spurious wake-ups in some edge cases.

nirandaperera (changes requested):
IMO we need to keep that cv notify after the extraction.

cpp/src/streaming/cudf/shuffler.cpp
madsbk: But we notify **all**. So if we have many consumers waiting, all of them wakes up

nirandaperera: Consumers (extractors) could miss the notification from the callback. Lets take 1 partition 2 consumers again. 
Say, by the time extract any is scheduled, the shuffle has completed message passing and the callbacks have already been called. Then consumers miss the cv notification. One consumer will bypass cv wait and consume the partition, but the other would still be waiting on the cv.

wence-: we only suspend if the predicate is not satisfied. in your example I think the predicate is satisfied so we won't suspend, no?

specifically the predicate is called with the lock held so if two people come in for one partition one of them gets the lock extracts the partition with the lock then unlocks which wakes the next person who sees that the predicate is true and returns because everything has been extracted

nirandaperera: Ah! good point @wence- . You are absolutely right. If we had two extract_any consumers for a single partition, it is true. 

Okay, let me see if I could counter. 

Say we have many consumers, and 2 partitions (0, 1). consumer 1 extracts PID 0, while other consumers  extracts any. Say, an other consumer locks the mutex first and extracts PID 0. Now, when consumer 1 locks the mutex, it would go into waiting, because not all partitions have been extracted yet, and PID 0 is not available. Later, some other consumer extracts PID 1. But consumer 1 will still be waiting, because no one has is notifying the cv, isnt it?

madsbk: Good catch, I have update the wait in `extract_async()!:
```c++
    co_await cv_.wait(lock, [this, pid]() {
        return ready_pids_.contains(pid) || extracted_pids_.contains(pid);
    });
```
A partition ID is always in exactly one of the two sets:
     - `ready_pids_`: partitions ready for extraction but not yet extracted.
     - `extracted_pids_`: partitions that have already been extracted.

Agree?

cpp/src/streaming/cudf/shuffler.cpp
```diff
@@ -89,79 +99,62 @@ void ShufflerAsync::insert_finished(std::vector<shuffler::PartID>&& pids) {
...
-        std::out_of_range
-    );
+    // Ensure that `pid` is owned by this rank.
+    {
+        auto pids = shuffler_.local_partitions();
+        RAPIDSMPF_EXPECTS(
+            std::ranges::find(pids, pid) != pids.end(),
+            "the pid isn't owned by this rank, see ShufflerAsync::partition_owner()",
+            std::out_of_range
+        );
```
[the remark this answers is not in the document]
madsbk: good idea, let's use `partition_owner`!

PR #559, opened 2025-10-06
Table chunk `is_spillable()`

madsbk:
Introduce a `is_spillable()` method to `TableChunk`:

```c++
    /**
     * @brief Indicates whether this chunk can be spilled.
     *
     * A table chunk is considered spillable if it was created from one of the following:
     *   - A device-owning source such as a `cudf::table`, `cudf::packed_columns`, or
     *     `PackedData`.
     *   - A `cudf::table_view` constructed with `is_exclusive_view == true`, indicating
     *     that the view is the sole representation of the underlying table and its
     *     associated owner exclusively manages the table's memory.
     *
     * In contrast, chunks constructed from non-exclusive `cudf::table_view` instances are
     * non-owning views of externally managed memory and therefore not spillable.
     *
     * @return `true` if the chunk can be spilled; otherwise, `false`.
     */
    [[nodiscard]] bool is_spillable() const;
```

### DIFF
--- a/python/rapidsmpf/rapidsmpf/streaming/cudf/table_chunk.pxd
+++ b/python/rapidsmpf/rapidsmpf/streaming/cudf/table_chunk.pxd
@@ -3,7 +3,7 @@
 
 from libc.stddef cimport size_t
 from libc.stdint cimport uint64_t
-from libcpp cimport bool
+from libcpp cimport bool as bool_t
 from libcpp.memory cimport unique_ptr
 from pylibcudf.libcudf.table.table_view cimport table_view as cpp_table_view
 from rmm.librmm.cuda_stream_view cimport cuda_stream_view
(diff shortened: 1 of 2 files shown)


nirandaperera (commented):
minor comment. I have a feeling that we can consolidate some of the class members.

cpp/include/rapidsmpf/streaming/cudf/table_chunk.hpp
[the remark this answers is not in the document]
madsbk: Not now, I like to keep it simple for now :)

cpp/include/rapidsmpf/streaming/cudf/table_chunk.hpp
```diff
@@ -51,25 +51,36 @@ class TableChunk {
...
+     * the @p owner will correctly free the associated device memory.
      */
     TableChunk(
         std::uint64_t sequence_number,
         cudf::table_view table_view,
         std::size_t device_alloc_size,
         rmm::cuda_stream_view stream,
-        OwningWrapper&& owner = {}
+        OwningWrapper&& owner,
+        bool is_exclusive_view
```
wence-, line 83: nitpick, bool arguments are never really very self-documenting.

What about:

```
enum class ExclusiveView : bool {
   NO,
   YES,
};

...
TableChunk(..., ExclusiveView::YES);
```
?

cpp/src/streaming/cudf/table_chunk.cpp
```diff
@@ -141,14 +159,18 @@ TableChunk TableChunk::spill_to_host(BufferResource* br) {
...
+            packed_data = std::make_unique<PackedData>(
+                std::move(packed_columns.metadata),
+                br->move(std::move(packed_columns.gpu_data), stream_)
+            );
         }
     }
+
     // Spill data to host memory.
     auto [res, _] = br->reserve(MemoryType::HOST, packed_data->data->size, false);
     packed_data->data = br->move(std::move(packed_data->data), res);
```
wence-, line 172: I guess not for this PR, but all the above methods calling `cudf::pack` are making allocations with reservations...

wence- (approved):
Trivial suggestion

PR #565, opened 2025-10-07
[MINOR] Fix Spill Manager remove spill function deadlock

nirandaperera:
Observed a deadlock recently, and there was a legit scenario where Spill manager is being shutdown, while another thread calls `remove_spill_function` and it hangs, because `pause` never completes. 

I think we dont _really_ need to wait for the background thread to pause. Indicating it to pause is sufficient.

### DIFF
--- a/cpp/src/buffer/spill_manager.cpp
+++ b/cpp/src/buffer/spill_manager.cpp
@@ -58,9 +58,9 @@ void SpillManager::remove_spill_function(SpillFunctionID fid) {
     }
     spill_functions_.erase(fid);
 
-    // Pause the spill thread if no spill functions are left.
+    // Pause asynchronously the spill thread if no spill functions are left.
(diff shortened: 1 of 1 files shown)


madsbk:
I mean, use `PausableThreadLoop::pause_nb()` instead of `PausableThreadLoop::pause()` everywhere?

nirandaperera:
> I mean, use `PausableThreadLoop::pause_nb()` instead of `PausableThreadLoop::pause()` everywhere?

@madsbk Yeah, that should be good IMO.

nirandaperera:
> > > I mean, use `PausableThreadLoop::pause_nb()` instead of `PausableThreadLoop::pause()` everywhere?
> > 
> > 
> > @madsbk Yeah, that should be good IMO.
> 
> Can you do it in this PR?

Sure will do.

PR #574, opened 2025-10-10
Introduce a streaming read_parquet node

### LINKED ISSUE #599: Ordering guarantees (or not) in rapidsmpf channels
I was looking through https://github.com/rapidsai/cudf/pull/20161 and noticed that it is making, in various places, assumptions about the sequence numbers of messages in a channel that are not guaranteed to hold.

Today, we only guarantee:

1. All messages have unique sequence numbers, the total order of messages in a channel is defined by order of the sequence numbers
2. The order that messages are _actually_ pushed into a channel is the order message are pulled out.

Note that if we have multiple coroutines pushing into the same channel, we do not provide guarantees on the order that those coroutines execute.

As a consequence the only thing a consumer of a channel can rely on is point (1).

In rapidsai/cudf#20161 we're making some stronger assumptions in various cases. Mostly to be able to match up messages from different channels.

1. sequence numbers are dense in `[0, ..., num_message_in_channel)`
2. Certain nodes are "implicitly known" to be single-producer coroutines, and as such messages come out in sequence number order.

As a consequence of these two assumptions we do "too much" work in some cases and "too little" in others.

Example 1: to match up chunks from channels in an `HConcat` (horizontally gluing tables together) we currently have to buffer in the worst case all the inputs from all the channels to match them up. We then assume that we can match things by the sequence number (rather than the order their sequence number induces).

Example 2: in a shuffle join, we assume that the left and right channels are delivering the hash partitioned outputs in the same order. This is true as an implementation detail today, but technically we can't rely on it.

I think we need to strengthen the guarantees on messages in channels.

Specifically, I think that _all_ channels should produce messages in sequence number order, and (additionally) sequence numbers should become an "implementation detail" (used to impose ordering, but the consumer should not care about them for ordering purposes).

I actually don't want to maintain uniqueness of sequence numbers (you can get order-preservation of the probe side of a join "for free" by allowing output ids from a given probe chunk to just be replicated, as long as you obey "channels are ordered", for example). There are also cases where I would like to relax the "sequence numbers are dense" constraint (I can't currently use the chunked parquet reader in IO because I can't guarantee that it would produce only a single output chunk).

Particularly, what I want to offer as a guarantee is:

1. A single producer pushing into a channel will always push that data in "in order"
1. If you have a single consumer from a channel, it always pulls data "in order"
1. Multiple consumers from a single channel will see a subset of the data "in order"
1. Multiple producers into a single channel will push their data in in a total order defined by sequence ids of the messages being inserted.

The first three are true today, the fourth is not. I propose that we implement a utility "lineariser" that a multi-producer node can use to linearise the inputs to its output channel.

We still need to maintain sequence numbers so we can implement the linearisation, but the consumer of channel messages should not need to use sequence numbers to reconstruct the order of channels.

Potential issues:

If we do this, a node that processes an input channel and wants to be "fault tolerant" can't easily "move on" to the next input (perhaps because it determines the first input is too large to process) and send it to the output channel and then go back to process the first input.

wence-:
The provided options are collective across the participating ranks. The input file list will be split (approximately evenly) across ranks.

The caller can control the number of rows that are read per chunk sent to the output channel, as well as the number producer tasks that may be waiting, suspended, having read a chunk. Increasing the number of producer tasks can increase pipelining and latency-hiding opportunities for the execution scheduler, at the cost of higher memory pressure.

To ensure that, in this multi-producer setup, chunks are still inserted into the output channel in increasing sequence number order, we implement a `Lineariser` utility that one should always use when sending from multiple producers. The idea is that it is created with one input channel per producer and each producer sends its sequence of chunks into the input channel in order, the lineariser buffers them and sends things out in global total order.

- Closes #599

### DIFF
--- a/cpp/src/streaming/cudf/parquet.cpp
+++ b/cpp/src/streaming/cudf/parquet.cpp
@@ -134,7 +134,7 @@ Node read_parquet(
         files.begin() + file_offset, files.begin() + file_offset + files_per_rank
     );
     cudf::io::parquet_reader_options local_options{options};
-    local_options.set_source(cudf::io::source_info(local_files));
+    local_options.set_source(cudf::io::source_info(std::move(local_files)));
     auto metadata = cudf::io::read_parquet_metadata(local_options.get_source());
     auto const local_num_rows = metadata.num_rows();
     auto skip_rows = options.get_skip_rows();
@@ -171,12 +171,12 @@ Node read_parquet(
         std::atomic<std::size_t> chunk_index{0};
         read_tasks.reserve(1 + num_producers);
         auto lineariser = std::make_shared<Lineariser>(ch_out, num_producers);
-        read_tasks.push_back(lineariser->drain(ctx));
         for (auto& ch_in : lineariser->get_inputs()) {
             read_tasks.push_back(
                 produce_chunks(ctx, ch_in, local_options, chunks, chunk_index)
             );
         }
+        read_tasks.push_back(lineariser->drain(ctx));
         coro_results(co_await coro::when_all(std::move(read_tasks)));
     }
     co_await ch_out->drain(ctx->executor());


cpp/include/rapidsmpf/streaming/cudf/parquet.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace rapidsmpf::streaming::node {
+
+/**
+ * @brief Asynchronously read parquet files into an output channel.
+ *
+ * @param ctx The execution context to use.
+ * @param ch_out Channel to which `TableChunk`s are sent.
+ * @param max_tickets Maximum number of tickets to throttle production of chunks. Up to
+ * this many tasks can start producing data simultaneously.
```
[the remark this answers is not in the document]
wence-: Added a check in the throttle ctor

cpp/src/streaming/cudf/parquet.cpp
```diff
@@ -0,0 +1,215 @@
...
+    RAPIDSMPF_EXPECTS(
+        files.size() < std::numeric_limits<int>::max(), "Trying to read too many files"
+    );
+    int files_per_rank =
+        static_cast<int>(files.size() / size + (rank < (files.size() % size)));
+    int file_offset = 0;
+    for (auto i = std::size_t{0}; i < rank; i++) {
+        file_offset +=
+            static_cast<int>(files.size() / size + (i < (files.size() % size)));
+    }
```
nirandaperera: Nit. I think we can drop the for-loop here. 
```c++
int file_offset = rank*(files.size()/size) + min(rank, files.size()%size));
```

cpp/src/streaming/cudf/parquet.cpp
```diff
@@ -0,0 +1,215 @@
...
+    auto options_skip_rows = options.get_skip_rows();
+    auto options_num_rows =
+        options.get_num_rows().value_or(std::numeric_limits<int64_t>::max());
+    std::uint64_t sequence_number = 0;
+    if (options_num_rows == 0) {
+        if (rank == 0) {
+            // Rank zero sends empty table of the correct shape/schema. Everyone else
+            // sends nothing.
+            cudf::io::parquet_reader_options empty_opts{options};
+            cudf::io::source_info source{files[0]};
```
nirandaperera: nit. Unused?

nirandaperera: nit. And isnt `empty_opts` redundant here?

cpp/src/streaming/cudf/parquet.cpp
```diff
@@ -0,0 +1,215 @@
...
+        std::ranges::move(
+            local_files.begin() + file_offset,
+            local_files.begin() + file_offset + nfiles,
+            std::back_inserter(chunk)
+        );
+        auto source = cudf::io::source_info(std::move(chunk));
+        auto metadata = cudf::io::read_parquet_metadata(source);
+        auto const source_num_rows = metadata.num_rows();
+        auto skip_rows = options_skip_rows;
+        options_skip_rows = std::max(0L, options_skip_rows - source_num_rows);
```
nirandaperera: nit. Unused?

cpp/src/streaming/core/node.cpp
```diff
@@ -7,13 +7,17 @@
 
 namespace rapidsmpf::streaming {
 
-void run_streaming_pipeline(std::vector<Node> nodes) {
-    auto results = coro::sync_wait(coro::when_all(std::move(nodes)));
+Node when_all_or_throw(std::vector<Node>&& nodes) {
```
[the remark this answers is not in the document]
wence-: Yes

cpp/src/streaming/cudf/parquet.cpp
wence-: Ah yes, I should write `empty_opts.set_source(source)`

madsbk (commented):
First pass, looks good. 
It would be good with some basic C++ testing

cpp/src/streaming/cudf/parquet.cpp
```diff
@@ -0,0 +1,215 @@
...
+                ctx->br()->stream_pool().get_stream(),
+                std::move(empty_opts),
+                0
+            ));
+        }
+        co_await ch_out->drain(ctx->executor());
+        co_return;
+    }
+    bool skipped_all{true};
+    for (file_offset = 0; file_offset < files_per_rank; file_offset += files_per_split) {
```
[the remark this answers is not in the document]
wence-: I think I am overthinking this. I should just set up a `source` for all the files I'm going to read and then launch one read_parquet_chunk for each slice of rows.

wence-:
Added C++ tests, ready for another look

cpp/tests/streaming/test_read_parquet.cpp
```diff
@@ -0,0 +1,237 @@
...
+INSTANTIATE_TEST_SUITE_P(
+    InvalidMaxTickets, StreamingReadParquetThrottle, ::testing::Values(-1, 0)
+);
+
+TEST_P(StreamingReadParquetThrottle, NonPositiveThrottleThrows) {
+    if (GlobalEnvironment->comm_->nranks() > 1) {
+        // Test is independent of size of communicator, so don't bother if we
+        // have more than one rank.
+        GTEST_SKIP();
+    }
```
madsbk: For convenience, it might be helpful to also run it in a multi-rank setup.
```suggestion
    if (GlobalEnvironment->comm_->rank() != 0) {
        // Test is independent of size of communicator.
        GTEST_SKIP() << "Test only runs on rank 0";
    }
```

cpp/src/streaming/cudf/parquet.cpp
```diff
@@ -0,0 +1,157 @@
...
+    co_await receipt;
+}
+}  // namespace
+
+Node read_parquet(
+    std::shared_ptr<Context> ctx,
+    std::shared_ptr<Channel> ch_out,
+    std::ptrdiff_t max_tickets,
+    cudf::io::parquet_reader_options options,
+    cudf::size_type num_rows_per_chunk
```
rjzamora: This `num_rows_per_chunk` argument is the only "knob" for tuning the chunk size, which is understandable.

We **may** have trouble using this variation of the `read_parquet` node in cudf-polars unless we implement dynamic partitioning, or we force the metadata sampling to gather **all** footer metadata before the streaming network is constructed. Recall that the statically-partitioned version of cudf-polars needs to know how many chunks the `Scan` operation will produce ahead of time.

We are already planning to do dynamic partitioning in the future, so temporarily collecting extra/redundant parquet metadata up-front probably isn't a big deal. Even so, I'm wondering if there is a more general way to configure the chunk size here?

For example, it would be great if we could set an upper limit on the row-count-per chunk, but could also ask for "N chunks per file" or "N files per chunk" (to temporarily satisfy the static partitioned case). Even if we do focus on a row-count-based configuration, it may make sense to align the chunks with row-group boundaries as long as the row-count doesn't exceed the upper limit.

rjzamora: >I still struggle to understand this statement. I understand that the output of a shuffle needs to specify ahead of time the number of partitions it is going to use, but the scan is an input.

Yeah, I'm hoping that this is *mostly* true. The only immediate problem I can think of is if the "plan" tells us we will get a single chunk, but the IO node actually produces 2 or more. Since we cannot dynamically inject a concatenation or shuffle yet, we will have an issue if we run into an operation that is not supported for multiple partitions.

>I am very happy to try different chunking strategies but they have to be amenable to the multi-threaded read approach I have here.

My naive hypothesis is that row-group aligned reads would be easier to do well, but I could be wrong.

cpp/include/rapidsmpf/streaming/core/lineariser.hpp
```diff
@@ -0,0 +1,140 @@
...
+ * Channels are guaranteed to deliver data in increasing sequence number order. When we
+ * have multiple producers we must ensure that they queue up their productions into the
+ * output channel in sequence number order. The `Lineariser` provides this interface by
+ * providing a per-producer "output" channel and buffering appropriately.
+ *
+ * @warning Individual producers promise to send messages into their channel in strictly
+ * increasing sequence number order. This bounds the buffering required in the
+ * `Lineariser` to the number of distinct producers. If this precondition is not met,
+ * behaviour is undefined.
+ */
```
madsbk: A small code example here would be good

cpp/src/streaming/cudf/parquet.cpp
```diff
@@ -0,0 +1,184 @@
...
+                 num_rows_to_read}
+            );
+            num_rows_to_read -= chunk_num_rows;
+            chunks.emplace_back(sequence_number++, skip_rows, chunk_num_rows);
+            skip_rows += chunk_num_rows;
+        }
+        std::vector<Node> read_tasks;
+        std::atomic<std::size_t> chunk_index{0};
+        read_tasks.reserve(1 + num_producers);
+        auto lineariser = std::make_shared<Lineariser>(ch_out, num_producers);
```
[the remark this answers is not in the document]
wence-: Done.

cpp/include/rapidsmpf/streaming/core/lineariser.hpp
```diff
@@ -0,0 +1,158 @@
...
+ *   // For best performance they should attempt to take "interleaved" ids from a shared
+ *   // task list. That is, don't have producer-0 produce the first K ids, producer-1 the
+ *   // next K, and so forth.
+ *   tasks.push_back(producer(ctx, ch_in, ...));
+ * }
+ * coro_results(co_await coro::when_all(std::move(tasks)));
+ * // ch_out will see inputs in global total order of sequence id.
+ * @endcode
+ */
+class Lineariser {
```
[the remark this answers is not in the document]
wence-: Done.

cpp/include/rapidsmpf/streaming/core/lineariser.hpp
```diff
@@ -0,0 +1,157 @@
...
+        }
+        while (!min_heap_.empty()) {
+            auto item = min_heap_.pop_top();
+            co_await ch_out_->send(std::move(item.value));
+            // And refill from the producer we just consumed from.
+            auto msg = co_await inputs_[item.producer]->receive();
+            if (!msg.empty()) {
+                min_heap_.emplace(item.producer, std::move(msg));
+            }
+        }
```
nirandaperera: Here, the loop is held back by the slowest producer. It pushes back all the other input channels. Could this potentially cause a deadlock, like we saw in the Union case (which had multiple consumers though)?

cpp/src/streaming/cudf/parquet.cpp
```diff
@@ -0,0 +1,184 @@
...
+    RAPIDSMPF_EXPECTS(
+        size == 1 || options.get_skip_rows() == 0,
+        "Skipping rows not yet supported in multi-rank execution"
+    );
+    auto files = source.filepaths();
+    RAPIDSMPF_EXPECTS(files.size() > 0, "Must have at least one file to read");
+    RAPIDSMPF_EXPECTS(
+        files.size() < std::numeric_limits<int>::max(), "Trying to read too many files"
+    );
+    // TODO: Handle case where multiple ranks are reading from a single file.
```
nirandaperera: nit, shall we open an issue to track these TODOs?

cpp/src/streaming/cudf/parquet.cpp
```diff
@@ -0,0 +1,184 @@
...
+    int files_per_rank =
+        static_cast<int>(files.size() / size + (rank < (files.size() % size)));
+    int file_offset = rank * (files.size() / size) + std::min(rank, files.size() % size);
+    auto local_files = std::vector(
+        files.begin() + file_offset, files.begin() + file_offset + files_per_rank
+    );
```
nirandaperera: can we use a `span` here?

nirandaperera (commented):
Had some questions mostly.

cpp/tests/test_allgather.cpp
```diff
@@ -121,7 +121,7 @@ TEST_P(AllGatherTest, basic_allgather) {
     std::vector<rapidsmpf::PackedData> results;
     results = allgather->wait_and_extract(ordered);
     EXPECT_TRUE(allgather->finished());
-    if (n_elements > 0 && n_inserts > 0) {
+    if (n_inserts > 0) {
```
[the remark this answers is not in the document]
wence-: Because I changed allgather to forward empty chunks

cpp/src/streaming/cudf/parquet.cpp
```diff
@@ -0,0 +1,184 @@
...
+        std::atomic<std::size_t> chunk_index{0};
+        read_tasks.reserve(1 + num_producers);
+        auto lineariser = std::make_shared<Lineariser>(ch_out, num_producers);
+        read_tasks.push_back(lineariser->drain(ctx));
+        for (auto& ch_in : lineariser->get_inputs()) {
+            read_tasks.push_back(
+                produce_chunks(ctx, ch_in, local_options, chunks, chunk_index)
+            );
+        }
+        coro_results(co_await coro::when_all(std::move(read_tasks)));
```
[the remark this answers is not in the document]
wence-: Good idea.

gforsyth (approved):

PR #577, opened 2025-10-11
Fix hang when a consumer throws

madsbk:
Fix hang when a consumer throws an uncatched exception and add tests for exceptions in both producer and consumer.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -34,8 +34,7 @@ ENV HISTFILE="/home/coder/.cache/._bash_history"
 ENV AWS_ROLE_ARN="arn:aws:iam::279114543810:role/nv-gha-token-sccache-devs"
 ENV SCCACHE_REGION="us-east-2"
 ENV SCCACHE_BUCKET="rapids-sccache-devs"
-# 2hr (1 minute longer than sccache-dist request timeout)
-ENV SCCACHE_IDLE_TIMEOUT=7200
+ENV SCCACHE_IDLE_TIMEOUT=0
 
 ###
 # sccache-dist configuration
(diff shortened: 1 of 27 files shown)


cpp/include/rapidsmpf/streaming/core/channel.hpp
```diff
@@ -173,7 +173,20 @@ class Channel {
...
+        // Ensure executor progresses until ring buffer is empty, thus ensuring a consumer
+        // that died prematurely (e.g., due to an uncatched exception) does not hang
+        // indefinitely.
+        while (!rb_.empty() && !rb_.is_shutdown()) {
+            co_await executor->yield();
+        }
+
+        if (!rb_.is_shutdown()) {
+            co_await rb_.shutdown();
+        }
```
nirandaperera: This looks very similar to `shutdown_drain` impl. 
https://github.com/jbaldwin/libcoro/blob/main/include/coro/ring_buffer.hpp#L304-L309

It's not very clear to me why `shutdown_drain` causes a hang and this change resolves it. Maybe add more details?

pentschev: I actually think I have a standalone reproducer now without RapidsMPF. I'll clean it up tomorrow and write an issue in the libcoro summarizing the problem and the potential fix from above.

wence-: Yeah, I think you're right. I have a simple repro if you want me to submit

pentschev: Opened https://github.com/jbaldwin/libcoro/issues/401 with the summary/repro/suggested solution.

pentschev (approved):
Since I'm not the author I'm approving. I'll leave it up for someone else to merge though.

PR #584, opened 2025-10-14
Add general ``PyObjectPayload`` API

rjzamora:
In order to simplify experimentation in cudf-polars, I'd like to wrap arbitrary Python objects in a `Message`. My intention is to use these messages to pass "metadata" through the streaming network at runtime. Eventually, we can use our learnings to design/implement an effective `Metadata` payload in C++. However, I'd like to have something more flexible for now.


python/rapidsmpf/rapidsmpf/streaming/core/pyobject.pyx
```diff
@@ -0,0 +1,269 @@
...
+    unique_ptr[cpp_PyObjectPayload] \
+        cpp_release_pyobject_from_message(cpp_Message) except +
+
+
+cdef class PyObjectPayload:
+    """
+    A payload that wraps an arbitrary Python object for streaming.
+
+    This allows passing any picklable Python object (dict, list, custom classes, etc.)
+    through streaming channels without needing to convert to cuDF tables.
```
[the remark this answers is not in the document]
rjzamora: Heh - Cursor auto-complete.

python/rapidsmpf/rapidsmpf/streaming/core/pyobject.pyi
```diff
@@ -0,0 +1,63 @@
...
+        obj
+            Any Python object to wrap.
+
+        Returns
+        -------
+        PyObjectPayload
+            A new payload wrapping the given object.
+        """
+
+    @classmethod
```
wence-: Docstrings go in the cython pyx, not the pyi (unfortunately).

wence- (changes requested):

wence-:
> And at least initially, I'm hesitant about sending around arbitrary objects, rather than plain data as structured objects (possibly in a serialized format like JSON or msgpack).

The advantage of arbitrary objects is you don't have to define a schema. Arguably if you just sent bytes around you could consume them in C++ (if you know the schema), in this scenario you can't.

wence-:
Question: is this metadata intended to be attached to individual chunks or is it "summary" data that corresponds to metadata for all the chunks through a channel (or something else...). If the former, I think it makes sense to attach it to the chunks themselves. We can do this by, for example, adding a type-erased slot to `TableChunk`s that uses the `OwningWrapper` to manage lifetimes (probably with a `release_metadata` mechanism to get ownership back into python before dropping the `TableChunk` reference).

If the latter then I would advocate separate channels with the type-erased thing we're playing around with here. Then it is obvious that one must receive from/send into the metadata channel.

python/rapidsmpf/rapidsmpf/tests/streaming/test_pyobject.py
```diff
@@ -0,0 +1,157 @@
...
+    assert payload_tuple.extract_object() == data_tuple
+
+
+def test_message_roundtrip() -> None:
+    """Test that PyObjectPayload can be wrapped in a message and extracted."""
+    original_data = {"user": "alice", "score": 100, "active": True}
+    payload1 = PyObjectPayload.from_object(sequence_number=42, obj=original_data)
+
+    # Wrap in message
+    msg = Message(payload1)
```
nirandaperera: Something's not very clear to me here. 
If this was a real send-recv scenario, where would the `payload1.extract_object()` be called? Its not sufficient to call `del payload1` here, isnt it? Because it's refcount has been incremented. I feel like there is a memory-leak in the send side.

nirandaperera (commented):
I had some minor suggestions and some clarification questions.

python/rapidsmpf/rapidsmpf/streaming/core/pyobject.pyx
```diff
@@ -0,0 +1,299 @@
...
+
+# Helper functions to release TypeErasedChunk from a message
+cdef extern from *:
+    """
+    #include <cstdint>
+    #include <memory>
+    #include <rapidsmpf/streaming/cudf/owning_wrapper.hpp>
+
+    namespace rapidsmpf {
+    namespace streaming {
```
madsbk:
```suggestion
    namespace rapidsmpf::streaming {
```

python/rapidsmpf/rapidsmpf/streaming/core/pyobject.pyx
```diff
@@ -0,0 +1,299 @@
...
+    @staticmethod
+    def from_object(uint64_t sequence_number, obj):
+        """
+        Create a PyObjectPayload from a Python object.
+
+        Parameters
+        ----------
+        sequence_number : int
+            Sequence number for this payload.
+        obj : Any
```
madsbk: No types in the docstring
```suggestion
        ----------
        sequence_number
            Sequence number for this payload.
        obj
```

python/rapidsmpf/rapidsmpf/streaming/core/pyobject.pyx
```diff
@@ -0,0 +1,299 @@
...
+        """
+        Move this PyObjectPayload into a Message.
+
+        This method is not typically called directly. Instead, it is invoked by
+        `Message.__init__()` when creating a new Message with this PyObjectPayload
+        as its payload.
+
+        Parameters
+        ----------
+        message : Message
```
madsbk:
```suggestion
        message
```

python/rapidsmpf/rapidsmpf/streaming/core/pyobject.pyx
```diff
@@ -0,0 +1,299 @@
...
+        """
+        Extract and consume the wrapped Python object.
+
+        This is a destructive operation - after calling this method, the payload
+        is no longer usable and subsequent calls will raise an error.
+
+        Returns
+        -------
+        Any
+            The Python object that was wrapped by this payload.
```
madsbk:
```suggestion
        The Python object that was wrapped by this payload.
```

python/rapidsmpf/rapidsmpf/tests/streaming/test_pyobject.py
wence-: The `OwningWrapper`, when it goes out of scope, calls the deleter which decrefs the pointer.

Issue #587, opened 2025-10-15
Possible deadlock in `CuptiMonitor`


pentschev:
There was a timeout in https://github.com/rapidsai/rapidsmpf/actions/runs/18522685933/job/52786573288?pr=585 with what looks like a deadlock in `CuptiMonitor`:

<details><summary>Stack</summary>

```
Memory Profiling
----------------
Disabled
+ python /__w/rapidsmpf/rapidsmpf/ci/timeout_with_stack.py 30 ./example_cupti_monitor

Process timed out after 30.0 seconds
Capturing stack traces for all processes...

Capturing C stack trace for process 81790:
Attaching to process 81790
[New LWP 81795]
[New LWP 81794]
[New LWP 81793]
[New LWP 81792]
[New LWP 81791]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/x86_64-linux-gnu/libthread_db.so.1".
0x000070b5d1c99d71 in ?? () from /usr/lib/x86_64-linux-gnu/libc.so.6

Thread 6 (Thread 0x70b5b19ff000 (LWP 81791) "cuda00001400006"):
#0  0x000070b5d1d1c4cd in poll () from /usr/lib/x86_64-linux-gnu/libc.so.6
#1  0x000070b5b1d1ce27 in ?? () from /usr/lib/x86_64-linux-gnu/libcuda.so.1
#2  0x000070b5b1e0c5f7 in ?? () from /usr/lib/x86_64-linux-gnu/libcuda.so.1
#3  0x000070b5b1d08bb3 in ?? () from /usr/lib/x86_64-linux-gnu/libcuda.so.1
#4  0x000070b5d1c9daa4 in ?? () from /usr/lib/x86_64-linux-gnu/libc.so.6
#5  0x000070b5d1d2aa34 in clone () from /usr/lib/x86_64-linux-gnu/libc.so.6

Thread 5 (Thread 0x70b5b0eaa000 (LWP 81792) "cuda-EvtHandlr"):
#0  0x000070b5d1d1c4cd in poll () from /usr/lib/x86_64-linux-gnu/libc.so.6
#1  0x000070b5b1d1ce27 in ?? () from /usr/lib/x86_64-linux-gnu/libcuda.so.1
#2  0x000070b5b1e0c5f7 in ?? () from /usr/lib/x86_64-linux-gnu/libcuda.so.1
[...]

PR #588, opened 2025-10-15
Release the `CuptiMonitor` mutex before joining thread

pentschev:
If the sampling thread executes `capture_memory_sample()` and `monitor_active_ == true`, but then simultaneously the application thread calls `stop_monitoring()` and acquires `mutex_` before `sampling_thread_` acquired the mutex a deadlock will occur since `sampling_thread_.join()` can never complete. To prevent this release the `mutex_` before `sampling_thread_.join()` is called.

Fixes https://github.com/rapidsai/rapidsmpf/issues/587

### DIFF
--- a/cpp/src/cupti.cpp
+++ b/cpp/src/cupti.cpp
@@ -100,8 +100,6 @@ void CuptiMonitor::start_monitoring() {
 }
 
 void CuptiMonitor::stop_monitoring() {
-    std::lock_guard<std::mutex> lock(mutex_);
-
     if (!monitoring_active_.load()) {
         return;
     }


cpp/src/cupti.cpp
```diff
@@ -95,20 +95,25 @@ void CuptiMonitor::start_monitoring() {
...
-    if (!monitoring_active_.load()) {
-        return;
-    }
+        if (!monitoring_active_.load()) {
+            return;
+        }
 
-    monitoring_active_.store(false);
+        monitoring_active_.store(false);
+    }
```
wence-: I don't think you need this lock at all since reads of `monitoring_active_` are not otherwise protected by the mutex.

So:

```
stop_monitoring() {
  if (!is_monitoring()) {
      return;
  }
  monitoring_active_.store(false);
  if (sampling_thread_.joinable()) {
      sampling_thread_.join();
  }
  {
    std::lock_guard ...;
    capture_memory_usage_impl();
  }
  unsubscribe();
}
```
?

PR #598, opened 2025-10-22
Refactor: `Message`

madsbk:
Moving the `Message` class into its own file. 
Pure refactor, no functional changes.

### DIFF
--- a/cpp/include/rapidsmpf/streaming/core/message.hpp
+++ b/cpp/include/rapidsmpf/streaming/core/message.hpp
@@ -73,7 +73,7 @@ class Message {
      * @return true if the payload is `typeid(T)`, false otherwise.
      */
     template <typename T>
-    [[nodiscard]] bool payload_type() const noexcept {
+    [[nodiscard]] bool holds() const noexcept {
         return payload_.type() == typeid(std::shared_ptr<T>);
     }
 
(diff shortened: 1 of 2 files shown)


cpp/include/rapidsmpf/streaming/core/message.hpp
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * @brief Compare the payload type.
+     *
+     * @tparam T Expected payload type.
+     * @return true if the payload is `typeid(T)`, false otherwise.
+     */
+    template <typename T>
+    [[nodiscard]] bool payload_type() const noexcept {
+        return payload_.type() == typeid(std::shared_ptr<T>);
+    }
```
wence-: *nit*: This function was renamed from `holds<T>()`. It now doesn't read as well, I think, since `payload_type()` suggests it returns a type, not a bool.

And so `if (msg.payload_type<T>())` doesn't make as much sense to me as a reader as `if (msg.holds<T>())`

madsbk: let's go back to `holds`

wence- (approved):

nirandaperera (approved):

PR #604, opened 2025-10-24
`TableChunk` copying

madsbk:
Implements `TableChunk::copy()`, which will be used in a follow-up _message-copy_ PR.


madsbk:
Thanks @pentschev

PR #609, opened 2025-10-28
Message copy

madsbk:
`Message` now takes a set of callbacks to handle copy of its payload.


cpp/include/rapidsmpf/streaming/core/message.hpp
```diff
@@ -7,28 +7,84 @@
...
+         *
+         * @param msg Reference to the message whose data size is queried.
+         * @param mem_type Target memory type to query.
+         * @return A pair (size, spillable) where:
+         *   - size: total size (in bytes) of the primary data for the given memory type.
+         *   - spillable: `true` if the message owns its buffers and releasing it frees
+         *     memory; otherwise `false`.
+         */
+        std::function<std::pair<size_t, bool>(Message const&, MemoryType)>
+            primary_data_size;
```
pentschev: Instead of `primary_data`, should we name it `payload`, just like we're doing with the `MetadataPayloadExchange`? Consistent terminology makes the code simpler to understand.

pentschev (approved):

PR #610, opened 2025-10-29
Memory Reservation python bindings

madsbk:
Adding python bindings to `MemoryReservation` and some related `BufferResource` methods.


pentschev (approved):
Minor nit, looks good.

PR #612, opened 2025-10-29
[MINOR] Adding a `RAPIDSMPF_DETAIL` mode

nirandaperera:
This PR adds a detail definition to the build. `RAPIDSMPF_DETAIL` def is available in c++.

### DIFF
--- a/cpp/src/shuffler/shuffler.cpp
+++ b/cpp/src/shuffler/shuffler.cpp
@@ -227,7 +227,7 @@ class Shuffler::Progress {
         {
             auto const t0_metadata_recv = Clock::now();
             RAPIDSMPF_NVTX_SCOPED_RANGE_VERBOSE("meta_recv");
-            int recv_any_iters =
+            [[maybe_unused]] int recv_any_iters =
                 0;  // this will be stripped off if RAPIDSMPF_VERBOSE_INFO is not set
             while (true) {
                 auto const [msg, src] = shuffler_.comm_->recv_any(metadata_tag);


cpp/CMakeLists.txt
```diff
@@ -37,6 +37,13 @@ rapids_cmake_write_version_file(include/rapidsmpf/version_config.hpp)
 # Set a default build type if none was specified
 rapids_cmake_build_type(Release)
 
+# Set RAPIDSMPF_DEBUG default based on build type
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+  set(RAPIDSMPF_DEBUG_DEFAULT ON)
+else()
+  set(RAPIDSMPF_DEBUG_DEFAULT OFF)
+endif()
+
```
pentschev: I would prefer that the default is always off, even when CMake build type is debug, thus requiring you to specify it manually via `--cmake-args`, for example. The reason for that is we are more likely to build in debug mode to do general debugging than to require the extended NVTX annotations.

Furthermore, I think it makes more sense to have a different name, debug invokes a meaning of trying to resolve a problem, whereas I think this is more about performance validation/investigation. How do you feel about `RAPIDSMPF_NVTX_DETAILED`?

nirandaperera: A `DEBUG` mode macro came up in several discussions. So, I would like this to be used for not just nvtx annotations.

pentschev: I think `DEBUG` is a more specific term IMO, I'd prefer `RAPIDSMPF_DETAIL`. In any case, I think we should only enable it when explicitly specified, and not just depend on `CMAKE_BUILD_TYPE=Debug` to enable it by default.

cpp/src/shuffler/shuffler.cpp
```diff
@@ -171,7 +171,9 @@ class Shuffler::Progress {
      * @return The progress state of the shuffler.
      */
     ProgressThread::ProgressState operator()() {
+#if RAPIDSMPF_DEBUG
         RAPIDSMPF_NVTX_SCOPED_RANGE("Shuffler.Progress", p_iters++);
+#endif
```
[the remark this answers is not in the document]
madsbk: Then call it `RAPIDSMPF_NVTX_SCOPED_RANGE_VERBOSE` ?

cpp/CMakeLists.txt
```diff
@@ -67,6 +68,7 @@ message(STATUS "  BUILD_SHARED_LIBS       : ${BUILD_SHARED_LIBS}")
 message(STATUS "  CUDA_STATIC_RUNTIME     : ${CUDA_STATIC_RUNTIME}")
 message(STATUS "  RAPIDSMPF_CLANG_TIDY    : ${RAPIDSMPF_CLANG_TIDY}")
 message(STATUS "  RAPIDSMPF_ASAN          : ${RAPIDSMPF_ASAN}")
+message(STATUS "  RAPIDSMPF_VERBOSE_INFO        : ${RAPIDSMPF_VERBOSE_INFO}")
```
KyleFromNVIDIA:
```suggestion
message(STATUS "  RAPIDSMPF_VERBOSE_INFO  : ${RAPIDSMPF_VERBOSE_INFO}")
```

PR #613, opened 2025-10-30
Content description of an object

madsbk:
Introducing:

```c++
/**
 * @brief Description of an object's content.
 *
 * A `ContentDescription` encapsulates resource-related information about an
 * object's content such as memory sizes, memory types, and spillability.
 *
 * In RapidsMPF, an object's *content* refers to the actual data associated
 * with the object, not its metadata or auxiliary information. Typically, this
 * content is represented by one or more `Buffer` instances that may reside in
 * different memory spaces (e.g., host or device). It is also this content that
 * is subject to spilling and that typically accounts for the majority of an
 * object's overall memory footprint.
 *
 * The spillability state of an object is treated as an all-or-nothing property.
 * While one could imagine an object with a mix of spillable and non-spillable
 * content, this distinction is intentionally simplified in RapidsMPF.
 */
struct ContentDescription;
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -61,7 +61,7 @@ repos:
     hooks:
       - id: cython-lint
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: 'v1.17.0'
+    rev: 'v1.18.2'
     hooks:
       - id: mypy
         args: ["--config-file=pyproject.toml",
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -185,6 +185,7 @@ if(RAPIDSMPF_HAVE_STREAMING)
             src/streaming/core/leaf_node.cpp
             src/streaming/core/node.cpp
             src/streaming/cudf/partition.cpp
+            src/streaming/cudf/parquet.cpp
             src/streaming/cudf/table_chunk.cpp
   )
 endif()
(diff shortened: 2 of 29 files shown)


cpp/include/rapidsmpf/buffer/content_description.hpp
```diff
@@ -0,0 +1,135 @@
...
+ContentDescription get_content_description(T const& obj) {
+    using U = std::remove_cvref_t<T>;
+    std::size_t size;
+    if constexpr (std::same_as<U, std::string>) {
+        size = obj.size();
+    } else {
+        size = sizeof(U);
+    }
+    return ContentDescription{{{MemoryType::HOST, size}}, true};
+}
```
wence-: Why do we need this? These things will never be on device, right? (certainly not the `std::string`).

cpp/include/rapidsmpf/buffer/content_description.hpp
```diff
@@ -0,0 +1,135 @@
...
+ * content, this distinction is intentionally simplified in RapidsMPF.
+ */
+struct ContentDescription {
+    /** @brief Description of zero-sized content that is spillable. */
+    ContentDescription() = default;
+
+    /**
+     * @brief Construct from a range of (MemoryType, size) pairs.
+     *
+     * Unspecified memory types becomes zero.
```
wence-: What does this mean?

cpp/include/rapidsmpf/buffer/content_description.hpp
madsbk: Everything that goes into a message must have a `ContentDescription`

pentschev (approved):
LGTM.

PR #620, opened 2025-10-31
Add an arbitrary payload chunk

wence-:
Useful for sending type-erased PyObjects through channels.

This is an alternative to #584 (cc @rjzamora). It's slightly simpler because we just use the `OwningWrapper` directly, rather having a `TypeErasedChunk` C++ class.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/streaming/chunks/arbitrary.pyx
+++ b/python/rapidsmpf/rapidsmpf/streaming/chunks/arbitrary.pyx
@@ -1,0 +1,173 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-License-Identifier: Apache-2.0
+
+from cpython.object cimport PyObject
+from cpython.ref cimport Py_DECREF, Py_INCREF
+from cython.operator cimport dereference as deref
+from libc.stdint cimport uint64_t
+from libcpp.memory cimport make_unique, unique_ptr
+from libcpp.utility cimport move
+
+from rapidsmpf.streaming.chunks.utils cimport py_deleter
+from rapidsmpf.streaming.core.message cimport Message, cpp_Message
+
+
+cdef extern from * nogil:
+    """
+    namespace {
+    rapidsmpf::streaming::Message cpp_to_message(
+        std::uint64_t sequence_number,
+        std::unique_ptr<rapidsmpf::streaming::OwningWrapper> obj
+    ) {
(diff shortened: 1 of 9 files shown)


python/rapidsmpf/rapidsmpf/tests/streaming/test_arbitrary_chunk.py
rjzamora: Does it make to have a simple/trivial test with a proper PyNode/Channel network?

python/rapidsmpf/rapidsmpf/streaming/chunks/arbitrary.pyx
```diff
@@ -0,0 +1,154 @@
...
+    }
+    }
+    """
+    cpp_Message cpp_to_message"to_message"(
+        uint64_t, unique_ptr[cpp_OwningWrapper]
+    ) except +
+
+
+cdef class ArbitraryChunk:
+    def __init__(self, object obj):
```
madsbk:
```suggestion
    def __init__(self, object obj not None):
```

python/rapidsmpf/rapidsmpf/streaming/chunks/arbitrary.pyx
```diff
@@ -0,0 +1,154 @@
...
+        return rapidsmpf::streaming::Message(sequence_number, std::move(obj));
+    }
+    }
+    """
+    cpp_Message cpp_to_message"to_message"(
+        uint64_t, unique_ptr[cpp_OwningWrapper]
+    ) except +
+
+
+cdef class ArbitraryChunk:
```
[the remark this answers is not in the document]
wence-: Done.

python/rapidsmpf/rapidsmpf/streaming/chunks/arbitrary.pyx
wence-: `None` is a valid payload (ok, kind of a dumb one, but still).

python/rapidsmpf/rapidsmpf/tests/streaming/test_arbitrary_chunk.py
```diff
@@ -0,0 +1,155 @@
...
+    def __init__(self, value: Any):
+        self.value = value
+
+
+def test_roundtrip_chunk(context: Context) -> None:
+    expect = Object(10)
+
+    got = ArbitraryChunk(expect).release()
+
+    assert got is expect
```
madsbk:
```suggestion
    expect = Object(10)
    got = ArbitraryChunk(expect).release()
    assert got is expect
```

madsbk (approved):
Looks good

PR #623, opened 2025-11-03
Message with content description

madsbk:
This is the new ctor:
```c++
    /**
     * @brief Construct a new message from a unique pointer to its payload.
     *
     * The message may optionally support deep-copy and spilling operations through a
     * user-provided `CopyCallback`. If no callback is provided, copy and spill
     * operations are disabled.
     *
     * @tparam T Type of the payload to store inside the message.
     * @param sequence_number Ordering identifier for the message.
     * @param payload Non-null unique pointer to the payload.
new->* @param content_description Description of the payload's content. When a copy
new->* callback is provided, this description must accurately reflect the content of the
new->* payload (e.g., per-memory-type sizes and spillable status).
new->* @param copy_cb Optional callback used to perform deep copies of the message. If
new->* `nullptr`, copying and spilling are disabled.
     *
     * @note Sequence numbers are used to ensure that when multiple producers send into
     * the same output channel, channel ordering is preserved. Specifically, the guarantee
     * is that `Channel`s always produce elements in increasing sequence number order. To
     * ensure this, single producers must promise to send into the channels in strictly
     * increasing sequence number order. Behaviour is undefined if not. To ensure
     * insertion into an output channel from multiple producers obeys this invariant, use
     * a `Lineariser`.
     *
     * This promise allows consumers to ensure ordering by buffering at most
     * `num_consumers` messages, rather than needing to buffer the entire channel input.
     *
     * @throws std::invalid_argument if @p payload is null.
     */
    template <typename T>
    Message(
        std::uint64_t sequence_number,
        std::unique_ptr<T> payload,
        ContentDescription content_description,
        CopyCallback copy_cb = nullptr
    )
```

### DIFF
--- a/cpp/include/rapidsmpf/streaming/core/message.hpp
+++ b/cpp/include/rapidsmpf/streaming/core/message.hpp
@@ -197,12 +197,8 @@ class Message {
      *
      * @see copy()
      */
-    [[nodiscard]] size_t copy_cost() const noexcept {
-        size_t ret = 0;
-        for (MemoryType mem_type : MEMORY_TYPES) {
-            ret += content_description().content_size(mem_type);
-        }
-        return ret;
+    [[nodiscard]] constexpr size_t copy_cost() const noexcept {
+        return content_description().content_size();
     }
 
     /**
--- a/cpp/include/rapidsmpf/buffer/content_description.hpp
+++ b/cpp/include/rapidsmpf/buffer/content_description.hpp
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 #pragma once
-#include <string>
+#include <numeric>
 #include <type_traits>
 #include <utility>
 
@@ -62,7 +62,7 @@ class ContentDescription {
         requires std::convertible_to<
             std::ranges::range_value_t<Range>,
             std::pair<MemoryType, std::size_t>>
-    explicit ContentDescription(Range&& sizes, Spillable spillable)
+    constexpr explicit ContentDescription(Range&& sizes, Spillable spillable)
         : spillable_(spillable == Spillable::YES) {
         content_sizes_.fill(0);
         for (auto&& [mem_type, size] : sizes) {
@@ -81,7 +81,7 @@ class ContentDescription {
      *
      * @param spillable Whether the content are spillable.
      */
-    ContentDescription(Spillable spillable = Spillable::NO)
+    constexpr ContentDescription(Spillable spillable = Spillable::NO)
         : ContentDescription({{}}, spillable) {}
 
     /**
@@ -90,7 +90,7 @@ class ContentDescription {
      * @param mem_type The memory type entry to access.
      * @return Reference to the size (in bytes) for the given memory type.
      */
-    [[nodiscard]] std::size_t& content_size(MemoryType mem_type) noexcept {
+    [[nodiscard]] constexpr std::size_t& content_size(MemoryType mem_type) noexcept {
         return content_sizes_[static_cast<std::size_t>(mem_type)];
     }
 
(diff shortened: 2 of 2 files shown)


cpp/include/rapidsmpf/streaming/core/message.hpp
```diff
@@ -188,47 +167,32 @@ class Message {
...
+        return content_description_;
     }
 
     /**
-     * @brief Query the size of the message's content for a given memory type.
-     *
-     * Invokes the registered `content_size` callback to compute the total size
-     * (in bytes) of the message's content portion stored in the specified memory
-     * space (e.g., host or device). This excludes any metadata or auxiliary information.
+     * @brief Returns the content description associated with the message.
```
pentschev:
```suggestion
     * @brief Returns the copy callback associated with the message.
```

cpp/tests/streaming/test_message.cpp
```diff
@@ -127,8 +122,14 @@ TEST_F(StreamingMessage, CopyWithCallbacks) {
...
+        };
         Message m1{
-            42, br->allocate(stream, br->reserve_or_fail(10, MemoryType::DEVICE)), cbs
+            42,
+            br->allocate(stream, br->reserve_or_fail(10, MemoryType::DEVICE)),
+            cd,
+            copy_cb
         };
         EXPECT_EQ(m1.copy_cost(), 10);
         auto res = br->reserve_or_fail(m1.copy_cost(), MemoryType::DEVICE);
```
pentschev: Same here (few lines below, GH doesn't let me comment there), validate `m2`.

pentschev (approved):
LGTM, thanks Mads.

cpp/include/rapidsmpf/streaming/core/message.hpp
```diff
@@ -26,64 +27,39 @@ namespace rapidsmpf::streaming {
...
-         * @param br Buffer resource used for memory allocations.
-         * @param reservation Memory reservation to consume during allocation.
-         * @return A new `Message` instance containing a deep copy of the payload.
-         */
-        std::function<
-            Message(Message const&, BufferResource* br, MemoryReservation& reservation)>
-            copy;
-    };
+    using CopyCallback = std::function<
+        Message(Message const&, BufferResource* br, MemoryReservation& reservation)>;
```
wence-: I think the `reservation` contains the relevant buffer resource pointer. So this callback should _only_ take the reservation (not the resource). In case we mix up and pass the wrong resource:

```suggestion
    using CopyCallback = std::function<
        Message(Message const&, MemoryReservation& reservation)>;
```

cpp/include/rapidsmpf/streaming/core/message.hpp
```diff
@@ -262,9 +225,9 @@ class Message {
      */
     [[nodiscard]] Message copy(BufferResource* br, MemoryReservation& reservation) const {
         RAPIDSMPF_EXPECTS(
-            callbacks_.copy, "message doesn't support `copy`", std::invalid_argument
+            copy_cb(), "message doesn't support `copy`", std::invalid_argument
         );
-        return callbacks_.copy(*this, br, reservation);
+        return copy_cb()(*this, br, reservation);
     }
```
wence-: Again here, only take the `reservation`, not the `br`.

cpp/include/rapidsmpf/streaming/core/message.hpp
```diff
@@ -26,64 +27,38 @@ namespace rapidsmpf::streaming {
...
-         * @param br Buffer resource used for memory allocations.
-         * @param reservation Memory reservation to consume during allocation.
-         * @return A new `Message` instance containing a deep copy of the payload.
-         */
-        std::function<
-            Message(Message const&, BufferResource* br, MemoryReservation& reservation)>
-            copy;
-    };
+    using CopyCallback =
+        std::function<Message(Message const&, MemoryReservation& reservation)>;
```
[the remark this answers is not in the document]
madsbk: Yes, the caller should be able to make a reservation up front before copying. Also, the caller decides where to copy.

cpp/src/streaming/chunks/partition.cpp
```diff
@@ -26,67 +26,43 @@ ContentDescription get_content_description(PartitionVectorChunk const& obj) {
...
-                ret.emplace_back(packed_data.copy(br, reservation));
+                pd.emplace_back(packed_data.copy(reservation));
             }
-            return Message(
-                msg.sequence_number(),
-                std::make_unique<PartitionVectorChunk>(std::move(ret)),
-                msg.callbacks()
-            );
+            auto chunk = std::make_unique<PartitionVectorChunk>(std::move(pd));
+            auto cd = get_content_description(*chunk);
```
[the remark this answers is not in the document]
madsbk: No, the location (host or device) of the chunk's content depend on the `reservation.mem_type()`.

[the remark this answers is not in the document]
nirandaperera: Oh okay. It was not immediately clear to me just by reading this. I was confused why ` get_content_description(...)` was called twice in a seemingly similar fashion. I think its best to add a small comment in the copy cb impls regarding this.

wence- (approved):

PR #627, opened 2025-11-04
Add a python implementation of TPCH query 9

wence-:


TomAugspurger (approved):

TomAugspurger:
The stream changes are probably from https://github.com/rapidsai/rmm/pull/2110/files#diff-83f78faf97018ac3ad1e8b43d42eb56d0a6acc442b499a75fb04a7838301174eR127 (cc @nirandaperera). https://docs.python.org/3/reference/datamodel.html#object.__hash__ mentions that objects implementing `__eq__` must also implement `__hash__` to be used in hashable collections like `set()`.

I'm not sure what the best way to uniquely identify a stream. I see https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM_1g5799ae8dd744e561dfdeda02c53e82df, but I don't know if that's an option for RMM.

If we only deal with *owning* streams at the Python level, then I think that `id(self)` is fine since the memory address of the python object will uniquely identify the CUDA stream. If we have a view on some other stream (Python or C++) we'd need a way to get the original stream.

madsbk:

PR #630, opened 2025-11-05
Collection for spillable messages

madsbk:
Adding a collection for messages that enables spilling. This collection will be used by the channels to enable spilling.

Any better name for `SpillableMessages` is very much welcome!

### DIFF
--- a/cpp/src/streaming/core/spillable_messages.cpp
+++ b/cpp/src/streaming/core/spillable_messages.cpp
@@ -23,7 +23,6 @@ Message SpillableMessages::extract(MessageId mid) {
 
     // If the item is being spilled, we block here until the spilling is done.
     std::unique_lock item_lock(item->mutex);
-    RAPIDSMPF_EXPECTS(item->message.has_value(), "message not found", std::out_of_range);
     return std::exchange(item->message, std::nullopt).value();
 }
 


cpp/include/rapidsmpf/streaming/core/spillable_messages.hpp
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * ### Example
+ * @code{.cpp}
+ * SpillableMessages messages;
+ * auto id = messages.insert(to_message(TableChunk{...}));
+ * messages.spill(id, ...);
+ * auto msg = messages.extract(id);
+ * @endcode
+ */
+class SpillableMessages {
```
madsbk: Any better name?

cpp/src/streaming/core/spillable_messages.cpp
```diff
@@ -0,0 +1,78 @@
...
+    items_.insert({counter_, std::make_shared<Item>(std::move(message))});
+    return counter_++;
+}
+
+Message SpillableMessages::extract(MessageId mid) {
+    std::unique_lock global_lock(global_mutex_);
+    std::shared_ptr<Item> item = extract_item(items_, mid).second;
+    content_descriptions_.erase(mid);
+    global_lock.unlock();
+
```
wence-: *question*: If someone else `extract`ed this message ID, I think that `item` at this point would be `nullptr`, or have we already `throw`n because `extract_item` throws if the key does not exist?

cpp/src/streaming/core/spillable_messages.cpp
```diff
@@ -0,0 +1,78 @@
...
+    // If the item is being spilled, we block here until the spilling is done.
+    std::unique_lock item_lock(item->mutex);
+    RAPIDSMPF_EXPECTS(item->message.has_value(), "message not found", std::out_of_range);
```
wence-: question: How can this occur? Noone else can have a reference to `item` at this point, right?

cpp/include/rapidsmpf/streaming/core/spillable_messages.hpp
```diff
@@ -0,0 +1,132 @@
...
+     * modify the container after this call.
+     *
+     * Use this snapshot to decide which messages to spill, but keep in mind
+     * that the information may no longer be accurate when the actual spill
+     * occurs. When calling `spill()`, the returned size reflects what was
+     * actually spilled.
+     *
+     * @return Copy of a map from `MessageId` to `ContentDescription`.
+     */
+    std::map<MessageId, ContentDescription> get_content_descriptions() const;
```
pentschev: Not sure if that would be useful, but would a method that does `spill` also immediately be helpful? For example, pass a function as argument that executes looking for the "perfect match" (e.g., largest message) and immediately does the spilling without risking invalidation of the returned value of `get_content_descriptions()`?

cpp/src/streaming/core/spillable_messages.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+    // Ensure the item still contains something to spill.
+    auto const& msg = item->message.value();
+    auto const old_cd = msg.content_description();
+    if (!old_cd.spillable() || old_cd.content_size(MemoryType::DEVICE) == 0) {
+        return 0;
+    }
+
+    // Spill item in-place.
+    auto res = br->reserve_or_fail(msg.copy_cost(), MemoryType::HOST);
```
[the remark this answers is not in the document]
madsbk: done

pentschev (approved):
Thanks Mads!

wence- (approved):

cpp/src/streaming/core/spillable_messages.cpp
madsbk: Yes, `extract_item()` will throw.

PR #631, opened 2025-11-05
Refactor: `Context.create_channel()`

madsbk:
Refactoring such that creating a new Channel has to go through a context.

This will become useful when introducing channel spilling where we need a centralized place (`Context`) to coordinate spilling.


pentschev (approved):
LGTM.

madsbk:
Thanks @pentschev and @KyleFromNVIDIA

Issue #632, opened 2025-11-05
Make `BufferResource.stream_pool` size configurable


TomAugspurger:
https://github.com/rapidsai/rapidsmpf/blob/da393b7e64e661504a5f2cbdde795bd4a4fcfee8/cpp/include/rapidsmpf/buffer/resource.hpp#L167 hardcodes the RMM CUDA stream pool size to 16. I'd like to make that configurable as part of https://github.com/rapidsai/cudf/issues/20484.

PR #633, opened 2025-11-05
Make CUDA stream pool size configurable

TomAugspurger:
This updates the C++ `BufferResource` and Cython wrappers to make the RMM CUDA stream pool size configurable. The default should be unchanged (16). A concrete RMM pool can still be passed in at the C++ layer. The Cython/Python wrapper allows passing in a pool size.

Closes https://github.com/rapidsai/rapidsmpf/issues/632


cpp/src/buffer/resource.cpp
```diff
@@ -26,12 +26,19 @@ BufferResource::BufferResource(
...
     std::unordered_map<MemoryType, MemoryAvailable> memory_available,
     std::optional<Duration> periodic_spill_check,
+    std::size_t stream_pool_size,
     std::shared_ptr<rmm::cuda_stream_pool> stream_pool,
     std::shared_ptr<Statistics> statistics
 )
     : device_mr_{device_mr},
       memory_available_{std::move(memory_available)},
-      stream_pool_{std::move(stream_pool)},
+      stream_pool_{
```
TomAugspurger: This bit was AI generated, but seems reasonable enough (to me).

cpp/include/rapidsmpf/buffer/resource.hpp
```diff
@@ -155,16 +155,19 @@ class BufferResource {
      * continuously checks and perform spilling based on the memory availability
      * functions. The value of `periodic_spill_check` is used as the pause between checks.
      * If `std::nullopt`, no periodic spill check is performed.
+     * @param stream_pool_size The number of streams to create in the CUDA stream pool.
+     * Only used if `stream_pool` is not provided.
```
bdice: This is a weird interface. Why not just accept a stream pool and let the user construct the stream pool if they want a custom size?

bdice: Yay! Do it!

TomAugspurger: I think https://github.com/rapidsai/rapidsmpf/pull/633/files#diff-049939465cbe3b3d4a7593bac281eecc840a6c0cfcf78042ae23b6389859190bR108 is the gymnastics you were referring to.

python/rapidsmpf/rapidsmpf/buffer/resource.pyx
```diff
@@ -183,6 +202,27 @@ cdef class BufferResource:
         if periodic_spill_check is not None:
             period = cpp_Duration(periodic_spill_check)
 
+        # Handle stream pool parameter
+        # If None, create a default pool with 16 streams
+        if stream_pool is None:
+            stream_pool = CudaStreamPool(pool_size=16)
```
TomAugspurger: I'd like to avoid this block entirely, but I wasn't sure how to essentially pass `None` down to the C++ constructor and let us use its default behavior. Maybe that's unavoidable.

But if we do need this, we need to make sure we match the behavior of the `BufferResource` constructor in C++. In particular, pass the non-blocking flag.

nirandaperera (approved):
LGTM

PR #636, opened 2025-11-05
Adding Fanout node

### LINKED ISSUE #560: [FEA] Support channel broadcast (streaming "fanout" node).

nirandaperera:
This PR adds `Fanout` Node. 

```c++

/**
 * @brief Fanout policy controlling how messages are propagated.
 */
enum class FanoutPolicy : uint8_t {
    /**
     * @brief Process messages as they arrive and immediately forward them.
     *
     * Messages are forwarded as soon as they are received from the input channel.
     * The next message is not processed until all output channels have completed
     * sending the current one, ensuring backpressure and synchronized flow.
     */
    BOUNDED,

    /**
     * @brief Forward messages without enforcing backpressure.
     *
     * In this mode, messages may be accumulated internally before being
     * broadcast, or they may be forwarded immediately depending on the
     * implementation and downstream consumption rate.
     *
     * This mode disables coordinated backpressure between outputs, allowing
     * consumers to process at independent rates, but can lead to unbounded
     * buffering and increased memory usage.
     *
     * @note Consumers might not receive any messages until *all* upstream
     * messages have been sent, depending on the implementation and buffering
     * strategy.
     */
    UNBOUNDED,
};

/**
 * @brief Broadcast messages from one input channel to multiple output channels.
 *
 * The node continuously receives messages from the input channel and forwards
 * them to all output channels according to the selected fanout policy, see
 * ::FanoutPolicy.
 *
 * Each output channel receives a shallow copy of the same message; no payload
 * data is duplicated. All copies share the same underlying payload, ensuring
 * zero-copy broadcast semantics.
 *
 * @param ctx The node context to use.
 * @param ch_in Input channel from which messages are received.
 * @param chs_out Output channels to which messages are broadcast.
 * @param policy The fanout strategy to use (see ::FanoutPolicy).
 *
 * @return Streaming node representing the fanout operation.
 *
 * @throws std::invalid_argument If an unknown fanout policy is specified.
 *
 * @note Since messages are shallow-copied, releasing a payload (`release<T>()`)
 * is only valid on messages that hold exclusive ownership of the payload.
 */
Node fanout(
    std::shared_ptr<Context> ctx,
    std::shared_ptr<Channel> ch_in,
    std::vector<std::shared_ptr<Channel>> chs_out,
    FanoutPolicy policy
);
```

Depends on https://github.com/rapidsai/rapidsmpf/pull/648

Closes #560


cpp/src/streaming/core/channel.cpp
```diff
@@ -8,13 +8,15 @@
 namespace rapidsmpf::streaming {
 
 coro::task<bool> Channel::send(Message msg) {
+    RAPIDSMPF_EXPECTS(!msg.empty(), "message cannot be empty");
```
madsbk: Add to doc

cpp/src/streaming/core/channel.cpp
```diff
@@ -8,13 +8,15 @@
...
     auto result = co_await rb_.produce(std::move(msg));
     co_return result == coro::ring_buffer_result::produce::produced;
 }
 
 coro::task<Message> Channel::receive() {
     auto msg = co_await rb_.consume();
     if (msg.has_value()) {
+        RAPIDSMPF_EXPECTS(!msg->empty(), "received empty message");
```
madsbk: Add to doc

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,276 @@
...
+        tasks.push_back(ch_out->send(msg.copy(res)));
+    }
+    coro_results(co_await coro::when_all(std::move(tasks)));
+}
+
+/**
+ * @brief Broadcast messages from one input channel to multiple output channels.
+ *
+ * @note Bounded fanout requires all the output channels to consume messages before the
+ * next message is sent/ consumed from the input channel.
```
madsbk:
```suggestion
 * next message is sent/consumed from the input channel.
```

cpp/src/streaming/core/channel.cpp
wence-: Why do you want this?

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,276 @@
...
+#include <coro/coro.hpp>
+
+namespace rapidsmpf::streaming::node {
+namespace {
+
+/**
+ * @brief Asynchronously send a message to multiple output channels.
+ *
+ * @param msg The message to broadcast. Each channel receives a shallow
+ * copy of the original message.
```
wence-: This is no longer true, right?

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,276 @@
...
+        }
+        logger.trace(
+            "recv_messages active size: ", state.recv_messages.size() - state.purge_idx
+        );
+    }
+
+    // Note: there will be some messages to be purged after the loop exits, but we don't
+    // need to do anything about them here
+    co_await tasks.yield_until_empty();
+    logger.debug("Unbounded fanout completed");
```
wence-: I wonder if this implementation is overly complicated. Would the following work:

The producer of the fanout pushes a copy of the `Message` into a `coro::queue` (one per output channel):

```
Node single_consumer(coro::queue<Message>& q, Channel ch_out) {
    while (true) {
        auto val = co_await q.pop();
        if (val != coro::queue_consume_result::consumed) {
           break;
        }
        co_await ch_out->send(std::move(*val));
   }
   co_await ch_out->drain(...);
}

Node fanout_producer(Channel ch_in, std::vector<coro::queue<Message>>& outputs) {
    while (true) {
        auto msg = co_await ch_in->receive();
        if (msg.empty()) {
            break;
        }
        for (int i = 0; i < num_outputs - 1; i++) {
            co_await outputs[i].push(msg.copy(...));
        }
        co_await outputs[num_outputs-1].push(std::move(msg)); // no need to copy this one.
    }
    // pseudo-code
    coro::when_all(q.shutdown_drain(...) for q in outputs);
}

Node fanout(Channel ch_in, std::vector<Channel> chs_out) {
   std::vector<Node> tasks;
   std::vector<coro::queue<Message>> queues(chs_out.size());
   tasks.push_back(fanout_producer(ch_in, queues));
   for (int i = 0; i < chs_out.size(); i++) {
       tasks.push_back(single_consumer(queues[i], chs_out[i]));
   }
   coro_results(co_await coro::when_all(std::move(tasks)));
}
```

python/rapidsmpf/rapidsmpf/streaming/core/fanout.pyx
```diff
@@ -0,0 +1,110 @@
...
+
+    cdef vector[shared_ptr[cpp_Channel]] _chs_out
+    owner = []
+    for ch_out in chs_out:
+        if not isinstance(ch_out, Channel):
+            raise TypeError("All elements in chs_out must be Channel instances")
+        owner.append(ch_out)
+        _chs_out.push_back((<Channel>ch_out)._handle)
+
+    cdef cpp_FanoutPolicy _policy = <cpp_FanoutPolicy>(<int>policy)
```
wence-: If you8 use the re-exported cdef enum this is unnecessary. Check some of the pylibcudf cython code to see how it's done there.

wence- (changes requested):
I wonder if there is a much simpler implementation lurking here.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,276 @@
...
+            break;
+        }
+
+        co_await send_to_channels(ctx.get(), msg, chs_out);
+        logger.debug("Sent message ", msg.sequence_number());
+    }
+
+    for (auto& ch : chs_out) {
+        co_await ch->drain(ctx->executor());
+    }
```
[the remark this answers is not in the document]
nirandaperera: @wence- 
This is how @madsbk 's initial impl did it. 
```c++
    std::vector<Node> tasks;
    tasks.reserve(chs_out.size());
    for (auto& ch_out : chs_out) {
        tasks.push_back(ch_out->drain(ctx->executor()));
    }
    coro_results(co_await coro::when_all(std::move(tasks)));
```
But then when I went through [`when_all` docs ](https://github.com/jbaldwin/libcoro/blob/main/README.md#when_all), I was put off by this comment, 
> Each task passed into when_all will initially be executed serially by the calling thread so it is recommended to offload the tasks onto an executor like coro::thread_pool or coro::io_scheduler so they can execute in parallel.

Felt there's no real difference with that and draining channels iteratively. 

Maybe use a `task_container`?

cpp/src/streaming/core/fanout.cpp
nirandaperera: My idea was using something like `atomic::wait` where a sender task can wait on the size of the deque. When its updated, we do a `notify_all` so all tasks update their target.

wence-: I still think a `coro::queue` is the right model here. How about the following idea for the consumer copying the messages:

- Producer of fanout pushes a pair of `(atomic<int>, Message)` into a `std::vector` where the atomic counter is initialised to the number of consumers.
- Each consumer has a `coro::queue<int>` to receive the indices of available messages.
- A consumer pulls an index out of its queue, obtains the "refcount" for the message by `fetch_sub(1, ...)` If this `1` we know we are the last consumer, so we can `std::move` the message. Otherwise, we must copy it.

I suppose that is racy because the refcount decrease and message copy are not an atomic operation.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,297 @@
...
+    Context* ctx, Message const& msg, std::vector<std::shared_ptr<Channel>>& chs_out
+) {
+    std::vector<coro::task<bool>> tasks;
+    tasks.reserve(chs_out.size());
+    for (auto& ch_out : chs_out) {
+        // do a reservation for each copy, so that it will fallback to host memory if
+        // needed
+        // TODO: change this
+        auto res = ctx->br()->reserve_or_fail(msg.copy_cost(), try_memory_types(msg)[0]);
+        tasks.push_back(ch_out->send(msg.copy(res)));
```
wence-: We own (or should own) the message, so one of these copies is redundant (the "last" output channel can just take ownership of the input message).

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,297 @@
...
+        // notify send_tasks to copy & send messages
+        co_await state.data_ready.notify_all();
+
+        // purge completed send_tasks. This will reset the messages to empty, so that they
+        // release the memory, however the deque is not resized. This guarantees that the
+        // indices are not invalidated.
+        // intentionally not locking the mtx here, because we only need to know a
+        // lower-bound on the last completed idx (ch_next_idx values are monotonically
+        // increasing)
+        size_t last_completed_idx = std::ranges::min(state.ch_next_idx);
```
wence-: This is UB, because another thread (one of the consumers) might be updating an entry in `state.ch_next_idx` simultaneously.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,297 @@
...
+                curr_recv_msg_sz = state.recv_messages.size();
+                return state.input_done || state.ch_next_idx[idx] < curr_recv_msg_sz;
+            });
+            if (state.input_done && state.ch_next_idx[idx] == curr_recv_msg_sz) {
+                // no more messages will be received, and all messages have been sent
+                break;
+            }
+        }
+
+        // now we can copy & send messages in indices [next_idx, end_idx)
```
wence-: These comments make no sense, because `end_idx` doesn't appear anywhere.

cpp/src/streaming/core/channel.cpp
nirandaperera: @wence- If we cant differentiate an empty message from a channel shutdown when calling `receive`, IMO we should not allow sending empty messages

cpp/src/streaming/core/fanout.cpp
nirandaperera: Good point. Let me move the msg to the last channel.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    // input_done is only set by this task, so reading without lock is safe here
+    while (!state.input_done) {
+        {
+            auto lock = co_await state.mtx.scoped_lock();
+            co_await state.request_data.wait(lock, [&] {
+                return std::ranges::any_of(state.ch_next_idx, [&](size_t next_idx) {
+                    return state.recv_messages.size() == next_idx;
+                });
```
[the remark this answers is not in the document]
nirandaperera: Yes, that's a good point. The only issue would be, cleaning up finished messages. Currently I use the `ch_next_idx` to find the boundary.

cpp/src/streaming/core/fanout.cpp
nirandaperera: Yes, but its strictly increasing. And we only need an approximate value here. So, a current running `min` may not be the exact min in `std::ranges::min`, but it will be strictly less than or equal. Consequence would be, not cleaning up all the finished messages. But I felt it was a better trade-off than trying to relock the mutex.

nirandaperera: Hmm. Thinking about this again, maybe I can do a `ranges::min_max` during `request_data.wait`, and purge until the `min` value. That would eliminate this.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,327 @@
...
+) {
+    RAPIDSMPF_EXPECTS(!chs_out.empty(), "output channels cannot be empty");
+
+    std::vector<coro::task<bool>> tasks;
+    tasks.reserve(chs_out.size());
+    for (size_t i = 0; i < chs_out.size() - 1; i++) {
+        // do a reservation for each copy, so that it will fallback to host memory if
+        // needed
+        // TODO: change this
+        auto res = ctx->br()->reserve_or_fail(msg.copy_cost(), try_memory_types(msg)[0]);
```
wence-: What does this TODO mean?

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+    auto& executor = *ctx->executor();
+    // schedule send tasks for each output channel
+    for (size_t i = 0; i < chs_out.size(); i++) {
+        tasks.emplace_back(executor.schedule(
+            unbounded_fo_send_task(*ctx, i, std::move(chs_out[i]), state)
+        ));
+    }
+    // schedule process input task
+    tasks.emplace_back(executor.schedule(
```
wence-: Let's cull some of these LLM comments.

cpp/src/streaming/core/fanout.cpp
wence-: OK, fair.

cpp/src/streaming/core/fanout.cpp
wence-: This is still no longer true.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,297 @@
...
+    UnboundedFanoutState state(chs_out.size());
+
+    // start send tasks for each output channel
+    coro::task_container<coro::thread_pool> tasks(ctx->executor());
+    for (size_t i = 0; i < chs_out.size(); i++) {
+        RAPIDSMPF_EXPECTS(
+            tasks.start(unbounded_fo_send_task(*ctx, i, chs_out[i], state)),
+            "failed to start send task"
+        );
+    }
```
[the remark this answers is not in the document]
nirandaperera: done

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,327 @@
...
+        // now we can copy & send messages in indices [next_idx, curr_recv_msg_sz)
+        // it is guaranteed that message purging will be done only on indices less than
+        // next_idx, so we can safely send messages without locking the mtx
+        for (size_t i = state->ch_next_idx[idx]; i < curr_recv_msg_sz; i++) {
+            auto const& msg = state->recv_messages[i];
+            RAPIDSMPF_EXPECTS(!msg.empty(), "message cannot be empty");
+
+            // make reservations for each message so that it will fallback to host memory
+            // if needed
+            // TODO: change this
```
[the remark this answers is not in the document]
nirandaperera: To add the changes from #648. Its addressed now.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,411 @@
...
+    ShutdownAtExit ch_in_shutdown{ch_in};
+    ShutdownAtExit chs_out_shutdown{chs_out};
+    co_await ctx->executor()->schedule();
+    auto& logger = ctx->logger();
+    auto state = std::make_shared<UnboundedFanoutState>(chs_out.size());
+
+    std::vector<Node> tasks;
+    tasks.reserve(chs_out.size() + 1);
+
+    auto& executor = *ctx->executor();
```
madsbk: Move to the top of the function

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,411 @@
...
+        // notify send_tasks to copy & send messages
+        co_await state->data_ready.notify_all();
+
+        // purge completed send_tasks. This will reset the messages to empty, so that
+        // they release the memory, however the deque is not resized. This guarantees
+        // that the indices are not invalidated. intentionally not locking the mtx
+        // here, because we only need to know a lower-bound on the last completed idx
+        // (ch_next_idx values are monotonically increasing)
+        while (state->purge_idx + 1 < last_completed_idx) {
+            state->recv_messages[state->purge_idx].reset();
```
madsbk: `Message::reset()` isn't thread-safe.

nirandaperera:
@madsbk I addressed your comments

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,430 @@
...
+ * @brief Wait for a data request from the send tasks.
+ *
+ * @param state The state of the unbounded fanout.
+ * @param last_completed_idx The index of the last completed message.
+ * @param latest_processed_idx The index of the latest processed message.
+ * @return True if the state is valid and can move forward, false otherwise (all send
+ * tasks are in an invalid state).
+ */
+auto wait_for_data_request(
+    UnboundedFanoutState& state, size_t& last_completed_idx, size_t& latest_processed_idx
```
madsbk: I would strongly suggest to return a std::tuple instead of the output parameters

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,430 @@
...
+    drain_tasks.reserve(chs_out.size());
+    for (auto& ch : chs_out) {
+        drain_tasks.emplace_back(ch->drain(ctx->executor()));
+    }
+    coro_results(co_await coro::when_all(std::move(drain_tasks)));
+    logger.trace("Completed bounded fanout");
+}
+
+/**
+ * @brief State for the unbounded fanout.
```
madsbk: I think it would be good with a description of the algorithm here.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,430 @@
...
+ *
+ * @param ctx The context to use.
+ * @param idx The index of the task
+ * @param ch_out The output channel to send messages to.
+ * @param state The state of the unbounded fanout.
+ * @return A coroutine representing the task.
+ */
+Node unbounded_fo_send_task(
+    Context& ctx,
+    size_t idx,
```
madsbk: Rename to something like `self`

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+/**
+ * @brief Process input messages and notify send tasks to copy & send messages.
+ *
+ * @param ctx The context to use.
+ * @param ch_in The input channel to receive messages from.
+ * @param state The state of the unbounded fanout.
+ * @return A coroutine representing the task.
+ */
+Node unbounded_fo_process_input_task(
```
madsbk: Rename to `unbounded_recv_task` ?

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,430 @@
...
+        // notify send_tasks to copy & send messages
+        co_await state->data_ready.notify_all();
+
+        // purge completed send_tasks. This will reset the messages to empty, so that
+        // they release the memory, however the deque is not resized. This guarantees
+        // that the indices are not invalidated. intentionally not locking the mtx
+        // here, because we only need to know a lower-bound on the last completed idx
+        // (ch_next_idx values are monotonically increasing)
+        while (purge_idx + 1 < last_completed_idx) {
+            state->recv_messages[purge_idx].reset();
```
madsbk: Do we need to purge here, could the last `unbounded_fo_send_task` not move the message out of `recv_messages`?

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,430 @@
...
+/**
+ * @brief State for the unbounded fanout.
+ */
+struct UnboundedFanoutState {
+    UnboundedFanoutState(size_t num_channels) : ch_next_idx(num_channels, 0) {}
+
+    coro::mutex mtx;
+    // notify send tasks to copy & send messages
+    coro::condition_variable data_ready;
+    // notify this task to receive more data from the input channel
```
madsbk:
```suggestion
    // notify recv task to receive more data from the input channel
```

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,430 @@
...
+                co_await set_ch_idx_invalid.set_channel_idx_invalid();
+                co_return;
+            }
+        }
+        logger.trace(
+            "sent ", idx, " [", state->ch_next_idx[idx], ", ", curr_recv_msg_sz, ")"
+        );
+
+        // now next_idx can be updated to end_idx, and if !input_done, we need to request
+        // parent task for more data
```
madsbk:
```suggestion
        // now next_idx can be updated to end_idx, and if !input_done, we need to request
        // the recv task for more data
```

madsbk (changes requested):

madsbk:
> If we don't want to consume the inputs until at least one output channel needs another message, we can use a semaphore.

This is important

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,441 @@
...
+    ShutdownAtExit c1{ch_in};
+    ShutdownAtExit c2{chs_out};
+    auto& logger = ctx->logger();
+    co_await ctx->executor()->schedule();
+    while (true) {
+        auto msg = co_await ch_in->receive();
+        if (msg.empty()) {
+            break;
+        }
+
```
wence-: *suggestion*:

```suggestion

        std::erase_if(chs_out, [](auto&& ch) { return ch->is_shutdown(); });
        if (chs_out.empty()) {
            break;
        }
```
So we don't end up sending (and hence copying) into closed channels.

And if everything is shut down we exit the loop.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,441 @@
...
+        size_t latest_processed_idx = InvalidIdx;
+
+        auto lock = co_await mtx.scoped_lock();
+        co_await request_data.wait(lock, [&] {
+            auto filtered_view = std::ranges::filter_view(ch_next_idx, [](size_t idx) {
+                return idx != InvalidIdx;
+            });
+
+            auto it = std::ranges::begin(filtered_view);  // first valid idx
+            auto end = std::ranges::end(filtered_view);  // end idx
```
wence-:
```suggestion
            auto it = std::ranges::begin(filtered_view);
            auto end = std::ranges::end(filtered_view);
```

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,441 @@
...
+            }
+
+            // notify send_tasks to copy & send messages
+            co_await data_ready.notify_all();
+
+            // purge completed send_tasks. This will reset the messages to empty, so that
+            // they release the memory, however the deque is not resized. This guarantees
+            // that the indices are not invalidated. intentionally not locking the mtx
+            // here, because we only need to know a lower-bound on the last completed idx
+            // (ch_next_idx values are monotonically increasing)
```
wence-: This comment makes no sense. We're not purging any send tasks, we're just releasing messages that have been handled.

`ch_next_idx` is not touched at all in this loop so the comment about the mutex doesn't seem to make much sense either.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,441 @@
...
+            while (purge_idx + 1 < last_completed_idx) {
+                recv_messages[purge_idx].reset();
+                purge_idx++;
+            }
```
wence-:
```suggestion
            while (purge_idx + 1 < last_completed_idx) {
                recv_messages[purge_idx++].reset();
            }
```

But question: why `purge_idx + 1 < last_completed_idx`? Suppose we have a deque with three messages `[A, B, C]`, `purge_idx` is currently zero. And `last_completed_idx` is two (which, to me, means every forwarder has processed `A, B, and C`).

But it seems that we will now only purge message `A`. (since `1 + 1 !< 2`).

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+/**
+ * @brief Broadcast messages from one input channel to multiple output channels.
+ *
+ * This is an all-purpose implementation that can support consuming messages by the
+ * channel order or message order. Output channels could be connected to
+ * single/multiple consumer nodes. A consumer node can decide to consume all messages
+ * from a single channel before moving to the next channel, or it can consume messages
+ * from all channels before moving to the next message. When a message has been sent
+ * to all output channels, it is purged from the internal deque.
```
wence-:
```suggestion
* In contrast to `bounded_fanout`, an unbounded fanout supports arbitrary 
* consumption orders of the output channels.
```

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,441 @@
...
+    std::shared_ptr<Channel> ch_in,
+    std::vector<std::shared_ptr<Channel>> chs_out
+) {
+    auto& executor = *ctx->executor();
+
+    ShutdownAtExit ch_in_shutdown{ch_in};
+    ShutdownAtExit chs_out_shutdown{chs_out};
+    co_await ctx->executor()->schedule();
+    auto& logger = ctx->logger();
+    auto fanout = std::make_unique<UnboundedFanout>(chs_out.size());
```
wence-: fanout never escapes this function, so can be stack allocated:


```suggestion
    auto fanout = UnboundedFanout(chs_out.size();
```

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,441 @@
...
+    RAPIDSMPF_EXPECTS(!chs_out.empty(), "output channels cannot be empty");
+
+    std::vector<coro::task<bool>> tasks;
+    tasks.reserve(chs_out.size());
+    for (size_t i = 0; i < chs_out.size() - 1; i++) {
+        // do a reservation for each copy, so that it will fallback to host memory if
+        // needed
+        auto res = ctx->br()->reserve_or_fail(msg.copy_cost(), try_memory_types(msg));
+        tasks.emplace_back(chs_out[i]->send(msg.copy(res)));
+    }
```
wence-: *question*: Should this copy and send pattern be scheduled? Otherwise we serialise all the copying and spilling.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,441 @@
...
+ * Additional considerations:
+ * - In the recv task loop, it also identifies the last completed message index by all
+ * send tasks. Message upto this index are no longer needed, and are purged from the
+ * cached messages.
+ * - When a send task fails to send a message, this means the channel may have been
+ * prematurely shut down. In this case, it sets its index to InvalidIdx. Recv task will
+ * filter out channels with InvalidIdx.
+ * - There two RAII helpers to ensure that the notification mechanisms are properly
+ * cleaned up when the unbounded fanout state goes out of scope/ encounters an error.
+ *
```
wence-: Please do some work on this explanation. It intermingles low-level implementation with algorithmic stuff. The latter is more important.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,441 @@
...
+            auto it = std::ranges::begin(filtered_view);  // first valid idx
+            auto end = std::ranges::end(filtered_view);  // end idx
+
+            if (it == end) {
+                // no valid indices, so all send tasks are in an invalid state
+                return true;
+            }
+
+            auto [min_it, max_it] = std::minmax_element(it, end);
+            last_completed_idx = *min_it;
```
wence-: "last" seems like the wrong verb here. I think this is the "lowest" completed index?

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+    /**
+     * @brief Send messages to multiple output channels.
+     *
+     * @param ctx The context to use.
+     * @param self Self index of the task
+     * @param ch_out The output channel to send messages to.
+     * @return A coroutine representing the task.
+     */
+    Node send_task(Context& ctx, size_t self, std::shared_ptr<Channel> ch_out) {
```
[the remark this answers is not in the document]
nirandaperera: Ah good point. I only took the idx in for logging. But I can see how that improves readability.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,441 @@
...
+                });
+                if (input_done && ch_next_idx[self] == curr_recv_msg_sz) {
+                    // no more messages will be received, and all messages have been sent
+                    break;
+                }
+            }
+
+            // now we can copy & send messages in indices [next_idx, curr_recv_msg_sz)
+            // it is guaranteed that message purging will be done only on indices less
+            // than next_idx, so we can safely send messages without locking the mtx
```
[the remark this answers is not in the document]
nirandaperera: If two threads are not modifying the same index of the deque, this is not an issue, isnt it? Currently the idea is multiple send tasks may read from an index `i`, but until all of send tasks have read `i`th index, it will not be updated. The last sent index vector is always updated under a mutex. So, I think the recv task and send tasks should not step on each others' toes.

cpp/src/streaming/core/fanout.cpp
nirandaperera: Ah! well, `lowest_completed_idx=min(per_ch_processed)` and `per_ch_processed` has the _next idx_. So, essentially it will reset 2 messages. But yeah, you are right. that `+ 1` is not required.

cpp/src/streaming/core/fanout.cpp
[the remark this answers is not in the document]
wence-: Just hold a vector<reference_wrapper<Message>> that you fill with the lock when woken and use that to get the messages

python/rapidsmpf/rapidsmpf/streaming/core/fanout.pyx
```diff
@@ -0,0 +1,77 @@
...
+    Parameters
+    ----------
+    ctx : Context
+        The node context to use.
+    ch_in : Channel
+        Input channel from which messages are received.
+    chs_out : list[Channel]
+        Output channels to which messages are broadcast.
+    policy : FanoutPolicy
+        The fanout strategy to use (see FanoutPolicy).
```
wence-: This isn't really documented in python I think. Can we have a sentence here describing the two options?

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,446 @@
...
+            auto lock = co_await mtx.scoped_lock();
+            self_next_idx = n_available_messages;
+            if (self_next_idx == recv_messages.size()) {
+                if (no_more_input) {
+                    // no more messages will be received, and all messages have been sent
+                    break;
+                } else {
+                    // request more data from the input channel
+                    lock.unlock();
+                    co_await request_data.notify_one();
```
wence-: I was reminded what was bugging me here. If there are multiple tasks running `notify_one` (or `notify_all`) on the same condvar, we can have lost wakeups unfortunately. I have tried to fix this in libcoro here: https://github.com/jbaldwin/libcoro/pull/416

wence-: I think the reason for that is that these tasks are typically not all firing at once. I think we are probably OK to merge now, but keep that in mind in case we see issues.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,446 @@
...
+ * input channel is fully consumed.
+ *
+ * Additional considerations:
+ * - In the recv task loop, it also identifies the lowest completed message index by all
+ * send tasks. Message upto this index are no longer needed, and are released from the
+ * cached messages deque.
+ * - When a send task fails to send a message, this means the channel may have been
+ * prematurely shut down. In this case, it sets a sential value to mark it as invalid.
+ * Recv task will filter out channels with the invalid sentinel value.
+ * - There two RAII helpers to ensure that the notification mechanisms are properly
```
madsbk:
```suggestion
 * - There are two RAII helpers to ensure that the notification mechanisms are properly
```

cpp/src/streaming/core/fanout.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+        auto lock = co_await mtx.scoped_lock();
+        co_await request_data.wait(lock, [&] {
+            auto filtered_view = std::ranges::filter_view(
+                per_ch_processed, [](size_t idx) { return idx != InvalidIdx; }
+            );
+
+            auto it = std::ranges::begin(filtered_view);  // advance to first valid idx
+            auto end = std::ranges::end(filtered_view);
+
```
madsbk:
```suggestion

```

madsbk (changes requested):

madsbk (approved):
LGTM, thanks @nirandaperera

PR #638, opened 2025-11-06
Ensure messages are dropped by refcount through channel sends

wence-:
The previous approach of defining local functions seems like it captures objects in a refcycle that is then only broken by the cyclic collector. This means that if we just sink messages without releasing their payload they might not get collected until the interpreter runs atexit teardown. If the destruction needs to deallocate device memory, the memory resource will already be gone leading to a segfault.

To fix this, use functools.partial to make the closures we pass to the C++ callbacks.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/streaming/core/channel.pyx
+++ b/python/rapidsmpf/rapidsmpf/streaming/core/channel.pyx
@@ -4,6 +4,7 @@
 from libcpp.memory cimport shared_ptr
 from libcpp.utility cimport move
 
+from functools import partial
 from rapidsmpf.streaming.core.context cimport Context, cpp_Context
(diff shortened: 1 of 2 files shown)


pentschev (approved):
> The previous approach of defining local functions seems like it captures objects in a refcycle that is then only broken by the cyclic collector.

Very interesting, I had no idea about this. Thanks for debugging and fixing! LGTM.

wence-:
> Very interesting, I had no idea about this. Thanks for debugging and fixing! LGTM.

I think it must be a consequence of the implementation in cython

pentschev (approved):
Ugh, this is indeed annoying that we need to handle empty messages differently. Thanks for fixing that. LGTM.

Issue #641, opened 2025-11-07
Make `BufferResource` usable as an rmm device_memory_resource


wence-:
The `BufferResource` holds our (adapted) RMM memory resource for device allocations, and also a pool of streams used for compute.

`rmm::device_buffer`s that are created from our memory resource and on a given stream hold two things that must be alive until their destructor runs:

1. A reference (non-owning) to the memory resource that allocated them
2. A `rmm::cuda_stream_view` (non-owning) to the stream that they were allocated on

As a consequence, the `BufferResource` needs to be alive until these destructors run, so that the `stream_view` and memory resources are still valid.

In C++ this is relatively easy to achieve (and the programming model makes it obvious that one must): create the `BufferResource` in `main` and then pass in to functions that use it. Its lifetime is then longer than that of any allocations made with it.

In contrast in Python a very "natural" thing to do is something like:

```
def run():
    resource = BufferResource(...)
    data = allocate_with(resource.get_mr(), resource.get_stream())
    return data
```

Now we're hosed, because nothing is keeping the `resource` alive and so when `data` is destroyed it will run its destruction on a `cuda_stream_view` that is already gone.

The issue of the memory resource going out of scope will be solved in an upcoming RMM release where we will use CCCL `any_resource` (owning) references in `device_buffer`s (stateful memory resources will be copied by shared_ptr-like semantics).

However, that still leaves the issue of the stream lifetime.

To fix this, I suggest that we make `BufferResource` implement the RMM `device_memory_resource` concepts so that it is directly usable as a memory resource. Then, any buffers in Python that are allocated with this resource will keep it alive until they are destroyed.

Since the BufferResource is now alive the stream pool, and hence streams, will also be alive for long enough. This also solves the problem for the C++ side of things.

PR #646, opened 2025-11-09
`Context`: use shared pointer for buffer resource

madsbk:
To simplify buffer resource lifetime management in Python, `Context` now holds a `std::shared_ptr` to its buffer resource instead of a raw pointer.

Previously, in Python, a `cpp_Context` could outlive its associated `cpp_BufferResource` because the former was kept alive by a shared pointer on the C++ side, while the latter was only referenced as a raw pointer and was freed when the Python `BufferResource` wrapper went out of scope.

### DIFF
--- a/cpp/src/streaming/core/context.cpp
+++ b/cpp/src/streaming/core/context.cpp
@@ -15,7 +15,7 @@ Context::Context(
     std::shared_ptr<Communicator> comm,
     std::shared_ptr<ProgressThread> progress_thread,
     std::unique_ptr<coro::thread_pool> executor,
-    BufferResource* br,
+    std::shared_ptr<BufferResource> br,
     std::shared_ptr<Statistics> statistics
 )
     : options_{std::move(options)},
@@ -34,7 +34,7 @@ Context::Context(
 Context::Context(
     config::Options options,
     std::shared_ptr<Communicator> comm,
-    BufferResource* br,
+    std::shared_ptr<BufferResource> br,
     std::shared_ptr<Statistics> statistics
 )
     : Context(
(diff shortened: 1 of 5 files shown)


pentschev (approved):
Left a question in form of a suggestion, but looks good, thanks for catching and fixing this!

cpp/src/streaming/core/context.cpp
```diff
@@ -34,7 +34,7 @@ Context::Context(
 Context::Context(
     config::Options options,
     std::shared_ptr<Communicator> comm,
-    BufferResource* br,
+    std::shared_ptr<BufferResource> br,
```
wence-: For all of these `std::move(br)` in the initialisation list of the ctor.

python/rapidsmpf/rapidsmpf/streaming/core/context.pyx
```diff
@@ -42,7 +42,6 @@ cdef class Context:
     ):
         self._comm = comm
         self._br = br
```
wence-: C++ context now keeps this alive so no need for a separate ref in python?

cpp/src/streaming/core/context.cpp
```diff
@@ -80,7 +80,7 @@ std::unique_ptr<coro::thread_pool>& Context::executor() noexcept {
 }
 
 BufferResource* Context::br() const noexcept {
-    return br_;
+    return br_.get();
 }
```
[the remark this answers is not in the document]
madsbk: Let's wait a bit. The rest of the C++ API uses raw pointers but I suspect that it might change with https://github.com/rapidsai/rapidsmpf/issues/641

PR #651, opened 2025-11-11
More python bindings

madsbk:
A lot of python bindings including a helper method that automatically extract and unspill a `TableChunk` (used in https://github.com/rapidsai/cudf/pull/20583):

```python
class TableChunk:
    ...
    def make_available_and_spill(
        self, BufferResource br not None, *, allow_overbooking
    ):
        """
        Make this table chunk available on device, spilling other data if necessary.

        Ensures that the data backing this table chunk is made available in device
        memory. If there is insufficient free memory to complete the operation, the
        buffer resource may spill other data until enough space has been freed.

        Parameters
        ----------
        br
            Buffer resource used for allocations and spill management.
        allow_overbooking
            Whether the memory reservation may temporarily exceed the current
            allocation limit.

        Returns
        -------
        A new table chunk with its data made available on device.

        Raises
        ------
        MemoryError
            If the allocation or spilling process fails to free enough memory.

        Warnings
        --------
        The original table chunk is released and must not be used after this call.

        Examples
        --------
        >>> # Make the data of an existing chunk available on device
        >>> chunk = chunk.make_available_and_spill(br, allow_overbooking=False)
        >>> chunk.table_view()
        """
```


python/rapidsmpf/rapidsmpf/streaming/cudf/table_chunk.pyx
```diff
@@ -291,6 +302,47 @@ cdef class TableChunk:
...
+
+        As part of the move, a copy or unpack operation may be performed,
+        using the associated CUDA stream for execution. After this call,
+        the current object is left in a moved-from state and should not be
+        accessed further except for reassignment, movement, or destruction.
+
+        Parameters
+        ----------
+        reservation
+            Memory reservation used for allocations, if needed.
```
pentschev:
```suggestion
            Memory reservation used for allocations, if making data available
            is needed.
```

python/rapidsmpf/rapidsmpf/streaming/core/spillable_messages.pyi
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-License-Identifier: Apache-2.0
+
+from rapidsmpf.buffer.content_description import ContentDescription
+from rapidsmpf.buffer.resource import BufferResource
+from rapidsmpf.streaming.core.message import Message
+
+class SpillableMessages:
+    def insert(self, message: Message) -> int: ...
+    def extract(self, mid: int) -> Message: ...
```
wence-:
```suggestion
    def insert(self, message: Message[Payload]) -> int: ...
    def extract(self, mid: int) -> Message[Payload]: ...
```

with appropriate import. Some typecheckers otherwise infer `Unknown` for the generic payload type.

python/rapidsmpf/rapidsmpf/streaming/cudf/table_chunk.pyx
```diff
@@ -330,3 +382,69 @@ cdef class TableChunk:
         True if the table chunk can be spilled, otherwise, False.
         """
         return deref(self.handle_ptr()).is_spillable()
+
+
+@singledispatch
```
wence-: *suggestion*: I think there's no need for `get_table_chunk` to accept a `Message`, since the thing it does in that case (`chunk = TableChunk.from_message(msg)`) is trivially available to the user.

Then this utility becomes a method of `TableChunk` that wraps up `make_available` with obtaining a reservation.

Could we instead implement a `to_device` utility method on `TableChunk`?

```
def to_device(self, BufferResource br not None, *, allow_overbooking):
    cdef MemoryReservation res = br.reserve_and_spill(...)
    return self.make_available(res)
```

wence- (changes requested):

python/rapidsmpf/rapidsmpf/streaming/cudf/table_chunk.pyx
madsbk: @wence- WDYT?

madsbk:
@wence- anything else?

PR #662, opened 2025-11-13
Include number of bytes spilled in nvtx payload

TomAugspurger:


nirandaperera:
I tested out the `HostBuffer` with @TomAugspurger's [reproducer](https://gist.github.com/TomAugspurger/d5b0d3b0e5765e448aa07a4fcc706171). 

image

Left is with HostBuffer and right is with std::vector. 
Now, `Buffer::allocate` time is insignificant. However, now `cudaMemcpyAsync` is taking up almost all the gains. The total time for each `insert` is more or less the same. :confused: 
Does `cudaMemcpyAsync` initialize memory before copying?

PR #665, opened 2025-11-14
[MINOR] Increase CI timeout to 10m

nirandaperera:
Currently CI test timeout is capped at 5m, but the test coverage has increased a lot during the past few months. We have seen some timeouts just because the test runs gets clipped by the timeout tool. 
So, this PR increases the timeout to 10m.


pentschev (approved):

bdice (approved):
Please write a PR description before merging so that the history of why this change occurred is preserved.

KyleFromNVIDIA (approved):

PR #679, opened 2025-11-20
Add missing `rrun` support to C++ benchmarks

pentschev:
Some C++ benchmarks are still missing `rrun` support, this change closes that gap.


pentschev:
Thanks Kyle and Mads!

PR #680, opened 2025-11-20
Add missing `rrun` support to Python benchmark

pentschev:
Expose functions to create UCXX communicators when launching with `rrun` to Python and support it with `bulk_mpi_shuffle` benchmark. Tests are not added in this PR because the existing `pytest` structure depends heavily on MPI, this will need to be done in a follow-up PR. Additionally, add a check for `rrun` and raise an exception in `streaming_benchmark` since it depends on `allreduce` that is currently implemented by MPI and we still don't have `allreduce` that's independent of MPI.

Depends on #679 .

### DIFF
--- a/ci/build_python.sh
+++ b/ci/build_python.sh
@@ -11,6 +11,8 @@ export CMAKE_GENERATOR=Ninja
 
 rapids-print-env
 
+rapids-generate-version > ./VERSION
+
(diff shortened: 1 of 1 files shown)


quasiben:
When this goes in I'll include the bulk_mpi_shuffle as part of our nightly measurement

PR #683, opened 2025-11-20
Add `AllReduce` collective

pentschev:
Implements an `AllReduce` collective on top of the `AllGather` implementation. This may or may not be an acceptable long-term implementation given the performance characteristics it delivers, see the `AllReduce` class docs for details, but it makes the implementation certainly easier to maintain.

```
/**
 * @brief AllReduce collective.
 *
 * The current implementation is built using `coll::AllGather` and performs
 * the reduction locally after allgather completes. Considering `R` is the number of
 * ranks, and `N` is the number of bytes of data, per rank this incurs `O(R * N)` bytes of
 * memory consumption and `O(R)` communication operations.
 *
 * Semantics:
 *  - Each rank calls `insert` exactly once to contribute data to the reduction.
 *  - Once all ranks call `insert`, `wait_and_extract` returns the
 *    globally-reduced `PackedData`.
 *
 * The actual reduction is implemented via a type-erased `ReduceOperator` that is
 * supplied at construction time. Helper factories such as
 * `detail::make_reduce_operator` (defaults to host-side) or
 * `detail::make_device_reduce_operator` (device-side) can be used to build
 * element-wise reductions over contiguous arrays.
 */
```

Both host-side and device-side reductions are supported, whether it happens on host or device depends on the operator being passed, for built-in operators there are `HostReduceOp` and `DeviceReduceOp`, and custom operators require defining a `ReduceOperator` with `ReduceOperatorType::Host` or `ReduceOperatorType::Device` to determine where it runs. Finally, the implementation ensures data is properly located before executing the reduction, any device data is moved to host when running a host reduction, and host data is moved to device when running a device reduction.

### DIFF
--- a/cpp/src/coll/allreduce.cpp
+++ b/cpp/src/coll/allreduce.cpp
@@ -66,10 +66,8 @@ void AllReduce::ensure_reduction_done() {
         if (finished_callback_) {
             finished_callback_();
         }
-        // Lock is released here at end of scope
     }
 
-    // Notify waiters after releasing the mutex to avoid "hurry up and wait"
     cv_.notify_all();
 }
 
@@ -88,31 +86,26 @@ bool AllReduce::finished() const noexcept {
 }
 
 PackedData AllReduce::wait_and_extract(std::chrono::milliseconds timeout) {
-    // If reduction is not done yet, wait on the condition variable
-    if (!reduction_done_.load(std::memory_order_acquire)) {
-        std::unique_lock lock(mutex_);
+    std::unique_lock lock(mutex_);
 
-        if (timeout.count() < 0) {
-            cv_.wait(lock, [this] {
-                return reduction_done_.load(std::memory_order_acquire);
-            });
-        } else {
-            bool completed = cv_.wait_for(lock, timeout, [this] {
-                return reduction_done_.load(std::memory_order_acquire);
-            });
-
-            if (!completed) {
-                RAPIDSMPF_FAIL(
-                    "AllReduce::wait_and_extract timed out waiting for reduction to "
-                    "complete",
-                    std::runtime_error
-                );
-            }
+    if (timeout.count() < 0) {
+        cv_.wait(lock, [this] {
+            return reduction_done_.load(std::memory_order_acquire);
+        });
+    } else {
+        bool completed = cv_.wait_for(lock, timeout, [this] {
+            return reduction_done_.load(std::memory_order_acquire);
+        });
+        if (!completed) {
+            RAPIDSMPF_FAIL(
+                "AllReduce::wait_and_extract timed out waiting for reduction to "
+                "complete",
+                std::runtime_error
+            );
         }
     }
 
     // Extract and return the result (this is destructive - can only be called once)
-    std::lock_guard lock(mutex_);
     RAPIDSMPF_EXPECTS(
         reduced_result_.has_value(),
         "AllReduce::wait_and_extract can only be called once",


KyleFromNVIDIA (approved):
Approved trivial CMake changes

cpp/include/rapidsmpf/coll/allreduce.hpp
```diff
@@ -0,0 +1,207 @@
...
+    std::shared_ptr<Statistics> statistics_;  ///< Statistics collection instance
+    ReduceKernel reduce_kernel_;  ///< Type-erased reduction kernel
+    std::function<void(void)> finished_callback_;  ///< Optional finished callback
+
+    AllGather gatherer_;  ///< Underlying allgather primitive
+
+    std::atomic<std::uint32_t> nlocal_insertions_{0};  ///< Number of local inserts
+    std::atomic<bool> reduced_computed_{
+        false
+    };  ///< Whether the reduction has been computed
```
wence-: Why do we need these at all? This object does no communication itself.

cpp/src/coll/device_kernels.cu
```diff
@@ -0,0 +1,180 @@
...
+    auto const count = acc_nbytes / sizeof(T);
+
+    // Input buffer: assume ready as provided by the communicator / AllReduce.
+    auto const* in_bytes = reinterpret_cast<std::byte const*>(in_buf->data());
+    auto const* in_ptr = reinterpret_cast<T const*>(in_bytes);
+
+    // Destination buffer: use write_access so the latest-write event is updated.
+    acc_buf->write_access([&](std::byte* acc_bytes, rmm::cuda_stream_view stream) {
+        auto* acc_ptr = reinterpret_cast<T*>(acc_bytes);
+        auto policy = rmm::exec_policy(stream);
```
wence-: nosync exec policy.

cpp/src/coll/allreduce.cpp
```diff
@@ -0,0 +1,122 @@
...
+        static_cast<std::size_t>(nlocal_insertions_.load(std::memory_order_acquire));
+    auto const n_per_rank = total / nranks;
+
+    // We allow non-uniform insertion counts across ranks but require that the local
+    // insertion count matches the per-rank contribution implied by the gather.
+    RAPIDSMPF_EXPECTS(
+        n_local == 0 || n_local == n_per_rank,
+        "AllReduce local insertion count does not match gathered contributions per rank",
+        std::runtime_error
+    );
```
wence-: This makes no sense. Either, everyone has to insert exactly `k` entries and the `i`th result is computed by reducing over the `PackedData` pieces that are sliced from `[i::k]`, or everyone inserts different numbers of entries (as is claimed is being done here) and then you can't do that.

cpp/src/coll/allreduce.cpp
```diff
@@ -0,0 +1,122 @@
...
+    //   [rank0:0..n_per_rank-1][rank1:0..n_per_rank-1]...[rankP-1:0..n_per_rank-1]
+    for (std::size_t k = 0; k < n_per_rank; ++k) {
+        // Start from rank 0's contribution for this logical insertion.
+        auto accum = std::move(gathered[k]);
+        for (std::size_t r = 1; r < nranks; ++r) {
+            auto idx = r * n_per_rank + k;
+            reduce_kernel_(accum, std::move(gathered[idx]));
+        }
+        results.emplace_back(std::move(accum));
+    }
```
wence-: Yes, this assumes that we have always have `k` insertions on every rank and then the reduction for output `i` is sliced from `gathered_results[i::k]`.

cpp/src/coll/device_kernels.cu
```diff
@@ -0,0 +1,180 @@
...
+
+        thrust::transform(policy, acc_ptr, acc_ptr + count, in_ptr, acc_ptr, op);
+    });
+}
```
wence-: stream-ordering not respected.

cpp/src/coll/allreduce.cpp
```diff
@@ -0,0 +1,122 @@
...
+
```
wence-: Yes, this assumes that we have always have `k` insertions on every rank and then the reduction for output `i` is sliced from `gathered_results[i::k]`.

cpp/include/rapidsmpf/coll/allreduce.hpp
```diff
@@ -0,0 +1,207 @@
...
+ * ranks, and `N` is the number of bytes of data, per rank this incurs `O(R * N)` bytes of
+ * memory consumption and `O(R)` communication operations.
+ *
+ * Semantics:
+ *  - Each rank may call `insert` any number of times with a local sequence number.
+ *  - Conceptually, the *k*-th insertion on each rank participates in a single
+ *    global reduction. That is, insertions are paired across ranks by their
+ *    local insertion order, not by sequence number values.
+ *  - Once all ranks call `insert_finished`, `wait_and_extract` returns one
+ *    globally-reduced `PackedData` per local insertion on this rank.
```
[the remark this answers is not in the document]
wence-: It makes sense to be able to insert multiple entries in allgather because you want the concatenation of all the inputs.

In allreduce, you want the accumulation of all the inputs, and I think we should just simplify and say "you should combine your local inputs before inserting them".

Matching what allgather does just because we happen to be using an allgather to implement the reduction is the wrong way to think about this, I think.

cpp/include/rapidsmpf/coll/allreduce.hpp
```diff
@@ -0,0 +1,207 @@
...
+    ReduceKernel reduce_kernel_;  ///< Type-erased reduction kernel
+    std::function<void(void)> finished_callback_;  ///< Optional finished callback
+
+    AllGather gatherer_;  ///< Underlying allgather primitive
+
+    std::atomic<std::uint32_t> nlocal_insertions_{0};  ///< Number of local inserts
+    std::atomic<bool> reduced_computed_{
+        false
+    };  ///< Whether the reduction has been computed
+    std::vector<PackedData> reduced_results_;  ///< Cached reduced results
```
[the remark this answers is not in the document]
pentschev: Removed in X as well.

cpp/src/coll/allreduce.cpp
pentschev: Also removed in https://github.com/rapidsai/rapidsmpf/pull/683/commits/b353aa35388c6c355e2d351505d2ada192f22c85.

cpp/include/rapidsmpf/coll/allreduce.hpp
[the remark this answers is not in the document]
pentschev: Not everything is already done, but the API has been updated to only allow one `PackedData` per rank in https://github.com/rapidsai/rapidsmpf/pull/683/commits/b353aa35388c6c355e2d351505d2ada192f22c85.

cpp/include/rapidsmpf/coll/allreduce.hpp
```diff
@@ -0,0 +1,191 @@
...
+ * @brief AllReduce collective.
+ *
+ * The current implementation is built using `coll::AllGather` and performs
+ * the reduction locally after allgather completes. Considering `R` is the number of
+ * ranks, and `N` is the number of bytes of data, per rank this incurs `O(R * N)` bytes of
+ * memory consumption and `O(R)` communication operations.
+ *
+ * Semantics:
+ *  - Each rank calls `insert` exactly once to contribute data to the reduction.
+ *  - Once all ranks call `insert_finished`, `wait_and_extract` returns the
```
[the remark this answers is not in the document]
pentschev: Yes, I hadn't finished making API changes after removing support for multiple insertions per rank. Removed it now.

cpp/include/rapidsmpf/coll/allreduce.hpp
```diff
@@ -0,0 +1,191 @@
...
+     *       that uses the same communicator, progress thread, and buffer resource.
+     */
+    AllReduce(
+        std::shared_ptr<Communicator> comm,
+        std::shared_ptr<ProgressThread> progress_thread,
+        OpID op_id,
+        BufferResource* br,
+        std::shared_ptr<Statistics> statistics = Statistics::disabled(),
+        ReduceKernel reduce_kernel = {},
+        std::function<void(void)> finished_callback = nullptr
```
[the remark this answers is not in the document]
pentschev: Lawrence is right, and since we build on `AllGather` we can't (easily) make that change. For now, let's follow what `AllGather` does.

cpp/include/rapidsmpf/coll/allreduce.hpp
```diff
@@ -0,0 +1,464 @@
...
+
+    if constexpr (Op == HostReduceOp::SUM) {
+        return [apply](PackedData& accum, PackedData&& incoming) {
+            if constexpr (std::is_same_v<T, bool>) {
+                apply(accum, std::move(incoming), [](T a, T b) {
+                    return static_cast<T>(a || b);
+                });
+            } else {
+                apply(accum, std::move(incoming), std::plus<T>{});
+            }
```
nirandaperera: I think `std::plus<bool>` behaves the same as logical OR

cpp/include/rapidsmpf/coll/allreduce.hpp
```diff
@@ -0,0 +1,464 @@
...
+    is_supported_reduce_op<T, OpEnum, Op>::value;
+
+/**
+ * @brief Create a host-based element-wise reduction operator for a given (T, Op).
+ *
+ * The operator assumes that the `PackedData::data` buffers reside in host memory
+ * and contain a contiguous array of @p T with identical sizes on all ranks.
+ */
+template <typename T, HostReduceOp Op>
+ReduceOperatorFunction make_host_reduce_operator_impl() {
```
nirandaperera: Best to put these `impl`s in a `detail` namespace.

cpp/src/coll/device_kernels.cu
```diff
@@ -0,0 +1,207 @@
...
+        // Safe to read from incoming buffer after the stream join above.
+        auto const* in_bytes = reinterpret_cast<std::byte const*>(in_buf->data());
+        auto const* in_ptr = reinterpret_cast<T const*>(in_bytes);
+
+        thrust::transform(policy, acc_ptr, acc_ptr + count, in_ptr, acc_ptr, op);
+    });
+}
+
+template <typename T, DeviceReduceOp Op>
+struct ReduceOperatorMaker;
```
nirandaperera: I prefer operator specialization like this, rather than the `if-else constexpr` in host operators. This is much cleaner.

cpp/src/coll/device_kernels.cu
```diff
@@ -0,0 +1,207 @@
...
+ReduceOperatorFunction
+make_device_reduce_operator_impl<unsigned long, DeviceReduceOp::MIN>() {
+    return make_reduce_operator_impl<unsigned long, DeviceReduceOp::MIN>();
+}
+
+template <>
+ReduceOperatorFunction
+make_device_reduce_operator_impl<unsigned long, DeviceReduceOp::MAX>() {
+    return make_reduce_operator_impl<unsigned long, DeviceReduceOp::MAX>();
+}
```
nirandaperera: again, I think we can use `concept`s here

cpp/include/rapidsmpf/coll/allreduce.hpp
```diff
@@ -0,0 +1,464 @@
...
+        );
+
+        auto* acc_bytes = acc_buf->exclusive_data_access();
+        auto* in_bytes = in_buf->exclusive_data_access();
+        auto* acc_ptr = reinterpret_cast<T*>(acc_bytes);
+        auto const* in_ptr = reinterpret_cast<T const*>(in_bytes);
+
+        for (std::size_t i = 0; i < count; ++i) {
+            acc_ptr[i] = op(acc_ptr[i], in_ptr[i]);
+        }
```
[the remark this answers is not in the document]
pentschev: I think we can't use `std::ranges::transform` with raw pointers, but we can use `std::transform`. Done in X.

cpp/include/rapidsmpf/coll/allreduce.hpp
pentschev: Yes, I thought so too, just wanted to be verbose, but removed now with B.

cpp/include/rapidsmpf/coll/allreduce.hpp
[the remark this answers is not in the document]
pentschev: I think this is not applicable anymore. The new for loop could be replaced by `std::for_each` I think, but it's probably more complicated than the for loop, so I left it as a for loop for now.

cpp/src/coll/allreduce.cpp
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    // Start with rank 0's contribution as the accumulator
+    auto accum = std::move(gathered[0]);
+
+    // Reduce contributions from all other ranks into the accumulator
+    for (std::size_t r = 1; std::cmp_less(r, nranks_); ++r) {
+        reduce_operator_(accum, std::move(gathered[r]));
+    }
```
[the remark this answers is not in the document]
pentschev: Please have a look at the changes discussed in https://github.com/rapidsai/rapidsmpf/pull/683/files#r2610331940. I think they cover most of what you suggested minus the `span`/`ranges`. Unfortunately, as I understand it, the problem will still remain for device code as I wrote in the linked discussion, but I'm open to other ideas.

cpp/include/rapidsmpf/coll/allreduce.hpp
```diff
@@ -0,0 +1,407 @@
...
+namespace detail {
+
+/**
+ * @brief Host-side elementwise operator working on raw bytes.
+ *
+ * The callable receives pointers to a single element of `accum` and `incoming`
+ * respectively and must combine them in-place into `accum`. The caller
+ * guarantees that both pointers reference `element_size` bytes.
+ */
+using HostByteOp = std::function<void(void* accum_elem, void const* incoming_elem)>;
```
nirandaperera: Based on the new changes proposed by @wence- [here](https://github.com/rapidsai/rapidsmpf/pull/683/changes#r2610414273), I think its best to change the reduction function to accept `PackedData&`  rather than individual elements here. I think an element-wise `std::function` would be a bit too inefficient. So, I think let's go ahead with Lawrence's suggestion.

nirandaperera (changes requested):
@pentschev sorry for the late revisit. But I think I'm more inclined to the aggregation operator definition @wence- suggested. I think element-wise aggregation might not be transparent enough for the compiler to optimize (due to std::function overhead).

cpp/src/coll/allreduce.cpp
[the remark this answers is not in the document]
pentschev: I've made changes to use span in https://github.com/rapidsai/rapidsmpf/pull/683/commits/d233ffa152fc3161775fe3874b63bc59ec429d8b, which I believe addresses this. @wence- @nirandaperera please have another look when you get the chance.

cpp/include/rapidsmpf/coll/allreduce.hpp
```diff
@@ -0,0 +1,423 @@
...
+     *
+     * Blocks until the allreduce operation completes and returns the
+     * globally reduced result.
+     *
+     * @param timeout Optional maximum duration to wait. Negative values mean
+     *        no timeout.
+     *
+     * @return The reduced packed data.
+     * @throws std::runtime_error If the timeout is reached.
+     */
```
madsbk:
```suggestion
    /**
     * @brief Wait for completion and extract the reduced data.
     *
     * Blocks until the allreduce operation completes and returns the
     * globally reduced result.
     *
     * This method is destructive and can only be called once. The first call
     * extracts data from the underlying AllGather and performs the reduction.
     * Subsequent calls will throw std::runtime_error because the underlying
     * data has already been consumed.
     *
     * @param timeout Optional maximum duration to wait. Negative values mean
     * no timeout.
     * @return The reduced packed data.
     *
     * @throws std::runtime_error If the timeout is reached or if this method is
     * called more than once.
     */
```

madsbk (commented):
Style: Do not indent after `@return` et al. We want:
```
 * @return True if all data and finish messages from all ranks have been
 * received and locally reduced.
```
Pro tip: ask your LLM not to indent after Doxygen `@param`, `@return`, `@throws`, and `@note`. That’s what I did 🙂

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -123,7 +127,9 @@ class BufferResource {
         RAPIDSMPF_EXPECTS(
             pinned_mr_, "no pinned memory resource is available", std::invalid_argument
         );
-        return *pinned_mr_;
+        return rmm::host_async_resource_ref{
+            cuda::mr::resource_ref<cuda::mr::host_accessible>{*pinned_mr_}
+        };
     }
```
nirandaperera: I'm still not sure about this change. Why is this not a problem in the `main`?

wence- (approved):
Thanks!

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -123,7 +123,9 @@ class BufferResource {
         RAPIDSMPF_EXPECTS(
             pinned_mr_, "no pinned memory resource is available", std::invalid_argument
         );
-        return *pinned_mr_;
+        // Upcast to HostMemoryResource& to get only host_accessible property
+        // (PinnedMemoryResource has both host_accessible and device_accessible)
+        return static_cast<HostMemoryResource&>(*pinned_mr_);
```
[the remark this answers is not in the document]
pentschev: Ok, I moved the implementation into the cpp file in https://github.com/rapidsai/rapidsmpf/pull/683/commits/969b6edf2fa9156a3e3be68b8131f5240a7925df, I think that's probably better than the current option casting types.

cpp/tests/test_allreduce.cpp
```diff
@@ -0,0 +1,787 @@
...
+        data[j] = this_rank;
+    }
+    auto packed = make_packed(br.get(), data.data(), data.size(), MemoryType::DEVICE);
+    allreduce.insert(std::move(packed));
+
+    auto result = allreduce.wait_and_extract();
+
+    // After wait_and_extract completes, callback should have been called exactly once
+    EXPECT_TRUE(callback_called.load(std::memory_order_acquire));
+    EXPECT_EQ(callback_count.load(std::memory_order_acquire), 1);
```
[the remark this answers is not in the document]
pentschev: I hope I didn't overwork this, but I think I've fixed to ensure `reduce_all` completes before executing `finished_callback` in https://github.com/rapidsai/rapidsmpf/pull/683/commits/a3fb2a786fd926e834951b3baa464c9a77d61e0e. With that I think this is not a problem anymore for `AllReduce`, but I've also applied your suggestion for `AllGather` via https://github.com/rapidsai/rapidsmpf/pull/683/commits/384d9d52fae1720cca6720beb2229484fd79753d, as you're right we should execute the callback before notifying.

cpp/src/coll/allreduce.cpp
```diff
@@ -34,8 +35,42 @@ AllReduce::AllReduce(
...
+void AllReduce::ensure_reduction_done() const {
+    // Fast path: check if reduction is already done.
+    if (reduction_done_.load(std::memory_order_acquire)) {
+        return;
+    }
+
+    // If we're here this should always be true, but check just in case.
+    if (!gatherer_.finished()) {
+        return;
+    }
```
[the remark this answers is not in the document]
pentschev: We can't loudly fail, I think. This is now executing on the progress thread, any failures wouldn't propagate back to the user, and we have no logging mechanism here.

cpp/src/coll/allreduce.cpp
```diff
@@ -34,8 +35,42 @@ AllReduce::AllReduce(
           op_id,
           br,
           std::move(statistics),
-          std::move(finished_callback)
-      } {}
+          [this]() { ensure_reduction_done(); }
+      },
+      finished_callback_{std::move(finished_callback)} {}
+
+void AllReduce::ensure_reduction_done() const {
```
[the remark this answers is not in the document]
pentschev: You're right, this was part of an initial attempt that wasn't right. Fixed now in https://github.com/rapidsai/rapidsmpf/pull/683/commits/fc4a90a21ac9e4a773036a40805af83ede33e2a7.

cpp/src/coll/allreduce.cpp
```diff
@@ -34,8 +35,42 @@ AllReduce::AllReduce(
...
+    // Fast path: check if reduction is already done.
+    if (reduction_done_.load(std::memory_order_acquire)) {
+        return;
+    }
```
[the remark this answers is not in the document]
pentschev: You're right, but we want to prevent ourselves from accidentally calling it again, no?

cpp/src/coll/allreduce.cpp
```diff
@@ -0,0 +1,161 @@
...
+        reduction_done_.store(true, std::memory_order_release);
+
+        // Call callback with mutex held to ensure consistent state
+        if (finished_callback_) {
+            finished_callback_();
+        }
+        // Lock is released here at end of scope
+    }
+
+    // Notify waiters after releasing the mutex to avoid "hurry up and wait"
```
wence-:
```suggestion
```

PR #685, opened 2025-11-21
Expose async collectives to Python

### LINKED ISSUE #682: Async Shuffler and AllGather bindings in Python
CuDF-Polars needs to construct `Shuffler` and `AllGather` objects at runtime within custom PyNode objects. My local experiments suggest that we need to use async objects, because we must be able to extract data from the collective operation without blocking the event loop. Otherwise, we can easily run into a race condition.

Note: I also see UCX listener setup errors when multiple collectives are created at the same time (I'd rather not debug these errors if the underlying problem is just that we need to use the async version of Shuffler and AllGather).

wence-:
This adds bindings for both the streaming node versions, and the object-based implementation.

- Closes #682

### DIFF
--- a/python/rapidsmpf/rapidsmpf/shuffler.pxd
+++ b/python/rapidsmpf/rapidsmpf/shuffler.pxd
@@ -37,6 +37,24 @@ cdef extern from "<rapidsmpf/shuffler/shuffler.hpp>" nogil:
         uint32_t wait_any() except +
         void wait_on(uint32_t pid) except +
         string str() except +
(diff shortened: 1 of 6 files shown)


madsbk (approved):
Looks good, I only have one suggestion

wence-:
> 1. Do you want further review on this, or do you feel it's ready?

I mean, take a look. I think it does what you're after.

rjzamora (approved):
Thanks @wence-!
Python API looks good. We can always follow-up if I missed something.

PR #687, opened 2025-11-22
Reduce lock contention in Shuffler `PostBox` 

### LINKED ISSUE #674: Reduce lock contention in spilling
As discussed in https://github.com/rapidsai/rapidsmpf/issues/657#issuecomment-3549102684, spill threads can spend a significant proportion of their time waiting for a lock while some other thread does its spilling:

Image

At the moment, that lock protects both

- spilling: https://github.com/rapidsai/rapidsmpf/blob/1df992ad5be382eb2580b73bbdf32a1134876d77/cpp/src/shuffler/shuffler.cpp#L818-L819
- extraction: https://github.com/rapidsai/rapidsmpf/blob/1df992ad5be382eb2580b73bbdf32a1134876d77/cpp/src/shuffler/shuffler.cpp#L771

As discussed in https://github.com/rapidsai/rapidsmpf/issues/657, the actual act of spilling buffers (allocating host memory, doing the host to device transfer) can take a substantial amount of time. *If* the lock is only there to protect attempting to spill the *same* buffer multiple times, we might be able to refactor the code split `postbox_spilling` into two distinct phases:

1. A phase to determine which set of buffers to spill, in order to reach some target amount of bytes spilled
2. A phase to actually spill that set of buffers identified in phase one.

That kind of spilling should only require a lock for phase 1.

However, because the same lock is used for extraction, things might be harder. Threads doing an `extract` (with the lock) might rely on some invariant like a buffer either being in device memory or host memory, but not in the process of being spilled. We might need to introduce a new "spilling" state, but I'm hazy on the details at this point.

This is related to the broader themes around spilling performance discussed in github.com/rapidsai/rapidsmpf/issues/657. As we improve the performance of spilling, lock contention ought to go down since the spill thread will spend less time actually spilling, and so will spend less time holding the lock with today's implementation.

cc @nirandaperera.

nirandaperera:
This PR introduces the following changes to `Shuffler PostBox`. 
- Currently `PostBox` has an internal mutex and to prevent extractions during spilling, its is also protected by an external mutex in the Shuffler. IINM, the idea there is, only to block extraction and still allow insertions during spilling. 
- This PR introduces finer grained locking to the postbox. It adds a mutex _per key_, and removes the external mutex in the shuffler. 
- It also now requires the `keys` during the postbox construction. This allows keys to be populated during the initialization. This allows us to remove the class level mutex in the PostBox. Consequence of this is, that the emptiness test has now become more complicated. Its handled by an atomic counter. 
- Additionally, we were stashing values as `unordered_map<ChunkID, Chunk>` map. This PR changes this to use `vector<Chunk>` because we will no longer be querying from ChunkID values. 

Todo: 
- [x]  Add multithreaded spilling test
- [ ] Do some nsys profiling  

profiling

Closes #674

### DIFF
--- a/cpp/include/rapidsmpf/shuffler/postbox.hpp
+++ b/cpp/include/rapidsmpf/shuffler/postbox.hpp
@@ -6,7 +6,9 @@
 
 #include <atomic>
 #include <functional>
+#include <list>
 #include <mutex>
+#include <random>
 #include <ranges>
 #include <string>
 #include <unordered_map>
@@ -50,6 +52,8 @@ class PostBox {
                 std::forward_as_tuple()
             );
         }
+        rng_ = std::mt19937(std::random_device{}());
+        dist_ = std::uniform_int_distribution<size_t>(0, keys.size() - 1);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


nirandaperera:
@TomAugspurger Can you run a profile with this PR?

nirandaperera:
@TomAugspurger Looking at the profile, it does seem like there are mutex lock regions corresponding to spill regions. I'm suspecting these could be insert/extract calls. I think one problem with this impl is, `insert/ extract` locks each key _per chunk_ while `spill` locks all the chunks. So, its quite possible that spilling receives more time. 
eg. pid 10 has 100 chunks and insert is trying to append another, while spill has lock and keeps on spilling for all 100 chunks. Since spilling takes long time, this unnecessarily delays insert. Hmm. I need to verify this. Let me check this again tomorrow.

madsbk:
I’m not sure it’s actually a problem that calls to `SpillManager::spill()` are serialized. We may eventually need to parallelize the work *inside* `spill()`, but it seems reasonable to wait if a spill operation is already in progress?

Have you tried increasing `periodic_spill_check`? [The default is only 1 ms](https://github.com/rapidsai/rapidsmpf/blob/4fb022455d3f9b72a5d786170ba617f9c626a945/cpp/include/rapidsmpf/memory/buffer_resource.hpp#L71), which might cause it to trigger far more often than needed.

madsbk:

PR #688, opened 2025-11-25
Memory reserve or wait

madsbk:
Introduce a new class, `MemoryReserveOrWait`, that provides asynchronous waiting for memory reservation requests.

```c++ 
    /**
     * @brief Attempts to reserve memory or waits until the reservation can be satisfied.
     *
     * This coroutine submits a memory reservation request and then suspends until
     * either sufficient memory becomes available or no progress is made within the
     * configured timeout.
     *
     * If the timeout expires before the request can be fulfilled, an empty
     * `MemoryReservation` is returned.
     *
     * @param size Number of bytes to reserve.
     * @param future_release_potential Estimated number of bytes the requester may release
     * in the future, used as a heuristic when selecting which eligible request to satisfy
     * first.
     * @return A `MemoryReservation` representing the allocated memory, or an empty
     * reservation if the timeout expires.
     *
     * @throw std::runtime_error If shutdown occurs before the request can be processed.
     */
    coro::task<MemoryReservation> reserve_or_wait(
        std::size_t size, std::size_t future_release_potential
    );
```

### DIFF
--- a/cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
+++ b/cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
@@ -1,5 +1,5 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES.
  * SPDX-License-Identifier: Apache-2.0
  */
 
--- a/cpp/src/streaming/core/memory_reserve_or_wait.cpp
+++ b/cpp/src/streaming/core/memory_reserve_or_wait.cpp
@@ -1,5 +1,5 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES.
  * SPDX-License-Identifier: Apache-2.0
  */
 
@@ -56,7 +56,7 @@ Node MemoryReserveOrWait::shutdown() {
     if (!reservation_requests.empty()) {
         std::vector<Node> nodes;
         for (Request const& request : reservation_requests) {
-            nodes.push_back(request.queue.shutdown_drain(ctx_->executor()));
+            nodes.push_back(request.queue.shutdown());
         }
         coro_results(co_await coro::when_all(std::move(nodes)));
     }
--- a/cpp/tests/streaming/test_memory_reserve_or_wait.cpp
+++ b/cpp/tests/streaming/test_memory_reserve_or_wait.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
  */
 
--- a/cpp/benchmarks/streaming/ndsh/utils.hpp
+++ b/cpp/benchmarks/streaming/ndsh/utils.hpp
@@ -1,5 +1,5 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES.
  * SPDX-License-Identifier: Apache-2.0
  */
 
--- a/cpp/include/rapidsmpf/memory/content_description.hpp
+++ b/cpp/include/rapidsmpf/memory/content_description.hpp
@@ -1,5 +1,5 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES.
  * SPDX-License-Identifier: Apache-2.0
  */
 #pragma once
--- a/cpp/include/rapidsmpf/streaming/core/context.hpp
+++ b/cpp/include/rapidsmpf/streaming/core/context.hpp
@@ -1,5 +1,5 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES.
  * SPDX-License-Identifier: Apache-2.0
  */
 
--- a/cpp/src/allgather/allgather.cpp
+++ b/cpp/src/allgather/allgather.cpp
@@ -1,5 +1,5 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES.
  * SPDX-License-Identifier: Apache-2.0
  */
 #include <algorithm>
--- a/cpp/src/streaming/core/context.cpp
+++ b/cpp/src/streaming/core/context.cpp
@@ -1,5 +1,5 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES.
  * SPDX-License-Identifier: Apache-2.0
  */
 
--- a/cpp/tests/streaming/base_streaming_fixture.hpp
+++ b/cpp/tests/streaming/base_streaming_fixture.hpp
@@ -1,5 +1,5 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES.
  * SPDX-License-Identifier: Apache-2.0
  */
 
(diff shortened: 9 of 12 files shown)


cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
```diff
@@ -0,0 +1,175 @@
...
+     *
+     * @return The total number of memory-check iterations executed so far.
+     */
+    [[nodiscard]] std::size_t periodic_memory_check_counter() const;
+
+  private:
+    /**
+     * @brief Represents a single memory reservation request.
+     *
+     * A `ResReq` is inserted into a sorted container and processed by
```
wence-: *nit*: Can we call this a `ReservationRequest` (or perhaps `Request` since it's a nested class)?

cpp/src/streaming/core/memory_reserve_or_wait.cpp
```diff
@@ -0,0 +1,229 @@
...
+    if (spawn_periodic_memory_check) {
+        // A previous periodic task may exist but is guaranteed to be either already
+        // finished or about to finish. This can happen when the last request was
+        // extracted and the task is in the process of exiting.
+        //
+        // We take ownership of that task here and await it below before proceeding,
+        // ensuring that at most one periodic task is active at any time.
+        previous_periodic_task = std::move(periodic_memory_check_task_);
+        periodic_memory_check_task_ =
+            ctx_->executor()->spawn_joinable(periodic_memory_check());
```
wence-:
```suggestion
        // We take ownership of that task here and await it below before proceeding,
        // ensuring that at most one periodic task is active at any time.
        previous_periodic_task = std::exchange(periodic_memory_check_task_,
            ctx_->executor()->spawn_joinable(periodic_memory_check());
```
?

cpp/src/streaming/core/memory_reserve_or_wait.cpp
```diff
@@ -0,0 +1,229 @@
...
+            // Try to reserve memory for the selected request.
+            auto [res, _] = ctx_->br()->reserve(mem_type_, it->size, no_overbooking);
+            if (res.size() == 0) {
+                continue;  // Memory is no longer available.
+            }
+
+            // Extract the selected request and push the reservation into its queue.
+            ResReq request = reservation_requests_.extract(it).value();
+            lock.unlock();
+            push_into_queue(request, std::move(res));
```
wence-: *issue*: Maybe? `push_into_queue` takes the request by reference. But that reference could be dead by the time the queue is accessed inside the task?

cpp/src/streaming/core/memory_reserve_or_wait.cpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+coro::task<MemoryReservation> MemoryReserveOrWait::reserve_or_wait(
+    std::size_t size, std::size_t future_release_potential
+) {
+    // First, check whether the requested memory is immediately available.
+    auto [res, _] = ctx_->br()->reserve(mem_type_, size, no_overbooking);
+    if (res.size() > 0) {
+        co_return std::move(res);
+    }
```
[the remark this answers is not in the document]
madsbk: > Might need a change in `reserve` to always return successfully for zero-sized allocations.

it does

cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
```diff
@@ -0,0 +1,175 @@
...
+        std::uint64_t sequence_number;
+
+        /// @brief Queue into which a reservation is pushed once the request is satisfied.
+        coro::queue<MemoryReservation>& queue;
+
+        /// @brief Lexicographic ordering.
+        friend bool operator<(ResReq const& a, ResReq const& b) {
+            return std::tie(a.size, a.future_release_potential, a.sequence_number)
+                   < std::tie(b.size, b.future_release_potential, b.sequence_number);
+        }
```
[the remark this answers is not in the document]
madsbk: now using:
```c++
        /// @brief Ordering by `size` and `sequence_number` (ascending).
        friend bool operator<(Request const& a, Request const& b) {
            return std::tie(a.size, a.sequence_number)
                   < std::tie(b.size, b.sequence_number);
        }
```

Note, we never use the `future_release_potential` ordering.

cpp/tests/streaming/test_memory_reserve_or_wait.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+class StreamingMemoryReserveOrWait : public BaseStreamingFixture {
+  public:
+    // Create a buffer resource that can be controlled by set_mem_avail()
+    // and get_mem_avail().
+    void SetUp() override {
+        auto dev_mem_available = [this]() -> std::int64_t {
+            return mem_avail_.load(std::memory_order_acquire);
+        };
+        SetUpWithThreads(1, {{MemoryType::DEVICE, dev_mem_available}});
```
nirandaperera: I think this test needs to be parameterized on multiple threads in the pool. I've uncovered several deadlocks that way

cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
```diff
@@ -0,0 +1,196 @@
...
+        /// @brief Monotonically increasing identifier used to preserve submission order.
+        std::uint64_t sequence_number;
+
+        /// @brief Queue into which a reservation is pushed once the request is satisfied.
+        coro::queue<MemoryReservation>& queue;
+
+        /// @brief Ordering by `size` and `sequence_number` (ascending).
+        friend bool operator<(Request const& a, Request const& b) {
+            return std::tie(a.size, a.sequence_number)
+                   < std::tie(b.size, b.sequence_number);
```
nirandaperera: When looking at the `periodic_memory_check` impl, I feel like the sequence_number is immaterial. We take the smallest/ range of leq requests based on size, and then try to find the one max future_release_potential. So, I feel like sequence_number order is not really required, and comparing on just the size is sufficient.

cpp/src/streaming/core/memory_reserve_or_wait.cpp
```diff
@@ -0,0 +1,258 @@
...
+                    co_return;
+                }
+            }
+            periodic_memory_check_counter_.fetch_add(1, std::memory_order_acq_rel);
+            co_await ctx_->executor()->yield();
+            if (Clock::now() - last_reservation_success > timeout_) {
+                // This is the only way out of the while-loop that doesn't shutdown
+                // the periodic memory check.
+                break;
+            }
```
nirandaperera: I'm missing something here. Shouldnt `timeout_` be evaluated for each request? `periodic_memory_check` is a long-running task per `MemoryReserveOrWait` instance. So, when the timeout expires, we pick the smallest request with the future_release_potential, and try to reserve that. BUT this could actually be the _last_  request that was added. And if the reservation fails, we inform it that it has failed. But in reality, it could have been attempted again later. 

What I'm trying to say is that, I feel like the Request should have a timestamp that tells us when it was created, and based on time-now, any expired requests should also be released.

nirandaperera (dismissed):

cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
madsbk: It is used as tie breaker so that each request is unique, it is also used on timeout:
```c++
        // Among the smallest requests, pick the one with the largest
        // future_release_potential. If multiple requests tie, we pick the oldest one,
        // since the set is ordered by size and then sequence_number (ascending).
        auto it = std::ranges::max_element(
            std::ranges::subrange(first, same_size_end),
            std::less<>{},
            &Request::future_release_potential
        );
```

cpp/src/streaming/core/memory_reserve_or_wait.cpp
```diff
@@ -0,0 +1,259 @@
...
+                    "could not push memory reservation"
+                );
+            }(queue, std::move(res))
+        );
+        RAPIDSMPF_EXPECTS(err, "cannot spawn push-into-queue task");
+    };
+
+    while (true) {
+        auto last_reservation_success = Clock::now();
+        while (true) {
```
nirandaperera: I have a bit of a concern with this double-while-loop. Consider the scenario where, memory is fully/almost fully reserved, and none of the request can be served. 
Then, IINM, this coroutine will wake up from the `yield` and very quickly `yeild` back in the next iteration, and continue to do so, until the timeout expires. Then on the outer loop, keep on draining the request queue. I feel like this is more-or-less a busy loop. 

Why didnt we consider a callback approach? Where we register a callback in BufferResource that will be called for every release. Then based on the callback call, we can wake up coroutines that are waiting in the request queue? 
I feel like we might not need this periodic mem check task that way? Then we would not have to worry about the busy-loops IMO. 
Basically "do nothing until someone releases memory".

madsbk: The main limitation of a pure callback-based approach is that `BufferResource` does not have full visibility into memory usage. It only sees allocations made via `BufferResource::allocate()`, not allocations performed by libcudf, which can represent a large share of the memory pressure and are often why a reservation cannot be satisfied.

I agree this is a valid concern. When the system is effectively out of memory, we can end up in a _wake up, immediately yield again_ pattern. That said, it may not be as bad as it sounds: `yield` gives all other coroutines a chance to run, and checking available memory is _very_ fast. The periodic memory check is therefore a pragmatic compromise. It keeps the logic simple, avoids coupling to a specific allocator, and still works when memory is released outside `BufferResource`’s control.

Longer term, a unified reservation system could enable a callback-driven approach and remove the need for periodic polling. Given the current constraints, however, the periodic check is the best trade-off.

cpp/src/streaming/core/memory_reserve_or_wait.cpp
```diff
@@ -0,0 +1,259 @@
...
+                }
+            }
+            periodic_memory_check_counter_.fetch_add(1, std::memory_order_acq_rel);
+            co_await ctx_->executor()->yield();
+            if (Clock::now() - last_reservation_success > timeout_) {
+                // This is the only way out of the while-loop that doesn't shutdown
+                // the periodic memory check.
+                break;
+            }
+            auto const max_size = memory_available();
```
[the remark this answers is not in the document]
wence-: That happens on line 204 where we found the eligible requests.

wence-:
I think this makes sense. With a query about shutdown.

cpp/src/streaming/core/memory_reserve_or_wait.cpp
```diff
@@ -0,0 +1,258 @@
...
+) {
+    // First, check whether the requested memory is immediately available.
+    auto [res, _] = ctx_->br()->reserve(mem_type_, size, no_overbooking);
+    if (res.size() > 0 || size == 0) {
+        co_return std::move(res);
+    }
+
+    // Use libcoro's queue to track completion of this reservation request.
+    // The queue will have at most one item: the fulfilled memory reservation.
+    coro::queue<MemoryReservation> request_queue{};
```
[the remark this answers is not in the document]
madsbk: My gut feeling is that this is an unnecessary complication, but we can consider it as a follow-up PR.

PR #689, opened 2025-11-25
bench-shuffle: discard output

madsbk:
Add a `-s` option to discard the the shuffle output


madsbk:
@quasiben can you try again with the `-s` option.

PR #696, opened 2025-12-01
Host buffer

madsbk:
Introduce `HostBuffer` that are backed by a `rmm::host_async_resource_ref` so it can use a host or pinned memory resource.


madsbk:
> Can we hold an `any_resource_ref` in the classes and then check against `device/host_accessible` property? I'm not sure if we might need the latest CCCL for that (we are using an old CCCL pinned from RMM)

I cannot get any checks on type erased references like `any_resource_ref` to match.

wence- (changes requested):

cpp/src/memory/host_buffer.cpp
```diff
@@ -0,0 +1,94 @@
...
+        );
+        stream_ = other.stream_;
+        mr_ = other.mr_;
+        span_ = std::exchange(other.span_, {});
+    }
+    return *this;
+}
+
+HostBuffer::~HostBuffer() noexcept {
+    deallocate_async();
```
nirandaperera: I think we need a `stream_.synchronize()` here.

cpp/include/rapidsmpf/memory/pinned_memory_resource.hpp
```diff
@@ -242,184 +243,4 @@ class PinnedMemoryResource {
...
-     *
-     * @tparam OtherBufferT The type of the other buffer to copy from. Eg.
-     * `rmm::device_buffer` or `PinnedHostBuffer`.
-     *
-     * @param other The other pinned host buffer to copy from.
-     * @param mr Memory resource to use for allocation and deallocation.
-     */
-    template <typename OtherBufferT>
-    PinnedHostBuffer(OtherBufferT const& other, std::shared_ptr<PinnedMemoryResource> mr)
-        : PinnedHostBuffer(other.data(), other.size(), other.stream(), std::move(mr)) {}
```
nirandaperera: @madsbk IMO we should port this constructor. It would be easy to create `HostBuffer` from any buffers here.

cpp/include/rapidsmpf/memory/host_memory_resource.hpp
```diff
@@ -0,0 +1,208 @@
...
+namespace rapidsmpf {
+
+/**
+ * @brief Host memory resource using standard CPU allocation.
+ *
+ * This resource allocates host memory using the ``new`` and ``delete`` operator. It is
+ * intended for use with `cuda::mr::resource` and related facilities, and advertises the
+ * `cuda::mr::host_accessible` property.
+ */
+class HostMemoryResource {
```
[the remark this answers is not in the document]
madsbk: Not necessarily. Since pinned host memory is accessible from the device, it can inherit from RMM’s `device_memory_resource`. But we may also experiment with other host-only memory resources, such as a non-pinned memory pool or mmap-backed buffers.

cpp/src/memory/host_buffer.cpp
madsbk: `HostBuffer` is stream ordered and assumes that it is backed by a stream-ordered memory resource. This is why `HostMemoryResource::deallocate()` does, in fact, synchronize the stream.

cpp/include/rapidsmpf/memory/host_memory_resource.hpp
```diff
@@ -0,0 +1,208 @@
...
+    ) {
+        return ::operator new(size, std::align_val_t{alignment});
+    }
+
+    /**
+     * @brief Deallocates memory using the given alignment.
+     *
+     * The default implementation synchronizes @p stream before deallocating the
+     * memory with the ``delete`` operator. This ensures that any in-flight CUDA
+     * operations using the memory complete before it is freed.
```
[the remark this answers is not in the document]
wence-: So this works as long as you only use the async interface, but if you rely on `deallocate_sync` then you get to keep both pieces.

nirandaperera:
@madsbk I've been thinking about this a bit. Now, if we are to use the same `HostBuffer` class for both pinned and host memory, its gonna be a bit messy to handle it in the `Buffer` class storage variant. 
I mean the main question would be, is pinned memory a separate memory type or its still the HOST memory type, but a separate allocation. MPO is it should be a separate memory type because it is stream-ordered from cuda 13+ (12.9+?). 

@wence- any thoughts?

madsbk:
But Buffer and HostBuffer is also stream ordered?

PR #702, opened 2025-12-03
`test_ray_ucxx_cluster_not_initialized` segfault occasionally

madsbk:
I have seen `test_ray_ucxx_cluster_not_initialized` segfault like here: https://github.com/rapidsai/rapidsmpf/actions/runs/19891515188/job/57039589326?pr=701


nirandaperera:
Yes, I saw this too yesterday.

pentschev:
That looks like a bug in Ray, @ayushdg any chance you have seen anything like that before?

@madsbk it seems Ray 2.52 is just 7 days old, seems roughly consistent with when this started to happen. Could you test 2.51.1 which is a month old to check if Valgrind reports the same? Alternatively, we could simply pin to 2.51.1 and observe if the error goes away.

nirandaperera:
@pentschev that's what I thought as well. We didnt see this instability in previous Ray releases. I only started seeing it this week, and we didint change anything from our end

pentschev:
> I haven't seen this before but it seems related to the ray update. To clarify does this happen only on `test_ray_ucxx_cluster_not_initialized` this test or any of the other ray tests?

I haven't seen other failures other than the one linked in the description, maybe @nirandaperera and @madsbk can clarify whether other tests have failed as well.

madsbk:
Seen again https://github.com/rapidsai/rapidsmpf/actions/runs/20085127429/job/57621224373?pr=727

ayushdg:
@madsbk Were you able to reproduce this locally?

PR #703, opened 2025-12-03
Add missing memory type parameters docs

pentschev:
Some parameters of MemoryType are not documented, causing failures with check-style.

### DIFF
--- a/cpp/include/rapidsmpf/memory/memory_type.hpp
+++ b/cpp/include/rapidsmpf/memory/memory_type.hpp
@@ -33,7 +33,13 @@ constexpr char const* to_string(MemoryType mem_type) {
     return MEMORY_TYPE_NAMES[static_cast<std::size_t>(mem_type)];
 }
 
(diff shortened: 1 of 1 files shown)


cpp/include/rapidsmpf/memory/memory_type.hpp
```diff
@@ -33,7 +33,13 @@ constexpr char const* to_string(MemoryType mem_type) {
...
 }
 
-/// @brief Overload to write type name to the output stream.
+/**
+ * @brief Overload to write type name to the output stream.
+ *
+ * @param os The output stream.
+ * @param mem_type The memory type to write name of to the output stream.
+ * @return The output stream.
+ */
```
pentschev: Here's a sample of the failure: https://github.com/rapidsai/rapidsmpf/actions/runs/19901829096/job/57047575935?pr=683#step:7:109

pentschev:
> Why didn't CI fail in the PR?

I have absolutely no idea.

PR #616, opened 2025-10-31
Add RapidsMPF launcher `rrun`

pentschev:
Introduce a new launcher for RapidsMPF without dependencies on `mpirun`. The new launcher currently named `rrun` (as in RapidsMPF run) introduces basic functionality to launch processes locally to run on multiple-GPUs while providing an extension to allow it to support different backends, the only existing backend supports file-based synchronization, in the future we plan to extend it to support other synchronization mechanisms including over-the-network via specialized service, Slurm, as well as Kubernetes and others that may be requested. Additionally, it will support automatic discovery and configuration of system topology per process (i.e., per GPU).

This initial implementation only supports local node multi-GPU, the implementation for multi-node multi-GPU already exists using SSH to spawn multiple processes but was split into a different PR to make this shorter.

PR #704, opened 2025-12-03
Support `rrun` binding to resources

pentschev:
Add new `--bind-to` argument to `rrun` that allows automatically binding to resources, with a default of `--bind-to all` (binds all supported resources, currently: CPU, NUMA/memory binding, UCX network devices). Includes a new tool `check_resource_binding` that can be used standalone or together with `rrun` to validate the actual resources that are bound to for each process individually.

### DIFF
--- a/cpp/tools/check_resource_binding.cpp
+++ b/cpp/tools/check_resource_binding.cpp
@@ -197,6 +197,11 @@ bool extract_gpu_info_from_json(
 
 }  // namespace
 
+/**
+ * @brief Print or validate binding configuration.
+ *
+ * See top of this file for more details.
+ */
 int main(int argc, char* argv[]) {
     std::string json_file;
     int gpu_id = -1;


bdice (approved):

cpp/src/bootstrap/utils.cpp
```diff
@@ -0,0 +1,231 @@
...
+#include <unistd.h>
+
+#include <rapidsmpf/bootstrap/utils.hpp>
+
+#if RAPIDSMPF_HAVE_NUMA
+#include <numa.h>
+#endif
+
+// NOTE: Do not use RAPIDSMPF_EXPECTS or RAPIDSMPF_FAIL in this file.
+// Using these macros introduces a CUDA dependency via rapidsmpf/error.hpp.
```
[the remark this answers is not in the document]
pentschev: Yes, @madsbk and I discussed this previously, see https://github.com/rapidsai/rapidsmpf/pull/616#issuecomment-3485708797, and decided against it (for now at least).

cpp/include/rapidsmpf/bootstrap/bootstrap.hpp
```diff
@@ -11,17 +11,10 @@
 #include <optional>
 #include <string>
 
-#include <rapidsmpf/config.hpp>
-#include <rapidsmpf/utils.hpp>
+#include <rapidsmpf/bootstrap/types.hpp>
```
madsbk: Include `#include <rapidsmpf/utils.hpp>` instead?

cpp/include/rapidsmpf/bootstrap/utils.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * @brief Get current NUMA node(s) for memory binding.
+ *
+ * Queries the NUMA node associated with the CPU the current process is running on.
+ * This is a best-effort approach and may not be accurate in all cases.
+ *
+ * @return Vector of NUMA node IDs. Empty if NUMA is not available or detection fails.
+ */
+std::vector<int> get_current_numa_nodes();
```
madsbk: I think it would be good to move them into `<rapidsmpf/utils.hpp>`, or possibly into a new header such as `system_info.hpp`. They will also be useful for `PinnedMemoryResource`, for example.

cpp/tools/check_resource_binding.cpp
```diff
@@ -0,0 +1,420 @@
+/**
```
madsbk: All of this string parsing will be very hard to maintain, writing it in Python might help a bit?

madsbk (approved):
Looks good, but I’m concerned about the string-parsing complexity, it will be very hard to maintain across different machines.

cpp/tools/check_resource_binding.cpp
```diff
@@ -0,0 +1,420 @@
...
+            }
+        }
+    }
+
+    return true;
+}
+
+}  // namespace
+
+int main(int argc, char* argv[]) {
```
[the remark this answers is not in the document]
pentschev: Writing it in Python would be ideal in an ideal world. For RapidsMPF that would work, but I think we may have other users that are not running Python, then that will not be useful for them, unfortunately. In the future we'll need to consider using a JSON library and remove all this complexity from here, but I intentionally tried to avoid introducing a new dependency for now because the support is still relatively limited and simple.

pentschev (commented):
Thanks Mads! Responded to and/or addressed your comments.

PR #710, opened 2025-12-04
Streaming Q4 implementation

TomAugspurger:
This implements TPCH query 4 using rapidsmpf.

The primary notable thing about q4 is the left semi join / `where exists` between the two tables:

```python
q = (
    # SQL exists translates to semi join in Polars API
    orders.join(
        (lineitem.filter(pl.col("l_commitdate") < pl.col("l_receiptdate"))),
        left_on="o_orderkey",
        right_on="l_orderkey",
        how="semi",
    )
    ...
)
```

I want to think a bit more about how to do this. Right now, I've implemented a version that broadcasts the smaller `orders` table and shuffles the larger `lineitem` table. We're currently unable to reuse the filtered joiner across chunks. cudf only supports probing with the `left` table (the smaller one in our case, `orders`). The build table must be the `right` table (the larger one in our case, `lineitem`). So I think we *have* to shuffle `lineitem` otherwise we'll get incorrect results (a key in `orders` matching against a key in `lineitem` multiple times, once per partition).

### DIFF
--- a/cpp/benchmarks/streaming/ndsh/q04.cpp
+++ b/cpp/benchmarks/streaming/ndsh/q04.cpp
@@ -32,15 +32,16 @@
 
 #include <rapidsmpf/communicator/communicator.hpp>
 #include <rapidsmpf/communicator/mpi.hpp>
+#include <rapidsmpf/integrations/cudf/bloom_filter.hpp>
 #include <rapidsmpf/nvtx.hpp>
 #include <rapidsmpf/streaming/coll/allgather.hpp>
 #include <rapidsmpf/streaming/core/channel.hpp>
 #include <rapidsmpf/streaming/core/context.hpp>
 #include <rapidsmpf/streaming/core/node.hpp>
+#include <rapidsmpf/streaming/cudf/bloom_filter.hpp>
 #include <rapidsmpf/streaming/cudf/parquet.hpp>
 #include <rapidsmpf/streaming/cudf/table_chunk.hpp>
 
-#include "bloom_filter.hpp"
 #include "concatenate.hpp"
 #include "groupby.hpp"
 #include "join.hpp"
@@ -305,9 +306,8 @@ int main(int argc, char** argv) {
     int device;
     RAPIDSMPF_CUDA_TRY(cudaGetDevice(&device));
     RAPIDSMPF_CUDA_TRY(cudaDeviceGetAttribute(&l2size, cudaDevAttrL2CacheSize, device));
-    auto const num_filter_blocks = rapidsmpf::ndsh::BloomFilter::fitting_num_blocks(
-        static_cast<std::size_t>(l2size)
-    );
+    auto const num_filter_blocks =
+        rapidsmpf::BloomFilter::fitting_num_blocks(static_cast<std::size_t>(l2size));
 
     for (int i = 0; i < arguments.num_iterations; i++) {
         rapidsmpf::OpID op_id{0};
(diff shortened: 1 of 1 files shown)


cpp/benchmarks/streaming/ndsh/join.cpp
```diff
@@ -344,6 +412,107 @@ streaming::Node inner_join_shuffle(
...
+        }
+        co_await ch_out->send(semi_join_chunk(
+            ctx,
+            left_table,
+            right_msg.release<streaming::TableChunk>(),
+            left_on,
+            right_on,
+            sequence++
+        ));
+    }
```
wence-: Given that you're broadcasting the left table, I would do:

```
left = broadcast(...);
CudaEvent event;
event.record(left.stream());
semi_join_chunk(...., &event, left.stream());
```

and then `semi_join_chunk` can do:

```
chunk = co_await ch_right->receive();
chunk = to_device(...);
// ensure left table is ready
event.stream_wait(chunk.stream());
// do all join on chunk.stream()
...
// ensure deallocation of left happens after join
cuda_stream_join(left.stream(), chunk.stream(), ...)
```
WDYT?

cpp/benchmarks/streaming/ndsh/q04.cpp
```diff
@@ -0,0 +1,612 @@
...
+
+            nodes.push_back(read_lineitem(
+                ctx, lineitem, 4, arguments.num_rows_per_chunk, arguments.input_directory
+            ));
+            nodes.push_back(
+                filter_lineitem(ctx, lineitem, filtered_lineitem)
+            );  // l_orderkey
+            nodes.push_back(read_orders(
+                ctx, order, 4, arguments.num_rows_per_chunk, arguments.input_directory
+            ));
```
wence-: The filter of the orders table is very selective. So I would recommend attempting to use a bloom filter to pre-filter `lineitem` before shuffling.

q03 has an example

cpp/benchmarks/streaming/ndsh/join.cpp
```diff
@@ -141,6 +142,76 @@ streaming::Node broadcast(
...
+                              ctx->br()->device_mr()
+    )
+                              ->release();
+
+    auto result_table = std::make_unique<cudf::table>(std::move(result_columns));
+    ctx->comm()->logger().debug(
+        "semi_join_chunk: result_table.num_rows()=", result_table->num_rows()
+    );
+    // Deallocation of the join indices will happen on chunk_stream, so add stream dep
+    cuda_stream_join(left_chunk.stream(), chunk_stream, &event);
```
wence-: Given that we don't reuse this `event` more than once, we could just use `cuda_stream_join` without the event argument and let it create one.

cpp/benchmarks/streaming/ndsh/join.cpp
```diff
@@ -342,6 +413,116 @@ streaming::Node inner_join_shuffle(
...
+            right_on,
+            sequence++,
+            &left_event
+        ));
+    }
+
+    co_await ch_out->drain(ctx->executor());
+}
+
+streaming::Node left_semi_join_shuffle(
```
wence-: Please add a docstring.

cpp/benchmarks/streaming/ndsh/join.cpp
```diff
@@ -342,6 +413,116 @@ streaming::Node inner_join_shuffle(
...
+    std::shared_ptr<streaming::Channel> ch_out,
+    std::vector<cudf::size_type> left_on,
+    std::vector<cudf::size_type> right_on
+) {
+    streaming::ShutdownAtExit c{left, right, ch_out};
+    ctx->comm()->logger().print("Shuffle left semi join");
+
+    co_await ctx->executor()->schedule();
+
+    while (true) {
```
wence-:
```suggestion
    while (!ch_out->is_shutdown()) {
```

But that might not work because I'm not sure we have a way to discard from a partially extracted shuffle.

cpp/benchmarks/streaming/ndsh/join.cpp
```diff
@@ -342,6 +413,116 @@ streaming::Node inner_join_shuffle(
...
+        if (right_msg.empty()) {
+            break;
+        }
+        co_await ch_out->send(semi_join_chunk(
+            ctx,
+            left_table,
+            right_msg.release<streaming::TableChunk>(),
+            left_on,
+            right_on,
+            sequence++,
```
wence-: I think this should be `right_msg.sequence_number()`.

cpp/benchmarks/streaming/ndsh/join.hpp
```diff
@@ -120,6 +120,65 @@ enum class KeepKeys : bool {
...
 
+/**
+ * @brief Perform a streaming left semi join between two tables.
+ *
+ * @note This performs a broadcast join, broadcasting the table represented by the `left`
+ * channel to all ranks, and then streaming through the chunks of the `right` channel.
+ * The `right` channel is required to provide hash-partitioned data in-order.
+ *
+ * @param ctx Streaming context.
+ * @param left Channel of `TableChunk`s in hash-partitioned order.
```
wence-: `left` need not be hash-partitioned I think, because we are broadcasting it
?

PR #711, opened 2025-12-04
Make unbounded fanout messages spillable 

### LINKED ISSUE #675: Make unbounded fanout state spillable
#636 Adds fanout node. It internally caches `Message`s in a deque. These messages needs to be registered with the spillable messages. 
cc @madsbk @rjzamora 

Ref: https://nvidia.slack.com/archives/C08469RJMRB/p1763413811933989

nirandaperera:
Closes #675

### DIFF
--- a/cpp/tests/streaming/test_fanout.cpp
+++ b/cpp/tests/streaming/test_fanout.cpp
@@ -11,6 +11,7 @@
 #include <cudf_test/table_utilities.hpp>
 
 #include <rapidsmpf/memory/buffer.hpp>
+#include <rapidsmpf/memory/pinned_memory_resource.hpp>
 #include <rapidsmpf/streaming/core/coro_utils.hpp>
 #include <rapidsmpf/streaming/core/fanout.hpp>
 #include <rapidsmpf/streaming/core/leaf_node.hpp>
@@ -545,9 +546,7 @@ class SpillingStreamingFanout : public BaseStreamingFixture {
                 {MemoryType::DEVICE, []() -> std::int64_t { return 0; }},
             };
         br = std::make_shared<rapidsmpf::BufferResource>(
-            mr_cuda,
-            rapidsmpf::BufferResource::PinnedMemoryResourceDisabled,
-            memory_available
+            mr_cuda, rapidsmpf::PinnedMemoryResource::Disabled, memory_available
         );
         auto options = ctx->options();
         ctx = std::make_shared<rapidsmpf::streaming::Context>(


cpp/src/streaming/core/spillable_messages.cpp
```diff
@@ -26,6 +25,26 @@ Message SpillableMessages::extract(MessageId mid) {
...
+
+    std::unique_lock item_lock(item->mutex);
+    RAPIDSMPF_EXPECTS(
+        item->message.has_value(),
+        "empty message " + std::to_string(mid),
+        std::out_of_range
+    );
+    auto const& msg = *item->message;
+    auto res = br->reserve_or_fail(msg.copy_cost(), msg.spillable_memory_types());
+    return msg.copy(res);
```
nirandaperera: I'm not 100% sure if this is good because every time we make a copy, we would have to lock-unlock two mutexes. 
Can we keep a copy in the fanout and make copies there, rather than copying messages from spillable messages? @madsbk WDYT?

cpp/include/rapidsmpf/memory/memory_type.hpp
```diff
@@ -33,6 +35,27 @@ constexpr std::array<char const*, MEMORY_TYPES.size()> MEMORY_TYPE_NAMES{
  */
 constexpr std::array<MemoryType, 1> SPILL_TARGET_MEMORY_TYPES{{MemoryType::HOST}};
 
+/**
+ * @brief Get the lower memory types than or equal to the @p mem_type .
+ *
+ * @param mem_type The memory type.
+ * @return A span of the lower memory types than the given memory type.
+ */
```
madsbk:
```suggestion
/**
 * @brief Get the memory types with preference lower than or equal to @p mem_type.
 *
 * The returned span reflects the predefined ordering used in @ref MEMORY_TYPES,
 * which lists memory types in decreasing order of preference.
 *
 * @param mem_type The memory type used as the starting point.
 * @return A span of memory types whose preference is lower than or equal to
 * the given type.
 */
```

cpp/include/rapidsmpf/streaming/core/channel.hpp
```diff
@@ -58,11 +58,22 @@ class Channel {
...
+     *
+     * Suspends if the channel is empty. Once the message id is received, the message can
+     * be extracted using the `SpillableMessages::extract` method. This could be useful
+     * when a node wants to consume a message from a channel, but leave it in the
+     * spillable messages container for later extraction.
+     *
+     * @return A coroutine that evaluates to the message id. If the channel is shut down,
+     * the message id will be `SpillableMessages::InvalidMessageId`.
+     */
+    coro::task<SpillableMessages::MessageId> receive_message_id();
```
madsbk: I think the use of `MessageId` inside a `Channel` should remain an implementation detail. We should avoid exposing it.

madsbk (changes requested):

cpp/include/rapidsmpf/streaming/core/channel.hpp
nirandaperera: I still think this is useful to expose. Because channels no longer own the `Message`s. So, intuitively, we should be able to get the message or the placeholder of the message. Later would be useful when `Node`s decide to leave the messages in `SpillableMessages` and use it later on.

madsbk: The thing is, Channels **do** own their messages. They may store them in a `SpillableMessages` as an implementation detail, but the Channel still owns the messages until a `Node` receives them.

If we give up this isolation, a Channel loses the freedom to change its internal strategy for storing messages. For example, it might later decide to use multiple `SpillableMessages` instances if contention becomes a problem.

I am not saying we should never go in that direction, but that discussion should be a separate issue with its own clear motivation.

cpp/src/streaming/core/channel.cpp
```diff
@@ -23,6 +23,11 @@ coro::task<Message> Channel::receive() {
     }
 }
 
+coro::task<SpillableMessages::MessageId> Channel::receive_message_id() {
+    auto msg_id = co_await rb_.consume();
+    co_return msg_id.has_value() ? *msg_id : SpillableMessages::InvalidMessageId;
+}
```
wence-: Like Mads, I think we don't want to expose this detail. But even if we did, carving out a special value for "no you didn't get a message" is not the right way. Use a `std::optional`.

cpp/src/streaming/core/fanout.cpp
```diff
@@ -207,8 +190,11 @@ struct UnboundedFanout {
         };
         co_await ctx.executor()->schedule();
 
+        auto spillable_messages = ctx.spillable_messages();
```
madsbk:
```suggestion
        // To make staged input messages spillable, insert them into the Context's
        // spillable_messages container while they are in transit.
        auto spillable_messages = ctx.spillable_messages();
```

cpp/src/streaming/core/fanout.cpp
```diff
@@ -230,12 +216,11 @@ struct UnboundedFanout {
             }
 
             for (auto const& msg : messages_to_send) {
```
madsbk:
```suggestion
            for (auto const msg_id : messages_to_send) {
```

PR #713, opened 2025-12-05
`SpillableMessages`: support message copy

madsbk:
```c++
    /**
     * @brief Create a deep copy of a message without removing it.
     *
     * This method duplicates the message identified by `mid` while leaving the
     * original message intact inside the container. The returned message is a
     * full deep copy of the payload. If the message is currently being spilled
     * by another thread, this call waits until spilling completes.
     *
     * @param mid Message identifier.
     * @param reservation Memory reservation used for allocating buffers during
     * the deep copy. The reservation also determines the memory type of the
     * returned message.
     *
     * @return A deep copy of the referenced `Message`.
     *
     * @throws std::out_of_range If the message has already been extracted.
     * @throws std::runtime_error If required memory cannot be allocated using
     * the provided reservation.
     */
    [[nodiscard]] Message copy(MessageId mid, MemoryReservation& reservation);
```

Based on https://github.com/rapidsai/rapidsmpf/pull/711

### DIFF
--- a/cpp/include/rapidsmpf/streaming/core/spillable_messages.hpp
+++ b/cpp/include/rapidsmpf/streaming/core/spillable_messages.hpp
@@ -82,7 +82,8 @@ class SpillableMessages {
      *
      * @return A deep copy of the referenced `Message`.
      *
-     * @throws std::out_of_range If the message has already been extracted.
+     * @throws std::out_of_range If the message has already been extracted or
+     * the message identifier is invalid.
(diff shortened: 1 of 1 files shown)


pentschev (approved):
Left a minor nit but LGTM.

nirandaperera:
> Just use `get_content_descriptions()` ?

That will create a copy of _all_ the content descriptions to access just one. That wouldnt be necessary either IMO. Why cant we take the BR as in the #711 PR?

PR #716, opened 2025-12-05
[MINOR] Change rounding method to round_decimal in q09.cpp

nirandaperera:
`cudf::round` is a deprecated method and raises a compilation warning

### DIFF
--- a/cpp/benchmarks/streaming/ndsh/q09.cpp
+++ b/cpp/benchmarks/streaming/ndsh/q09.cpp
@@ -467,7 +467,7 @@ rapidsmpf::streaming::Node sort_by(
     auto chunk =
         rapidsmpf::ndsh::to_device(ctx, msg.release<rapidsmpf::streaming::TableChunk>());
     auto table = chunk.table_view();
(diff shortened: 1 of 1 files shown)


cpp/benchmarks/streaming/ndsh/q09.cpp
```diff
@@ -467,7 +467,7 @@ rapidsmpf::streaming::Node sort_by(
     auto chunk =
         rapidsmpf::ndsh::to_device(ctx, msg.release<rapidsmpf::streaming::TableChunk>());
     auto table = chunk.table_view();
-    auto rounded = cudf::round(
+    auto rounded = cudf::round_decimal(
```
wence-: you can't just do this because it requires the input column to be decimal already which it is not

nirandaperera: Ah, yeah, looks like I was mistaken. 
https://docs.rapids.ai/api/libcudf/legacy/group__transformation__unaryops#ga0c3cc731433773ab9e36aca8d0d25912

PR #717, opened 2025-12-05
Include spill amounts in NVTX payloads

TomAugspurger:
This updates our spilling to include

1. the amount requested for spilling in postbox_spilling(), as a payload in the existing NVTX range.
2. the amount actually spilled in postbox_spilling() as a marker at the end of postbox_spilling().

This will slightly increase the size of output, by including a payload in the existing `postbox_spilling` range and adding a new NVTX marker for the amount actually spilled. I think this should be negligible. Looking at a recent cudf-polars run of query 4 at SF-1K, we made ~100-150 calls to postbox_spilling.

Finally, I've included a small python / SQL script for generating a spill report from the nsys report file:

```
❯ python -m rapidsmpf._spill_report slow-spill.nsys-rep --force-overwrite
Postbox Spilling Summary
================================================================================

         Metric | Count |       Avg |       Std |       Min |        Max |     Total
----------------+-------+-----------+-----------+-----------+------------+----------
       Duration |   391 |   3.97 ms |   9.81 ms | 912.00 ns |   63.36 ms |    1.55 s
Bytes requested |   391 | 41.02 MiB | 23.20 MiB | 19.31 KiB | 107.79 MiB | 15.66 GiB
  Bytes spilled |   391 |  2.75 MiB |  7.32 MiB |       0 B |  37.93 MiB |  1.05 GiB

Spill satisfaction  : 6.7%
Effective throughput: 691.65 MiB/s
```

I think having something like this would be useful. I don't think that having it be a submodule inside rapidsmpf makes sense, but perhaps as a console script installed with rapidsmpf?

### DIFF
--- a/cpp/include/rapidsmpf/nvtx.hpp
+++ b/cpp/include/rapidsmpf/nvtx.hpp
@@ -55,6 +55,10 @@ struct rapidsmpf_domain {
 
 // implement the func range macro with a value
 #define RAPIDSMPF_NVTX_FUNC_RANGE_IMPL_WITH_VAL(val)              \
+    static_assert(                                                \
+        std::is_arithmetic_v<decltype(val)>,                      \
+        "Value must be integral or floating point type"           \
+    );                                                            \
     nvtx3::scoped_range_in<rapidsmpf_domain> RAPIDSMPF_CONCAT(    \
         _rapidsmpf_nvtx_range, __LINE__                           \
     ) {                                                           \


python/rapidsmpf/rapidsmpf/_spill_report.py
```diff
@@ -0,0 +1,284 @@
+#!/usr/bin/env python3
```
madsbk: What about creating a new submodule for tools like `rapidsmpf.tools`?

python/rapidsmpf/rapidsmpf/_spill_report.py
wence-: FWIW (nit), I don't like this `_impl` for "private" files. I much prefer to define public things via `__all__`. Someone is going to import this anyway :)

cpp/include/rapidsmpf/nvtx.hpp
```diff
@@ -53,25 +53,53 @@ struct rapidsmpf_domain {
...
     }(msg)
 
-// Macro overloads of RAPIDSMPF_NVTX_FUNC_RANGE
-#define RAPIDSMPF_NVTX_FUNC_RANGE_IMPL() NVTX3_FUNC_RANGE_IN(rapidsmpf_domain)
+// implement the func range macro with a value
+#define RAPIDSMPF_NVTX_FUNC_RANGE_IMPL_WITH_VAL(val)              \
+    nvtx3::scoped_range_in<rapidsmpf_domain> RAPIDSMPF_CONCAT(    \
+        _rapidsmpf_nvtx_range, __LINE__                           \
+    ) {                                                           \
+        nvtx3::event_attributes {                                 \
```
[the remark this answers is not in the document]
TomAugspurger: Thanks. I had to change the `((val))` to just `(val)` and that seems to work.

PR #725, opened 2025-12-09
Use static linkage for CUDA runtime

bdice:
## Summary
- Set `CUDA_STATIC_RUNTIME=ON` by default in `cpp/CMakeLists.txt`
- Remove `cuda-cudart` from run requirements in conda recipes

With static linking of the CUDA runtime, the runtime is embedded in the binaries and the `cuda-cudart` package is not needed at runtime.

Part of https://github.com/rapidsai/build-planning/issues/235


KyleFromNVIDIA (approved):

robertmaynard (changes requested):
We can drop all the compute sanitizer suppressions  in `cpp/compute-sanitizer-suppressions.xml`

wence-:
> We can drop all the compute sanitizer suppressions in `cpp/compute-sanitizer-suppressions.xml`

What? why?

pentschev:
> > > We can drop all the compute sanitizer suppressions in `cpp/compute-sanitizer-suppressions.xml`
> > 
> > 
> > What? why?
> 
> Sorry, to be clear the suppressions have `libcudart.so` in the stack will need to be reviewed. Since we move to `libcudart_static.a` the stack traces will be different and I expect will need to be regenerated.

But symbols will still exist somewhere, so that means we can remove the module name but still want to keep the suppressions, and this will actually make it more difficult now since before we could do "suppress all CUDA symbols" and now we need to list them by names.

pentschev:
> > But symbols will still exist somewhere
> Yes. My first comment wasn't correct. We can't remove the suppressions, we just have to make the detection significantly more loose when looking at the callstack

I understood the correction you've made to the first comment. However, I think what you're saying is we have to make detection significantly _tighter_ now, because we can't anymore just ignore `libcudart.so` for example, we have to instead explicitly list all the function names that come from `libcudart.so`, we could use `cuda*` wildcard for example but that would also suppress any functions coming from any other libraries that begin with `cuda*`.

dependencies.yaml
```diff
@@ -257,7 +257,6 @@ dependencies:
     common:
       - output_types: [conda]
         packages:
-          - cuda-cudart-dev
```
KyleFromNVIDIA, line 362: Is the `cuda` dependency group supposed to be for build time or runtime? Or both? We still need `cuda-cudart-dev` for build time.

bdice: Generally we've been using that for build time dependencies, I think. I've been using this to mean "all the conda packages needed for building from source in a provided conda environment" because the runtime conda packages are all defined by the recipes.

KyleFromNVIDIA (approved):

KyleFromNVIDIA:
Even when I comment out the suppressions file entirely, CI still passes. @pentschev Do you know if the suppressions are even still relevant?

pentschev (approved):

robertmaynard (approved):

pentschev:
> Even when I comment out the suppressions file entirely, CI still passes. @pentschev Do you know if the suppressions are even still relevant?

For completeness: suppressions are still relevant but I don't think they're covered in CI. However, that's far from a priority right now so let's just move and ignore them for the moment.

PR #728, opened 2025-12-10
Enable clang-tidy checks in `build.sh`

pentschev:
At the moment clang-tidy checks only get verified in the `check-style` CI job, normally requiring to push changes and let CI run to then find out whether it breaks clang-tidy. While not a perfect solution, enabling clang-tidy in `build.sh` allows catching failures early during development and before waiting for CI to run, at minimal/no extra build cost. The ideal solution would be to enable it in `pre-commit`, but since it requires building at least parts of the library that may be too expensive for `pre-commit`.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -316,7 +316,7 @@ dependencies:
           - matrix:
               arch: x86_64
             packages:
-              - ray-default>=2.49
+              - ray-default>=2.49,<2.52
           - matrix:
               arch: aarch64
             packages:
@@ -436,7 +436,7 @@ dependencies:
           - matrix:
               arch: x86_64
             packages:
-              - ray-default>=2.49
+              - ray-default>=2.49,<2.52
           - matrix:
               arch: aarch64
             packages:
--- a/conda/environments/all_cuda-129_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-129_arch-x86_64.yaml
@@ -47,7 +47,7 @@ dependencies:
 - pytest<9.0.0
 - python>=3.10,<3.14
 - rapids-build-backend>=0.4.0,<0.5.0
-- ray-default>=2.49
+- ray-default>=2.49,<2.52
 - rmm==26.2.*,>=0.0.0a0
 - scikit-build-core>=0.10.0
 - sphinx-autobuild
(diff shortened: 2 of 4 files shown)


madsbk:
I mean, provide a convenient command line option for people who are tired of clang-tidy errors in CI, but do not make it mandatory.

madsbk:
> I don't think this is a valid reason, you're most definitely creating an environment using a conda/environments file which will bring tidy automatically.

You have more trust in the build system and in the setups we will need to reproduce issues than I do :)

> So in the interest of meeting in the middle, how about providing an option to disable clang-tidy instead for the cases one may be worried about longer build times?

Alright, that sounds reasonable.

conda/recipes/librapidsmpf/recipe.yaml
```diff
@@ -65,10 +65,12 @@ cache:
       - ${{ compiler("c") }}
       - ${{ compiler("cxx") }}
       - ${{ compiler("cuda") }} =${{ cuda_version }}
+      - clang==20.1.4
```
bdice: We should use a separate CI job for clang-tidy checks, see cudf for an example. Our primary conda C++ build pipeline shouldn’t do this work, because it will make dependencies hard to manage in certain situations we’ve encountered in the past.

pentschev: Does https://github.com/rapidsai/rapidsmpf/pull/728/commits/d16e44a27ed474c99d33b62ea0d413b55d998dee satisfy your suggestion?

bdice (approved):

dependencies.yaml
```diff
@@ -165,11 +165,19 @@ dependencies:
...
+        packages:
+          - &clang_ver clang==20.1.4
+          - &clang_tools_ver clang-tools==20.1.4
   build-cpp:
     common:
       - output_types: conda
         packages:
           - c-compiler
+          - *clang_ver
+          - *clang_tools_ver
```
bdice: I thought some more after approving. Are these needed? Why? The `clang_tidy` list is already in the "all" environment so these should be included in developer environments. I don't see a reason to duplicate them here.

bdice (changes requested):

pentschev:
@madsbk could you or someone else approve on behalf of @rapidsai/rapidsmpf-cpp-codeowners as well? We'll end up required that because of some minor clang-tidy fixes in the CUPTI code, see https://github.com/rapidsai/rapidsmpf/pull/728/commits/003f36df9fb5aa3ab9965ccb3a3c457a8500935f.

madsbk (approved):

bdice (approved):

PR #729, opened 2025-12-10
Formalize Buffer Memory Types and Storage Mapping

madsbk:
This PR defines the memory type of a `Buffer` and clarifies how each `MemoryType` maps to its underlying storage implementation (`rmm::device_buffer` or `HostBuffer`).

This work prepares for the introducing of a new memory type, `MemoryType::PINNED_HOST`, which will be added in a follow-up PR.

### DIFF
--- a/cpp/include/rapidsmpf/memory/buffer.hpp
+++ b/cpp/include/rapidsmpf/memory/buffer.hpp
@@ -26,24 +26,20 @@ namespace rapidsmpf {
 /**
  * @brief Buffer representing device or host memory.
  *
- * A `Buffer` holds either device memory or host memory. The memory type is fixed at
- * construction and determines which internal storage type is used. Buffers are
- * stream ordered and have an associated CUDA stream, see `stream()`. All work that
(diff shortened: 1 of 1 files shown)


cpp/include/rapidsmpf/memory/buffer.hpp
```diff
@@ -276,11 +279,13 @@ class Buffer {
      * stream (or established ordering with it) for correctness.
      *
      * @param device_buffer Unique pointer to a device buffer. Must be non-null.
+     * @param mem_type The memory type underlying @p device_buffer.
      *
      * @throws std::invalid_argument If @p device_buffer is null.
+     * @throws std::invalid_argument If @p mem_type it's suitable for device buffers.
```
pentschev:
```suggestion
     * @throws std::invalid_argument If @p mem_type is not suitable for device buffers.
```

wence- (approved):

pentschev (approved):
Thanks Mads.

Issue #615, opened 2025-10-30
Feature improvements for multi-rank read-parquet


wence-:
After #574 we have a multi-threaded multi-rank read parquet node.

It has a few missing features (especially when running in multiple ranks)

1. skip_rows/num_rows specification not supported when running with more than one rank
2. If the number of files is fewer than the number of ranks, some ranks get no portions of the table. If that file is large, this is not great
3. Files are split up by row count, not byte count (perhaps aligned to row group boundaries)
4. Parquet metadata for all the files is read by every chunk read. We should switch the the hybrid scan reader to avoid this.

PR #736, opened 2025-12-11
Support large-file splitting between ranks in ``read_parquet``

rjzamora:
Addresses item 2 in https://github.com/rapidsai/rapidsmpf/issues/615

**General Approach**: When the number of files is smaller than the number of ranks, we estimate the number of chunks we need to produce for each file, and use this estimate to determine the (row-group-aligned) boundary between ranks.

This PR does not attempt to optimize metadata collection or to align chunks with row-group boundaries. I'd like that to separate that work for now.

### DIFF
--- a/cpp/src/streaming/cudf/parquet.cpp
+++ b/cpp/src/streaming/cudf/parquet.cpp
@@ -164,6 +164,219 @@ std::pair<std::int64_t, std::int64_t> compute_split_range(
 }
 
 /**
+ * @brief Result of chunk assignment containing work distribution for producers.
+ */
+struct ChunkAssignment {
+    std::vector<std::vector<ChunkDesc>> chunks_per_producer;
+    bool rank_has_assigned_work;
+};
+
+/**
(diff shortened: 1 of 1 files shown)


cpp/src/streaming/cudf/parquet.cpp
```diff
@@ -153,70 +259,178 @@ Node read_parquet(
...
-    {
-        std::vector<std::string> chunk_files;
-        auto const nchunk_files = std::min(num_files - file_offset, files_per_chunk);
-        std::ranges::copy_n(
-            local_files.begin() + static_cast<std::int64_t>(file_offset),
-            static_cast<std::int64_t>(nchunk_files),
-            std::back_inserter(chunk_files)
+    bool rank_has_assigned_work = false;  // Track if this rank was assigned work
+
+    if (files.size() >= size) {
```
madsbk: I think it would be very helpful to move the two cases into separate helper functions, with docstrings describing the two different approaches.

PR #737, opened 2025-12-11
Fix NDSH q09 to use round_decimal

bdice:
Update the q09 benchmark to use `round_decimal` instead of the deprecated `round` function.


bdice:
We need some solution to not use deprecated APIs. libcudf is removing this because rounding binary floats to decimal places is not well-defined. I think requiring a cast is the best outcome but you're welcome to implement some other form of "rounding" by calling some libcudf math operators.

PR #738, opened 2025-12-11
Implement Q1 and Q3 and refactor to provide utilities

wence-:
This introduces a bunch of useful (mostly chunkwise) utilities in the ndsh benchmarks as well as implementations of Q1 and Q3.

New features:
- chunkwise groupby
- chunkwise sortby
- bloom filter construction
- write parquet

### DIFF
--- a/python/librapidsmpf/pyproject.toml
+++ b/python/librapidsmpf/pyproject.toml
@@ -71,3 +71,10 @@ regex = "(?P<value>.*)"
 
 [project.entry-points."cmake.prefix"]
 librapidsmpf = "librapidsmpf"
+
+[tool.pydistcheck]
+select = [
+    "distro-too-large-compressed",
+]
+output_file_size_precision = 4
+max_allowed_size_compressed = '75M'


cpp/src/streaming/cudf/parquet.cpp
```diff
@@ -90,9 +91,21 @@ Node produce_chunks(
         co_await ctx->executor()->schedule();
         // TODO: This reads the metadata ntasks times.
         // See https://github.com/rapidsai/cudf/issues/20311
-        auto sent = co_await ticket->send(
-            read_parquet_chunk(ctx, stream, chunk_options, chunk.sequence_number)
-        );
+        auto [msg, exception] = [&]() -> std::pair<Message, std::exception_ptr> {
```
TomAugspurger: Would you recommend using this pattern in all of our utilities calling cudf operations?

wence-:
> q1 and q3 we need to change the timestamp type in the read from timestamp_D to timestamp_ms. q9 not sure what I broke, will check

Did that.

For q9 at SF3k, you need `--use-shuffle-join` (or else turn on spilling with `--spill-device-limit`, but I think that will be much slower).

cpp/src/streaming/cudf/parquet.cpp
wence-: No, this is because we can't use the `ShutdownAtExit` utility because `produce_chunks` is used in a multi-producer setup. Usually we use `ShutdownAtExit` to shutdown channels if an exception is not caught. But it _also_ shuts down the channel when the coroutine finishes.

But if we have two `produce_chunks` into the same output channel, we don't want to shut down the output channel if one `produce_chunks` is finished, only if an exception occurred.

cpp/benchmarks/streaming/ndsh/q03.cpp
```diff
@@ -0,0 +1,727 @@
...
+    cudf::size_type num_rows_per_chunk,
+    std::string const& input_directory
+) {
+    auto files = rapidsmpf::ndsh::detail::list_parquet_files(
+        rapidsmpf::ndsh::detail::get_table_path(input_directory, "customer")
+    );
+    auto options = cudf::io::parquet_reader_options::builder(cudf::io::source_info(files))
+                       .columns({"c_custkey"})  // 0
+                       .build();
+    auto filter_expr = [&]() -> std::unique_ptr<rapidsmpf::streaming::Filter> {
```
TomAugspurger: Just for my understanding: is there a reason to make this a lambda function that we immediately call?

PR #739, opened 2025-12-11
Force BUILD_TESTING flag for CTest

bdice:
Ensures BUILD_TESTING remains ON even if libcoro sets it to OFF, allowing CTest to generate test lists properly.

This is one fix needed to make `test-rapidsmpf-cpp` work in a unified pip devcontainer. There is another issue with MPI+CUDA tests that I am not investigating yet.


pentschev:
https://github.com/rapidsai/rapidsmpf/pull/741 fixed the issue, do we still need this or should we close it?

bdice:
Still needed, this is a different requirement. Merging.

PR #742, opened 2025-12-12
Python bindings for `PinnedMemoryResource`

madsbk:
... and enable pinned memory in `bench_shuffle.cpp`, `bench_streaming_shuffle.cpp`, and `q09.cpp`.

### DIFF
--- a/cpp/benchmarks/bench_shuffle.cpp
+++ b/cpp/benchmarks/bench_shuffle.cpp
@@ -638,7 +638,7 @@ int main(int argc, char** argv) {
     rapidsmpf::BufferResource br{
         mr,
         args.pinned_mem_disable ? nullptr
-                                : std::make_shared<rapidsmpf::PinnedMemoryResource>(),
+                                : rapidsmpf::PinnedMemoryResource::make_if_available(),
         std::move(memory_available)
     };
 
--- a/cpp/benchmarks/streaming/bench_streaming_shuffle.cpp
+++ b/cpp/benchmarks/streaming/bench_streaming_shuffle.cpp
@@ -364,7 +364,7 @@ int main(int argc, char** argv) {
     auto br = std::make_shared<rapidsmpf::BufferResource>(
         mr,
         args.pinned_mem_disable ? nullptr
-                                : std::make_shared<rapidsmpf::PinnedMemoryResource>(),
+                                : rapidsmpf::PinnedMemoryResource::make_if_available(),
         std::move(memory_available)
     );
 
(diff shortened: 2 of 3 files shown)


cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -51,21 +51,19 @@ class BufferResource {
...
 
-    /// @brief Sentinel value used to disable pinned host memory.
-    static constexpr auto PinnedMemoryResourceDisabled = nullptr;
-
     /**
      * @brief Constructs a buffer resource.
      *
      * @param device_mr Reference to the RMM device memory resource used for device
      * allocations.
+     * @param memory_available Optional functions that report available memory for each
```
pentschev:
```suggestion
```

pentschev: Seems like this line was separated from the rest below.

bdice (approved):
Memory resource changes look fine. I'd like to make a game plan for moving these features to RMM and/or using CCCL directly once we get the CCCL 3.2 update finished.

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
```diff
@@ -188,11 +196,15 @@ cdef class BufferResource:
         # TODO: once RMM is migrating to CCCL (copyable) any_resource,
         # rather than the any_resource_ref reference type, we don't
         # need to keep this alive here.
-        self._mr = device_mr
+        self._device_mr = device_mr
+        self._pinned_mr = pinned_mr
+        cdef shared_ptr[cpp_PinnedMemoryResource] _pinned_mr
+        if self._pinned_mr is not None:
+            _pinned_mr = self._pinned_mr._handle
```
pentschev: Do we need to `self._pinned_mr.reset()` in `__dealloc__`?

README.md
```diff
@@ -121,18 +121,30 @@ int main(int argc, char** argv) {
...
+    // should use its own Communicator backed by its own MPI communicator.
     std::shared_ptr<rapidsmpf::Communicator> comm =
-        std::make_shared<rapidsmpf::MPI>(MPI_COMM_WORLD);
+        std::make_shared<rapidsmpf::MPI>(MPI_COMM_WORLD, options);
+
+    // Create a statistics instance for the shuffler that tracks useful information.
+    auto stats = std::make_shared<rapidsmpf::Statistics>();
+
+    // Then a progress thread where the shuffler event loop executes is created. A single
+    // progress thread may be used by multiple shufflers simultaneously.
```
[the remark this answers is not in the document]
madsbk: Let's address the README update in another PR. This is just copy-pasted.

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
pentschev: Right, but there only its own reference is reset, the reference here may still remain. I think as is `__dealloc__` will call `self._handle.reset()` which will not guarantee `self._pinned_mr` will be reset until any potential cyclic references are broken and the garbage collector runs, so to be safe I think we should call `self._pinned_mr.reset()` as well, otherwise this reference may still outlive `PinnedMemoryResource`.

PR #743, opened 2025-12-12
System info utils

madsbk:


cpp/CMakeLists.txt
```diff
@@ -274,7 +274,7 @@ target_link_libraries(
   rapidsmpf
   PUBLIC rmm::rmm cudf::cudf CCCL::CCCL $<TARGET_NAME_IF_EXISTS:ucxx::ucxx>
          $<TARGET_NAME_IF_EXISTS:libcoro>
-  PRIVATE $<$<BOOL:${RAPIDSMPF_HAVE_NUMA}>:numa>
+  PRIVATE numa
```
pentschev: I don't know if we can do this, I don't think there's a suitable package in PyPI for NUMA, that was the primary reason this was made optional.

cpp/src/system_info.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+int get_current_numa_node_id() noexcept {
+    static const int ret = [] {
+        if (!numa_available()) {
+            return 0;
+        }
+        return numa_node_of_cpu(sched_getcpu());
+    }();
+    return ret;
+}
```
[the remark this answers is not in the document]
madsbk: It would have to be its own shared library? Or header-only. I am not sure it is worth it?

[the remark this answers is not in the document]
madsbk: Make sense, something like: ee2f91f5b4be4ff5d3f3ca30d3c274bab1137738 ?

cpp/include/rapidsmpf/system_info.hpp
```diff
@@ -0,0 +1,80 @@
...
+namespace rapidsmpf {
+
+/**
+ * @brief Get the total amount of host (main) memory.
+ *
+ * @return Total host memory in bytes.
+ *
+ * @note On WSL and in containerized environments, the returned value
+ * reflects the memory visible to the Linux kernel instance, which may
+ * differ from the host system's physical memory.
```
pentschev:
```suggestion
 * @brief Get the total amount of system memory.
 *
 * @return Total host memory in bytes.
 *
 * @note On WSL and in containerized environments, the returned value
 * reflects the memory visible to the Linux kernel instance, which may
 * differ from the physical memory of the host.
```

pentschev: It is slightly confusing to use the term "host" when referring to system memory because of the note talking about host from the perspective of containers. Let's change to make a more clear distinction.

PR #744, opened 2025-12-12
Reorder `bench_shuffle` barriers

pentschev:
Reorder all barriers inside `do_run`, making the requirements more self-evident, removing barrier requirements from 3 positions in the code to only 2. This is also required for `rrun` to execute safely, otherwise synchronization issues may occur causing crossover of messages between iterations when the operation ID remains the same as is the case in `bench_shuffle`.


pentschev:
Thanks Tom!

PR #746, opened 2025-12-14
q09: fix `cudf::transform()` call.

madsbk:
`cudf::transform()` now takes a `null_policy` argument.


bdice (approved):

bdice:
Cross-linking breaking change from https://github.com/rapidsai/cudf/pull/20206.

PR #752, opened 2025-12-16
ndsh: add `--no-pinned-host-memory`

madsbk:
Enable spilling to pinned host memory by default.


madsbk:
> I'm doing a full rebuild now. But does this require CUDA 13?

No, but you should also get an error, not a segfault if it isn't supported. Sounds like a bug, can you reproduce?

madsbk:
@TomAugspurger, I can reproduce on DXG15, which has CUDA 12.2.

I have added an explicit check:
```c++
    RAPIDSMPF_EXPECTS(
        arguments.no_pinned_host_memory || is_pinned_memory_resources_supported(),
        "Pinned host memory is not supported on this system. CUDA 12.6 is one of the "
        "requirements, but additional platform or driver constraints may apply. "
        "If needed, use `--no-pinned-host-memory` to disable pinned host memory, noting "
        "that this may significantly degrade spilling performance.",
        std::invalid_argument
    );
```

PR #753, opened 2025-12-16
Basic streaming read_parquet benchmark

wence-:
Adds a simple benchmark that just reads parquet files and immediately discards them.

Might be useful for @rjzamora to benchmark read_parquet changes in isolation.


madsbk (approved):
We should add smoke tests for each ndsh benchmark at some point

wence-:
Updated to latest

PR #757, opened 2025-12-16
[MINOR] Prevent overbooking overflow

nirandaperera:
Currently, when memory is not available, the overbooking value it returns, is relative to the memory resource. ie. it tells how much of memory is overbooked in total in the memory resource. 
My intuition was, this should be relative to the size you request. ie. how much was overbooked out of the size. So, essentially overbooking should return `min(overbooking, size)`


nirandaperera:
Overbooking requires to show the total wrt to the mem type because that is the amount to the spilled essentially.
So, closing this PR as this is expected

PR #759, opened 2025-12-18
Add a glossary to the documentation

TomAugspurger:
I've added a `glossary` page to the docs (at the top level with "Quickstart", "Background", "API Reference", and "Glossary"). The intent is to have a single place for defining all the concepts we have floating around.

There's quite a bit of overlap with the API provided by rapidsmpf (classes like `BufferResource`). I've tried to keep the glossary definitions very brief and very high level, leaving implementation details to the API reference.

### DIFF
--- a/docs/source/background/actors.md
+++ b/docs/source/background/actors.md
@@ -20,13 +20,14 @@ rapidsmpf::streaming::Actor accumulator(
         if (msg.empty()) {
             break;
         }
+        auto seq = msg.sequence_number();
         auto chunk = co_await msg.release<rapidsmpf::streaming::TableChunk>()
                             .make_available(ctx);
         total += chunk.table_view().num_rows();
 
         // Forward the chunk downstream.
         co_await ch_out->send(
-            streaming::to_message(msg.sequence_number(), std::move(chunk))
+            streaming::to_message(seq, std::move(chunk))
         );
     }
 }


docs/source/glossary.md
```diff
@@ -0,0 +1,83 @@
...
+  - `HOST`: Regular system memory
+
+Message
+  A type-erased container for data payloads passed through {term}`Channel`s. Messages wrap arbitrary payload types (such as cuDF tables or buffers) along with metadata including a sequence number for ordering. Messages support deep-copy operations and can be spilled to different memory types when memory pressure occurs.
+
+Network
+  A directed graph of {term}`Node`s connected by {term}`Channel`s representing a streaming data processing pipeline. Nodes are relational operators on data, and channels are the edges connecting operations in the workflow. The network remains in place for the duration of a workflow, with nodes continuously processing data as it flows through.
+
+Node
+  A coroutine-based asynchronous operator in a streaming pipeline. Nodes receive from zero or more input {term}`Channel`s, perform computation, and send to zero or more output channels. They can be local (operating on data within a single rank) or collective (coordinating across multiple ranks). Multiple nodes execute concurrently via a thread pool executor.
```
madsbk:
```suggestion
  A coroutine-based asynchronous operator in a streaming pipeline. Nodes receive from zero or more input {term}`Channel`s, perform computation, and send to zero or more output channels. From the network’s perspective, nodes are always local operators. A node may internally use a {term}`Communicator` to perform inter-rank communication, but this coordination is fully encapsulated within the node and is not visible to the surrounding network or channels. Multiple nodes execute concurrently via a thread pool executor.
```

madsbk (approved):

PR #760, opened 2025-12-18
Add streaming benchmark validator

TomAugspurger:
This adds a CLI validator for our streaming benchmarks. Our expected results come from running a SQL query against the data with DuckDB. The actual results comes from running the C++ binary.

488c8a7d499f46ff11d8a9f175c8730858751402 has the validator changes.

2fc29d395a097b8ce66b273c737738b9adc20e94 is a preliminary commit that updates our pre-commit config to validate our python files under `ci/`, along with the changes to get the current scripts passing.

### DIFF
--- a/conda/recipes/rapidsmpf/recipe.yaml
+++ b/conda/recipes/rapidsmpf/recipe.yaml
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION.
 # SPDX-License-Identifier: Apache-2.0
 schema_version: 1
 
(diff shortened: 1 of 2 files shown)


ci/test_python.sh
```diff
@@ -57,5 +58,13 @@ rapids-logger "Pytest RapidsMPF (MPI+UCXX)"
 rapids-logger "Pytest RapidsMPF (UCXX polling mode)"
 RAPIDSMPF_UCXX_PROGRESS_MODE=polling ./ci/run_pytests.sh --disable-mpi
 
+rapids-logger "Validate NDSH benchmarks"
+python ./ci/checks/validate_ndsh.py \
```
bdice: Why is this in `test_python.sh`? It seems to run librapidsmpf C++ benchmarks. I would expect `test_python.sh` to run benchmarks of the Python `rapidsmpf` package. It would be perfectly okay to add a Python dependency to the C++ test environment for running and validating these benchmarks.

TomAugspurger: > It would be perfectly okay to add a Python dependency to the C++ test environment for running and validating these benchmarks.

Yep, the Python dependency was the reason. If adding the Python dependency to the C++ job is preferable then we can do that.

dependencies.yaml
```diff
@@ -302,6 +302,14 @@ dependencies:
           - openmpi >=5.0  # See <https://github.com/rapidsai/rapidsmpf/issues/17>
           - valgrind
           - gdb
+          # NDSH validation dependencies
+          - duckdb
+          - numpy
+          - nvidia-ml-py>=12
```
wence-: What do we need the nvml bindings for for checking correctness?

wence- (approved):
Tiny suggestions

PR #762, opened 2025-12-19
[MINOR] Fix build warnings

nirandaperera:
I had some build errors in latest `main` with an updated conda env


TomAugspurger (approved):
I think we'll need to wait for the post-RMM/CCCL update to make some progress before CI will pass.

PR #764, opened 2025-12-22
Host buffer with a type-erased owner

### LINKED ISSUE #761: Take ownership of a pinned host buffer in `HostBuffer`
Currently, `cudf::` methods returns pinned host memory allocations as `rmm::device_buffer`s when we pass the `PinnedMemoryResource` as a device mr. In HostBuffer, we can not take the ownership of this, because rmm device buffer can not be released. https://github.com/rapidsai/rmm/issues/2189

Workaround.
Add a `unique_ptr<rmm::device_buffer> owner == nullptr` class member to `HostBuffer` to handle such scenarios. Could be a bit consfusing to have a device buffer in HostBuffer though.

nirandaperera:
This PR allows us to create a `HostBuffer` with a any container that owns data in contiguous. Currently looking at std::vector and rmm::device_buffer that holds pinned host memory. 

Fixes #761


cpp/include/rapidsmpf/memory/host_buffer.hpp
```diff
@@ -139,10 +154,73 @@ class HostBuffer {
...
+    /**
+     * @brief Construct a `HostBuffer` by taking ownership of a
+     * `std::vector<std::uint8_t>`.
+     *
+     * The buffer takes ownership of the vector's memory. The vector is moved into
+     * internal storage and will be destroyed when the `HostBuffer` is destroyed.
+     *
+     * @note The memory is regular host memory (not pinned). Async operations on this
+     * buffer may require synchronization.
+     *
```
madsbk: I don't think it has to be non-pinned? Also, the stream semantic is the same as all the other ctors/factories?

```suggestion
```

cpp/src/memory/host_buffer.cpp
```diff
@@ -91,6 +112,53 @@ HostBuffer HostBuffer::from_uint8_vector(
...
+    std::span<std::byte> span{ptr, size};
+
+    std::unique_ptr<void, OwnedStorageDeleter> owned_storage{
+        ptr, [shared_vec_ = std::move(shared_vec)](void*) mutable { shared_vec_.reset(); }
+    };
+
+    return HostBuffer{span, stream, std::move(mr), std::move(owned_storage)};
+}
+
+HostBuffer HostBuffer::from_owned_rmm_pinned_host_buffer(
```
madsbk: What about rename to `from_rmm_device_buffer()` and then only check that the buffer is accessible from the host?

cpp/include/rapidsmpf/memory/host_buffer.hpp
```diff
@@ -8,10 +8,13 @@
 #include <cstdint>
 #include <cstdlib>
 #include <cstring>
+#include <functional>
+#include <memory>
 #include <span>
```
changchunkao88: ## A second-level heading

cpp/include/rapidsmpf/memory/host_buffer.hpp
```diff
@@ -8,10 +8,13 @@
...
 #include <vector>
 
 #include <rmm/cuda_stream_view.hpp>
+#include <rmm/device_buffer.hpp>
 #include <rmm/resource_ref.hpp>
```
changchunkao88: include <rmm/resource_ref.hpp>  // Resource management utilities for RAPIDS Memory Manager (RMM)

cpp/include/rapidsmpf/memory/host_buffer.hpp
```diff
@@ -8,14 +8,18 @@
...
 
 #include <rapidsmpf/error.hpp>
 #include <rapidsmpf/memory/host_memory_resource.hpp>
+#include <rapidsmpf/memory/pinned_memory_resource.hpp>
```
changchunkao88: include <rapidsmpf/memory/host_memory_resource.hpp>

changchunkao88 (changes requested):

cpp/src/memory/host_buffer.cpp
```diff
@@ -91,6 +112,53 @@ HostBuffer HostBuffer::from_uint8_vector(
...
+        pinned_host_buffer != nullptr,
+        "pinned_host_buffer must not be null",
+        std::invalid_argument
+    );
+
+    RAPIDSMPF_EXPECTS(
+        ptr_to_memory_type(pinned_host_buffer->data()) == MemoryType::PINNED_HOST,
+        "pinned_host_buffer must be a pinned host buffer",
+        std::invalid_argument
+    );
```
madsbk: I think we should drop `ptr_to_memory_type()` and instead check that `pinned_host_buffer->data()` is accessible from the host? I doesn't have to be pinned memory!

madsbk (approved):

PR #765, opened 2025-12-30
Fix `non_device_size` calculation in unpack_and_concat

TomAugspurger:
I believe this is intended to reflect the number of bytes not in device memory. We'll want to examine the other usage of `non_device_size` (in `br->reserve_device_memory_and_spill`) to ensure that we're OK with this change. IIUC, we'll end up reserving additional bytes equal to the number of bytes that were not in GPU memory.


nirandaperera (approved):
nice catch @TomAugspurger LGTM

PR #774, opened 2026-01-09
Use `TopologyDiscovery` from cuCascade

pentschev:
Add cuCascade as a dependency to RapidsMPF, allowing removal of `TopologyDiscovery` implementation from RapidsMPF and use the one from cuCascade.


cpp/tools/check_resource_binding.cpp
```diff
@@ -30,7 +30,8 @@
 
 #include <rapidsmpf/bootstrap/utils.hpp>
 #include <rapidsmpf/system_info.hpp>
-#include <rapidsmpf/topology_discovery.hpp>
+
+#include <memory/topology_discovery.hpp>
```
wence-: Opened https://github.com/NVIDIA/cuCascade/issues/21 to ask cucascade to move stuff to a namespaced subdirectory.

cmake/thirdparty/get_cucascade.cmake
```diff
@@ -0,0 +1,70 @@
...
+    set(_result "$ENV{CONDA_PREFIX}/include")
+  else()
+    set(_result "${CMAKE_PREFIX_PATH}/include")
+  endif()
+  set(${out_var}
+      "${_result}"
+      PARENT_SCOPE
+  )
+endfunction()
+
```
KyleFromNVIDIA: This is not being used anywhere (and its implementation is not good practice anyway). Please remove it.

PR #779, opened 2026-01-12
Address RMM setup in ``bootstrap_dask_cluster``

rjzamora:
**Proposed Changes**:
- Adds a new `rapidsmpf.integrations.core.setup_rmm_pool` function to setup the RMM pool for a single "worker".
  - The logic is essentially copied from dask-cuda
  - **Motivation**: We do not want to rely on Dask/Dask-CUDA for this logic as we branch out to other deployment strategies.
- Adds a call to the new `setup_rmm_pool` function in `rapidsmpf.integrations.core.rmpf_worker_setup`
- Adds test coverage to `rapidsmpf/python/rapidsmpf/rapidsmpf/tests/test_dask.py`

@pentschev - Let me know what you think about "copying" the dask-cuda logic like this. Feel free to suggests something different.

@quasiben - Perhaps this makes multi-node deployment a bit easier? (i.e. we rely less on logic within `LocalCUDACluster`)

### DIFF
--- a/python/rapidsmpf/rapidsmpf/integrations/core.py
+++ b/python/rapidsmpf/rapidsmpf/integrations/core.py
@@ -670,6 +670,42 @@ def spill_func(
     return ctx.spill_collection.spill(amount, stream=DEFAULT_STREAM, device_mr=mr)
 
 
+def _parse_pool_size(
+    value: float | None, total_device_memory: int, *, alignment_size: int = 256
+) -> int | None:
+    """
+    Parse a pool size value, supporting fractions of total device memory.
+
+    Parameters
+    ----------
+    value
+        Can be:
+        - None: Returns None
+        - A float in (0, 1]: Interpreted as fraction of total device memory
+        - A float > 1: Interpreted as byte count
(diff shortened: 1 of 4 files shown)


pentschev:
Copying is something I despise because they're going to diverge eventually and we'll have multiple places to update. However, I understand our alternatives are very limited:

1. We need a separate repository to placee these functions (I know this isn't going to happen, so we can rule it out immediately); or
2. Depend directly on Dask-CUDA.

Therefore, our only alternative to copying is really to depend on Dask-CUDA and I think we don't want to do that. Thus unfortunately I think we have no other options.

pentschev (approved):
I'm preemptively approving this PR after pointing out some pitfalls in https://github.com/rapidsai/rapidsmpf/pull/779/files#r2687445927, for example. I don't want to block this in case this becomes important, but it's important to keep those in mind that users can easily shoot themselves in the foot. I'm also happy to review if you want to try and address my suggestions trying to make this a bit less error prone.

rjzamora:
Closing as stale

PR #781, opened 2026-01-12
Re-enable and extend chunked pack benchmarks

nirandaperera:
Since https://github.com/NVIDIA/cccl/pull/7006 is merged, we should be able to reenable chunked pack benchmarks that use pinned memory. 

This PR also extends the pack benchmarks to copy the packed data to a destination HostBuffer (host/ pinned). This gives a more appropriate picture for spilling. 

## Latest results 
https://docs.google.com/spreadsheets/d/1yXiB3aFZO8GUD4dAVnwh7o9zjzXKwaSvGYGZphf9jgQ/edit?usp=sharing

## Previous results

Workstation RTX A6000 driver 580.105.08  CUDA 13.0 
image

PDX H100  driver 535.216.03   CUDA 13.1  (using cuda-compat) 
image

Looking at these results, **if we consider the spilling scenario where we pack and copy to host/ pinned host memory**,  for a 1GB table, 
| A6000                                     |           | H100                                      |           |
|-------------------------------------------|----------:|-------------------------------------------|----------:|
| BM_Pack_device_copy_to_pinned_host        | 22,760.87 | BM_Pack_device_copy_to_pinned_host        | 45,062.92 |
| BM_ChunkedPack_device_copy_to_pinned_host | 22,483.63 | BM_ChunkedPack_device_copy_to_pinned_host | 43,336.96 |
| BM_ChunkedPack_device_copy_to_host        | 21,823.79 | BM_ChunkedPack_pinned_copy_to_pinned_host | 22,014.05 |
| BM_Pack_device_copy_to_host               | 21,011.26 | BM_ChunkedPack_device_copy_to_host        | 20,057.96 |
| BM_ChunkedPack_pinned_copy_to_pinned_host | 11,564.92 | BM_ChunkedPack_pinned_copy_to_host        | 14,565.09 |
| BM_ChunkedPack_pinned_copy_to_host        | 11,346.53 | BM_Pack_device_copy_to_host               | 14,189.53 |
| BM_Pack_pinned_copy_to_pinned_host        |  9,027.17 | BM_Pack_pinned_copy_to_pinned_host        |  7,902.45 |
| BM_Pack_pinned_copy_to_host               |  8,462.00 | BM_Pack_pinned_copy_to_host               |    869.39 |


nirandaperera:
The latest results on my workstation. I need to verify the results in a H100 machine as well

image

cpp/benchmarks/bench_pack.cpp
```diff
@@ -148,7 +207,7 @@ void run_chunked_pack(
                 )
             );
             RAPIDSMPF_CUDA_TRY(cudaMemcpyAsync(
-                static_cast<std::uint8_t*>(destination.data()) + offset,
+                reinterpret_cast<std::uint8_t*>(destination.data()) + offset,
```
madsbk: isn't `destination.data()` a `std::byte*` already?

cpp/benchmarks/bench_pack.cpp
```diff
@@ -101,8 +214,9 @@ static void BM_Pack_pinned(benchmark::State& state) {
         cuda_mr, rmm::percent_of_free_device_memory(40)
     };
```
wence-: Unused.

cpp/benchmarks/bench_pack.cpp
```diff
@@ -196,18 +346,47 @@ static void BM_ChunkedPack_device(benchmark::State& state) {
...
-static void BM_ChunkedPack_pinned(benchmark::State& state) {
-    state.SkipWithMessage("Skipping until cudf#20886 is fixed");
-    /*     if (!rapidsmpf::is_pinned_memory_resources_supported()) {
+static void BM_ChunkedPack_device_copy_to_host(benchmark::State& state) {
+    auto const table_size_mb = static_cast<std::size_t>(state.range(0));
+    auto const table_size_bytes = table_size_mb * MB;
+
+    // Bounce buffer size: max(1MB, table_size / 10)
+    auto const bounce_buffer_size = std::max(MB, table_size_bytes / 10);
+
```
wence-: I think this is a bad model for the bounce buffer size. I don't think we want to scale it with the table size, but rather have a _fixed_ size bounce buffer. That way, if we're using a putative fixed size pinned host resource each chunk neatly fits into a block from that host resource.

wence- (commented):
A few comments

cpp/benchmarks/bench_pack.cpp
nirandaperera: Okay, let me run with a fixed sized bounce buffer. The reason why I didnt go ahead with that previously was, fixed size buffer resource proposed in cucascade is 1MB. I felt like there will be too many calls to cudamemcpyasync. But I should have run a benchmark, rather than assuming things.

cpp/benchmarks/bench_pack.cpp
```diff
@@ -73,20 +77,126 @@ static void BM_Pack_device(benchmark::State& state) {
...
+    rapidsmpf::PinnedMemoryResource pinned_mr;
+    run_pack(state, table_size_mb, cuda_mr, pinned_mr, stream);
+}
+
+/**
+ * @brief Runs the cudf::pack and copy the packed data to a host buffer
+ * @param state The benchmark state
+ * @param table_size_mb The size of the table in MB
+ * @param table_mr The memory resource for the table
+ * @param pack_mr The memory resource for the packed data
```
pentschev:
```suggestion
 * @param pack_mr The memory resource for the packed data
 * @param dest_mr The memory resource for the destination data
```

cpp/benchmarks/bench_pack.cpp
pentschev: I think this is still unresolved, or did the code change here? Still looks like a variable bounce buffer size based on the input size, where Lawrence suggests using a fix buffer size which I agree seems more realistic an easier to manage.

cpp/benchmarks/bench_pack.cpp
```diff
@@ -221,107 +333,714 @@ static void BM_ChunkedPack_pinned(benchmark::State& state) {
...
+
+    state.SetBytesProcessed(
+        static_cast<std::int64_t>(state.iterations())
+        * static_cast<std::int64_t>(table_size)
+    );
+    state.counters["table_size_mb"] =
+        static_cast<double>(table_size) / static_cast<double>(MB);
+    state.counters["num_rows"] = nrows;
+    state.counters["bounce_buffer_mb"] =
+        static_cast<double>(bounce_buffer_size) / static_cast<double>(MB);
```
pentschev: This function is named `run_chunked_pack_without_bounce_buffer`, but here's a bounce buffer size, why?

cpp/benchmarks/bench_pack.cpp
```diff
@@ -221,107 +333,714 @@ static void BM_ChunkedPack_pinned(benchmark::State& state) {
...
+
+/**
+ * @brief Custom argument generator for pack benchmarks.
+ *
+ * Configures benchmarks to run with various table sizes ranging from 1MB to 4GB.
+ *
+ * @param b The benchmark to configure with arguments.
+ */
+void PackArguments(benchmark::internal::Benchmark* b) {
+    // Test different table sizes in MB (minimum 1MB as requested)
```
pentschev: As requested by whom/where? I was going to ask if there's use for smaller sizes too to show cases where performance is bad and should be avoided (if that's the case).

nirandaperera:
Based on the [latest results in H100 and NVL4](https://docs.google.com/spreadsheets/d/1yXiB3aFZO8GUD4dAVnwh7o9zjzXKwaSvGYGZphf9jgQ/edit?usp=sharing), following were my conclusions. 

## `chunked_pack`ing bounce buffer size 

In both H100 & NVL4, when packing a 1GB device table to device/ pinned host, bandwidth increases as we increase the size of the bounce buffer (this corresponds to the number of `cudamemcpyasync` calls to the output buffer offsets).
- In pinned host, the packing bw flattens after 8MB in H100 and 32MB in NVL4. 

So, **when using `chunked_pack`, we should try to minimize the number of async copies to the output buffer.**

PS: this the main reason why I picked 10% of the table size as the bounce buffer for next set of experiments. 

## Best strategy to pack a table to a destination memory type

### To device 
- `pack` is strictly better than `chunked_pack`. Also, when packing, we make a device reservation before calling the operation. So, chunked_pack is, irrelevant. 
- Therefore, use `pack` with device mr

### To pinned host 
- `pack`, and `chunked_pack` to offset output buffer yield similar performance. `chunked_pack` with a device bounce buffer, is slightly worse. 
- Therefore, use `pack` with pinned mr (as a device mr) 

### To host
- Based on the results (best first) 
    - `pack` to device and copy to host, `chunked_pack` with a device bounce buffer (1.5-1.7x faster than pinned)
    - `pack` to pinned and copy to host, `chunked_pack` with a pinned bounce buffer 
- Since chunked_pack offers less device/ pinned memory consumption, we let's forget about `pack`ing.
- So the strategy would be, 
   - First try to reserve device memory for the estimated size of the table (this reduces the number of subsequent memcpys to host buffer) 
   - If the actual device available (considering overbooking) > 1MB, allocate that and use for chunked_pack
   - Else, try with pinned memory the same way. 
   - If we can't truly reserve at least 1MB pinned memory, throw. 
   - Since chunked_pack requires at least a 1MB bounce buffer, for small tables, use `pack` to device and copy.

madsbk:

PR #785, opened 2026-01-14
Add device PCI Bus ID information to `check_resource_binding`

pentschev:
The device PCI Bus ID information is useful to provide better understanding to what physical device is used for a given rank/device ID. Additionally reorganize the code a bit providing separation between detection and printing information.


madsbk (approved):
Looks good, thanks @pentschev

PR #796, opened 2026-01-19
Pass objects safely from Python coroutines to C++ coroutines

pentschev:
In `Channel`, currently `c_msg` is created in the stack of `recv()`, under normal circumstances this is fine, as we're certain that the message will still be in scope when `cpp_channel_recv()` is called and it is passed by reference. However, an exception being raised may cause `recv` to go out-of-context and invalidate `c_msg`, which leads to a segfault when `_channel_recv_task` is later executed. One way to prevent that issue is to use a `shared_ptr` passed by value which is guaranteed to have its refcount increased and thus making the local copy inside of `recv` unnecessary to keep the object safely accessible.

Similar cases occur in `AllGather` and `Shuffler` and they are also addressed using the same approach.

### DIFF
--- a/cpp/.clang-format
+++ b/cpp/.clang-format
@@ -36,6 +36,7 @@ BreakBeforeBinaryOperators: NonAssignment
 BreakBeforeBraces: Custom
 BreakStringLiterals: true
 ColumnLimit: 90
+CommentPragmas: 'SPDX-(FileCopyrightText|License-Identifier): '
 DerivePointerAlignment: false
 IndentCaseBlocks: true
 IndentCaseLabels: false
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -1,6 +1,6 @@
 # =================================================================================
 # cmake-format: off
-# SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 # cmake-format: on
 # =================================================================================
--- a/cpp/benchmarks/CMakeLists.txt
+++ b/cpp/benchmarks/CMakeLists.txt
@@ -1,6 +1,6 @@
 # =================================================================================
 # cmake-format: off
-# SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 # cmake-format: on
 # =================================================================================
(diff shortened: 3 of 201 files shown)


python/rapidsmpf/rapidsmpf/streaming/core/channel.pyx
```diff
@@ -283,18 +283,20 @@ cdef class Channel:
         A `Message` if a message is available, otherwise ``None`` if the channel is
         shut down and empty.
         """
-        cdef cpp_Message c_msg
+        cdef shared_ptr[cpp_Message] c_msg_ptr = (
```
madsbk: Keep the name `c_msg`.

```suggestion
        cdef shared_ptr[cpp_Message] c_msg = (
```

python/rapidsmpf/rapidsmpf/streaming/core/channel.pyx
```diff
@@ -283,18 +283,20 @@ cdef class Channel:
...
+            shared_ptr[cpp_Message](new cpp_Message())
+        )
         ret = asyncio.get_running_loop().create_future()
         Py_INCREF(ret)
         with nogil:
             cpp_channel_recv(
                 ctx._handle,
                 self._handle,
-                c_msg,
+                c_msg_ptr,
```
madsbk:
```suggestion
                c_msg

File: python/rapidsmpf/rapidsmpf/streaming/coll/shuffler.pyx [hunk 3]
```diff
@@ -56,17 +56,23 @@ cdef extern from * nogil:
 
     coro::task<void> _extract_any_async_task(
         rapidsmpf::streaming::ShufflerAsync *shuffle,
-        std::optional<std::pair<std::uint32_t, std::vector<rapidsmpf::PackedData>>>
-            &output
+        std::shared_ptr<
+            std::optional<
```
madsbk: What about dropping `std::optional` and use `shared_ptr == nullptr` instead?

pentschev: Actually I fooled myself on this. We allocate the `shared_ptr` before passing it to the C++ coroutine and there it is used as a return value, at that point it will never be a `nullptr` so we can't use it as a sentinel to determine whether it has content or not, thus the optional is required to determine whether it has a value. I reverted that change now, but please let me know if you have something different in mind that I'm failing to understand.

python/rapidsmpf/rapidsmpf/streaming/core/channel.pyx
```diff
@@ -283,7 +284,12 @@ cdef class Channel:
         A `Message` if a message is available, otherwise ``None`` if the channel is
         shut down and empty.
         """
-        cdef cpp_Message c_msg
+        # Use a shared_ptr here for safety, if an exception occurs this coroutine may
+        # go out of scope and destroy objects in its stack before the C++ coroutine
+        # executes, leading to a segfault.
+        cdef shared_ptr[cpp_Message] c_msg = (
+            shared_ptr[cpp_Message](new cpp_Message())
+        )
```
wence-: *nit* (throughout): Can we use `cdef shared_ptr[cpp_Message] c_msg = make_shared[cpp_Message]()` ?

wence- (approved):
One trivial comment

PR #801, opened 2026-01-20
Split ndsh validation into run and validate

TomAugspurger:
This changes `validate_ndsh` to have two subcommands:

1. `run`: generates the results by running the by-hand queries and and expected by running duckdb.
2. `validate`: compares the results and expecte

This will make it a bit easier to reuse this script in other contexts (e.g. using cudf-polars & pre-computed results).

cc @quasiben

### DIFF
--- a/cpp/include/rapidsmpf/utils/string.hpp
+++ b/cpp/include/rapidsmpf/utils/string.hpp
@@ -79,7 +79,7 @@ std::string format_nbytes(
  * @brief Format a time duration as a human-readable string.
  *
  * Converts a duration given in seconds into a scaled string representation
- * using common time units such as ns, µs, ms, s, min, h, and d.
+ * using common time units such as ns, us, ms, s, min, h, and d.
  *
  * The duration is accepted as a @c double to support both fractional seconds
  * and very large values without overflow.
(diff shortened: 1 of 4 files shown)


cpp/scripts/validate_ndsh.py
```diff
@@ -82,6 +87,26 @@ def discover_benchmarks(
     return sorted(benchmarks)
 
 
+def discover_parquet_files(directory: Path) -> dict[str, Path]:
+    """
+    Discover parquet files matching the qDD.parquet pattern.
+
+    Returns a dict mapping query_name (e.g., 'q03') to file path.
```
pentschev: Let's maybe follow the same pattern of the rest of our Python code, adding distinct docstring `Parameters`/`Returns` sections?

cpp/scripts/validate_ndsh.py
```diff
@@ -316,27 +341,27 @@ def compare_parquet(
...
     *,
     reuse_expected: bool = False,
     reuse_output: bool = False,
 ) -> bool:
     """
-    Validate a single benchmark.
+    Run a single benchmark and generate expected results.
 
-    Returns True if validation passes, False otherwise.
+    Returns True if both operations succeed, False otherwise.
```
pentschev: Ditto on docstring sections.

PR #803, opened 2026-01-21
String utilities

madsbk:
Introduce new utilities and refactor existing ones for consistent formatting and parsing of time intervals and byte counts.

Also renamed the option `memory_reserve_timeout_ms -> memory_reserve_timeout`.


madsbk:
> Overall this looks good, basing my understanding on the tests it looks right, however, I've admittedly not tried to understand all the highly complex regexes in the code as they're far too advanced for my regex knowledge.

I had some help from a LLM :) that is also why I added a lot of tests!

Issue #817, opened 2026-01-23
Align on standard ETL Metadata schema


rjzamora:
In order to make adaptive decisions about shuffling/broadcasting/etc during run time, we need to keep track of some basic metadata. In cudf-polars, this metadata is propagated through the network using a distinct `Channel` between every node. After https://github.com/rapidsai/rapidsmpf/pull/811, we will be able to use the same `Channel` for both data and metadata payloads. However, in order for cudf-polars to **use** any C++ Join/GroupBy nodes, we will (probably) still need cudf-polars and RapidsMPF to align on a common/standard metadata schema.

At the very minimum, the channel metadata should keep track of the expected local chunk count, and whether the local data coming through the channel is already broadcasted (duplicated) across all ranks. In order to avoid "re-shuffling" data unnecessarily, we also want to track the "partitioning" status of the data being sent into an actor.

## Current Metadata Schema

The current `Metadata` [definition in cudf-polars](https://github.com/rapidsai/cudf/blob/4a8cdb232e4d5a8cfd00783dbb23c79b2cd77307/python/cudf_polars/cudf_polars/experimental/rapidsmpf/utils.py#L90) looks like:

```python
class Metadata:
    """Metadata payload for an ETL workload."""

    local_count: int
    """Local chunk-count estimate for the current rank."""
    global_count: int | None
    """Global chunk-count estimate across all ranks."""
    partitioning: HashPartitioned | None
    """How the data is partitioned, or None if not partitioned."""
    duplicated: bool
    """Whether the data is duplicated (identical) on all workers."""

class HashPartitioned:
    """Hash-partitioning metadata.

    columns: tuple[str, ...]
    """Columns the data is hash-partitioned on."""
    scope: Literal["local", "global"]
    """Whether data is partitioned locally (within a rank) or globally (across all ranks)."""
    count: int
    """The modulus used for hash partitioning (number of partitions)."""
```

## Partitioning metadata

An important "open" question is how to best encode partitioning status of the data being sent through a channel. For now, the `partitioning` attribute can either be empty (`None`), or it can be set to `HashPartitioned`. This makes it easy to keep track of data that has already been globally shuffled. However, it does not make it easy to distinguish between data that was directly shuffled into N_g global chunks, and data that was shuffled between R ranks, and then N_l local chunks.

PR #819, opened 2026-01-24
Introduce standard ``ChannelMetadata`` definition

rjzamora:
Closes https://github.com/rapidsai/rapidsmpf/issues/817

**Goal**: Define a standard channel-metadata definition for cudf-based workflows. This metadata should be sufficient for dynamic Join/GroupBy decisions at runtime, and should be adoptable by cudf-polars.

**Proposal**: We need to keep track of estimated chunk counts, broadcast (duplication) status, and hierarchical partitioning status. I think the trickiest part to get right is the hierarchical partitioning. We need something that cleanly covers dynamic Join/GroupBy planning, and will be easily extended to Sort-based workflows. I spent a lot of time thinking about this, and feel that we must track both `local` and `inter_rank` partitioning, and that the `local` partitioning specification is allowed to `INHERIT` (i.e. preserve the partitioning performed at the parent level). Since we *may* want to create additional levels of partitioning in the future (e.g. rank groups), we don't want to make it impossible to extend from a strict two-level design.

Rough outline:

```c++
// Hash partitioning scheme.
struct HashScheme {
    std::vector<std::string> columns;  ///< Columns to hash on.
    std::int64_t modulus;  ///< Hash modulus (number of partitions).
};

// Partitioning-type enum "label"
enum class SpecType : std::uint8_t {
    NONE,  ///< No partitioning at this level.
    INHERIT,  ///< Inherit from parent level.
    HASH,  ///< Hash partitioning.
    // Can add "other" flavors later (e.g. ORDERED)
};

// Partitioning-specification polymorphic container
struct PartitioningSpec {
    SpecType type = SpecType::NONE;  ///< The type of partitioning.
    ... factories/etc ...
};

// Hierarchical partitioning metadata
struct Partitioning {
    PartitioningSpec inter_rank;  ///< Distribution across ranks (Context::comm()).
    PartitioningSpec local;  ///< Distribution within a rank ("single" comm).
};

// Channel-level metadata describing the "tabular" data stream
struct ChannelMetadata {
    std::uint64_t local_count{};  ///< Local chunk-count estimate for this rank.
    Partitioning partitioning;  ///< How the data is partitioned.
    bool duplicated{};  ///< Whether data is duplicated on all workers.
};
```

### DIFF
--- a/python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyi
+++ b/python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyi
@@ -9,12 +9,6 @@ from typing import Literal
 from rapidsmpf.streaming.core.message import Message
 
 class HashScheme:
-    """
-    Hash partitioning scheme.
-
-    Rows are distributed by ``hash(table[column_indices]) % modulus``.
-    """
-
     def __init__(self, column_indices: tuple[int, ...], modulus: int) -> None: ...
     @property
     def column_indices(self) -> tuple[int, ...]: ...
@@ -26,13 +20,6 @@ class HashScheme:
 PartitioningSpecValue = HashScheme | None | Literal["inherit"]
 
 class Partitioning:
-    """
-    Hierarchical partitioning metadata for a data stream.
-
-    Describes how data flowing through a channel is partitioned at multiple
-    levels of the system hierarchy.
-    """
-
     def __init__(
         self,
         inter_rank: PartitioningSpecValue = None,
@@ -46,13 +33,6 @@ class Partitioning:
     def __repr__(self) -> str: ...
 
 class ChannelMetadata:
-    """
-    Channel-level metadata describing the data stream.
-
-    Contains information about chunk counts, partitioning, and duplication
-    status for the data flowing through a channel.
-    """
-
     def __init__(
         self,
         local_count: int,


cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * @brief Type tag for PartitioningSpec.
+ */
+enum class SpecType : std::uint8_t {
+    NONE,  ///< No partitioning at this level.
+    ALIGNED,  ///< Aligned with parent level.
+    HASH,  ///< Hash partitioning.
+};
+
```
wence-: I guess if I were just in C++, I would do:

```
struct None {
};
struct Aligned {
};
struct Hash {
  std::vector<cudf::size_type> indices;
  int modulus;
};

using PartitionSpec = std::variant<None, Aligned, Hash>;
```
But those are hard to use in python bindings :(

cpp/tests/streaming/test_channel_metadata.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    // Equality (avoid inline construction to prevent GCC 14.x false positive)
+    ChannelMetadata m_same{4, p, true};
+    ChannelMetadata m_diff{8, p, true};
+    EXPECT_EQ(m, m_same);
+    EXPECT_NE(m, m_diff);
+}
+
+TEST_F(StreamingChannelMetadata, MessageRoundTrip) {
+    // ChannelMetadata round-trip (avoid copy to prevent GCC 14.x false positive)
```
wence-: Please explain what these comments mean.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -0,0 +1,208 @@
...
+        if not message.empty():
+            raise ValueError("cannot move into a non-empty message")
+        message._handle = cpp_to_message_channel_metadata(
+            sequence_number, move(self.release_handle())
+        )
+
+    cdef void _check_handle(self) except *:
+        """Raise ValueError if handle has been released."""
+        if not self._handle:
+            raise ValueError("ChannelMetadata has been released (handle moved out)")
```
wence-: suggestion:

parse, don't validate.

```
cdef const cpp_ChannelMetadata *handle_ptr(self):
    if not self._handle:
        raise ValueError("ChannelMetadata is uninitialized, has it been released?")
    return self._handle.get()
```

And then use `self.handle_ptr()` whereever you want to access `self._handle`.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -0,0 +1,208 @@
...
+    def duplicated(self) -> bool:
+        self._check_handle()
+        return deref(self._handle).duplicated
+
+    def __eq__(self, other):
+        if not isinstance(other, ChannelMetadata):
+            return NotImplemented
+        self._check_handle()
+        (<ChannelMetadata>other)._check_handle()
+        return deref(self._handle) == deref((<ChannelMetadata>other)._handle)
```
wence-: e.g. this would become:
```
return deref(self.handle_ptr()) == deref((<ChannelMetadata>other).handle_ptr())
```

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /**
+     * @brief Create a spec indicating alignment with the parent level.
+     * @return A PartitioningSpec with type ALIGNED.
+     */
+    static PartitioningSpec aligned() {
+        return {.type = SpecType::ALIGNED, .hash = std::nullopt};
+    }
+
```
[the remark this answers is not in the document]
rjzamora: I think you are right that "aligned" is the wrong word. This type of partitioning is meant to signal that the partitioning has not been disturbed at this level, and that we are simply preserving whatever partitioning was performed at the parent level. Perhaps a better word is "passthrough"?

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyi
```diff
@@ -0,0 +1,73 @@
...
+# SPDX-License-Identifier: Apache-2.0
+"""Type stubs for channel_metadata module."""
+
+from __future__ import annotations
+
+from typing import Literal
+
+from rapidsmpf.streaming.core.message import Message
+
+class HashScheme:
```
wence-: Let's remove the docstrings from the typestub, since they don't show up in docs anyway and duplicate the source of truth in the cython binding

wence- (approved):

PR #827, opened 2026-01-27
Introduce the coroutine `reserve_memory()` and use it in the ndsh workflows

madsbk:
Introduce a reserve memory helper function:
```c++
/**
 * @brief Reserve memory using the context memory reservation mechanism.
 *
 * Submits a memory reservation request for the configured memory type and suspends
 * until the request is satisfied. If no pending reservation request can be satisfied
 * within the configured `"memory_reserve_timeout"`, the behavior depends on
 * @p allow_overbooking.
 *
 * This is a convenience helper that returns only the `MemoryReservation`. If more
 * control is required, for example inspecting the amount of overbooking, callers
 * should use `MemoryReserveOrWait` directly, such as
 * `ctx.memory(MemoryType::DEVICE).reserve_or_wait_or_overbook(size, net_memory_delta)`.
 *
 * Priority and progress semantics are identical to
 * `MemoryReserveOrWait::reserve_or_wait()`. In particular, @p net_memory_delta is used as
 * a heuristic to prefer eligible requests that are expected to reduce memory pressure
 * sooner. Smaller values have higher priority.
 *
 * @param ctx Node context used to obtain the memory reservation handle.
 * @param size Number of bytes to reserve.
 * @param net_memory_delta Estimated net change in memory usage after the reservation is
 * allocated and the dependent operation completes. Smaller values have higher priority.
 * @param mem_type Memory type for which to reserve memory.
 * @param allow_overbooking Controls the behavior when no progress is possible within the
 * configured timeout:
 * - If set to `AllowOverbooking::YES`, the call may overbook memory when forcing
 *   progress.
 * - If set to `AllowOverbooking::NO`, the call fails if no progress is possible.
 * - If not provided, the default behavior is determined by the configuration option
 *   `"allow_overbooking_by_default"`.
 * @return The allocated memory reservation.
 *
 * @throws std::runtime_error If shutdown occurs before the request can be processed.
 * @throws std::overflow_error If no progress is possible within the timeout and
 * `allow_overbooking` resolves to `AllowOverbooking::NO`.
 *
 * @code{.cpp}
 * // Reserve memory inside a node:
 * auto res = co_await reserve_memory(
 *     ctx,
 *     1024,
 *     0,  // net_memory_delta
 *     MemoryType::DEVICE,
 *     AllowOverbooking::YES
 * );
 * EXPECT_EQ(res.size(), 1024);
 *
 * // Disable overbooking and fail if no progress is possible:
 * auto res2 = co_await reserve_memory(
 *     ctx,
 *     2048,
 *     0,  // net_memory_delta
 *     MemoryType::DEVICE,
 *     AllowOverbooking::NO
 * );
 * @endcode
 *
 * @see MemoryReserveOrWait::reserve_or_wait()
 * @see MemoryReserveOrWait::reserve_or_wait_or_overbook()
 * @see MemoryReserveOrWait::reserve_or_wait_or_fail()
 */
coro::task<MemoryReservation> reserve_memory(
    std::shared_ptr<Context> ctx,
    std::size_t size,
    std::int64_t net_memory_delta,
    MemoryType mem_type = MemoryType::DEVICE,
    std::optional<AllowOverbooking> allow_overbooking = std::nullopt
);
```

### DIFF
--- a/cpp/benchmarks/streaming/ndsh/utils.cpp
+++ b/cpp/benchmarks/streaming/ndsh/utils.cpp
@@ -110,10 +110,11 @@ streaming::Node consume_channel(
 coro::task<streaming::TableChunk> to_device(
     std::shared_ptr<streaming::Context> ctx, streaming::TableChunk&& chunk
 ) {
-    auto const net_memory_delta = static_cast<std::int64_t>(chunk.make_available_cost());
     co_return chunk.make_available(
         co_await streaming::reserve_memory(
-            ctx, chunk.make_available_cost(), net_memory_delta
+            ctx,
+            chunk.make_available_cost(),
+            streaming::MemoryReserveOrWait::missing_net_memory_delta
         )
     );
 }
--- a/cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
+++ b/cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
@@ -33,6 +33,19 @@ class Context;
 class MemoryReserveOrWait {
   public:
     /**
+     * @brief Sentinel indicating that `net_memory_delta` estimation has not yet been
+     * implemented.
+     *
+     * This value is used when a reasonable estimate of the net memory delta is not
+     * yet available. Any use of this sentinel should be treated as a TODO, since
+     * providing a concrete estimate enables better spilling and scheduling
+     * decisions.
+     *
+     * @see reserve_or_wait()
+     */
+    static constexpr std::int64_t missing_net_memory_delta = 0;
+
+    /**
      * @brief Constructs a `MemoryReserveOrWait` instance.
      *
      * If no reservation request can be satisfied within the timeout specified by


cpp/benchmarks/streaming/ndsh/utils.hpp
```diff
@@ -79,20 +79,27 @@ namespace detail {
...
+ * @param chunk Chunk to ensure is available on device. The input chunk is consumed
+ * and left in a moved-from state.
+ * @return A new `TableChunk` that is available on device.
  *
- * @return New `TableChunk` on device
- * @throws std::overflow_error if overbooking is not allowed and not enough memory is
- * available to reserve.
+ * @throws std::runtime_error If shutdown occurs before the reservation can be processed.
+ * @throws std::overflow_error If no progress is possible within the timeout and
+ * overbooking is disabled.
```
pentschev: Again here, similar to a discussion on a Python PR from a few days ago, [`overflow_error`](https://en.cppreference.com/w/cpp/error/overflow_error.html) seem like an odd choice, according to the docs it's purpose is:
> It can be used to report arithmetic overflow errors (that is, situations where a result of a computation is too large for the destination type).

I think the more appropriate choice from C++ standard exceptions would be [`bad_alloc`](https://en.cppreference.com/w/cpp/memory/new/bad_alloc), or we should consider adding some custom RapidsMPF exceptions to report these more "niche" errors.

madsbk:
Thanks @pentschev, will submit a _error_ PR shortly

PR #832, opened 2026-01-29
Move to_device coroutine to TableChunk

wence-:
This utility is generically useful outside of the benchmarks. For naming consistency, we rename it to `make_available`.


madsbk (approved):
Thanks @wence-

PR #834, opened 2026-01-30
Migrate bloom filter and related utilities into rapidsmpf proper

wence-:
So that we can start using this in cudf-polars, migrate the streaming bloom filter construction and application into the main library, rather than just having it in benchmarks.

To hide the cuco bloom filter implementation we introduce an `AlignedBuffer` object that represents an allocation of a certain size with a particular alignment. RMM's `device_buffer` unfortunately does not offer this, and to use the `device_vector<T>` would require us to either expose the cuco type in the API, or bake our own and then have strict aliasing violations.

Using a bloom filter is a two-step process. We construct a (cheaply copyable) `BloomFilter` type to track the number of filter blocks and the hash seed, and then do `build` to build a filter from a channel, followed by `apply` to apply the filter to a channel.

### DIFF
--- a/cpp/include/rapidsmpf/streaming/cudf/bloom_filter.hpp
+++ b/cpp/include/rapidsmpf/streaming/cudf/bloom_filter.hpp
@@ -56,7 +56,6 @@ struct BloomFilter {
      * @param ch_out Output channel receiving a single message containing the bloom
      * filter.
      * @param tag Disambiguating tag to combine filters across ranks.
-     *
      * @return Coroutine representing the construction of the bloom filter.
      */
     [[nodiscard]] Node build(
--- a/cpp/benchmarks/streaming/ndsh/q21.cpp
+++ b/cpp/benchmarks/streaming/ndsh/q21.cpp
@@ -92,7 +92,11 @@ rapidsmpf::streaming::Node read_nation(
         auto stream = ctx->br()->stream_pool().get_stream();
         auto owner = new std::vector<std::any>;
         constexpr auto name = "SAUDI ARABIA";
-        owner->push_back(std::make_shared<cudf::string_scalar>(name, true, stream));
+        owner->push_back(
+            std::make_shared<cudf::string_scalar>(
+                name, /* is_valid = */ true, stream, ctx->br()->device_mr()
+            )
+        );
         owner->push_back(
             std::make_shared<cudf::ast::literal>(
                 *std::any_cast<std::shared_ptr<cudf::string_scalar>>(owner->at(0))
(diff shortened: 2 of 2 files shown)


cpp/src/streaming/cudf/bloom_filter.cpp
```diff
@@ -0,0 +1,143 @@
...
+        cuda_stream_join(chunk_stream, stream, &event);
+        // Reservation for the mask construction and guess at output size.
+        auto res = co_await ctx_->memory(MemoryType::DEVICE)
+                       ->reserve_or_wait(
+                           static_cast<std::size_t>(chunk.table_view().num_rows())
+                                   // TODO: no magic numbers: the hashing algorithm in
+                                   // `contains` below returns an int64 column.
+                                   * (1 + sizeof(std::int64_t))
+                               // Guess at how selective the filter is.
+                               + chunk.data_alloc_size(MemoryType::DEVICE) / 2,
```
madsbk: Should this be configurable?

cpp/src/memory/aligned_buffer.cpp
```diff
@@ -0,0 +1,72 @@
...
+    return data_;
+};
+
+std::size_t AlignedBuffer::size() const noexcept {
+    return size_;
+};
+
+rmm::device_async_resource_ref AlignedBuffer::mr() noexcept {
+    return mr_;
+};
```
madsbk:
```suggestion
}
```

cpp/include/rapidsmpf/streaming/cudf/bloom_filter.hpp
```diff
@@ -0,0 +1,79 @@
...
+        : ctx_{std::move(ctx)}, seed_{seed}, num_filter_blocks_{num_filter_blocks} {}
+
+    /**
+     * @brief Build a bloom filter from the input channel.
+     *
+     * @param ch_in Input channel of `TableChunk`s to build bloom filter for.
+     * @param ch_out Output channel receiving a single message containing the bloom
+     * filter.
+     * @param tag Disambiguating tag to combine filters across ranks.
+     *
```
madsbk:
```suggestion
```

cpp/src/streaming/cudf/bloom_filter.cpp
```diff
@@ -0,0 +1,143 @@
...
+        if (msg.empty()) {
+            break;
+        }
+        auto chunk = msg.release<TableChunk>();
+        chunk = co_await chunk.make_available(
+            ctx_, -static_cast<std::int64_t>(chunk.data_alloc_size(MemoryType::DEVICE))
+        );
+        build_event.stream_wait(chunk.stream());
+        filter->add(chunk.table_view(), chunk.stream(), mr);
+        cuda_stream_join(stream, chunk.stream(), &event);
```
madsbk: Add a comment on the cuda stream sync strategic.

madsbk (commented):
It would be good with a test of `BloomFilter` as well.

cpp/src/memory/aligned_buffer.cpp
```diff
@@ -0,0 +1,72 @@
...
+      data_{std::exchange(other.data_, nullptr)} {}
+
+AlignedBuffer& AlignedBuffer::operator=(AlignedBuffer&& other) {
+    if (this != &other) {
+        RAPIDSMPF_EXPECTS(
+            !data_,
+            "cannot move into an already initialized aligned_buffer",
+            std::invalid_argument
+        );
+    }
```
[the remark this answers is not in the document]
wence-: Dunno, but I moved the setting inside the condition.

madsbk (approved):
Looks good, thanks @wence-

PR #840, opened 2026-02-02
Removing packed_columns from TableChunk

nirandaperera:
This PR removes the `packed_columns` class member from `TableChunk`. 
Now, if a `PackedData` is provided with `DEVICE Buffer`, it will be trivially unpacked and made available.

### DIFF
--- a/cpp/tests/streaming/test_table_chunk.cpp
+++ b/cpp/tests/streaming/test_table_chunk.cpp
@@ -216,6 +216,15 @@ TEST_P(StreamingTableChunk, FromPackedDataOn) {
     EXPECT_TRUE(chunk.is_spillable());
     EXPECT_THROW((void)chunk.table_view(), std::invalid_argument);
     EXPECT_EQ(chunk.make_available_cost(), size);
+
+    auto chunk2 = chunk.make_available(
+        br->reserve_or_fail(chunk.make_available_cost(), MemoryType::DEVICE)
+    );
+    EXPECT_FALSE(chunk.is_available());
+    EXPECT_TRUE(chunk2.is_available());
+    EXPECT_TRUE(chunk2.is_spillable());
+    EXPECT_EQ(chunk2.make_available_cost(), 0);
+    CUDF_TEST_EXPECT_TABLES_EQUIVALENT(chunk2.table_view(), expect);
 }
 
 TEST_F(StreamingTableChunk, DeviceToDeviceCopy) {


madsbk:
> This PR removes the `packed_columns` class member from `TableChunk`. Now, if a `PackedData` is provided with `DEVICE Buffer`, it will be trivially unpacked and made available.

But this also means that spilling now uses `cudf::pack()` instead of a simple `cudaMemcpyAsync()`. Are we sure this extra overhead is neglectable?

wence-:
> > This PR removes the `packed_columns` class member from `TableChunk`. Now, if a `PackedData` is provided with `DEVICE Buffer`, it will be trivially unpacked and made available.
> 
> But this also means that spilling now uses `cudf::pack()` instead of a simple `cudaMemcpyAsync()`. Are we sure this extra overhead is neglectable?

I don't understand why that would be required. If the backing store is `PackedData` surely we can just memcpy the buffer. After all `PackedData` is just `cudf::packed_columns` in a trenchcoat.

cpp/include/rapidsmpf/memory/buffer.hpp
```diff
@@ -45,6 +45,13 @@ namespace rapidsmpf {
  */
 class Buffer {
     friend class BufferResource;
+    friend void buffer_copy(
+        Buffer& dst,
+        Buffer const& src,
+        std::size_t size,
+        std::ptrdiff_t dst_offset,
+        std::ptrdiff_t src_offset
+    );
```
wence-: This is an antipattern we should try and avoid. I see you do this because you want `buffer_copy` to have internal access to `Buffer` (so that you can access `latest_write_event_`). But that is reaching deep into the internals and could easily break.

Please instead propose an API that captures the _semantics_ of what you want to do.

cpp/src/streaming/cudf/table_chunk.cpp
```diff
@@ -152,67 +166,36 @@ TableChunk TableChunk::copy(MemoryReservation& reservation) const {
...
+                // packed_columns and then we move the packed_columns' gpu_data to a new
+                // host buffer.
+
+                // TODO: use `cudf::chunked_pack()` with a bounce buffer. Currently,
+                // `cudf::pack()` allocates device memory we haven't reserved.
+                auto packed_columns = cudf::pack(table_view(), stream(), br->device_mr());
+                auto packed_data = std::make_unique<PackedData>(
+                    std::move(packed_columns.metadata),
+                    br->move(std::move(packed_columns.gpu_data), stream())
+                );
```
wence-: If we have `PackedData` then the data layout is already that of `cudf::packed_columns`, so we should just memcpy the buffer, and only fall back to `cudf::pack` in the case where we have a cudf::table as the backing store.

nirandaperera:
> > I don't understand why that would be required. If the backing store is `PackedData` surely we can just memcpy the buffer. After all `PackedData` is just `cudf::packed_columns` in a trenchcoat.
> 
> True, but what is the point of this PR then?

@madsbk Now there are only 2 storages, `cudf::table` or `PackedData`, rather than 3. If the device data is in `PackedData` and copy reservation is host/pinned, it will use `buffer_copy`. If device data is in cudf::table, it'll use pack.

madsbk (commented):
Overall looks good, thanks @nirandaperera

madsbk (changes requested):

wence- (approved):

PR #843, opened 2026-02-03
Add cudf table packing utilities with memory type support

nirandaperera:
Introduces `pack.hpp` and `pack.cpp` with utilities for packing cudf tables into contiguous buffers with explicit memory type control. This enables packing directly to device, pinned host, or host memory using the rapidsmpf memory reservation system.

Depends on #840 

### Changes

**New files:**
- `cpp/include/rapidsmpf/integrations/cudf/pack.hpp`
- `cpp/src/integrations/cudf/pack.cpp`
- `cpp/tests/test_cudf_pack.cpp`

**API:**
| Function | Description |
|----------|-------------|
| `detail::pack<MemoryType>()` | Templated pack with compile-time memory type |
| `pack()` | Runtime dispatch based on reservation's memory type |
| `chunked_pack()` | Uses `cudf::chunked_pack` with a bounce buffer for memory-efficient packing to host |

**Refactored:**
- `TableChunk::copy()` now uses `detail::pack<>` for host memory destinations


cpp/src/integrations/cudf/pack.cpp
```diff
@@ -0,0 +1,266 @@
...
+                    cudaMemcpyDefault,
+                    stream
+                ));
+            });
+        }
+
+        return std::make_unique<PackedData>(
+            std::move(packed_columns.metadata), std::move(dest_buf)
+        );
+    }
```
wence-: Is this extra complexity worth it? Why can't we just go down the chunked_pack route below?

cpp/tests/test_cudf_pack.cpp
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// test chunked pack and unpack with device bounce buffer
+TEST_P(CudfPackTest, ChunkedPackAndUnpackDevice) {
+    cudf::table expect = random_table_with_index(seed, num_rows_, 0, 1000);
+
+    auto cudf_packed = cudf::pack(expect, stream_, br_->device_mr());
+    std::size_t packed_size = cudf_packed.gpu_data->size();
+
+    auto reservation = br_->reserve_or_fail(packed_size, mem_type_);
```
wence-: Again, what is `cudf::pack` doing here?

cpp/tests/test_cudf_pack.cpp
```diff
@@ -0,0 +1,208 @@
...
+    if (!rapidsmpf::is_pinned_memory_resources_supported()) {
+        GTEST_SKIP() << "Skipping test for non-pinned memory type";
+    }
+
+    cudf::table expect = random_table_with_index(seed, num_rows_, 0, 1000);
+
+    auto cudf_packed = cudf::pack(expect, stream_, br_->device_mr());
+    std::size_t packed_size = cudf_packed.gpu_data->size();
+
+    auto reservation = br_->reserve_or_fail(packed_size, mem_type_);
```
wence-: Again

cpp/tests/test_cudf_pack.cpp
```diff
@@ -0,0 +1,208 @@
...
+        }
+    );
+
+    cudf::table expect = random_table_with_index(seed, num_rows, 0, 1000);
+
+    // Get packed size using cudf::pack (we need device memory for this estimation).
+    auto cudf_packed = cudf::pack(expect, stream, br->device_mr());
+    std::size_t packed_size = cudf_packed.gpu_data->size();
+
+    auto reservation = br->reserve_or_fail(packed_size, MemoryType::HOST);
```
wence-: Why are we doing this?

cpp/src/integrations/cudf/pack.cpp
```diff
@@ -0,0 +1,266 @@
...
+        std::move(packed_columns.metadata), std::move(host_buffer)
+    );
+}
+
+constexpr size_t cudf_chunked_pack_min_buffer_size = 1024 * 1024;  // 1MB
+
+// Handle the case where packing allocates slightly more than the input size. This can
+// occur because cudf uses aligned allocations, which may exceed the requested size. To
+// accommodate this, we allow some wiggle room.
+void handle_packed_size_overflow(
```
wence-:
```suggestion
static void handle_packed_size_overflow(
```

cpp/src/integrations/cudf/pack.cpp
```diff
@@ -0,0 +1,266 @@
...
+    // packed table is now in rmm::device_buffer. We need to move it to a
+    // HostBuffer as Pinned memory allocations are only used with HostBuffers.
+
+    auto host_buffer = reservation.br()->move(std::move(packed_columns.gpu_data), stream);
+    return std::make_unique<PackedData>(
+        std::move(packed_columns.metadata), std::move(host_buffer)
+    );
+}
+
+constexpr size_t cudf_chunked_pack_min_buffer_size = 1024 * 1024;  // 1MB
```
wence-: Expose this as a `static constexpr` property of `cudf::chunked_pack`.


```suggestion
constexpr size_t cudf_chunked_pack_min_buffer_size = 1024 * 1024;
```

cpp/src/integrations/cudf/pack.cpp
```diff
@@ -0,0 +1,266 @@
...
+    cudf::table_view const& table,
+    rmm::cuda_stream_view stream,
+    MemoryReservation& reservation
+) {
+    RAPIDSMPF_EXPECTS(
+        reservation.mem_type() == MemoryType::PINNED_HOST,
+        "pack<PINNED_HOST> requires a pinned host memory reservation",
+        std::invalid_argument
+    );
+    // benchmarks shows that using cudf::pack with pinned mr is sufficient.
```
wence-: remind me: do these already exist in the repo?

cpp/src/integrations/cudf/pack.cpp
```diff
@@ -0,0 +1,266 @@
...
+    // try device memory first
+    auto [dev_res, dev_ob] =
+        br->reserve(MemoryType::DEVICE, est_size, AllowOverbooking::YES);
+
+    auto dev_avail = max_availble(dev_res.size(), dev_ob);
+    if (dev_avail >= cudf_chunked_pack_min_buffer_size) {
+        rmm::device_buffer bounce_buffer(dev_avail, stream, br->device_mr());
+        dev_res.clear();
+        return chunked_pack(table, stream, bounce_buffer, br->device_mr(), reservation);
+    }
```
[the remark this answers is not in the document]
wence-: I thought we had determined that an appropriate chunk size was 32MiB (or maybe 64MiB). So I think we should always just try and allocate that.

cpp/src/integrations/cudf/pack.cpp
wence-: Also, I'm not sure this is a good name

wence- (changes requested):
I am a bit concerned that the logic here is more complicated than it needs to be.

Do we really need special case code for tables less than 1MiB in size, for example?

cpp/include/rapidsmpf/memory/memory_type.hpp
```diff
@@ -66,6 +68,53 @@ static_assert(std::ranges::equal(
...
+ *
+ * @return A span of memory types that are device accessible.
+ */
+constexpr std::span<MemoryType const> device_accessible_memory_types() noexcept {
+    return std::span{MEMORY_TYPES}.first<2>();
+}
+
+static_assert(std::ranges::equal(
+    device_accessible_memory_types(),
+    std::array{MemoryType::DEVICE, MemoryType::PINNED_HOST}
```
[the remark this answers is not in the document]
nirandaperera: I wanted to make sure that if we meddle with `MEMORY_TYPES` array in the future, we fail statically if somethings not right.

cpp/src/integrations/cudf/pack.cpp
```diff
@@ -0,0 +1,266 @@
...
+    cudf::table_view const& table,
+    rmm::cuda_stream_view stream,
+    MemoryReservation& reservation
+) {
+    RAPIDSMPF_EXPECTS(
+        reservation.mem_type() == MemoryType::DEVICE,
+        "pack<DEVICE> requires a device memory reservation",
+        std::invalid_argument
+    );
+    auto packed_columns = cudf::pack(table, stream, reservation.br()->device_mr());
```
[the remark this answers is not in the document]
nirandaperera: The issue would be, `mr()` could be one of the following,
 ```c++
    rmm::device_async_resource_ref device_mr_;
    std::shared_ptr<PinnedMemoryResource> pinned_mr_;
    HostMemoryResource host_mr_;
```
These dont have a common base class. So, we might have to use a variant. I personally think this is an overkill. 

Besides, callers pass the reservation. So, throwing an invalid arg exception in `pack<device>` if `reservation` is not device is sufficient IMO.

PR #844, opened 2026-02-05
Add hybrid Slurm support to rrun with PMIx-based coordination

pentschev:
This PR adds hybrid Slurm support for `rrun`, enabling RapidsMPF to run without MPI. This is achieved by reusing `SlurmBackend` that provides a passthrough mode.

The new hybrid Slurm execution model allows running one task per node, multiple GPUs per task with parent-mediated coordination:
* Root parent launches rank 0 first to obtain UCXX address, then broadcasts via PMIx to parents in all nodes
* `rrun` parents coordinate via PMIx then spawn non-root ranks with the root rank address

The hybrid execution mode requires specifying the number of tasks/ranks (`-n`) directly to `rrun` so it knows it should run in hybrid mode and know how many tasks/ranks per node it should launch.

Usage example:

```bash
  srun \
      --mpi=pmix \
      --nodes=2 \
      --ntasks-per-node=1 \
      --cpus-per-task=144 \
      --gpus-per-task=4 \
      --gres=gpu:4 \
      rrun -n 4 ./benchmarks/bench_shuffle -C ucxx
```

Requires https://github.com/rapidsai/rapidsmpf/pull/775


cpp/tools/rrun.cpp
```diff
@@ -1010,14 +1318,18 @@ int launch_ranks_fork_based(
         if (WIFEXITED(status)) {
             int code = WEXITSTATUS(status);
             if (code != 0) {
-                std::cerr << "[rrun] Rank " << (static_cast<std::size_t>(rank_offset) + i)
+                std::cerr << "[rrun] Rank "
+                          << (static_cast<size_t>(rank_offset)
```
madsbk:
```suggestion
                          << (static_cast<std::size_t>(rank_offset)
```

cpp/tools/rrun.cpp
```diff
@@ -950,15 +1280,21 @@ int launch_ranks_fork_based(
...
 
-    // Launch ranks
-    for (int local_rank = 0; local_rank < ranks_per_task; ++local_rank) {
+    // Include pre-launched process first (e.g. rank 0 in Slurm hybrid).
+    if (pre_launched_process) {
+        processes.insert(processes.begin(), std::move(*pre_launched_process));
+    }
+
+    // Launch remaining ranks (skip rank 0 if we already have it in pre_launched_process)
+    int start_local_rank = pre_launched_process.has_value() ? 1 : 0;
```
madsbk: `start_local_rank` is computed after `std::move(*pre_launched_process)`. This works because moving the value does not _empty_ the optional, but  this confused me :) 
Consider computing `start_local_rank` before the `if (pre_launched_process)` block to make the intent clear.

cpp/tools/rrun.cpp
```diff
@@ -783,17 +1078,14 @@ int launch_ranks_fork_based(
...
-    forwarders.reserve(static_cast<std::size_t>(ranks_per_task) * 2);
 
-    // Helper to start a forwarder thread for a given fd
-    auto start_forwarder = [&](int fd, int rank, bool to_stderr) {
-        if (fd < 0) {
-            return;
-        }
-        forwarders.emplace_back([fd, rank, to_stderr, &cfg, suppress_output]() {
+    // Helper to create a forwarder thread for a given fd (returns default thread if fd <
+    // 0).
```
madsbk:
```suggestion
    // Helper to create a forwarder thread for a given fd (returns default thread 
    // if fd < 0).
```

cpp/tools/rrun.cpp
```diff
@@ -783,17 +1078,14 @@ int launch_ranks_fork_based(
...
+    auto make_forwarder = [&](int fd, int rank, bool to_stderr) -> std::thread {
+        if (fd < 0)
+            return {};
```
madsbk:
```suggestion
        if (fd < 0) {
            return {};
        }
```

cpp/tools/rrun.cpp
```diff
@@ -843,21 +1133,24 @@ int launch_ranks_fork_based(
...
+            pid,
+            make_forwarder(fd_out, global_rank, false),
+            make_forwarder(fd_err, global_rank, true)
+        );
     }
 
-    // Start a signal-waiting thread to forward signals.
-    std::thread([signal_set, &pids, suppress_output]() mutable {
+    // Forward signals to all processes (including pre-launched).
+    std::thread([signal_set, &processes, suppress_output]() mutable {
```
madsbk: Might be safer to copy here?
```suggestion
    std::thread([signal_set, processes, suppress_output]() mutable {
```

cpp/tools/rrun.cpp
```diff
@@ -683,6 +716,255 @@ int setup_launch_and_cleanup(
...
+
+    if (is_root_parent) {
+        if (cfg.verbose) {
+            std::cout << "[rrun] Publishing root address via PMIx (hex-encoded, "
+                      << encoded_address.size() << " chars)" << std::endl;
+        }
+        backend->put("rapidsmpf_root_address", encoded_address);
+    }
+
+    backend->sync();
```
madsbk: If think this will hang if rank0 return early because of a `root address file was empty` error?

cpp/tools/rrun.cpp
```diff
@@ -683,6 +716,255 @@ int setup_launch_and_cleanup(
...
+            backend->get("rapidsmpf_root_address", std::chrono::seconds{30});
+        if (cfg.verbose) {
+            std::cout << "[rrun] Retrieved root address via PMIx (hex-encoded, "
+                      << encoded_address.size() << " chars)" << std::endl;
+        }
+    }
+
+    // Write into the local FileBackend kv store so children can get() it.
+    // The raw address from UCXX is hex-encoded, but the FileBackend stores
+    // opaque bytes.  The UCXX bootstrap Path 3 stores the *raw* address via
```
madsbk: What are these _paths_ referring to? Is there a overview somewhere?

cpp/src/bootstrap/ucxx.cpp
```diff
@@ -35,22 +55,57 @@ std::shared_ptr<ucxx::UCXX> create_ucxx_comm(
...
+            addr_ofs << encoded_address << std::endl;
+            addr_ofs.close();
+            if (std::rename(temp_path.c_str(), address_file->c_str()) != 0) {
+                std::remove(temp_path.c_str());
+                throw std::runtime_error(
+                    "Failed to rename root address file to: " + *address_file
+                );
+            }
+
+            auto verbose = getenv_optional("RAPIDSMPF_VERBOSE");
```
madsbk: Is `RAPIDSMPF_VERBOSE` a new option?

madsbk (changes requested):
I’m still a bit concerned that we’re adding quite a bit of complexity to accommodate a misconfigured system with incorrect hardware bindings. It might be more appropriate to require users to fix the underlying configuration instead 🙂

That said, if we want to support this many modes/paths, it would really help maintainability to separate them into distinct files. This way, you only have to follow a single code path when debugging a specific mode.

cpp/src/bootstrap/ucxx.cpp
pentschev: No, my mistake. Removed.

cpp/tools/rrun.cpp
pentschev: It shouldn't, `sync()` means "make the address available" (it's not a barrier). If something goes wrong, there's a 30 second timeout just below to prevent exactly this case.

cpp/tools/rrun.cpp
pentschev: Oh sorry, this is outdated, the different paths were reworked and simplified, we use to have 4 decision paths but they were simplified in https://github.com/rapidsai/rapidsmpf/pull/844/commits/c2122de76cdc56faf3695b74677136cf468f2ebe.

cpp/tools/rrun.cpp
pentschev: No, this wouldn't work. `processes` is a `std::vector<LaunchedProcesses>` that contains `std::thread` members that are move-only.

pentschev (commented):
> I’m still a bit concerned that we’re adding quite a bit of complexity to accommodate a misconfigured system with incorrect hardware bindings.

While I agree having misconfigured system is not ideal, it's a reality we need to be able to deal with. In any case, to clarify, there are more uses than misconfigured systems. Other uses include:
- Interactive sessions: a multi task Slurm session is not suitable/very difficult to run interactively.
- Generating profiles: with a single task we're able to generate multi-process/multi-GPU profiles instead of an individual profile per process/GPU that need to be stitched together.

> It might be more appropriate to require users to fix the underlying configuration instead 🙂

Based on my own experience as a user and the experience helping others getting setup properly, I strongly disagree here. It's not always easy to know what exactly is required for the user to do for proper configuration, nor if user-mode changes suffice to fix a given issue. Even for ourselves who have in-depth knowledge of a lot of the parts in those systems, many of those issues are not obvious and often take days to identify and resolve. Having an alternative path, even if it is just for a proof-of-concept of what the application is expected to look like is fundamental.

> That said, if we want to support this many modes/paths, it would really help maintainability to separate them into distinct files. This way, you only have to follow a single code path when debugging a specific mode.

That is a very fair point, I have attempted previously to move as much of the complexity into `SlurmBackend`, but obviously that's not enough. While I can't think of a "perfect modular" organization of files, I have made an attempt in https://github.com/rapidsai/rapidsmpf/pull/844/commits/89b60970317490ac6bd4d9689c8c8973a0322b0d to give each part its own file, which hopefully makes the code considerably more understandable than a monolithic implementation. Please let me know if you think this is sufficient or if there's more we need.

madsbk (approved):
Thanks @pentschev, LGTM.

pentschev:
Thanks both for the reviews!

PR #850, opened 2026-02-06
refactor: build wheels and conda packages using Python limited API

gforsyth:
General testing done in #773 

Opening a new PR to not disrupt any ongoing downstream testing using artifacts from the original draft PRs.

Resolves #773 

xref https://github.com/rapidsai/build-planning/issues/42

Ops-Bot-Merge-Barrier: true

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -53,7 +53,6 @@ jobs:
       date: ${{ inputs.date }}
       script: ci/build_python.sh
       sha: ${{ inputs.sha }}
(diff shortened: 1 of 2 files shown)


wence-:
OK, it is currently not possible to build rapidsmpf with the limited API due to lifetime issues of nested closures in cython when using the limited API. A "fix" is slated to be applied in cython 3.3. But we can also probably work around by not using nested closures, I will try that.

gforsyth:
Cherry-picked @wence- commit removing the nested closures in #878

wence- (approved):
Is there a plan for this to be the default (and hence developers will also build limited API by default)?

PR #853, opened 2026-02-10
[Python] Add ``OrderScheme`` partitioning metadata

rjzamora:
Depends on https://github.com/rapidsai/rapidsmpf/pull/993 (adds the C++ API)
Adds mechanism to track ordering within ``ChannelMetadata``. Needed in cudf-polars to keep track of sorted data.

### Proposed API

```c++
@dataclass(frozen=True, slots=True)
class OrderKey:
    """Sort key: column index, direction, and null ordering (``pylibcudf`` enums)."""

    column_index: int
    order: plc.types.Order
    null_order: plc.types.NullOrder

class OrderScheme:
    def __init__(
        self,
        keys: Sequence[OrderKey],
        boundaries: TableChunk,
        *,
        strict_boundaries: bool = False,
    ) -> None: ...
    @property
    def keys(self) -> tuple[OrderKey, ...]: ...
    @property
    def strict_boundaries(self) -> bool: ...
    @property
    def num_boundaries(self) -> int: ...
    def replace_keys(self, new_keys: Sequence[OrderKey]) -> OrderScheme: ...
    def boundaries_aligned_with(self, other: OrderScheme) -> bool: ...
    # Shallow equality: keys + strict_boundaries + boundary shape only.
    def __eq__(self, other: object) -> bool: ...
    def __repr__(self) -> str: ...
```


rjzamora:
@TomAugspurger - I'd be interested to get your thoughts on the specific contract proposed here.

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -31,6 +33,38 @@ struct HashScheme {
...
+ * - Partition i (0 < i < N-1): boundaries[i-1] <= values < boundaries[i]
+ * - Partition N-1: values >= boundaries[N-2]
+ *
+ * Comparisons use the specified column orders and null ordering.
+ */
+struct OrderScheme {
+    std::vector<cudf::size_type> column_indices;  ///< Column indices to order on.
+    std::vector<cudf::order> orders;  ///< Sort order per column (ASCENDING/DESCENDING).
+    std::vector<cudf::null_order>
+        null_orders;  ///< Null ordering per column (BEFORE/AFTER).
```
TomAugspurger: These all need to be the same length. Does rapidsmpf have a convention for how to store these (cc @madsbk)? As three separate vectors with a promise that they're the same length? Or would it be better to have a `vector<OrderStruct>` where that `OrderStruct` is a tuple with`column_indices, orders, null_orders`.

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -31,6 +33,38 @@ struct HashScheme {
...
+/**
+ * @brief Order-based partitioning scheme for sorted/range-partitioned data.
+ *
+ * Data is partitioned by value ranges based on predetermined boundaries.
+ * For N partitions, there are N-1 boundary rows:
+ * - Partition 0: values < boundaries[0]
+ * - Partition i (0 < i < N-1): boundaries[i-1] <= values < boundaries[i]
+ * - Partition N-1: values >= boundaries[N-2]
+ *
+ * Comparisons use the specified column orders and null ordering.
```
TomAugspurger: Requesting some more docs:

1. What's the behavior for sorting by multiple columns: presumably lexicographic?
2. Are there any restrictions on data types here?

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -31,6 +33,38 @@ struct HashScheme {
...
+        null_orders;  ///< Null ordering per column (BEFORE/AFTER).
+    std::shared_ptr<TableChunk> boundaries;  ///< N-1 boundary rows for N partitions.
+    bool strict_boundary{false};  ///< Sort keys disjoint across chunks.
+
+    /**
+     * @brief Equality comparison.
+     *
+     * @note Two OrderSchemes are equal if they have the same column indices,
+     * orders, null_orders, strict_boundary flag, and boundary values. Boundary
+     * comparison currently uses table shape only (full content comparison TBD).
```
TomAugspurger: Looks like a todo here. Solving that here, or file a separate issue to track it?

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -9,6 +9,42 @@
...
+    if (this_view.num_rows() != other_view.num_rows()
+        || this_view.num_columns() != other_view.num_columns())
+    {
+        return false;
+    }
+
+    // Note: Full content comparison would require device-side comparison.
+    // For now, we consider tables with same dimensions as potentially equal.
+    // A more complete implementation would use cudf utilities for comparison.
+    return true;
```
TomAugspurger: This seems thorny.

Do we want to avoid overloading `==` here entirely? I suspect that if we do ever need to compare two OrderScheme structs in a way that includes a comparison of the boundary values, then we'll need an API that provides a `stream` and `mr`.

cpp/tests/streaming/test_channel_metadata.cpp
```diff
@@ -129,3 +226,58 @@ TEST_F(StreamingChannelMetadata, MessageRoundTrip) {
...
+    EXPECT_TRUE(msg_m.holds<ChannelMetadata>());
+    auto released = msg_m.release<ChannelMetadata>();
+    EXPECT_EQ(released.local_count, 8);
+    EXPECT_TRUE(released.duplicated);
+    EXPECT_EQ(released.partitioning.inter_rank.type, PartitioningSpec::Type::ORDER);
+    EXPECT_EQ(released.partitioning.inter_rank.order->column_indices.size(), 2);
+    EXPECT_EQ(released.partitioning.inter_rank.order->column_indices[0], 0);
+    EXPECT_EQ(released.partitioning.inter_rank.order->column_indices[1], 1);
+    EXPECT_EQ(released.partitioning.inter_rank.order->orders[0], cudf::order::ASCENDING);
+    EXPECT_EQ(released.partitioning.inter_rank.order->orders[1], cudf::order::DESCENDING);
```
TomAugspurger: Are these assert duplicates of some of the ones from previous tests? Or is there something about going through `ChannelMetadata` that uses a different code path?

cpp/tests/streaming/test_channel_metadata.cpp
```diff
@@ -129,3 +226,58 @@ TEST_F(StreamingChannelMetadata, MessageRoundTrip) {
...
+}
+
+TEST_F(StreamingChannelMetadata, MessageRoundTripWithOrderSchemeStrictBoundary) {
+    OrderScheme os{
+        {0, 1},
+        {cudf::order::ASCENDING, cudf::order::DESCENDING},
+        {cudf::null_order::BEFORE, cudf::null_order::AFTER},
+        nullptr
+    };
+    os.strict_boundary = true;
```
TomAugspurger: Why do this here and not when building the struct a line above?

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyi
```diff
@@ -17,7 +20,31 @@ class HashScheme:
     def __eq__(self, other: object) -> bool: ...
     def __repr__(self) -> str: ...
 
-PartitioningSpecValue = HashScheme | None | Literal["inherit"]
+class OrderScheme:
+    def __init__(
+        self,
+        column_indices: tuple[int, ...],
+        orders: tuple[Literal["ascending", "descending"], ...],
```
TomAugspurger: `plc.types.Order`

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyi
```diff
@@ -17,7 +20,31 @@ class HashScheme:
...
+        null_orders: tuple[Literal["first", "last"], ...],
```
TomAugspurger: `plc.types.NullOrder`

(though the enum members in `NullOrder` are `AFTER` and `BEFORE`  so make sure they're used the same).

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyi
```diff
@@ -17,7 +20,31 @@ class HashScheme:
     def __eq__(self, other: object) -> bool: ...
     def __repr__(self) -> str: ...
 
-PartitioningSpecValue = HashScheme | None | Literal["inherit"]
+class OrderScheme:
+    def __init__(
+        self,
+        column_indices: tuple[int, ...],
```
TomAugspurger: A tuple, or a list? Or maybe `collections.abc.Sequence`?

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyi
TomAugspurger: And you'll need to update the return type on the properties.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyi
```diff
@@ -17,7 +20,31 @@ class HashScheme:
...
+        strict_boundary: bool = False,
+    ) -> None: ...
+    @property
+    def column_indices(self) -> tuple[int, ...]: ...
+    @property
+    def orders(self) -> tuple[Literal["ascending", "descending"], ...]: ...
+    @property
+    def null_orders(self) -> tuple[Literal["first", "last"], ...]: ...
+    @property
+    def has_boundaries(self) -> bool: ...
```
TomAugspurger: Do we need this? Is it just `order_scheme.boundaries is not None`?

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,6 +102,148 @@ cdef class HashScheme:
...
+        tuple null_orders,
+        TableChunk boundaries = None,
+        *,
+        bint strict_boundary = False,
+    ):
+        cdef vector[int32_t] cols
+        cdef vector[cpp_order] ords
+        cdef vector[cpp_null_order] nulls
+
+        if len(column_indices) != len(orders):
```
TomAugspurger: Shouldn't this be done in C++?

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,6 +102,148 @@ cdef class HashScheme:
         return f"HashScheme({self.column_indices!r}, {self.modulus})"
 
 
+cdef cpp_order _str_to_order(str s) except *:
```
TomAugspurger: If you change `orders` and `null_orders` to use the enums these can all be removed.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyi
```diff
@@ -17,7 +20,31 @@ class HashScheme:
...
+    @property
+    def strict_boundary(self) -> bool: ...
+    def get_boundaries_table(self) -> plc.Table | None: ...
```
TomAugspurger: And just confirming: do you anticipate needing to check the boundaries table from Python? It's nice to be able to get, but the implementation is complicated enough that it's worth checking whether we'll actually need it.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -2,13 +2,56 @@
 # SPDX-License-Identifier: Apache-2.0
 """Channel metadata types for streaming pipelines."""
 
+from cuda.bindings.cyruntime cimport cudaStream_t
```
TomAugspurger: I don't see this import elsewhere. Do we need it?

In https://github.com/rapidsai/rapidsmpf/blob/c7ecfe59678b27cc7e991b8ed9bff25d3249c483/python/rapidsmpf/rapidsmpf/streaming/core/context.pyx#L14, we use `Stream._from_cudaStream_t`. But I'm not sure about the type.

TomAugspurger (commented):
High-level, I think this makes sense as a way to support what we need out of cudf-polars. Left some comments and questions on the implementation.

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
rjzamora: Okay - Added more info to the docstring

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
rjzamora: I updated this comment a bit. I think its "okay" for us to do a "shallow" equality and document that this is the case. We can implement a separate API for strict equality later (or just let cudf-polars do this pylibcudf).

cpp/src/streaming/cudf/channel_metadata.cpp
rjzamora: Agree that this is thorny. I don't think we should ever require `==` to enforce "deep" equality here. We need a separate API for that.

cpp/tests/streaming/test_channel_metadata.cpp
rjzamora: Yeah, I'm not sure it's important, but I was thinking we need to check that the attributes are correct after round-tripping through `to_message -> release`.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyi
rjzamora: Yeah, I'm thinking we will want a sort-based join implementation in cudf-polars that will need to use the boundaries of the "sorted" join side to sort the other side of the join. We may also use this information to coordinate boundary-exchange for window-based operations.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
rjzamora: In C++, `Partitioning` / `PartitioningSpec` wraps `HashScheme` or `OrderScheme`. `OrderScheme` is intentionally a plain struct, so we don’t enforce parallel-vector length in a C++ constructor. However, we **do** validate parallel-vector length in `PartitioningSpec::from_order`.

The problem is that Python can still build an `OrderScheme` with mismatched lengths before it is ever passed into `PartitioningSpec::from_order`. When that scheme is used to build a real `PartitioningSpec` object (e.g. when embedding it in `ChannelMetadata`), C++ will throw.  If we want only one place for validation, we could probably add a C++ `OrderScheme` factory and have the bindings call that.

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
madsbk: No, we don’t have anything formal, but I would prefer `vector<OrderStruct>`.

rjzamora: Okay, makes sense. I combined these into a single `std::vector<OrderKey> keys` to track everything in one place.

rjzamora:
**Update**:

I revised this PR a few times, and I think it's ready for review. I think I'd like the reviewed in two phases:

1. **Does the public-facing `OrderScheme` API makes sense?** 
2. **Do the internal details/implementation make sense?** It may make sense for someone with stronger C++ and/or Cython skills to take over this PR or replace it if that are many issues. My primary goal here was to prototype what I think we need in cudf-polars.

One open question I have about the internal implementation is: **Should the sort "boundaries" be tracked as a _spillable_ `unique_ptr<TableChunk>` internally?** I decided the answer is "yes" for now. However, since `ChannelMetadata` is typically copied when it is pushed into a Channel, this means we need to copy this underlying `TableChunk` often as it moves through the pipeline.

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -49,10 +117,12 @@ struct PartitioningSpec {
         NONE,  ///< No partitioning information at this level.
         INHERIT,  ///< Partitioning is inherited from parent level unchanged.
         HASH,  ///< Hash partitioning.
+        ORDER,  ///< Order/range partitioning.
     };
 
     Type type = Type::NONE;  ///< The type of partitioning.
     std::optional<HashScheme> hash;  ///< Valid only when type == HASH.
+    std::optional<OrderScheme> order;  ///< Valid only when type == ORDER.
```
wence-, line 156: *question*: Can one have both `hash` and `order` for the same partitioningspec?

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,19 +4,103 @@
...
 
+#include <rapidsmpf/memory/memory_type.hpp>
 #include <rapidsmpf/streaming/cudf/channel_metadata.hpp>
 
 namespace rapidsmpf::streaming {
 
+PartitioningSpec PartitioningSpec::from_order(OrderScheme o) {
+    if (o.keys.empty()) {
+        throw std::invalid_argument("OrderScheme: keys must not be empty");
+    }
```
wence-:
```suggestion
    RAPIDSMPF_EXPECTS(!o.keys.empty(), "OrderScheme: keys must not be empty", std::invalid_argument);
```

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,19 +4,103 @@
...
+    if (this_has != other_has) {
+        return false;
+    }
+    if (!this_has) {
+        return true;
+    }
+    // Both have boundaries: shape only (see doc in header).
+    auto tv = boundaries->table_view();
+    auto ov = other.boundaries->table_view();
+    return tv.num_rows() == ov.num_rows() && tv.num_columns() == ov.num_columns();
```
wence-: table view may not always be available, use the `shape` method on tablechunk.

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,19 +4,103 @@
...
+        .inter_rank = inter_rank.clone(reservation),
+        .local = local.clone(reservation),
+    };
+}
+
+ChannelMetadata ChannelMetadata::clone(MemoryReservation& reservation) const {
+    return ChannelMetadata{local_count, partitioning.clone(reservation), duplicated};
+}
+
+ContentDescription content_description_for(ChannelMetadata const& m) {
```
wence-: This should be called `get_content_description` to align with all other Message values.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pxd
```diff
@@ -68,12 +89,40 @@ cdef class HashScheme:
...
+    cdef cpp_OrderScheme* _view        # non-owning (used when non-NULL)
+    cdef object _owner                 # keepalive for non-owning mode
+
+    @staticmethod
+    cdef OrderScheme from_cpp(cpp_OrderScheme scheme)
+
+    @staticmethod
+    cdef OrderScheme view_of(cpp_OrderScheme* ptr, object owner)
+
+    cdef cpp_OrderScheme* _get(self)
```
wence-: No, this looks horrible, what is going on?

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pxd
```diff
@@ -68,12 +89,40 @@ cdef class HashScheme:
...
+    cdef object _owner                         # keepalive for non-owning mode
 
     @staticmethod
     cdef Partitioning from_handle(unique_ptr[cpp_Partitioning] handle)
 
+    @staticmethod
+    cdef Partitioning view_of(cpp_Partitioning* ptr, object owner)
+
+    cdef cpp_Partitioning* _get(self)
+
```
wence-, line 120: Why do we need this?

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyi
```diff
@@ -4,20 +4,62 @@
...
+        null_order: plc.types.NullOrder,
+    ) -> None: ...
+    @property
+    def column_index(self) -> int: ...
+    @property
+    def order(self) -> plc.types.Order: ...
+    @property
+    def null_order(self) -> plc.types.NullOrder: ...
+    def __eq__(self, other: object) -> bool: ...
+    def __repr__(self) -> str: ...
```
wence-: This could just be a `@dataclass` in the type system, I think.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -2,35 +2,182 @@
...
+            }
+            return rapidsmpf::streaming::PartitioningSpec::none();
+        };
+        rapidsmpf::streaming::Partitioning part{
+            clone_spec(p.inter_rank), clone_spec(p.local)
+        };
+        return std::make_unique<rapidsmpf::streaming::ChannelMetadata>(
+            local_count, std::move(part), duplicated
+        );
+    }
```
wence-: Why do we need all of this given the existence of all of the `clone` methods on the C++ objects?

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +206,232 @@ cdef class HashScheme:
...
+cdef cpp_order _object_to_cpp_order(object o) except *:
+    """Map pylibcudf.types.Order (int-valued enum) to libcudf order."""
+    cdef int v = int(o)
+    if v == <int>cpp_order.ASCENDING:
+        return cpp_order.ASCENDING
+    if v == <int>cpp_order.DESCENDING:
+        return cpp_order.DESCENDING
+    raise ValueError(
+        f"Invalid order: {o!r}; use pylibcudf.types.Order.ASCENDING or Order.DESCENDING"
+    )
```
wence-: What, no, just accept a cimported `pylibcudf.types.order` (or whatever it's called) and get this for free.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +206,232 @@ cdef class HashScheme:
...
+    def column_index(self) -> int:
+        return self._key.column_index
+
+    @property
+    def order(self):
+        return PyOrder(<int>self._key.order)
+
+    @property
+    def null_order(self):
+        return PyNullOrder(<int>self._key.null_order)
```
wence-: No, we should talk in pylibcudf types.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +206,232 @@ cdef class HashScheme:
...
+    are not compared.
+
+    Parameters
+    ----------
+    keys
+        Sequence of ``OrderKey`` objects (one per sort column).
+    boundaries
+        Optional ``TableChunk`` of N-1 boundary rows for N partitions.
+    strict_boundary
+        If True, sort keys do not span partition interiors. Default False.
```
wence-: This needs better explanation of the meaning. And I think should also be documented properly in the C++ object.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +206,232 @@ cdef class HashScheme:
...
+        *,
+        bint strict_boundary = False,
+    ):
+        cdef vector[cpp_OrderKey] cpp_keys
+        for key in keys:
+            if not isinstance(key, OrderKey):
+                raise TypeError(
+                    f"keys must contain OrderKey objects, got {type(key).__name__}"
+                )
+            cpp_keys.push_back((<OrderKey>key)._key)
```
wence-:
```suggestion
        cdef vector[cpp_OrderKey] cpp_keys = list(keys)
```

Should work, I think.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +206,232 @@ cdef class HashScheme:
...
+        if cpp_keys.empty():
+            raise ValueError("OrderScheme: keys must not be empty")
+        self._get().keys = move(cpp_keys)
+        if boundaries is not None:
+            self._get().boundaries = move(boundaries.release_handle())
+        self._get().strict_boundary = strict_boundary
```
wence-: Why can we not just make the C++ object in one fell swoop?

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -2,35 +2,182 @@
...
+    void _ensure_boundaries_available(
+        rapidsmpf::streaming::OrderScheme* s,
+        rapidsmpf::MemoryReservation& res
+    ) {
+        if (!s->boundaries->is_available()) {
+            s->boundaries = std::make_unique<rapidsmpf::streaming::TableChunk>(
+                s->boundaries->make_available(res)
+            );
+        }
+    }
```
wence-: Why do we need all of this machinery? Could we make it simpler by just requiring that that the boundaries table is always available. It should be tiny right?

wence- (changes requested):

wence-:
I don't think I understand how you are using this new OrderScheme object. Can you show some examples?

In particular, I don't understand what `strict_boundaries` means at all. I also think things can be significantly simplified by requiring that the `boundaries` table is _not_ spilled. It should be tiny (?) so not being able to spill it is not very problematic I think.

rjzamora:
>I don't understand what strict_boundaries means at all.

If we are sorting on `"name"`, it is technically possible for the same "name" to fall into multiple adjacent partitions if there are many members of that group. In that case, we wound't be able to do a "chunkwise" GroupBy operation on `"name"`. This `strict_boundaries` attribute just indicates whether we are sorted **and** have a unique group-to-chunk mapping.

>I also think things can be significantly simplified by requiring that the boundaries table is not spilled. It should be tiny (?) so not being able to spill it is not very problematic I think.

Yeah. This was my main question for reviewers (i.e. you). Boundaries may have thousands of rows, and we currently copy metadata somewhat frequently. This is probably small enough to stay in device memory, but I was afraid of *starting* with that assumption in case we discovered issues. I'm happy either way.

wence-:
> Boundaries may have thousands of rows, and we currently copy metadata somewhat frequently. This is probably small enough to stay in device memory, but I was afraid of starting with that assumption in case we discovered issues. I'm happy either way.

Can I see how you _use_ `boundaries`? So we can think about things.

rjzamora:
>Can I see how you use boundaries? So we can think about things.

I don't have any code that is using this yet, but this is essentially the [boundaries we would use to find the split points](https://github.com/rapidsai/cudf/blob/e81ad54ed957866c235cf0720fc3ef8986aa39b0/python/cudf_polars/cudf_polars/experimental/rapidsmpf/collectives/sort.py#L315) we call `insert_split` to insert each chunk into a sort-based shuffle. If we keep track of this information (along with the other ordering metadata in OrderScheme) we can "align" two sides of a join when one side is already sorted. We can can also use this information to implement MergeSorted pretty easily.

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
rjzamora: I suppose you could, but hash is ignored unless type is HASH and order is ignored unless type is ORDER. In the absence of Cython, I would never design it this way, but I remember we landed on this after I struggled with HashScheme.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
rjzamora: Okay, yeah. Agree we can start with boundaries as non-spillable and revisit if that ends up being a problem (though, I doubt it will be).

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pxd
rjzamora: So we didn’t really have to think hard about ownership with the `HASH` flavor of `Partitioning` (it’s small copy-friendly metadata). Now `ORDER` can carry a uniquely-owned `TableChunk` (`boundaries`). Exposing `ChannelMetadata.partitioning` as a cheap view means we have to deal with Python lifetime issues. This is how I got things “working”, but happy to revisit if we’d rather copy or revise the general approach.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
rjzamora: Cython build was failing when I tried this.

rjzamora:
@wence-  did some cleanup here. You mentioned offline that I should remove the Python bits or open a separate PR with the C++ pieces only. Let me know if you still think that's necessary and I'll probably do it tomorrow AM.

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -31,6 +36,63 @@ struct HashScheme {
...
+ *
+ * `keys[i]` is the i-th sort column; ordering is lexicographic by `keys[0]`,
+ * then `keys[1]`, and so on.
+ *
+ * When `boundaries` is set, its columns must align with `keys`
+ * (same count and compatible dtypes). Mismatched dtypes are a usage error.
+ *
+ * `strict_boundaries`: when true, every row in a chunk belongs to a single partition's
+ * half-open key range (partition keys do not straddle chunk interiors). When false,
+ * a chunk may contain keys spanning multiple partitions (e.g. before a shuffle).
```
wence-: I don't think this example is helpful.


```suggestion
 * a chunk may contain keys spanning multiple partitions.
```

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -31,6 +36,63 @@ struct HashScheme {
...
+    /// See struct-level note on `strict_boundaries` semantics.
+    bool strict_boundaries{false};
+
+    /**
+     * @brief Shallow metadata equality (not semantic boundary value equality).
+     *
+     * Returns true when `keys` and `strict_boundaries` match, and boundary tables
+     * are consistent in the weak sense: both absent, or both present with the
+     * same shape from `TableChunk::shape()` (not `table_view()`, which requires an
+     * available device table).
```
wence-:
```suggestion
     * same shape.
```

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,19 +4,73 @@
...
-        [](Message const& msg, MemoryReservation& /* reservation */) -> Message {
+        cd,
+        [](Message const& msg, MemoryReservation& /*reservation*/) -> Message {
             auto copy = std::make_unique<ChannelMetadata>(msg.get<ChannelMetadata>());
-            return Message{msg.sequence_number(), std::move(copy), {}, msg.copy_cb()};
+            auto copy_cd = get_content_description(*copy);
+            return Message{
+                msg.sequence_number(), std::move(copy), copy_cd, msg.copy_cb()
+            };
         }
```
wence-, line 117: The default ctor for contentdescription gives you an unspillable message, so:

Not applicable as is because I selected the wrong lines.
```suggestion
Message to_message(std::uint64_t sequence_number, std::unique_ptr<ChannelMetadata> m) {
return Message{
sequence_number,
std::move(m),
        {},
        [](Message const& msg, MemoryReservation& /*reservation*/) -> Message {
auto copy = std::make_unique<ChannelMetadata>(msg.get<ChannelMetadata>());
            return Message{
                msg.sequence_number(), std::move(copy), {}, msg.copy_cb()
            };
```

cpp/tests/streaming/test_channel_metadata.cpp
```diff
@@ -69,6 +159,27 @@ TEST_F(StreamingChannelMetadata, PartitioningScenarios) {
     EXPECT_EQ(p_twostage.inter_rank.hash->modulus, 4);
     EXPECT_EQ(p_twostage.local.hash->modulus, 8);
 
+    // Order-based partitioning (range partitioned / sorted)
+    Partitioning p_ordered{
+        PartitioningSpec::from_order(
+            OrderScheme{{{0, cudf::order::ASCENDING, cudf::null_order::AFTER}}, nullptr}
```
wence-: I think it should be invalid to construct an orderscheme with no boundary table. All the actual consumers of the orderscheme will need the boundaries table.

rjzamora: I was originally imagining the case that we know something is sorted (maybe Polars tells us?), but we don't know the boundaries yet. I think we can still use this information to do something "smart" during runtime. However, maybe that's a lowering-time optimization anyway.

wence-: Ah ok, via `set_sorted`.

rjzamora: Just a note that I will submit a PR to support `set_sorted`/`MapFunction(name="hint_sorted")` soon (probably once this PR is in). I think it makes sense to require a boundaries table, and so we need to collect these boundaries to define the `OrderScheme` metadata whenever there is a `MapFunction(name="hint_sorted")` in the plan.

My current sketch lowers `MapFunction(name="hint_sorted")` to a `hint_sorted` actor (via an intermediate `HintSorted` IR node) that separately consumes both the child of the `MapFunction` and the **projected** child (only the ordered-by columns; with the projection pushed into the `Scan` when the child is a parquet scan). In my sketches, this allows us to avoid buffering the data in memory when `set_sorted` is applied immediately after `scan_parquet`. We can potentially optimize this even further to use only parquet metadata when min/max statistics are available.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pxd
```diff
@@ -64,11 +92,25 @@ cdef class HashScheme:
...
 
+cdef class OrderKey:
+    cdef cpp_OrderKey _key
+
+    @staticmethod
+    cdef OrderKey from_cpp(cpp_OrderKey key)
+
+
+cdef class OrderScheme:
+    cdef shared_ptr[cpp_OrderScheme] _handle
```
wence-: I think, since an `OrderScheme` is copyable, we can have a `cdef cpp_OrderScheme _handle` here

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pxd
wence-: the tablechunk is `shared_ptr` now, so I think this is moot?

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyi
```diff
@@ -4,20 +4,56 @@
...
     def column_indices(self) -> tuple[int, ...]: ...
     @property
     def modulus(self) -> int: ...
     def __eq__(self, other: object) -> bool: ...
     def __repr__(self) -> str: ...
 
-PartitioningSpecValue = HashScheme | None | Literal["inherit"]
+@dataclass(frozen=True, slots=True)
+class OrderKey:
+    """Sort key: column index, direction, and null ordering (``pylibcudf`` enums)."""
```
wence-:
```suggestion
    """Sort key: column index, direction, and null ordering."""
```

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyi
```diff
@@ -4,20 +4,56 @@
...
+    @property
+    def strict_boundaries(self) -> bool: ...
+    @property
+    def num_boundaries(self) -> int: ...
+    def with_keys(self, new_keys: Sequence[OrderKey]) -> OrderScheme: ...
+    def boundaries_aligned_with(
+        self, other: OrderScheme, br: BufferResource
+    ) -> bool: ...
+    # Shallow equality: keys + strict_boundaries + boundary shape only.
+    def __eq__(self, other: object) -> bool: ...
```
wence-:
```suggestion
```

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +49,158 @@ cdef class HashScheme:
...
+        ret._key = key
+        return ret
+
+    @property
+    def column_index(self) -> int:
+        return self._key.column_index
+
+    @property
+    def order(self):
+        return plc_types.Order(<int>self._key.order)
```
wence-: This cast should be unnecessary, it should happen automatically, because `.order` is defined as the plc enum.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +49,158 @@ cdef class HashScheme:
...
+
+    @property
+    def null_order(self):
+        return plc_types.NullOrder(<int>self._key.null_order)
```
wence-: Likewise here.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
wence-: Oh, how about:


```suggestion
        cdef vector[cpp_OrderKey] cpp_keys
        for key in keys:
            cpp_keys.push_back((<OrderKey?>key)._key)
```

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -170,7 +319,7 @@ cdef class ChannelMetadata:
     def from_message(Message message not None):
         """Construct by consuming a Message (message becomes empty)."""
         return ChannelMetadata.from_handle(
-            cpp_channel_metadata_from_message(move(message._handle))
+            _channel_metadata_from_message(move(message._handle))
```
wence-: Why did this change?

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -193,9 +342,9 @@ cdef class ChannelMetadata:
 
     @property
     def partitioning(self) -> Partitioning:
-        cdef Partitioning ret = Partitioning.__new__(Partitioning)
-        ret._handle = make_unique[cpp_Partitioning](self.handle_ptr().partitioning)
-        return ret
+        if not self._handle:
+            raise ValueError("ChannelMetadata is uninitialized, has it been released?")
+        return Partitioning.from_cpp(self._handle.get().partitioning)
```
wence-: This diff makes no sense.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +49,158 @@ cdef class HashScheme:
...
+
+    def with_keys(self, object new_keys) -> OrderScheme:
+        """Return a new ``OrderScheme`` with updated key column indices."""
+        cdef vector[cpp_OrderKey] cpp_keys
+        for key in new_keys:
+            if not isinstance(key, OrderKey):
+                raise TypeError(
+                    f"keys must contain OrderKey objects, got {type(key).__name__}"
+                )
+            cpp_keys.push_back((<OrderKey>key)._key)
```
wence-:
```suggestion
            cpp_keys.push_back((<OrderKey?>key)._key)
```

wence- (changes requested):
I feel like the cython bindings want a bit of a tidy up now we've changed how the C++ works.

rjzamora:
>I feel like the cython bindings want a bit of a tidy up now we've changed how the C++ works.

Thanks for the quick review @wence- ! I addressed your suggestions, but let me know if you're looking for a larger overhaul here.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +49,148 @@ cdef class HashScheme:
...
+        ret._key = key
+        return ret
+
+    @property
+    def column_index(self) -> int:
+        return self._key.column_index
+
+    @property
+    def order(self):
+        return plc_types.Order(self._key.order)
```
wence-: I think this manual construction of a `Order` object is not required?

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +49,148 @@ cdef class HashScheme:
...
+
+    @property
+    def null_order(self):
+        return plc_types.NullOrder(self._key.null_order)
```
wence-: Same here.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pxd
```diff
@@ -20,14 +26,31 @@ cdef extern from "<rapidsmpf/streaming/cudf/channel_metadata.hpp>" \
         cpp_HashScheme(vector[int32_t], int) except +
         bool_t operator==(const cpp_HashScheme&)
 
+    cdef cppclass cpp_OrderKey "rapidsmpf::streaming::OrderKey":
+        int32_t column_index
```
wence-, line 32:
```suggestion
        cpp_OrderKey(int32_t, cpp_order, cpp_null_order) noexcept
        int32_t column_index
```

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pxd
```diff
@@ -20,14 +26,31 @@ cdef extern from "<rapidsmpf/streaming/cudf/channel_metadata.hpp>" \
...
+        cpp_order order
+        cpp_null_order null_order
+        bool_t operator==(const cpp_OrderKey&)
```
wence-:
```suggestion
        bool_t operator==(const cpp_OrderKey&) noexcept
```

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pxd
```diff
@@ -20,14 +26,31 @@ cdef extern from "<rapidsmpf/streaming/cudf/channel_metadata.hpp>" \
...
+
+    cdef cppclass cpp_OrderScheme "rapidsmpf::streaming::OrderScheme":
+        vector[cpp_OrderKey] keys
```
wence-, line 42:
```suggestion
        cpp_OrderScheme() noexcept
        cpp_OrderScheme(vector[cpp_orderKey], shared_ptr[TableChunk], bool) except +
        vector[cpp_OrderKey] keys
```

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +49,148 @@ cdef class HashScheme:
...
+
+    def __init__(
+        self,
+        int column_index,
+        cpp_order order,
+        cpp_null_order null_order,
+    ):
+        self._key.column_index = column_index
+        self._key.order = order
+        self._key.null_order = null_order
```
wence-:
```suggestion
        self._key = cpp_OrderKey(column_index, order, null_order)
```

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +49,148 @@ cdef class HashScheme:
...
+
+    @staticmethod
+    cdef OrderKey from_cpp(cpp_OrderKey key):
+        cdef OrderKey ret = OrderKey.__new__(OrderKey)
+        ret._key = key
+        return ret
+
+    @property
+    def column_index(self) -> int:
```
wence-, line 105: Please document (throughout)

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +49,148 @@ cdef class HashScheme:
...
+        bint strict_boundaries = False,
+    ):
+        cdef vector[cpp_OrderKey] cpp_keys
+        for key in keys:
+            cpp_keys.push_back((<OrderKey?>key)._key)
+        if cpp_keys.empty():
+            raise ValueError("OrderScheme: keys must not be empty")
+        self._handle = _make_order_scheme(
+            move(cpp_keys), move(boundaries.release_handle()), strict_boundaries
+        )
```
wence-:
```suggestion
        self._handle = cpp_OrderScheme(move(cpp_keys), move(boundaries.release_handle()), strict_boundaries)
```

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +49,148 @@ cdef class HashScheme:
...
+    @staticmethod
+    cdef OrderScheme from_cpp(cpp_OrderScheme scheme):
+        cdef OrderScheme ret = OrderScheme.__new__(OrderScheme)
+        ret._handle = move(scheme)
+        return ret
+
+    @property
+    def keys(self) -> tuple:
+        cdef int i
+        cdef int n = <int>self._handle.keys.size()
```
wence-: Don't think you need the explicit cast.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -92,38 +202,66 @@ cdef class Partitioning:
...
-            self._handle.reset()
+        _apply_spec(self._data.inter_rank, inter_rank)
+        _apply_spec(self._data.local, local)
 
     @staticmethod
-    cdef Partitioning from_handle(unique_ptr[cpp_Partitioning] handle):
+    cdef Partitioning from_cpp(cpp_Partitioning data):
         cdef Partitioning ret = Partitioning.__new__(Partitioning)
-        ret._handle = move(handle)
+        ret._data = move(data)
```
wence-: No, let us consistently call the wrapped C++ thing a `_handle`.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -92,38 +202,66 @@ cdef class Partitioning:
...
+    static rapidsmpf::streaming::OrderScheme
+    _make_order_scheme(
+        std::vector<rapidsmpf::streaming::OrderKey> keys,
+        std::unique_ptr<rapidsmpf::streaming::TableChunk> boundaries,
+        bool strict_boundaries
+    ) {
+        return rapidsmpf::streaming::OrderScheme{
+            std::move(keys), std::move(boundaries), strict_boundaries
+        };
+    }
```
wence-: This function doesn't need to exist with the additions I propose.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -7,30 +7,20 @@ from libc.stdint cimport int32_t, uint64_t
...
-        return std::make_unique<rapidsmpf::streaming::ChannelMetadata>(
-            msg.release<rapidsmpf::streaming::ChannelMetadata>()
-        );
-    }
-    }
-    """
-    unique_ptr[cpp_ChannelMetadata] cpp_channel_metadata_from_message(
-        cpp_Message
-    ) except +
+import pylibcudf.types as plc_types
```
wence-: Don't move this `cdef extern from` to the end of the file.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +49,148 @@ cdef class HashScheme:
...
+cdef class OrderKey:
+    """A single sort key: column index, sort direction, and null placement."""
+
+    def __init__(
+        self,
+        int column_index,
+        cpp_order order,
+        cpp_null_order null_order,
+    ):
+        self._key.column_index = column_index
```
wence-: Call the handle `_handle` for uniformity with all the wrapped types.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -59,28 +49,148 @@ cdef class HashScheme:
...
         raise TypeError(
-            f"Expected HashScheme, None, or 'inherit', got {type(obj).__name__}"
+            f"Expected HashScheme, OrderScheme, None, or 'inherit', "
+            f"got {type(obj).__name__}"
         )
 
 
-cdef object _from_spec(cpp_PartitioningSpec spec):
-    """Convert PartitioningSpec to Python object."""
+cdef object _from_spec(cpp_PartitioningSpec& spec):
```
wence-:
```suggestion
cdef object _from_spec(const cpp_PartitioningSpec& spec):
```

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -148,13 +286,14 @@ cdef class ChannelMetadata:
...
-        self._handle = make_unique[cpp_ChannelMetadata](
-            local_count, part, duplicated
-        )
+            self._handle = make_unique[cpp_ChannelMetadata](
+                local_count, (<Partitioning>partitioning)._data, duplicated
+            )
+        else:
+            self._handle = make_unique[cpp_ChannelMetadata](
+                local_count, cpp_Partitioning(), duplicated
+            )
```
wence-: This change was unnecessary.

python/rapidsmpf/rapidsmpf/tests/streaming/test_channel_metadata.py
```diff
@@ -23,7 +68,111 @@ def test_hash_scheme() -> None:
...
+
+
+def test_order_scheme(context: Context) -> None:
+    """Test OrderScheme construction, properties, equality, and repr."""
+    o1 = _two_key_order_scheme(context)
+    assert o1.keys == (
+        OrderKey(0, plc.types.Order.ASCENDING, plc.types.NullOrder.BEFORE),
+        OrderKey(1, plc.types.Order.DESCENDING, plc.types.NullOrder.AFTER),
+    )
+    assert o1.strict_boundaries is False
```
wence-:
```suggestion
    assert not o1.strict_boundaries
```

python/rapidsmpf/rapidsmpf/tests/streaming/test_channel_metadata.py
```diff
@@ -23,7 +68,111 @@ def test_hash_scheme() -> None:
...
+    assert o1.num_boundaries == 1
+    assert "OrderScheme" in repr(o1)
+
+    assert o1.boundaries_aligned_with(_two_key_order_scheme(context), context.br())
+
+    o_strict = _two_key_order_scheme(context, strict_boundaries=True)
+    assert o_strict.strict_boundaries is True
```
wence-:
```suggestion
    assert o_strict.strict_boundaries
```

python/rapidsmpf/rapidsmpf/tests/streaming/test_channel_metadata.py
```diff
@@ -23,7 +68,111 @@ def test_hash_scheme() -> None:
...
+    assert o_strict.boundaries_aligned_with(
+        _two_key_order_scheme(context, strict_boundaries=True), context.br()
+    )
+
+    with pytest.raises(TypeError, match="OrderKey"):
+        OrderScheme(
+            cast(
+                "Sequence[OrderKey]",
+                [(0, plc.types.Order.ASCENDING, plc.types.NullOrder.BEFORE)],
+            ),
```
wence-: Use type-ignore rather than an invalid cast.

python/rapidsmpf/rapidsmpf/tests/streaming/test_channel_metadata.py
```diff
@@ -101,6 +257,58 @@ def test_message_roundtrip() -> None:
...
+    m = ChannelMetadata(
+        local_count=8,
+        partitioning=Partitioning(order_scheme, "inherit"),
+        duplicated=True,
+    )
+    msg_m = Message(42, m)
+    assert msg_m.sequence_number == 42
+    got_m = ChannelMetadata.from_message(msg_m)
+    assert got_m.local_count == 8
+    assert got_m.duplicated is True
```
wence-:
```suggestion
    assert got_m.duplicated
```

wence- (changes requested):

rjzamora:
Thanks (again) for the review @wence- !

wence- (approved):

PR #858, opened 2026-02-13
Set rapidsmpf asyncio loop to use py_executor

### LINKED ISSUE #855: Set default executor in `run_streaming_pipeline`
As noted by @wence- in https://github.com/rapidsai/cudf/pull/21382#discussion_r2792635493, cudf-polars' use of `asyncio.to_thread` ends up offloading blocking function calls to a thread from a new `ThreadPoolExecutor` created by asyncio.

We'd (maybe?) prefer that those functions run on the `py_executor` passed into `run_streaming_pipeline`. I haven't thought through all the consequences of this, but it seems reasonable.

Because rapidsmpf creates and controls the event loop, we should be able to get a running loop in the `runner` closure called by `asynico.run` around https://github.com/rapidsai/rapidsmpf/blob/42eca8b1cfaf2d6338ae5b8cda9adf8ed7aef787/python/rapidsmpf/rapidsmpf/streaming/core/node.pyx#L254, and call [`loop.set_default_executor`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.set_default_executor). Then calls to `asyncio.to_thread` will run on `py_executor`.

TomAugspurger:
This updates the asyncio event loop managed by rapidsmpf to use the user-provided `py_executor`. This means that any Python Actor provided by the user that includes a

```
result = asyncio.to_thread(blocking_function, ...)
```

will run in the user-provieded thread pool, rather than an additional one spawned by asyncio.

Closes #855

### DIFF
--- a/python/rapidsmpf/rapidsmpf/streaming/core/node.pyx
+++ b/python/rapidsmpf/rapidsmpf/streaming/core/node.pyx
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES.
 # SPDX-License-Identifier: Apache-2.0
 
 from cython.operator cimport dereference as deref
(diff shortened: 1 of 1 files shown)


pentschev (approved):
LGTM, thanks @TomAugspurger .

TomAugspurger:
The issues Lawrence identified in https://github.com/rapidsai/rapidsmpf/pull/858#discussion_r2812241344 I think preclude this change. There isn't even a public API for checking the number of threads in a `ThreadPoolExecutor`, so we couldn't even check that and error if it's too small.

https://github.com/rapidsai/cudf/issues/22379 might provide an alternative on the cudf-polars side.

wence-:
See #1023.

PR #862, opened 2026-02-17
Expose `local_partitions` of shuffler to Python

wence-:
Currently when we construct a shuffler and then want to extract the local partitions in order, we need to know the partition assignment function. The default is round-robin assignment and so we just hard-code that (since the assignment function is not exposed in Python today). However, if we were to import a shuffle object from C++, this hard-coded assignment might be wrong. Instead, we should just ask the shuffler for its assignment.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/streaming/coll/shuffler.pyx
+++ b/python/rapidsmpf/rapidsmpf/streaming/coll/shuffler.pyx
@@ -339,8 +339,7 @@ cdef class ShufflerAsync:
 
         Returns
         -------
-        list[int]
-            Partition IDs owned by this shuffler.
+        Partition IDs owned by this shuffler.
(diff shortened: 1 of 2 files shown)


python/rapidsmpf/rapidsmpf/streaming/coll/shuffler.pyx
```diff
@@ -331,3 +332,20 @@ cdef class ShufflerAsync:
...
             return None
+
+    def local_partitions(self):
+        """
+        Return the partition IDs owned by this rank.
+
+        Returns
+        -------
+        list[int]
+            Partition IDs owned by this shuffler.
```
madsbk:
```suggestion
        Partition IDs owned by this shuffler.
```

PR #863, opened 2026-02-17
Remove multi-message support in shuffle chunks

madsbk:
To significantly simplify the code, I suggest removing the ability for shuffle chunks to contain multiple messages.

Instead, if we want to concatenate messages, this should be handled at a higher level, for example by using multiple shuffle operations.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/tests/test_shuffler.py
+++ b/python/rapidsmpf/rapidsmpf/tests/test_shuffler.py
@@ -65,7 +65,6 @@ def test_shuffler_single_nonempty_partition(
 
     my_partitions = shuffler.local_partitions()
     expected_partitions = set(my_partitions)
-    shuffler.insert_finished(list(range(total_num_partitions)))
 
     local_outputs = []
     extracted_partitions = set()


cpp/src/shuffler/postbox.cpp
```diff
@@ -14,14 +14,8 @@ namespace rapidsmpf::shuffler::detail {
...
 void PostBox<KeyType>::insert(Chunk&& chunk) {
-    // check if all partition IDs in the chunk map to the same key
-    KeyType key = key_map_fn_(chunk.part_id(0));
-    for (size_t i = 1; i < chunk.n_messages(); ++i) {
-        RAPIDSMPF_EXPECTS(
-            key == key_map_fn_(chunk.part_id(i)),
-            "PostBox.insert(): all messages in the chunk must map to the same key"
-        );
-    }
+    // Single-message chunks only - get the key for the one partition
```
pentschev:
```suggestion
```

cpp/src/shuffler/shuffler.cpp
```diff
@@ -491,31 +476,20 @@ void Shuffler::insert_into_ready_postbox(detail::Chunk&& chunk) {
 void Shuffler::insert(detail::Chunk&& chunk) {
     {
         std::lock_guard const lock(outbound_chunk_counter_mutex_);
-        // There are multiple partitions in the chunk. So, increment the counter for
-        // each partition.
-        for (size_t i = 0; i < chunk.n_messages(); ++i) {
-            ++outbound_chunk_counter_[chunk.part_id(i)];
-        }
+        // Single-message chunk - increment counter for the one partition
```
pentschev:
```suggestion
```

pentschev (approved):
Looks good, left some suggestions to remove unnecessary comments only.

pentschev (approved):

nirandaperera (approved):
LGTM. I wish we had kept the `ChunkBatch` API when this was initially added. Then maybe we could have kept some of the concat code for later use.

PR #871, opened 2026-02-19
Statistics cleanup

madsbk:
Wherever we previously took a `statistics` argument, we now use the BufferResource’s statistics when available.

### DIFF
--- a/cpp/src/integrations/cudf/partition.cpp
+++ b/cpp/src/integrations/cudf/partition.cpp
@@ -257,9 +257,7 @@ std::vector<PackedData> unspill_partitions(
         ret.emplace_back(std::move(metadata), br->move(std::move(data), reservation));
     }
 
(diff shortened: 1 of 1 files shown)


cpp/src/integrations/cudf/partition.cpp
```diff
@@ -234,16 +226,17 @@ std::vector<PackedData> spill_partitions(
...
 }
 
 std::vector<PackedData> unspill_partitions(
     std::vector<PackedData>&& partitions,
     BufferResource* br,
-    AllowOverbooking allow_overbooking,
-    std::shared_ptr<Statistics> statistics
+    AllowOverbooking allow_overbooking
 ) {
     auto const start_time = Clock::now();
```
nirandaperera:
```suggestion
    auto statistics = br->statistics();
    auto const start_time = Clock::now();
```

PR #872, opened 2026-02-19
Reimplement allreduce using a recursive doubling scheme

wence-:
Replaces the previous allgather-based implementation with one based on pair-wise recursive doubling message exchange.

The API is also adapted since the implementation is "one-shot". The `AllReduce` constructor now takes both the local input buffer to be combined, and the output buffer to write results into. This change allows us to provide Buffers that satisfy any particular alignment requirements that the combining operator might have. It also means that the allreduce now does no internal allocations.

The user is responsible for ensuring that the memory type and sizes of the buffers match, and that the combining operator works with that memory type.

To help with the implementation, I added two new methods to the Communicator `test` to check if a future is complete in a non-blocking fashion, and `wait_all` (though this latter is not used so I could remove it).

I also added more documentation on the pre- and post-conditions in the Communicator, and write-after-read hazards when checking for `is_latest_write_done` on `Buffer`s.

### DIFF
--- a/cpp/include/rapidsmpf/coll/allreduce.hpp
+++ b/cpp/include/rapidsmpf/coll/allreduce.hpp
@@ -135,6 +135,11 @@ class AllReduce {
      * `Buffer` you obtain from this function, you _must_ obtain the correct stream from
      * the `Buffer` itself.
      *
+     * @warning There may be outstanding stream-ordered work _reading_ from the first
+     * `Buffer` when this function returns (not tracked by the buffer's
+     * `Buffer::latest_write_event()`). If you want to pass it to a non-stream-ordered API
+     * that writes to the buffer you _must_ synchronise its stream first.
+     *
      * @throws std::runtime_error If the timeout is reached or if this method is
      * called more than once.
      */


cpp/include/rapidsmpf/coll/allreduce.hpp
```diff
@@ -190,24 +113,29 @@ class AllReduce {
...
+     * the `Buffer` itself.
      *
      * @throws std::runtime_error If the timeout is reached or if this method is
      * called more than once.
      */
-    [[nodiscard]] PackedData wait_and_extract(
-        std::chrono::milliseconds timeout = std::chrono::milliseconds{-1}
-    );
+    [[nodiscard]] std::pair<std::unique_ptr<Buffer>, std::unique_ptr<Buffer>>
+    wait_and_extract(std::chrono::milliseconds timeout = std::chrono::milliseconds{-1});
```
madsbk: Use `std::optional<std::chrono::milliseconds>` or maybe accept a `Options` instead?

madsbk (commented):
Thanks @wence- , this is very good. I only have minor stuff

cpp/include/rapidsmpf/coll/allreduce.hpp
madsbk: Fair enough

cpp/src/coll/allreduce.cpp
```diff
@@ -3,98 +3,111 @@
...
+AllReduce::~AllReduce() noexcept {
+    if (function_id_.is_valid() && progress_thread_) {
+        auto const phase = phase_.load(std::memory_order_acquire);
+        if (phase != Phase::ResultAvailable && phase != Phase::Done) {
+            // If we get here and we hadn't finished the event loop then there are in
+            // flight messages that will be lost forever and potentially be matched
+            // incorrectly so there's really nothing we can do.
+            comm_->logger().warn("Destroying AllReduce before waiting for extraction");
+            std::terminate();
+        }
```
madsbk:
```suggestion
        // If we get here and we hadn't finished the event loop then there are in
        // flight messages that will be lost forever and potentially be matched
        // incorrectly so there's really nothing we can do.
        RAPIDSMPF_EXPECTS_FATAL(
            phase == Phase::ResultAvailable || phase == Phase::Done, 
            "Destroying AllReduce before waiting for extraction"
        );
```

PR #876, opened 2026-02-20
Add ucxx as a runtime dependency

Matt711:
`ucxx` is currently listed as a build and test dependency of `rapidsmpf`, but not as a runtime dependency. However, `rapidsmpf's cython imports from `ucxx`, so it must be importable at runtime.

- xref https://github.com/rapidsai/cudf/pull/21028 which is how we found this bug.


rjzamora (approved):
Change makes sense to me (once James' suggestion is addressed).
I know I was a bit hesitant earlier, but we are already dependent on libucxx at the c++ level - The ucxx dependency shouldn't be a big deal.

Another note: The CI failure in `test_dask_cudf_join` is certainly unrelated - It looks like there may be a "p2p" but in dask-cudf :/

rjzamora:
Thanks @jameslamb! Are you happy with this PR? (LGTM)

PR #878, opened 2026-02-23
Promote nested closures in run_actor_network to top-level

wence-:
When compiling Cython with the limited API enabled, nested function definitions participate in reference cycles. Hence, if they close over any C++ objects, destruction can happen "out of order" with respect to the (RAII-collected) owner of the C++ object.

To solve this problem, lift the nested functions to top-level. This is probably a good thing to do anyway because these functions don't need to be recreated every time.

Additionally, simplify the nesting to remove some of the complexity. The decorator's job is now to return a partially-evaluated PyActor constructor.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
+++ b/python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
@@ -7,11 +7,12 @@ from libcpp.utility cimport move
 
 import asyncio
 import inspect
-from collections.abc import Awaitable, Iterable, Iterator, Mapping
-from functools import wraps
+from collections.abc import Awaitable, Iterable, Mapping
+from functools import partial, wraps
 
 from rapidsmpf.streaming.core.channel import Channel
 from rapidsmpf.streaming.core.context import Context
+from rapidsmpf.streaming.core.context cimport Context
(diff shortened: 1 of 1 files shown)


python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
```diff
@@ -83,13 +84,56 @@ class PyActor(Awaitable[None]):
...
+    if not inspect.iscoroutinefunction(func):
+        raise TypeError(f"`{func.__qualname__}` must be an async function")
+    return wraps(func)(partial(PyActor, func, extra_channels))
 
-    def __await__(self) -> Iterator[None]:
-        return self._coro.__await__()
+
+async def run_py_actors(py_actors):
+    """Await all ``py_actors`` concurrently."""
+    return await asyncio.gather(*py_actors)
```
TomAugspurger: Question: do we need this (maybe for some Cython-related reason I don't understand), or can we inline it with

```python
py_future = py_executor.submit(asyncio.run, asyncio.gather(*py_actors))
```

I'm fine with either, mostly just curious.

madsbk (commented):
Looks good!

PR #880, opened 2026-02-24
Statistics: decouple formatters from stats

madsbk:
And enable formatting of multiple stats, and a `max` value.

### DIFF
--- a/cpp/include/rapidsmpf/statistics.hpp
+++ b/cpp/include/rapidsmpf/statistics.hpp
@@ -141,11 +141,10 @@ class Statistics {
     /**
      * @brief Generates a formatted report of all collected statistics.
      *
-     * Registered formatters are applied when all their required statistics have been
-     * recorded. Multi-stat formatters are skipped if any of their stats are missing.
-     *
-     * Any statistic not covered by an applied formatter is shown as a plain numeric
-     * value.
+     * Every registered formatter always produces an entry. If all its required
+     * statistics have been recorded the formatter renders the values; otherwise the
+     * entry reads "No data collected". Statistics not covered by any formatter are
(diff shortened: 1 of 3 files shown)


cpp/src/statistics.cpp
```diff
@@ -39,48 +40,70 @@ std::shared_ptr<Statistics> Statistics::disabled() {
...
+}
+
 void Statistics::add_bytes_stat(std::string const& name, std::size_t nbytes) {
-    add_stat(name, nbytes, [](std::ostream& os, std::size_t count, double val) {
-        os << format_nbytes(val);
-        if (count > 1) {
-            os << " (avg " << format_nbytes(val / count) << ")";
-        }
-    });
+    if (!exist_report_entry_name(name)) {
```
TomAugspurger: What's the benefit to the `!exists_report_entry_name(name)` here? IIUC, `register_formatter` contains the same check?

I guess this is where the "before constructing arguments to register_formatter" comes in?

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -19,8 +23,47 @@ namespace rapidsmpf {
...
 /**
- * @class Statistics
  * @brief Track statistics across rapidsmpf operations.
+ *
+ * Two distinct naming concepts are used throughout this class:
+ *
+ * - **Stat name**: the key used to identify an individual `Stat` accumulator,
+ *   as passed to `add_stat()`, `get_stat()`, `add_bytes_stat()`, and
+ *   `add_duration_stat()`. A stat name may be used exclusively — accumulated
+ *   via `add_stat()` and retrieved via `get_stat()` — without ever registering
```
wence-: What does "exclusively" mean here?

PR #883, opened 2026-02-25
Statistics: add JSON export

madsbk:
Adds `write_json()` to `Statistics` for exporting collected stats and memory records as structured JSON.

### DIFF
--- a/cpp/src/statistics.cpp
+++ b/cpp/src/statistics.cpp
@@ -55,11 +55,6 @@ Statistics::Stat Statistics::get_stat(std::string const& name) const {
 }
 
 void Statistics::add_stat(std::string const& name, double value) {
-    RAPIDSMPF_EXPECTS(
-        !has_json_unsafe_chars(name),
-        "stat name cannot contains characters that require JSON escaping: " + name,
-        std::invalid_argument
-    );
     if (!enabled()) {
         return;
     }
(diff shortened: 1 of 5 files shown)


madsbk:
> Basically fine but with some small suggestions.
> 
> As soon as we consider doing anything slightly more complicated with the json, we should (via self-checkout) start using nlohmann/json

@wence-, I am removing all the escape handling. This is overkill! You are right, if we need more advance JSON, we should use a library like  nlohmann/json. 

For now, we just say that it is illegal to use special chars in stat names, which I think is totally reasonable :)

docs/source/statistics.md
```diff
@@ -35,3 +37,38 @@ Statistics:
...
+### JSON (`write_json()`)
+
+JSON output contains raw numeric values for all statistics. Registered
+formatters (which produce human-readable strings such as "1.0 KiB" or "3.5 ms"
+in the text report) are not applied — values remain as plain numbers to keep
+the output machine-parseable. For example, a bytes statistic that reads
+`"3.00001e+09"` is three billion bytes; the text report would show `"2.79 GiB"`
+for the same figure.
+
+Raw units: memory sizes are in **bytes** (float), timings are in **seconds** (float).
```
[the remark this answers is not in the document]
madsbk: This will have  to be followed up. Currently, we have no way to know what kind of type (bytes, duration, count) a specific Stat is. They are all stored as doubles.

cpp/src/statistics.cpp
```diff
@@ -127,6 +141,12 @@ Statistics::MemoryRecorder::MemoryRecorder(
     : stats_{stats}, mr_{mr}, name_{std::move(name)} {
     RAPIDSMPF_EXPECTS(stats_ != nullptr, "the statistics cannot be null");
     RAPIDSMPF_EXPECTS(mr != nullptr, "the memory resource cannot be null");
+    RAPIDSMPF_EXPECTS(
+        !has_json_unsafe_chars(name_),
+        "memory record name cannot contains characters that require JSON escaping: "
+            + name_,
+        std::invalid_argument
+    );
```
wence-: Same here.

docs/source/statistics.md
```diff
@@ -21,6 +21,8 @@ Statistics are available in both C++ and [Python](#api-statistics).
 
 ## Example Output
 
+### Text (`report()`)
+
```
wence-: Add restriction that statistic names must only contain printable ascii characters?

wence- (approved):
Approving with one question about where we want to do the name validation.

PR #884, opened 2026-02-25
Let the communicator own a progress thread

wence-:
We currently only use a single progress thread for driving the event loop of multiple collectives. In service of simplifying the API and in preparation for splitting the Communicator out of the streaming `Context` object (so that you can use the same local computational context in multiple different collective contexts), we now always give a Communicator a progress thread and use that progress thread when creating collectives.

A followup will remove the explicit passing of a progress thread to the collectives.

While here, remove some tests of the shuffler that were inappropriately intimate: they needed to introspect internals that are not part of the advertised public API.

So that we know how many progress threads are available, we now _always_ require that you provide a progress thread when constructing a communicator.

### DIFF
--- a/cpp/include/rapidsmpf/communicator/ucxx_utils.hpp
+++ b/cpp/include/rapidsmpf/communicator/ucxx_utils.hpp
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#pragma once
+
 #include <memory>
 
 #include <mpi.h>
--- a/cpp/benchmarks/bench_comm.cpp
+++ b/cpp/benchmarks/bench_comm.cpp
@@ -283,9 +283,8 @@ int main(int argc, char** argv) {
     // Initialize configuration options from environment variables.
     rapidsmpf::config::Options options{rapidsmpf::config::get_environment_variables()};
 
-    auto stats = std::make_shared<rapidsmpf::Statistics>();
-    // We'll only measure the last run.
-    stats->disable();
+    // We'll only measure the last run, so start disabled.
+    auto stats = rapidsmpf::Statistics::disabled();
     auto progress_thread = std::make_shared<rapidsmpf::ProgressThread>(stats);
     std::shared_ptr<Communicator> comm;
     if (args.comm_type == "mpi") {


cpp/include/rapidsmpf/communicator/mpi.hpp
```diff
@@ -245,6 +249,13 @@ class MPI final : public Communicator {
         return logger_;
     }
 
+    /**
+     * @copydoc Communicator::progress_thread
+     */
+    [[nodiscard]] std::shared_ptr<ProgressThread> progress_thread() override {
```
madsbk:
```suggestion
    [[nodiscard]] std::shared_ptr<ProgressThread> progress_thread() const override {
```

cpp/src/communicator/mpi.cpp
```diff
@@ -96,14 +96,20 @@ void check_mpi_thread_support() {
...
-    : comm_{comm}, logger_{this, std::move(options)} {
-    int rank;
-    int nranks;
-    RAPIDSMPF_MPI(MPI_Comm_rank(comm_, &rank));
-    RAPIDSMPF_MPI(MPI_Comm_size(comm_, &nranks));
-    rank_ = rank;
-    nranks_ = nranks;
+MPI::MPI(MPI_Comm comm, config::Options options, std::shared_ptr<Statistics> statistics)
+    : comm_{comm},
+      rank_{[&]() {
```
madsbk:
```suggestion
      rank_{[comm]() {
```

madsbk (changes requested):
I think we also need to update the Python integration.

In `python/rapidsmpf/rapidsmpf/integrations/core.py` around line 778, `rmpf_worker_setup` still does:

```python
ctx = WorkerContext(
    br=br,
    progress_thread=ProgressThread(comm, statistics), 
    comm=comm,
    ...
)
```

Since the communicator now constructs and owns a `ProgressThread` internally, this ends up creating a second progress thread that runs alongside the one inside the communicator. 

We should expose a `Communicator.progress_thread` property in the Python bindings and use that instead of constructing a new `ProgressThread`.

cpp/tests/main/mpi.cpp
```diff
@@ -26,7 +26,7 @@ void Environment::SetUp() {
     options_ = rapidsmpf::config::Options(rapidsmpf::config::get_environment_variables());
 
     comm_ = std::make_shared<rapidsmpf::MPI>(mpi_comm_, options_);
-    progress_thread_ = std::make_shared<rapidsmpf::ProgressThread>(comm_->logger());
+    progress_thread_ = comm_->progress_thread();
```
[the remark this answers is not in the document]
wence-: Done.

cpp/include/rapidsmpf/communicator/ucxx_utils.hpp
```diff
@@ -1,11 +1,14 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <memory>
```
madsbk:
```suggestion
#pragma once 

#include <memory>
```

PR #891, opened 2026-02-27
Add "contiguous" partition-assignment variation

rjzamora:
Right now, we only offer "round_robin" partition assignment during shuffling. I propose that we add a "contiguous" variation for sorting. When we sort data using the streaming engine, we want adjacent partitions to be assigned to "adjacent" ranks.

This PR includes a rough sketch.


cpp/include/rapidsmpf/streaming/coll/shuffler.hpp
```diff
@@ -15,6 +16,20 @@
...
+
+/**
+ * @brief Factory that creates a ShufflerAsync with contiguous partition assignment.
+ *
+ * @param ctx The streaming context to use.
+ * @param op_id Unique operation ID for this shuffle.
+ * @param total_num_partitions Total number of partitions.
+ * @return A ShufflerAsync that assigns contiguous partition ID ranges to ranks.
+ */
+[[nodiscard]] std::unique_ptr<ShufflerAsync> make_shuffler_async_contiguous(
```
madsbk: Do we need this factory? Why not just call the ctor with `contiguous()`?

cpp/src/streaming/coll/shuffler.cpp
```diff
@@ -83,6 +83,21 @@ coro::task<void> insert_and_notify(
...
+)
+    : ShufflerAsync(
+          std::move(ctx),
+          op_id,
+          total_num_partitions,
+          partition_assignment == PartitionAssignment::contiguous
+              ? shuffler::Shuffler::PartitionOwner{&shuffler::Shuffler::contiguous}
+              : shuffler::Shuffler::PartitionOwner{&shuffler::Shuffler::round_robin}
+      ) {}
+
```
wence-:
```suggestion
```

nirandaperera (changes requested):
some suggestions. Also I think we should have a test to use `contiguous` partitioner for the Shuffle.

robertmaynard (approved):

PR #895, opened 2026-03-02
Split the streaming context and communicator apart

wence-:
The streaming `Context` object represents a _local_ bag of computational resources. We previously bundled it with the `Communicator` that describes the multi-GPU collective context that an actor is working in.

In exploring various algorithmic optimisations in cudf-polars, it has become clear that we will want to use the _same_ `Context` object in multiple different collective contexts.

To facilitate this, split the communicator and context apart. Actors that are collective over a communicator now advertise this explicitly in their type signature by requiring a `Communicator`. Ones that operate purely on local data do similarly by _not_ requiring a `Communicator`.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/streaming/coll/allgather.pyx
+++ b/python/rapidsmpf/rapidsmpf/streaming/coll/allgather.pyx
@@ -91,7 +91,7 @@ cdef class AllGather:
 
     @property
     def comm(self):
-        """"
+        """
         Get the communicator used by the allgather.
 
         Returns


python/rapidsmpf/rapidsmpf/streaming/core/context.pyx
```diff
@@ -43,8 +43,8 @@ cdef class Context:
 
     Parameters
     ----------
-    comm
-        The communicator to use.
+    logger
+        The logger to use.
```
madsbk: Update the example in the docs

python/rapidsmpf/rapidsmpf/streaming/coll/allgather.pyx
```diff
@@ -77,16 +78,28 @@ cdef class AllGather:
...
+                ctx._handle, comm._handle, op_id
             )
 
     def __dealloc__(self):
         with nogil:
             self._handle.reset()
 
+    @property
+    def comm(self):
+        """"
```
madsbk:
```suggestion
        """
```

PR #920, opened 2026-03-18
Simplify Shuffler postboxes

wence-:
The postbox containing chunks to send to remote ranks does not need to
be stratified by partition ID since we only care about the destination
rank which is encoded by applying the partition ownership function to
the partition ID (stored on the chunk).

Similarly, we no longer need the received chunks (ready for the user)
to be stratified by both partition ID and chunk ID, we only need the
partition ID. This is a consequence of no longer sending ACK messages
for metadata.

Take advantage of these facts to simplify the implementation. The
to-send postbox becomes a mutex-protected std::vector, the received
postbox a mutex-protected std::unordered_map<PartID,
std::vector<...>>. As a consequence, we can avoid a bunch of useless
churn in the spilling phases: we can spill chunks in-place in the
containers (when holding the mutex).

### DIFF
--- a/cpp/src/shuffler/postbox.cpp
+++ b/cpp/src/shuffler/postbox.cpp
@@ -39,14 +39,11 @@ bool ChunksToSend::empty() const {
 }
 
 std::string ChunksToSend::str() const {
-    if (empty()) {
-        return "ChunksToSend()";
-    }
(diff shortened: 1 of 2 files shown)


madsbk (approved):
Looks good, I only have minor stuff

PR #938, opened 2026-04-02
[MINOR] Improve error message for duplicate option keys

nirandaperera:


TomAugspurger (commented):
Do you have an example of what caused this? Or maybe a test? I don't immediately see what the new error message is describing.

TomAugspurger:
Ah, I see. The map collecting the options already had that key, just with a different casing.

In that case, maybe "option already *set*" would be slightly nicer than *available*.

TomAugspurger (approved):

PR #940, opened 2026-04-03
Migrate RMM usage to CCCL MR design

bdice:
## Summary

- Rewrite `RmmResourceAdaptor` as a thin shell inheriting `cuda::mr::shared_resource<detail::RmmResourceAdaptorImpl>`, with all mutable state in the impl class for copyable shared ownership.
- Replace `device_memory_resource*` with `rmm::device_async_resource_ref` for non-owning references and `cuda::mr::any_resource` for owning storage.
- Remove `rmm::mr::owning_wrapper` usage (removed in RMM 26.06).
- Update `pool_memory_resource` usage (no longer a template; upstream passed as `device_async_resource_ref`).
- Replace `set_current_device_resource(ptr)` with `set_current_device_resource_ref(ref)`.
- Update test resources to satisfy CCCL resource concept (`allocate_sync`, `deallocate_sync`, `operator==`, `get_property`).
- Update Cython bindings (`.pxd`/`.pyx`) to use `device_async_resource_ref` instead of `device_memory_resource`.
- Point `get_cucascade.cmake` to local cuCascade source directory (companion cuCascade PR: https://github.com/NVIDIA/cuCascade/pull/98).

Depends on https://github.com/rapidsai/rmm/pull/2361.
Depends on https://github.com/rapidsai/ucxx/pull/636.
Depends on https://github.com/rapidsai/cudf/pull/22008.

## Notes

- Depends on cuCascade migration: https://github.com/NVIDIA/cuCascade/pull/98
- The `get_cucascade.cmake` change to use `SOURCE_DIR` is a development convenience and should be updated to point to the merged cuCascade commit before this PR is finalized.


cmake/thirdparty/get_cucascade.cmake
bdice, line 1: rapidsmpf should be pinned to a specific commit of cuCascade (the tip of PR 98) until https://github.com/NVIDIA/cuCascade/pull/98 is merged.

cpp/benchmarks/streaming/ndsh/utils.cpp
```diff
@@ -130,8 +130,7 @@ streaming::Actor consume_channel(
 
 std::pair<std::shared_ptr<streaming::Context>, std::shared_ptr<Communicator>>
 create_context(ProgramOptions& arguments, RmmResourceAdaptor* mr) {
-    rmm::mr::set_current_device_resource(mr);
-    rmm::mr::set_current_device_resource_ref(mr);
+    rmm::mr::set_current_device_resource(*mr);
```
bdice: We may be able to pass `RmmResourceAdaptor` by value rather than by pointer here.

cpp/benchmarks/streaming/bench_streaming_shuffle.cpp
```diff
@@ -353,21 +353,21 @@ int main(int argc, char** argv) {
...
 
-    auto const mr_stack = set_current_rmm_stack(args.rmm_mr);
+    set_current_rmm_resource(args.rmm_mr);
     auto stat_enabled_mr = set_device_mem_resource_with_stats();
     std::unordered_map<rapidsmpf::MemoryType, rapidsmpf::BufferResource::MemoryAvailable>
         memory_available{};
     if (args.device_mem_limit_mb >= 0) {
         memory_available[rapidsmpf::MemoryType::DEVICE] = rapidsmpf::LimitAvailableMemory{
-            stat_enabled_mr.get(), args.device_mem_limit_mb << 20
+            &stat_enabled_mr, args.device_mem_limit_mb << 20
```
bdice: We may be able to pass by value rather than by pointer here.

cpp/benchmarks/utils/rmm_stack.hpp
bdice, line 1: We don't really need this to be a "stack" anymore. Previously that was required to keep the MRs alive but now adaptors will keep their upstreams alive.

Action: more renaming / cleanup?

cpp/benchmarks/bench_memory_resources.cpp
```diff
@@ -151,7 +151,7 @@ void BM_DeviceToHostCopyInclAlloc(benchmark::State& state) {
     auto device_mr = std::make_unique<rmm::mr::cuda_memory_resource>();
```
bdice, line 151: This probably doesn't need to be owned by `unique_ptr` anymore.

(Pattern repeats.)

cpp/include/rapidsmpf/rmm_resource_adaptor.hpp
```diff
@@ -28,9 +24,20 @@ namespace rapidsmpf {
  * This adaptor implements:
  * - Memory usage tracking.
  * - Fallback memory resource support upon out-of-memory in the primary resource.
+ *
+ * This class is copyable and shares ownership of its internal state via
+ * `cuda::mr::shared_resource`.
  */
-class RmmResourceAdaptor final : public rmm::mr::device_memory_resource {
+class RmmResourceAdaptor
+    : public cuda::mr::shared_resource<detail::RmmResourceAdaptorImpl> {
```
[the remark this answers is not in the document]
bdice: Really we have two types of memory resources:
- **Stateless:** trivially copyable, movable, etc. Example: `cuda_memory_resource{}`.
- **Stateful:** adaptors and/or nontrivial base resources. Examples: any adaptor (statistics, logging, pool) or `cuda_async_memory_resource` which holds a CUDA driver pool object.

So yes, all adaptors use this pattern, but it's not _just_ adaptors, it's anything that has shared state. The way we handle stateful resources is PIMPL-like, with some nuances.

- `AdaptorImpl` is a private/detail class that holds the upstream `any_resource` and/or adaptor state as members,
- `Adaptor` is the public API, it uses `cuda::mr::shared_resource` which forwards allocation methods to the `Impl` class.

With that split, copying stateful resources is cheap (just the cost of copying a `shared_ptr`). Crucially, we don't _actually_ use `shared_ptr`, we use `cuda::mr::shared_resource<Impl>` and that does the magic for us. It has its own shared-pointer-like implementation.

cpp/src/memory/buffer_resource.cpp
```diff
@@ -75,7 +75,9 @@ std::shared_ptr<BufferResource> BufferResource::from_options(
 }
 
 rmm::device_async_resource_ref BufferResource::device_mr() const noexcept {
-    return device_mr_;
+    return rmm::device_async_resource_ref{
+        const_cast<cuda::mr::any_resource<cuda::mr::device_accessible>&>(device_mr_)
```
bdice, line 79: We need this method to be `const` but you can't create a `resource_ref` from an `any_resource` that is `const`. We decided in RMM to just `const_cast` to solve this, because we didn't want to mark the `any_resource` member as `mutable`.

cpp/src/rmm_resource_adaptor.cpp
```diff
@@ -4,29 +4,72 @@
...
+    if (this == &other) {
+        return true;
+    }
+    // Manual comparison of optionals to avoid recursive constraint satisfaction in
+    // CCCL 3.2. std::optional::operator== triggers infinite concept checking when the
+    // wrapped type (rmm::device_async_resource_ref) inherits from CCCL's concept-based
+    // resource_ref.
+    // TODO: Revert this after the RMM resource ref types are replaced with
+    // plain cuda::mr ref types. This depends on
+    // https://github.com/rapidsai/rmm/issues/2011.
```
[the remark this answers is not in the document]
bdice: Good catch, this can be removed now. I'm pretty sure this can be reduced to:

```cpp
bool RmmResourceAdaptorImpl::operator==(
    RmmResourceAdaptorImpl const& other
) const noexcept {
    return this == std::addressof(other);
}
```

[See this example pattern.](https://github.com/rapidsai/rmm/blob/6348f6c5c41b432a062b150c1c25271a75cad32f/cpp/include/rmm/mr/detail/failure_callback_resource_adaptor_impl.hpp#L47-L50)

cpp/src/rmm_resource_adaptor.cpp
```diff
@@ -4,29 +4,81 @@
...
 
 namespace rapidsmpf {
 
-ScopedMemoryRecord RmmResourceAdaptor::get_main_record() const {
+// ---------------------------------------------------------------------------
+// detail::RmmResourceAdaptorImpl
+// ---------------------------------------------------------------------------
+namespace detail {
+
+RmmResourceAdaptorImpl::RmmResourceAdaptorImpl(
```
bdice: RMM uses a separate `.cpp` file for the `Impl` classes to avoid accidental overlap between private and public APIs. Either way is fine.

cpp/tests/test_rmm_resource_adaptor.cpp
```diff
@@ -24,94 +26,142 @@
...
         allocs.erase(ptr);
     }
 
-    [[nodiscard]] bool do_is_equal(
-        rmm::mr::device_memory_resource const& other
-    ) const noexcept override {
+    [[nodiscard]] bool operator==(
+        throw_at_limit_resource_impl const& other
+    ) const noexcept {
         return this == &other;
```
bdice, line 64:
```suggestion
        return this == std::addressof(other);
```

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
```diff
@@ -216,18 +217,16 @@ cdef class BufferResource:
         # checked cast requires the GIL
         stats_handle = (<Statistics?>statistics)._handle
 
-        # Keep MR alive because the C++ BufferResource stores a raw pointer.
-        # TODO: once RMM is migrating to CCCL (copyable) any_resource,
-        # rather than the any_resource_ref reference type, we don't
-        # need to keep this alive here.
+        # Stored for the Python device_mr/pinned_mr property accessors.
+        # The C++ BufferResource owns the resource via any_resource.
         self._device_mr = device_mr
```
bdice, line 225: Previously these members were for keeping the MR alive. However, now it seems like they're used for the property accessors. I updated the comment.

cpp/tests/test_rmm_resource_adaptor.cpp
```diff
@@ -94,24 +119,24 @@ TEST(RmmResourceAdaptor, NoFallbackUsedIfNotNecessary) {
...
 }
 
 TEST(RmmResourceAdaptor, RejectsNonOutOfMemoryExceptions) {
     throw_at_limit_resource<std::logic_error> primary_mr{1_MiB};
     throw_at_limit_resource<rmm::out_of_memory> fallback_mr{8_MiB};
-    RmmResourceAdaptor mr{primary_mr, fallback_mr};
+    RmmResourceAdaptor mr(primary_mr, fallback_mr);
 
-    EXPECT_THROW(mr.allocate_sync(2_MiB), std::logic_error);
+    EXPECT_THROW((void)mr.allocate_sync(2_MiB), std::logic_error);
```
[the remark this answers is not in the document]
bdice: This same proposal came up recently and I did some research: https://github.com/rapidsai/rmm/pull/2301#discussion_r2984403279

I concluded that the current state is the best option we have.

> I did some reading on this. Apparently `std::ignore` is really only meant to ignore members during tuple unpacking. The [cppreference docs](https://en.cppreference.com/w/cpp/utility/tuple/ignore.html) state:
> 
> > `std::ignore` is intended for use with [`std::tie`](https://en.cppreference.com/w/cpp/utility/tuple/tie.html) when unpacking a [`std::tuple`](https://en.cppreference.com/w/cpp/utility/tuple.html), as a placeholder for the arguments that are not used, but can be used for any unwanted assignment.
> > 
> > Some code guides recommend using `std::ignore` to avoid warnings from unused return values of [[[nodiscard]](https://en.cppreference.com/w/cpp/language/attributes/nodiscard.html)] functions, even though an assignment isn't required.
> > 
> > For ignoring values not requiring assignment, one may cast to `void`. For variables that have names, but whose value is unused, one may cast those to `void` or declare those variables with [[[maybe_unused]](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused.html)].
> 
> The `std::ignore` name even requires including `<tuple>`, suggesting to me that this is probably the wrong direction. I confirmed with a little more googling: we might actually have the best thing with `(void)`.

bdice (commented):
@nirandaperera Thanks for your help with this so far. I would like to ask if you can help with the remaining refactoring steps here. I will handle the cuCascade pinning soon (just need to get cuDF artifacts built and verify that the cuCascade tests pass, then I will pin to that commit).

cpp/benchmarks/streaming/ndsh/utils.cpp
```diff
@@ -129,9 +129,8 @@ streaming::Actor consume_channel(
 }
 
 std::pair<std::shared_ptr<streaming::Context>, std::shared_ptr<Communicator>>
-create_context(ProgramOptions& arguments, RmmResourceAdaptor* mr) {
+create_context(ProgramOptions& arguments, RmmResourceAdaptor mr) {
```
wence-: Should this take an r-value reference?
```suggestion
create_context(ProgramOptions& arguments, RmmResourceAdaptor&& mr) {
```

cpp/benchmarks/utils/rmm_utils.hpp
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/**
+ * @brief Create a statistics-enabled device memory resource wrapping the current
+ * device resource, and set it as the current device resource.
+ *
+ * @return A RmmResourceAdaptor (shared ownership) for accessing statistics.
+ */
+[[nodiscard]] inline rapidsmpf::RmmResourceAdaptor set_device_mem_resource_with_stats() {
+    rapidsmpf::RmmResourceAdaptor adaptor{cudf::get_current_device_resource_ref()};
```
wence-:
```suggestion
    rapidsmpf::RmmResourceAdaptor adaptor{rmm::mr::get_current_device_resource()};
```

?

cpp/benchmarks/utils/rmm_utils.hpp
```diff
@@ -0,0 +1,58 @@
+/**
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <string>
+
+#include <cudf/utilities/memory_resource.hpp>
```
wence-:
```suggestion
```

cpp/benchmarks/bench_memory_resources.cpp
wence-: Let's do that in a followup, can you open an issue?

cpp/src/rmm_resource_adaptor.cpp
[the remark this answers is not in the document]
wence-: Can we do this please?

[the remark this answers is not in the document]
wence-: Although wait, wouldn't that say two adaptorimpl resources are only the same if they are the same pointer? Perhaps that is what we want.

cpp/src/rmm_resource_adaptor.cpp
```diff
@@ -110,35 +169,64 @@ void RmmResourceAdaptor::do_deallocate(
...
+    auto* ptr = allocate(cuda::stream_ref{cudaStream_t{nullptr}}, bytes, alignment);
+    RAPIDSMPF_CUDA_TRY(cudaStreamSynchronize(cudaStream_t{nullptr}));
+    return ptr;
+}
+
+void RmmResourceAdaptorImpl::deallocate_sync(
+    void* ptr, std::size_t bytes, std::size_t alignment
+) noexcept {
+    deallocate(cuda::stream_ref{cudaStream_t{nullptr}}, ptr, bytes, alignment);
+}
```
wence-, line 182: Question: must one implement the `sync` concept?

cpp/src/rmm_resource_adaptor.cpp
```diff
@@ -4,29 +4,86 @@
...
 #include <rapidsmpf/error.hpp>
 #include <rapidsmpf/rmm_resource_adaptor.hpp>
+#include <rapidsmpf/utils/misc.hpp>
 
 namespace rapidsmpf {
 
-ScopedMemoryRecord RmmResourceAdaptor::get_main_record() const {
+// ---------------------------------------------------------------------------
+// detail::RmmResourceAdaptorImpl
+// ---------------------------------------------------------------------------
```
wence-:
```suggestion
```

cpp/src/rmm_resource_adaptor.cpp
```diff
@@ -110,35 +169,64 @@ void RmmResourceAdaptor::do_deallocate(
...
+) noexcept {
+    deallocate(cuda::stream_ref{cudaStream_t{nullptr}}, ptr, bytes, alignment);
+}
+
+}  // namespace detail
+
+// ---------------------------------------------------------------------------
+// RmmResourceAdaptor (thin shell delegating to shared_resource<Impl>)
+// ---------------------------------------------------------------------------
+
```
wence-:
```suggestion
```

cpp/src/statistics.cpp
```diff
@@ -154,29 +154,23 @@ Statistics::~Statistics() noexcept {
...
 }
 
-// Setting `mr_ = nullptr` disables memory profiling.
-Statistics::Statistics(bool enabled) : enabled_{enabled}, mr_{nullptr} {}
+// Leaving `mr_` as std::nullopt disables memory profiling.
+Statistics::Statistics(bool enabled) : enabled_{enabled} {}
 
 Statistics::Statistics(
-    RmmResourceAdaptor* mr, std::shared_ptr<PinnedMemoryResource> pinned_mr
+    RmmResourceAdaptor mr, std::shared_ptr<PinnedMemoryResource> pinned_mr
```
wence-, line 161: todo (followup): migrate the pinned memory resource to the same copyable any_resource setup.

cpp/src/statistics.cpp
```diff
@@ -154,29 +154,23 @@ Statistics::~Statistics() noexcept {
...
-        mr != nullptr,
-        "when enabling memory profiling, `mr` cannot be nullptr",
-        std::invalid_argument
-    );
-}
+    : enabled_{true}, mr_{std::move(mr)}, pinned_mr_{std::move(pinned_mr)} {}
 
 std::shared_ptr<Statistics> Statistics::from_options(
-    RmmResourceAdaptor* mr,
+    RmmResourceAdaptor mr,
```
wence-, line 166: OK, so at the call site we make a copy...

cpp/src/statistics.cpp
```diff
@@ -154,29 +154,23 @@ Statistics::~Statistics() noexcept {
...
     config::Options options,
     std::shared_ptr<PinnedMemoryResource> pinned_mr
 ) {
     bool const statistics = options.get<bool>("statistics", [](auto const& s) {
         return parse_string<bool>(s.empty() ? "False" : s);
     });
-    return statistics ? std::make_shared<Statistics>(mr, std::move(pinned_mr))
+    return statistics ? std::make_shared<Statistics>(std::move(mr), std::move(pinned_mr))
```
wence-, line 173: ... that is moved into the `Statistics` here.

cpp/tests/test_allgather.cpp
```diff
@@ -295,9 +296,8 @@ TEST_F(BaseAllGatherTest, opid_reuse) {
     std::unique_ptr<AllGather> allgather;
     constexpr rapidsmpf::OpID op_id = 0;
     if (this_rank == 0) {
-        delayed_mr = std::make_unique<DelayedMemoryResource>(
-            mr.get(), std::chrono::milliseconds(500)
-        );
+        delayed_mr =
+            std::make_unique<DelayedMemoryResource>(*mr, std::chrono::milliseconds(500));
         // Recreate the buffer resource and allgather with the delayed MR.
```
wence-: the unique ptr was only to keep the mr alive. Instead we can just rely on it being trivially copyable.

cpp/tests/test_rmm_resource_adaptor.cpp
```diff
@@ -24,94 +26,142 @@
...
+    [[nodiscard]] bool operator==(
+        throw_at_limit_resource_impl const& other
+    ) const noexcept {
+        return this == std::addressof(other);
+    }
+
+    void* allocate_sync(
+        std::size_t bytes, std::size_t alignment = rmm::CUDA_ALLOCATION_ALIGNMENT
+    ) {
+        return allocate(cuda::stream_ref{cudaStream_t{nullptr}}, bytes, alignment);
```
wence-: missing stream sync

cpp/tests/test_rmm_resource_adaptor.cpp
```diff
@@ -24,94 +26,142 @@
...
+    std::unordered_set<void*> const& allocs() const {
+        return this->get().allocs;
+    }
+};
+
 TEST(RmmResourceAdaptor, TracksAllocationsAcrossResources) {
     throw_at_limit_resource<rmm::out_of_memory> primary_mr{1_MiB};
     throw_at_limit_resource<rmm::out_of_memory> fallback_mr{4_MiB};
-    RmmResourceAdaptor mr{primary_mr, fallback_mr};
+    RmmResourceAdaptor mr(primary_mr, fallback_mr);
```
wence-: necessary?

cpp/tests/test_rmm_resource_adaptor.cpp
```diff
@@ -24,94 +26,142 @@
...
 
-    EXPECT_THROW(mr.allocate_sync(8_MiB), rmm::out_of_memory);
+    EXPECT_THROW((void)mr.allocate_sync(8_MiB), rmm::out_of_memory);
 }
 
 TEST(RmmResourceAdaptor, RejectsNonOutOfMemoryExceptions) {
     throw_at_limit_resource<std::logic_error> primary_mr{1_MiB};
     throw_at_limit_resource<rmm::out_of_memory> fallback_mr{8_MiB};
-    RmmResourceAdaptor mr{primary_mr, fallback_mr};
+    RmmResourceAdaptor mr(primary_mr, fallback_mr);
```
wence-: necssary?

cpp/tests/test_rmm_resource_adaptor.cpp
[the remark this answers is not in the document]
wence-: This is an incorrect reading. Defect report https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2968r2.html, backported all the way to C++11, explicitly blesses this usage (`std::ignore = function_marked_nodiscard();`). Please follow the prevailing style here.

cpp/tests/test_rmm_resource_adaptor.cpp
```diff
@@ -342,7 +392,7 @@ TEST(RmmResourceAdaptorScopedMemory, NestedScopedTracksAllocsAndDeallocs) {
 }
 
 TEST(RmmResourceAdaptorScopedMemory, NestedDeallocationYieldsNegativeStats) {
-    rapidsmpf::RmmResourceAdaptor mr{cudf::get_current_device_resource_ref()};
+    rapidsmpf::RmmResourceAdaptor mr(cudf::get_current_device_resource_ref());
```
wence-: Throughout: why are these changes required?

wence- (dismissed):

cpp/benchmarks/bench_memory_resources.cpp
bdice: https://github.com/rapidsai/rapidsmpf/issues/981

cpp/src/rmm_resource_adaptor.cpp
bdice: Yes, the `resource` concept is the `synchronous_resource` concept plus the asynchronous `allocate` / `deallocate` APIs.

https://github.com/NVIDIA/cccl/blob/18bc09c3530ddf2bd76741af08834a4d1da580e9/libcudacxx/include/cuda/__memory_resource/resource.h#L60-L71

cpp/tests/test_rmm_resource_adaptor.cpp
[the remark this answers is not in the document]
bdice: Thanks for the correction. Perhaps we should propose a change to cppreference.

cpp/tests/test_rmm_resource_adaptor.cpp
bdice: I believe they were required earlier to work around some missing CCCL constructors, which have now been resolved. We can revert these now.

cpp/src/rmm_resource_adaptor.cpp
wence-: Ugh, we should petition to remove this

cpp/include/rapidsmpf/rmm_resource_adaptor.hpp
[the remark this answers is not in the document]
nirandaperera: Thanks @bdice. 
My only concern about this is, now `Adapter` is backed by a shared-ptr like state. Callers need to be explicitly aware of it to ensure that its `move`d everywhere possible (to prevent unnecessary atomic increments). What if we made resources non-copyable and non-moveable, and give a CRTP like `std::enable_shared_from_this<SomeResource>` that creates a `shared_resource<SomeResource>`, which can be used with `any_resource<Properties>`? This way, nothing is stashed in `detail` ns IMO.

cpp/src/statistics.cpp
nirandaperera: Yep. Already created an issue #978

cpp/src/statistics.cpp
nirandaperera: @bdice See, callers have to go through @wence- 's realization. 
this is what I meant [here](https://github.com/rapidsai/rapidsmpf/pull/940/changes#r3118707033) 😅

cpp/src/statistics.cpp
nirandaperera: @wence- I think MRs need not be class members of Statistics object. I will remove it here #979

cpp/tests/test_allgather.cpp
nirandaperera: Sorry I've missed this.

cpp/src/rmm_resource_adaptor.cpp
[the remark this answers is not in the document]
nirandaperera: I'll fix this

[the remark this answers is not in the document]
nirandaperera: Seems like both the following still trigger infinite concept checking, 
```c++
return primary_mr_ == other.primary_mr_ && fallback_mr_ == other.fallback_mr_; 
// OR 
return primary_mr_ == other.primary_mr_ && get_fallback_resource() == other.get_fallback_resource();
```

cpp/include/rapidsmpf/rmm_resource_adaptor.hpp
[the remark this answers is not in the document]
bdice: At this point I think the CCCL memory resource design is fixed. You're welcome to ask the CCCL team about non-copyable / non-movable types, or you can create your own wrapper.

If you want to avoid refcount increments, pass around `ref` types. Creating a ref should not increment the refcount of a shared resource. If that resource is stored by a container like `rmm::device_buffer`, that will refcount only when the ref is "reified" into an `any_resource` (when the number of owners grows).

bdice (commented):
Thanks for the fixes @nirandaperera! I think this is fine to merge when the rapidsmpf team is ready. @wence- requested changes but it may be expedient to merge this to unblock builds rather than waiting for further review -- there are several leftover tasks, so any further changes that are needed could be combined with those.

nirandaperera (approved):

cpp/tests/test_rmm_resource_adaptor.cpp
```diff
@@ -94,24 +119,24 @@ TEST(RmmResourceAdaptor, NoFallbackUsedIfNotNecessary) {
 
 TEST(RmmResourceAdaptor, NoFallbackProvidedThrowsOnOOM) {
     throw_at_limit_resource<rmm::out_of_memory> primary_mr{1_MiB};
-    RmmResourceAdaptor mr{primary_mr};
+    RmmResourceAdaptor mr(primary_mr);
 
-    EXPECT_THROW(mr.allocate_sync(8_MiB), rmm::out_of_memory);
+    EXPECT_THROW((void)mr.allocate_sync(8_MiB), rmm::out_of_memory);
```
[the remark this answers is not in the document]
pentschev: Nit2: Why is this marked resolved? It looks like `std::ignore = ` is more appropriate.

pentschev (approved):
Reemphasized on nit (totally fine to slip), and giving a loose approval to be understood as "nothing looks strikingly wrong", since I haven't really followed much of the CCCL discussions and migration so far.

cpp/tests/test_rmm_resource_adaptor.cpp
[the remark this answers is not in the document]
nirandaperera: @pentschev mainly due to this https://github.com/rapidsai/rapidsmpf/pull/940#discussion_r3098388388

But then again, seems like everyone has agreed to use `std::ignore`. Let me add it to #977

PR #943, opened 2026-04-03
Execute PyActors in asyncio.TaskGroup

mroeschke:
Similar what was done in cudf_polars, https://github.com/rapidsai/cudf/pull/21858, now that Python 3.11 is the minimum version, it would be good to start running PyActors in an `asyncio.TaskGroup` to have the nice benefit of canceling other PyActor if one fails.

Once difference is that `asyncio.TaskGroup`s return `ExceptionGroup`s instead of regular `Exceptions` like

```python
  |   File "/conda/envs/rapidsmpf-dev/lib/python3.14/asyncio/taskgroups.py", line 174, in _aexit
  |     raise BaseExceptionGroup(
  |     ...<2 lines>...
  |     ) from None
  | ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "rapidsmpf/streaming/core/actor.pyx", line 143, in _await_actor
    |     return await actor
    |     ^^^^^^^
    |   File "rapidsmpf/streaming/core/actor.pyx", line 105, in run
    |     return await coro
    |     
    |   File "/rapidsmpf/python/rapidsmpf/rapidsmpf/tests/streaming/test_exceptions.py", line 24, in task_that_throws
    |     raise RuntimeError("Throwing in task")
    | RuntimeError: Throwing in task
    +------------------------------------
```

which I think is OK as it provides a little more context about the exception.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/examples/streaming/basic_example.py
+++ b/python/rapidsmpf/rapidsmpf/examples/streaming/basic_example.py
@@ -29,7 +29,9 @@ from rapidsmpf.streaming.cudf.table_chunk import TableChunk
 from rapidsmpf.utils.cudf import cudf_to_pylibcudf_table
 
 if TYPE_CHECKING:
-    from rapidsmpf.streaming.core.actor import CppActor, PyActor
+    from collections.abc import Awaitable
+
+    from rapidsmpf.streaming.core.actor import CppActor
     from rapidsmpf.streaming.core.channel import Channel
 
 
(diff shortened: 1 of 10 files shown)


madsbk (approved):
Looks good

wence- (changes requested):

madsbk:
> I am looking at the PyActor stuff again and I am still a bit confused as to why we need this class at all. PyActor does two things:
> 
>     1. augment the decorated function so that channels are automatically shut down on exceptions
> 
>     2. provide a way of distinguishing from the `CppActor`s that must be on the libcoro event loop.
> 
> 
> (1) is important, but other than the need for (2) a `PyActor` is really just any Python async function. Given that we can distinguish CppActors because those all _do_ inherit from a given class I wonder why we don't just have the `define_actor` decorator just handle point (2).

I think that might work fine.

wence- (approved):

PR #952, opened 2026-04-09
Forward channel metadata through bloom filter apply

wence-:
This is needed to use the apply method in cudf polars since the first thing we do is expect metadata on the output channel. It is safe to forward metadata because filter application acts chunkwise.


pentschev (approved):
This looks good. Can we add a test to confirm the metadata is being forwarded too? Anyway, approving.

madsbk (approved):

wence-:
Added a test

pentschev (approved):

PR #953, opened 2026-04-09
Move away from deprecated `ucxx::Address::getString()`

pentschev:
Use the new `ucxx::Address::getStringView()` and remove the deprecated use of `ucxx::Address::getString()` that are treated as errors in RapidsMPF.

### DIFF
--- a/cpp/src/communicator/ucxx_utils.cpp
+++ b/cpp/src/communicator/ucxx_utils.cpp
@@ -67,7 +67,7 @@ std::shared_ptr<UCXX> init_using_mpi(
         root_listener_address = comm->listener_address();
         root_worker_address_str =
             std::get<std::shared_ptr<::ucxx::Address>>(root_listener_address.address)
-                ->getString();
+                ->getStringView();
     }
     broadcast_listener_address(mpi_comm, root_worker_address_str);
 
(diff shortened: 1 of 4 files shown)


madsbk (approved):

cpp/src/communicator/ucxx_utils.cpp
```diff
@@ -67,7 +67,7 @@ std::shared_ptr<UCXX> init_using_mpi(
         root_listener_address = comm->listener_address();
         root_worker_address_str =
             std::get<std::shared_ptr<::ucxx::Address>>(root_listener_address.address)
-                ->getString();
+                ->getStringView();
```
jakirkham, line 70: Noticed that in replacing `getString` with `getStringView`, the result is assigned directly to `root_worker_address_str` though the return type changes from `std::string` to `std::string_view`

cpp/src/communicator/ucxx_utils.cpp
jakirkham, line 59: Think this would need to change to `std::string_view` as well

Captured this and a couple other minor changes in PR: 
https://github.com/rapidsai/rapidsmpf/pull/954

jakirkham (commented):
Thanks Peter! 🙏

Something caught my eye below that may also need a change as well. Have picked this up in a linked PR. Happy to discuss further

PR #954, opened 2026-04-09
Accept & work with `std::string_view` in `ucxx_utils.cpp`

jakirkham:
Updates `ucxx_utils.cpp` to accept any `std::string_view` supporting type in `broadcast_listener_address`. Also now that `init_using_mpi`'s `root_worker_address_str` is assigned a value from `getStringView`, its type needs to match `std::string_view` (or a copy to `std::string` would be needed). Go ahead and align the type given anywhere `root_worker_address_str` is used a `std::string_view` is accepted. Lastly ensure C++ standard library `#include`s are added to cover the types used in `ucxx_utils.cpp`.

### DIFF
--- a/cpp/src/communicator/ucxx_utils.cpp
+++ b/cpp/src/communicator/ucxx_utils.cpp
@@ -31,7 +31,9 @@ namespace {
  * @param listener_address object containing the listener address of the root,
  * which will be read from in rank 0 and stored to in all other ranks.
  */
-void broadcast_listener_address(MPI_Comm mpi_comm, std::string_view root_worker_address_str) {
+void broadcast_listener_address(
+    MPI_Comm mpi_comm, std::string_view root_worker_address_str
+) {
     std::size_t address_size{root_worker_address_str.size()};
 
     RAPIDSMPF_MPI(
--- a/cpp/src/communicator/ucxx_utils.cpp
+++ b/cpp/src/communicator/ucxx_utils.cpp
@@ -40,9 +40,13 @@ void broadcast_listener_address(
         MPI_Bcast(&address_size, sizeof(address_size), MPI_UINT8_T, 0, mpi_comm)
     );
 
-    RAPIDSMPF_MPI(
-        MPI_Bcast(reinterpret_cast<void*>(root_worker_address_str.data()), address_size, MPI_UINT8_T, 0, mpi_comm)
-    );
+    RAPIDSMPF_MPI(MPI_Bcast(
+        reinterpret_cast<void*>(root_worker_address_str.data()),
+        address_size,
+        MPI_UINT8_T,
+        0,
+        mpi_comm
+    ));
 }
 
 }  // namespace
--- a/cpp/src/communicator/ucxx_utils.cpp
+++ b/cpp/src/communicator/ucxx_utils.cpp
@@ -41,7 +41,7 @@ void broadcast_listener_address(
     );
 
     RAPIDSMPF_MPI(MPI_Bcast(
-        reinterpret_cast<void*>(root_worker_address_str.data()),
+        const_cast<char*>(root_worker_address_str.data()),
         address_size,
         MPI_UINT8_T,
         0,
--- a/cpp/src/communicator/ucxx_utils.cpp
+++ b/cpp/src/communicator/ucxx_utils.cpp
@@ -41,7 +41,7 @@ void broadcast_listener_address(
     );
 
     RAPIDSMPF_MPI(
-        MPI_Bcast(root_worker_address_str.data(), address_size, MPI_UINT8_T, 0, mpi_comm)
+        MPI_Bcast(reinterpret_cast<void*>(root_worker_address_str.data()), address_size, MPI_UINT8_T, 0, mpi_comm)
     );
 }
 
--- a/cpp/src/communicator/ucxx_utils.cpp
+++ b/cpp/src/communicator/ucxx_utils.cpp
@@ -3,6 +3,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <cstdint>
+#include <memory>
+#include <optional>
+#include <string>
+#include <utility>
+
 #include <cuda_device_runtime_api.h>
 #include <ucxx/listener.h>
 
--- a/cpp/src/communicator/ucxx_utils.cpp
+++ b/cpp/src/communicator/ucxx_utils.cpp
@@ -40,8 +40,6 @@ void broadcast_listener_address(
         MPI_Bcast(&address_size, sizeof(address_size), MPI_UINT8_T, 0, mpi_comm)
     );
 
-    root_worker_address_str.resize(address_size);
-
     RAPIDSMPF_MPI(
         MPI_Bcast(root_worker_address_str.data(), address_size, MPI_UINT8_T, 0, mpi_comm)
     );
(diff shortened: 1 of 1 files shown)


cpp/src/communicator/ucxx_utils.cpp
```diff
@@ -25,15 +31,15 @@ namespace {
  * @param listener_address object containing the listener address of the root,
  * which will be read from in rank 0 and stored to in all other ranks.
  */
-void broadcast_listener_address(MPI_Comm mpi_comm, std::string& root_worker_address_str) {
+void broadcast_listener_address(
+    MPI_Comm mpi_comm, std::string_view root_worker_address_str
+) {
     std::size_t address_size{root_worker_address_str.size()};
```
jakirkham, line 37: Here we find the `address_size`, which is simply `root_worker_address_str.size()`

cpp/src/communicator/ucxx_utils.cpp
```diff
@@ -25,15 +31,15 @@ namespace {
...
 
     RAPIDSMPF_MPI(
         MPI_Bcast(&address_size, sizeof(address_size), MPI_UINT8_T, 0, mpi_comm)
     );
 
-    root_worker_address_str.resize(address_size);
```
jakirkham, line 35: However we later call `resize` on `root_worker_address_str` using the size measured before. This seems unneeded. So have dropped it

Looks like this traces back to PR: https://github.com/rapidsai/rapidsmpf/pull/80

Not seeing discussion of this point in that PR

@nirandaperera looks like you authored this code. Am I missing something or is this ok to drop?

jakirkham (commented):
It looks like there was a `resize` call on a string below. However the net effect should be a no-op. So dropped it

More details below. Would appreciate if those more familiar with this code could double check we are not missing something

cpp/src/communicator/ucxx_utils.cpp
```diff
@@ -25,17 +31,17 @@ namespace {
...
-
     RAPIDSMPF_MPI(
-        MPI_Bcast(root_worker_address_str.data(), address_size, MPI_UINT8_T, 0, mpi_comm)
+        MPI_Bcast(reinterpret_cast<void*>(root_worker_address_str.data()), address_size, MPI_UINT8_T, 0, mpi_comm)
```
jakirkham: It looks like [`MPI_Bcast`]( https://docs.open-mpi.org/en/v5.0.x/man-openmpi/man3/MPI_Bcast.3.html ) requires a `void*`. IOW it is untyped and non-`const`. The former is expected. However the latter is unfortunate. Though IIUC the memory is treated as read-only as it is a message that is broadcasted. So think this should be safe in the end

cpp/src/communicator/ucxx_utils.cpp
```diff
@@ -25,18 +31,22 @@ namespace {
...
         MPI_Bcast(&address_size, sizeof(address_size), MPI_UINT8_T, 0, mpi_comm)
     );
 
-    root_worker_address_str.resize(address_size);
-
-    RAPIDSMPF_MPI(
-        MPI_Bcast(root_worker_address_str.data(), address_size, MPI_UINT8_T, 0, mpi_comm)
-    );
+    RAPIDSMPF_MPI(MPI_Bcast(
+        const_cast<char*>(root_worker_address_str.data()),
```
jakirkham, line 44: It looks like [`MPI_Bcast`]( https://docs.open-mpi.org/en/v5.0.x/man-openmpi/man3/MPI_Bcast.3.html ) requires a `void*`. IOW it is untyped and non-`const`. The former is expected. However the latter is unfortunate. Though IIUC the memory is treated as read-only as it is a message that is broadcasted. So think this should be safe in the end

cpp/src/communicator/ucxx_utils.cpp
jakirkham: Moved to this thread: https://github.com/rapidsai/rapidsmpf/pull/954#discussion_r3057429381

cpp/src/communicator/ucxx_utils.cpp
pentschev: The `resize()` here is required. This function is executed by all ranks, the "root" rank has a valid `root_worker_address_str` with the actual address, all other ranks have probably an empty `root_worker_address_str`, and thus the `resize` requirement. In your [comment below](https://github.com/rapidsai/rapidsmpf/pull/954/files#r3057429381) you question the non-`const` case correctly, and there's a reason for it: the exact same code will execute on all ranks, the "root" rank (in this case `0`, defined by the 4th argument to `MPI_Bcast`) will read the value of `buffer` and send a broadcast message to all other ranks, other ranks will also execute `MPI_Bcast` but in "receive-mode" meaning they will retrieve that value from the "root" rank and store into their local `root_worker_address_str`.

With that said, I don't think we can actually change this here to `std::string_view`, since all ranks (except for the "root" rank) need to write to it.

jakirkham: Interesting thank you for the context Peter! 🙏

In that case we may need to simply change this line to explicitly construct a `std::string` instead

https://github.com/rapidsai/rapidsmpf/blob/5e6c8cfa35cbe9fc87e4f8b06bbf37efca162624/cpp/src/communicator/ucxx_utils.cpp#L68-L70

pentschev: But what would be the benefit of changing it there, rather than leaving the implementation of `broadcast_listener_address` as it is currently written in `main`? At one point or another we'll need to create a `std::string` from the `std::string_view` returned by `getStringView()`, since `broadcast_listener_address` is the function that contains that requirement to operate on a `std::string` it seems more natural to me leaving that operation for the time the function is called and argument is converted.

jakirkham: Think there may be a misunderstanding here

Have put my alternative proposed fix with explanation in PR: https://github.com/rapidsai/rapidsmpf/pull/955

Happy to discuss further as needed

pentschev: Sorry, I misspoke. The conversion is done implicitly when assigning to `root_worker_address_str`. Still, I don't think we need any other changes and also not #955. IOW, I'm arguing that what we currently have in `main` is still a better choice among all the proposals and thus I propose continuing _without_ changes from #954 and #955.

jakirkham: No worries

Summarizing our discussion in PR https://github.com/rapidsai/rapidsmpf/pull/955

There are some tricky cases with `std::string` and `std::string_view` as to whether a conversion will happen or not. However the specific case here is able to use `std::string`'s `operator=` with the `std::string_view` argument. As a result no explicit construction is needed in this case

This is best captured in this comment: https://github.com/rapidsai/rapidsmpf/pull/955#issuecomment-4217414861

Though the rest of the discussion may be worthwhile for others to read

jakirkham:
As noted by Peter above: https://github.com/rapidsai/rapidsmpf/pull/954#discussion_r3057477720

The mutation of `std::string` is intended and important to the behavior of these functions. Given this the use of `std::string_view` wouldn't be appropriate

So closing this PR

jakirkham:
Also thank you again Peter for taking the time to explain to me how this code is working! 🙏

pentschev:
No worries, John. The MPI code is indeed not always immediately obvious, thanks for raising those questions.

PR #955, opened 2026-04-09
Convert `ucx_utils` `std::string_view` to `std::string`

jakirkham:
Recently the use of `getString` here was changed to `getStringView` ( https://github.com/rapidsai/rapidsmpf/pull/953 ). However the return type would also change from `std::string` to `std::string_view`. Given `root_worker_address_str` is `std::string` and conversion from `std::string_view` to `std::string` needs to be done explicitly, this explicitly converts the result of `getStringView` to `std::string`.

### DIFF
--- a/cpp/src/communicator/ucxx_utils.cpp
+++ b/cpp/src/communicator/ucxx_utils.cpp
@@ -65,9 +65,10 @@ std::shared_ptr<UCXX> init_using_mpi(
         );
 
         root_listener_address = comm->listener_address();
-        root_worker_address_str =
+        root_worker_address_str = std::string{
             std::get<std::shared_ptr<::ucxx::Address>>(root_listener_address.address)
-                ->getStringView();
+                ->getStringView()
+        };
     }
     broadcast_listener_address(mpi_comm, root_worker_address_str);
 


pentschev:
> Given `root_worker_address_str` is `std::string` and conversion from `std::string_view` to `std::string` needs to be done explicitly

How so? [C++17 and above have an assignment operator (6) that will copy the `std::string_view`](https://en.cppreference.com/w/cpp/string/basic_string/operator=.html).

jakirkham:
It's a good question. Am honestly not sure.

Had expected that to work. However in some cases it appears to require explicit construction. 

Here is a simple example:

```cpp
#include <string>
#include <string_view>


std::string copy_string(std::string_view sv) {
    std::string s;
    s = sv;         // Error needs explicit constructor
    return sv;
}
```

Trying [this in Godbolt errors]( https://godbolt.org/z/9P1deKocr )

Found various threads (mostly on Reddit and Stack Overflow) on the topic when searching. Did just see [this one from the ISO C++ Google Group in 2017 before they moved the list]( https://groups.google.com/a/isocpp.org/g/std-discussion/c/YVGIEJOt_E0 ). Maybe it has some answers

If you have thoughts on behavior differences here, would be interested to hear them

pentschev:
The error is in your code. The conversion `s = sv` succeeds as expected, the error is in `return sv`:

```c++
<source>:8:12: error: could not convert 'sv' from 'std::string_view' {aka 'std::basic_string_view<char>'} to 'std::string' {aka 'std::__cxx11::basic_string<char>'}
```

You're trying to return the original `sv` argument on a function that expects `std::string` return type. You wanted to write `return s`.

jakirkham:
Oops you are right I had a typo

What I should have had, but didn't was the following

```cpp
#include <string>
#include <string_view>


std::string copy_string(std::string_view sv) {
    std::string s;
    s = sv;               // OK
    std::string s2 = sv;  // Error needs explicit constructor
    return s2;
}
```

Here [it is in Godbolt]( https://godbolt.org/z/Y6f7EcTY5 )

This matches what is discussed in the ISO C++ thread

IOW what was already here is ok because the declaration and the assignment are separate (avoiding an implicit conversion attempt and doing a copy). However one can't use `=` in the declaration line to copy-construct `std::string` (though using `{...}` is fine)

pentschev:
I see, in that case it looks like we would in fact need the explicit conversion, however, in our current use case it seems unnecessary, I don't see any benefits in performing an explicit conversion just for the sake of it when the standard is explicitly support of the behavior as written in `main` and everything is working (otherwise we would already have seen compiler errors before). With that said, for now I would still keep my recommendation to leaving things as they currently are.

jakirkham:
Yes though to be clear this would fix the compiler issue above

```diff
 #include <string>
 #include <string_view>
 
 
 std::string copy_string(std::string_view sv) {
     std::string s;
     s = sv;               // OK
-    std::string s2 = sv;  // Error needs explicit constructor
+    std::string s2{sv};   // OK, using explicit constructor
     return s2;
 }
```

jakirkham:
Thanks Peter for the discussion, feedback, and insights here! 🙏

Think I'm getting a better handle on the RapidsMPF code and how `std::string_view` fits into it

Was originally concerned that we had a bug here and was trying to be proactive in addressing it quickly (since it came out of changes I added to UCXX)

Though after our discussion and playing with `std::string_view` more in this context, agree that either the code in `main` or this PR should work equally well. Expect there is not much difference between the two and a good compiler brings them to similar places

Happy to go with your recommendation to close. Thanks again!

pentschev:
Thanks for raising that John!

PR #957, opened 2026-04-10
Split out utilities in allgather and simplify completion tracking

wence-:
This is mostly preparatory work for introducing a sparse alltoall that we will use for neighbourhood ghost exchanges to implement rolling windows. The chunk and postbox classes are made available separately.

Additionally, I simplify the allgather a bit by removing API surface that is unused outside of tests.


cpp/include/rapidsmpf/coll/utils.hpp
```diff
@@ -0,0 +1,362 @@
...
+     * @note We use the finish marker chunk ID to encode the number of
+     * insertions on the originating rank.
+     */
+    Chunk(ChunkID id, Rank destination);
+
+  public:
+    /// @brief Sentinel destination for chunks with unknown destination.
+    ///
+    /// A received chunk will have an invalid destination.
+    static constexpr Rank INVALID_RANK = -1;
```
madsbk: Avoid relying on Rank being signed?
```suggestion
    static constexpr Rank INVALID_RANK = std::numeric_limits<Rank>::max();
```

madsbk (approved):
Looks good

cpp/include/rapidsmpf/coll/utils.hpp
wence-: I don't think we're going to change the signedness of rank?

madsbk: Maybe not, up to you

wence-: Done.

PR #958, opened 2026-04-10
Gather statistics

madsbk:
Adds support for gathering statistics across ranks:

* `Statistics::serialize()` / `Statistics::deserialize()`
  Binary serialization round-trip for statistics data.
* `Statistics::copy()`
  Deep copy returning `shared_ptr`.
* `Statistics::merge()`
  Combines two or more `Statistics` objects, summing counts/values, taking the max of maxima, and preserving formatters.
* `coll::gather_statistics()`
  Blocking collective that gathers serialized statistics from all non-root ranks to the root rank, returning `vector<shared_ptr<Statistics>>` for easy merging.


cpp/src/coll/gather_statistics.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include <thread>
+#include <vector>
+
+#include <rapidsmpf/coll/gather_statistics.hpp>
+#include <rapidsmpf/error.hpp>
+#include <rapidsmpf/utils/misc.hpp>
+
+namespace rapidsmpf::coll {
+
+std::vector<std::shared_ptr<Statistics>> gather_statistics(
```
wence-, line 17: If the serialized representation of the statistics is a constant size, you could use AllReduce for this.

madsbk: It might not be constant size. E,g, if some ranks spills and some don't, then the number of `Stat` might differ

nirandaperera, line 17: I'm wondering if we should encapsulate this in a `gather` collective? 😅

cpp/src/coll/gather_statistics.cpp
```diff
@@ -0,0 +1,66 @@
...
+    if (rank != root) {
+        // Serialize and send to root.
+        auto serialized = stats->serialize();
+        auto msg = std::make_unique<std::vector<std::uint8_t>>(std::move(serialized));
+        auto future = comm->send(std::move(msg), root, tag);
+        // Poll until send completes.
+        while (!comm->test(future)) {
+            std::this_thread::yield();
+        }
+        std::ignore = comm->release_sync_host_data(std::move(future));
```
nirandaperera: is this really required. `~future` handles this, doesnt it?

cpp/src/statistics.cpp
```diff
@@ -390,6 +391,126 @@ void Statistics::write_json(std::filesystem::path const& filepath) const {
...
+    //   Per stat (sorted by name):
+    //     [name_length: uint64] [name: char[]] [count: uint64] [value: double] [max:
+    //     double]
+    std::size_t total = sizeof(std::uint64_t);  // num_stats
+    for (auto const& [name, stat] : stats_) {
+        total += sizeof(std::uint64_t)  // name_length
+                 + name.size()  // name bytes
+                 + sizeof(std::uint64_t)  // count
+                 + sizeof(double)  // value
+                 + sizeof(double);  // max
```
nirandaperera: I'm wondering if we should add a `Stat::serialize(std::unit8_t* out)` method. And I think  these `sizeof`s reduces to `sizeof(Stat)`. 
Say, in the future if we add a new stat field (say, `min`), we'll have many places to go and change usages.

cpp/src/statistics.cpp
```diff
@@ -390,6 +391,126 @@ void Statistics::write_json(std::filesystem::path const& filepath) const {
...
+        std::string name(reinterpret_cast<char const*>(ptr), name_len);
+        ptr += name_len;
+
+        std::uint64_t count{};
+        double value{};
+        double max{};
+        read(count);
+        read(value);
+        read(max);
+        ret->stats_.emplace(std::move(name), Stat(count, value, max));
```
nirandaperera: same here. I think we can move this to `size_t Stat::deserialize(uint8_t const* in)` method.

cpp/src/statistics.cpp
```diff
@@ -390,6 +391,126 @@ void Statistics::write_json(std::filesystem::path const& filepath) const {
...
+    ret->formatters_ = formatters_;
+
+    for (auto const& [name, stat] : other->stats_) {
+        auto [it, inserted] = ret->stats_.try_emplace(name, stat);
+        if (!inserted) {
+            it->second = Stat(
+                it->second.count() + stat.count(),
+                it->second.value() + stat.value(),
+                std::max(it->second.max(), stat.max())
+            );
```
nirandaperera: same here, `Stat Stat::merge(const & other)`?

nirandaperera (changes requested):

cpp/src/coll/gather_statistics.cpp
madsbk: Let’s hold off on adding a general gather collective until we have a clear high-performance use case where a simple send-everything-to-root approach isn’t sufficient.

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -206,6 +207,67 @@ class Statistics {
      */
     void write_json(std::filesystem::path const& filepath) const;
 
+    /**
+     * @brief Creates a deep copy of this Statistics object.
+     *
+     * @note Memory records are not copied.
```
pentschev, line 213: Why? Is it too much data? Is that not useful for final stats?

cpp/src/coll/gather_statistics.cpp
```diff
@@ -0,0 +1,66 @@
...
+    // overtaking on the same (rank, tag) pair, so consecutive calls with the
+    // same op_id cannot interfere.
+    std::vector<std::shared_ptr<Statistics>> ret;
+    ret.reserve(safe_cast<std::size_t>(nranks - 1));
+
+    Rank received = 0;
+    while (received < nranks - 1) {
+        auto [msg, sender] = comm->recv_any(tag);
+        if (msg) {
+            ret.push_back(Statistics::deserialize(*msg));
```
pentschev, line 57: Not sure if there's use for it, certainly not in the current implementation, but you could move that into the rank's own bucket too, accounting for the exclusion of the root.

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -230,40 +304,74 @@ class Statistics {
...
+        /**
+         * @brief Returns the serialized size of this Stat in bytes.
+         *
+         * We size each field individually rather than using `sizeof(Stat)` to
+         * avoid platform-dependent struct padding.
+         *
+         * @return The number of bytes needed to serialize this Stat.
+         */
+        [[nodiscard]] static constexpr std::size_t serialized_size() noexcept {
+            return sizeof(std::uint64_t) + sizeof(double) + sizeof(double);
```
pentschev, line 340: Why keep this implementation in the header file when you moved all others to the cpp file?

cpp/src/statistics.cpp
madsbk: Good point, done

cpp/src/coll/gather_statistics.cpp
madsbk: True, removed,

cpp/src/statistics.cpp
```diff
@@ -36,6 +37,68 @@ using Names2DArray = std::array<NamesArray, rapidsmpf::MEMORY_TYPES.size()>;
...
+    in = read(in, max);
+    return {Stat(count, value, max), in};
+}
+
+Statistics::Stat Statistics::Stat::merge(Stat const& other) const {
+    return Stat(count_ + other.count_, value_ + other.value_, std::max(max_, other.max_));
+}
+
+// -- Statistics --------------------------------------------------------------
+
```
pentschev:
```suggestion
}

```

cpp/src/statistics.cpp
```diff
@@ -390,6 +453,112 @@ void Statistics::write_json(std::filesystem::path const& filepath) const {
...
+    ptr = read(ptr, num_stats);
+
+    for (std::uint64_t i = 0; i < num_stats; ++i) {
+        std::uint64_t name_len{};
+        ptr = read(ptr, name_len);
+        RAPIDSMPF_EXPECTS(
+            ptr + name_len <= end,
+            "truncated Statistics serialization data",
+            std::invalid_argument
+        );
```
pentschev, line 518:
```suggestion
```

pentschev: I think this is a duplicate of lines 501-505, no?

pentschev (commented):
Thanks Mads, left some minor comments and suggestions.

cpp/include/rapidsmpf/statistics.hpp
madsbk: Just to support `constexpr`. I was thinking we might want to use it in some `static_assert`.

nirandaperera:
@madsbk Currently, we gather and clear stats for each iteration. So, ultimately in the frontends we'd have to call something like, `ctx.statistics.clear()` to clear? 
Either way, this `gathering and clearing` is a bit fuzzy IMO. Can/ should we use some sorta scoping mechanism for stats like we do to memory records?

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -230,40 +304,74 @@ class Statistics {
...
+         * @brief Deserializes a Stat from a byte buffer.
+         *
+         * @param in Pointer to the input buffer.
+         * @param end Pointer past the end of the input buffer.
+         * @return A pair of the deserialized Stat and a pointer past the last byte
+         * read.
+         * @throws std::invalid_argument If the data is truncated.
+         */
+        [[nodiscard]] static std::pair<Stat, std::uint8_t const*> deserialize(
+            std::uint8_t const* in, std::uint8_t const* end
```
nirandaperera: why dont we accept a `std::span<uint8_t> const` here?

cpp/src/statistics.cpp
```diff
@@ -36,6 +37,66 @@ using Names2DArray = std::array<NamesArray, rapidsmpf::MEMORY_TYPES.size()>;
...
+std::pair<Statistics::Stat, std::uint8_t const*> Statistics::Stat::deserialize(
+    std::uint8_t const* in, std::uint8_t const* end
+) {
+    // Read a value from `ptr` and return a pointer past the bytes read.
+    auto const read = [end]<typename T>(std::uint8_t const* ptr, T& val) {
+        RAPIDSMPF_EXPECTS(
+            ptr + sizeof(T) <= end,
+            "truncated Stat serialization data",
+            std::invalid_argument
+        );
```
nirandaperera: If we accept a span as input arg to the method, we can trivially check this size as `serialized_size() <= in.size()`. 

Then lets return the number of elements read (ie. serialized_size())

nirandaperera (changes requested):
Let's use std::span for deserialize Stat

cpp/include/rapidsmpf/statistics.hpp
madsbk: Good point, span all the way! :)

cpp/include/rapidsmpf/statistics.hpp
madsbk: It’s not particularly useful. Memory records are tied to the local `RmmResourceAdaptor`, which tracks allocations for a specific rank’s memory resource. Merging them across ranks would mix measurements from different memory pools, which isn’t meaningful.

I think the useful pattern is to report each rank’s memory records separately (they’re already available locally), and only merge aggregate statistics (counts, values, maxes) that are semantically additive across ranks.

cpp/src/coll/gather_statistics.cpp
madsbk: I was also wondering whether we should return a map instead of a vector, but honestly, let’s not complicate things right now.

Later, we could generate a more complete report that includes per-rank stats and memory records, along with combined statistics, and potentially highlight any outliers with warnings.

nirandaperera (approved):
LGTM

cpp/src/statistics.cpp
pentschev: @madsbk you marked this as resolved, but it still seems like this is a duplicate of lines 501-505. Am I missing something?

madsbk: Yes, sorry, I should have mentioned that. They are not the same: one checks the fixed-size POD type (`sizeof(T)`), while this one checks the variable-length name string (`name_len` bytes). The `read()` lambda only handles fixed-size fields; the name requires a separate bounds check since its length is data-dependent.

pentschev: Ah ok, I could have read that better. Thanks for clarifying.

pentschev (approved):
LGTM

PR #959, opened 2026-04-10
Implement a sparse alltoall exchange pattern

wence-:
In this collective every rank advertises the destination ranks it will send
to and the source ranks it will receive from (these have to match up
collective, no error is provided). The caller can then insert messages to
particular ranks, followed by a final insert_finished() call.

On the receive side, after waiting for completion, we can extract received
messages by rank. The receive side message order is defined by the
insertion order on the send side. That is, if rank-A inserts messages in
order [A0, A1, A2] to rank-B, then when rank-B calls `extract(rank-A)` it
will see the same order (even if the messages were sent in a different order).

### DIFF
--- a/cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
+++ b/cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
@@ -1,0 +1,177 @@
+/**
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <atomic>
+#include <chrono>
+#include <condition_variable>
+#include <cstdint>
+#include <functional>
+#include <memory>
+#include <mutex>
+#include <unordered_map>
+#include <vector>
+
+#include <rapidsmpf/coll/utils.hpp>
+#include <rapidsmpf/communicator/communicator.hpp>
+#include <rapidsmpf/memory/buffer_resource.hpp>
+#include <rapidsmpf/memory/packed_data.hpp>
+#include <rapidsmpf/progress_thread.hpp>
+
+namespace rapidsmpf::coll {
+
+/**
+ * @brief Sparse all-to-all collective over explicit source and destination peer sets.
+ *
+ * Each rank may send zero or more messages to ranks listed in `dsts` and
+ * receives zero or more messages from ranks listed in `srcs`. Sender order is defined by
+ * the local order of calls to `insert(dst, ...)` for each destination rank.
+ *
+ * This object is logically collective over the communicator and identified by `op_id`.
+ * Local extraction is only valid after `wait()` has completed.
+ */
+class SparseAlltoall {
+  public:
+    /**
+     * @brief Construct a sparse all-to-all collective instance.
+     *
+     * @param comm Communicator for the collective.
+     * @param op_id Collective operation identifier.
+     * @param br Buffer resource used for allocations.
+     * @param srcs Ranks this rank will receive from.
+     * @param dsts Ranks this rank will send to.
+     * @param finished_callback Optional callback invoked exactly once when the collective
+     * is locally complete. The callback should be fast and non-blocking. Ideally it
+     * should only be used to signal a thread to do the actual work of extraction. Note in
+     * particular that the callback should not extract any data.
+     *
+     * @note It is safe to reuse the `op_id` as soon as `wait` has completed
+     * locally or the `finished_callback` has been invoked.
+     *
+     * @note The caller promises that inserted buffers are stream-ordered with respect
+     * to their own stream, and extracted buffers are likewise guaranteed to be stream-
+     * ordered with respect to their own stream.
+     *
+     * @note Collectively the src and dst pairs of participating ranks must be consistent
+     * (not checked for). That is if rank-A advertises that rank-B is in its dst set,
+     * rank-B must advertise that rank-A is in its src set. If we ever need to relax this
+     * restriction we could have each rank advertise its send set and bootstrap the
+     * two-sided information using the non-blocking consensus algorithm of Hoefler,
+     * Siebert, and Lumsdaine, ACM SIGPLAN (2010),
+     * https://dl.acm.org/doi/10.1145/1837853.1693476.
+     */
+    SparseAlltoall(
+        std::shared_ptr<Communicator> comm,
+        OpID op_id,
+        BufferResource* br,
+        std::vector<Rank> srcs,
+        std::vector<Rank> dsts,
+        std::function<void()>&& finished_callback = nullptr
+    );
+
+    ~SparseAlltoall() noexcept;
+
+    SparseAlltoall(SparseAlltoall const&) = delete;
+    SparseAlltoall& operator=(SparseAlltoall const&) = delete;
+    SparseAlltoall(SparseAlltoall&&) = delete;
+    SparseAlltoall& operator=(SparseAlltoall&&) = delete;
+
+    /**
+     * @brief Gets the communicator associated with this SparseAlltoall.
+     *
+     * @return Shared pointer to communicator.
+     */
+    [[nodiscard]] std::shared_ptr<Communicator> const& comm() const noexcept;
+
+    /**
+     * @brief Insert data to send to a destination rank.
+     *
+     * The order the destination rank obtains the sent data is given by the insertion
+     * order on the send side. If inserting concurrently to the same destination, the
+     * caller must establish a total order of the insertions, otherwise the reconstruction
+     * order on the receive side is unspecified.
+     *
+     * @note Concurrent insertion by multiple threads is supported, the caller must ensure
+     * that `insert_finished()` is called _after_ all `insert()` calls have completed.
+     *
+     * @param dst Destination rank. Must be present in the constructor's `dsts`.
+     * @param packed_data Packed payload and metadata to send.
+     */
+    void insert(Rank dst, PackedData&& packed_data);
+
+    /**
+     * @brief Indicate that no more data will be inserted for any destination.
+     *
+     * Must be called exactly once.
+     *
+     * @note If multiple threads are `insert()`ing, you must establish a happens-before
+     * relationship between the completion of all `insert()`s and the final call to
+     * `insert_finished()`.
+     */
+    void insert_finished();
+
+    /**
+     * @brief Wait for local completion.
+     *
+     * @param timeout Optional timeout. Negative values mean no timeout.
+     *
+     * @throws std::runtime_error If the timeout is reached.
+     */
+    void wait(std::chrono::milliseconds timeout = std::chrono::milliseconds{-1});
+
+    /**
+     * @brief Extract all received messages from a source rank.
+     *
+     * The returned vector is ordered by the sender's local insertion order.
+     *
+     * @param src Source rank. Must be present in the constructor's `srcs`.
+     * @return All messages received from `src`.
+     *
+     * @throws std::logic_error If extracting before the collective is complete.
+     */
+    [[nodiscard]] std::vector<PackedData> extract(Rank src);
+
+  private:
+    struct SourceState {
+        std::uint64_t expected_count{0};
+        std::uint64_t received_count{0};
+        std::vector<std::unique_ptr<detail::Chunk>> chunks;
+
+        [[nodiscard]] bool ready() const noexcept {
+            return expected_count > 0 && expected_count == received_count;
+        }
+    };
+
+    void send_ready_messages();
+    void receive_metadata_messages();
+    void receive_data_messages();
+    void complete_data_messages();
+    [[nodiscard]] bool containers_empty() const;
+    [[nodiscard]] ProgressThread::ProgressState event_loop();
+
+    std::shared_ptr<Communicator> comm_;
+    BufferResource* br_;
+    std::vector<Rank> srcs_;
+    std::vector<Rank> dsts_;
+    std::unordered_map<Rank, std::atomic<std::uint64_t>> next_ordinal_per_dst_;
+
+    mutable std::mutex mutex_;
+    std::condition_variable cv_;
+    OpID op_id_;
+    std::atomic<bool> locally_finished_{false};
+    bool can_extract_{false};
+
+    detail::PostBox outgoing_{};
+    std::unordered_map<Rank, std::vector<std::unique_ptr<detail::Chunk>>>
+        incoming_by_src_;
+    std::vector<std::unique_ptr<detail::Chunk>> receive_posted_;
+    std::vector<std::unique_ptr<Communicator::Future>> receive_futures_;
+    std::vector<std::unique_ptr<Communicator::Future>> fire_and_forget_;
+    std::unordered_map<Rank, SourceState> source_states_;
+    std::function<void()> finished_callback_;
+    ProgressThread::FunctionID function_id_;
+};
+
+}  // namespace rapidsmpf::coll
--- a/cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
+++ b/cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
@@ -93,8 +93,10 @@ class SparseAlltoall {
      * caller must establish a total order of the insertions, otherwise the reconstruction
      * order on the receive side is unspecified.
      *
-     * @note Concurrent insertion by multiple threads is supported, the caller must ensure
-     * that `insert_finished()` is called _after_ all `insert()` calls have completed.
+     * @note Concurrent insertion by multiple threads is supported.
+     *
+     * @note the caller must ensure that `insert_finished()` is called _after_ all
+     * `insert()` calls have completed.
      *
      * @param dst Destination rank. Must be present in the constructor's `dsts`.
      * @param packed_data Packed payload and metadata to send.
@@ -129,6 +131,9 @@ class SparseAlltoall {
      * @param src Source rank. Must be present in the constructor's `srcs`.
      * @return All messages received from `src`.
      *
+     * @note Concurrent extraction is supported, behaviour is undefined if two threads
+     * attempt to extract data from the same source.
+     *
      * @throws std::logic_error If extracting before the collective is complete.
      */
     [[nodiscard]] std::vector<PackedData> extract(Rank src);
@@ -144,11 +149,17 @@ class SparseAlltoall {
         }
     };
 
+    /// @brief Send all ready messages.
     void send_ready_messages();
+    /// @brief Post receives for any outstanding metadata messages.
     void receive_metadata_messages();
+    /// @brief Post receives for expected data messages.
     void receive_data_messages();
+    /// @brief Complete receives for data messages.
     void complete_data_messages();
+    /// @brief @return true if all internal containers are empty.
     [[nodiscard]] bool containers_empty() const;
+    /// @brief @return Progress the communication state and return the progress state.
     [[nodiscard]] ProgressThread::ProgressState event_loop();
 
     std::shared_ptr<Communicator> comm_;
--- a/cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
+++ b/cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
@@ -142,7 +142,8 @@ class SparseAlltoall {
     struct SourceState {
         std::uint64_t expected_count{0};
         std::uint64_t received_count{0};
-        std::vector<std::unique_ptr<detail::Chunk>> chunks;
+        std::vector<std::unique_ptr<detail::Chunk>> chunks{};
+        std::vector<std::unique_ptr<detail::Chunk>> incoming{};
 
         [[nodiscard]] bool ready() const noexcept {
             return expected_count > 0 && expected_count == received_count;
@@ -175,8 +176,6 @@ class SparseAlltoall {
     bool can_extract_{false};
 
     detail::PostBox outgoing_{};
-    std::unordered_map<Rank, std::vector<std::unique_ptr<detail::Chunk>>>
-        incoming_by_src_;
     std::vector<std::unique_ptr<detail::Chunk>> receive_posted_;
     std::vector<std::unique_ptr<Communicator::Future>> receive_futures_;
     std::vector<std::unique_ptr<Communicator::Future>> fire_and_forget_;
--- a/cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
+++ b/cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
@@ -47,6 +47,12 @@ class SparseAlltoall {
      * should only be used to signal a thread to do the actual work of extraction. Note in
      * particular that the callback should not extract any data.
      *
+     * @throws std::out_of_range If either `srcs` or `dsts` have invalid values. All
+     * source and destination ranks must be in `[0, ..., comm->nranks())`, and not equal
+     * to the current rank.
+     * @throws std::invalid_argument If the rank lists are not unique.
+     * @throws std::logic_error If the communicator or buffer resource pointers are null.
+     *
      * @note It is safe to reuse the `op_id` as soon as `wait` has completed
      * locally or the `finished_callback` has been invoked.
      *
--- a/cpp/src/coll/sparse_alltoall.cpp
+++ b/cpp/src/coll/sparse_alltoall.cpp
@@ -1,0 +1,261 @@
+/**
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <algorithm>
+#include <atomic>
+#include <chrono>
+#include <memory>
+#include <mutex>
+#include <utility>
+#include <vector>
+
+#include <rapidsmpf/coll/sparse_alltoall.hpp>
+#include <rapidsmpf/error.hpp>
+#include <rapidsmpf/memory/buffer_resource.hpp>
+
+namespace rapidsmpf::coll {
+
+SparseAlltoall::SparseAlltoall(
+    std::shared_ptr<Communicator> comm,
+    OpID op_id,
+    BufferResource* br,
+    std::vector<Rank> srcs,
+    std::vector<Rank> dsts,
+    std::function<void()>&& finished_callback
+)
+    : comm_{std::move(comm)},
+      br_{br},
+      srcs_{std::move(srcs)},
+      dsts_{std::move(dsts)},
+      op_id_{op_id},
+      finished_callback_{std::move(finished_callback)} {
+    RAPIDSMPF_EXPECTS(comm_ != nullptr, "the communicator pointer cannot be null");
+    RAPIDSMPF_EXPECTS(br_ != nullptr, "the buffer resource pointer cannot be null");
+    auto const size = comm_->nranks();
+    auto const self = comm_->rank();
+    for (auto src : srcs_) {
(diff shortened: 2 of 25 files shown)


wence-:
I can split these into bits for review purposes if that is useful

cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
```diff
@@ -0,0 +1,177 @@
...
+    /**
+     * @brief Insert data to send to a destination rank.
+     *
+     * The order the destination rank obtains the sent data is given by the insertion
+     * order on the send side. If inserting concurrently to the same destination, the
+     * caller must establish a total order of the insertions, otherwise the reconstruction
+     * order on the receive side is unspecified.
+     *
+     * @note Concurrent insertion by multiple threads is supported, the caller must ensure
+     * that `insert_finished()` is called _after_ all `insert()` calls have completed.
```
pentschev: They seem like two separate notes or am I misreading? It seems that the multi-threaded insertion comment and `insert_finished()` requirement are independent from each other.

pentschev: This relates to the note in `insert_finished`, they are indeed dependent. I would rephrase this slightly to make dependence clearer:

```suggestion
     * @note Concurrent insertion by multiple threads is supported, the caller must ensure
     * all `insert()` calls (concurrent or not) have completed before calling
     * `insert_finished()`.
```

python/rapidsmpf/rapidsmpf/coll/sparse_alltoall.pyx
```diff
@@ -0,0 +1,159 @@
...
+
+from rapidsmpf.coll.sparse_alltoall cimport cpp_SparseAlltoall, milliseconds_t
+from rapidsmpf.communicator.communicator cimport Communicator, Rank
+from rapidsmpf.memory.buffer_resource cimport (BufferResource,
+                                               cpp_BufferResource)
+from rapidsmpf.memory.packed_data cimport (PackedData, cpp_PackedData,
+                                           packed_data_vector_to_list)
+
+
+cdef class SparseAlltoall:
```
pentschev, line 19: I know this isn't common, but especially with Python free threading do we want to also add notes about multithreading support here?

python/rapidsmpf/rapidsmpf/tests/streaming/test_sparse_alltoall.py
```diff
@@ -0,0 +1,154 @@
...
+            plc.Table(
+                [plc.Column.from_array(np.array([11], dtype=np.int32), stream=stream)]
+            ),
+        )
+        assert_eq(
+            unpack_table(distributed_context, results[1]),
+            plc.Table(
+                [plc.Column.from_array(np.array([29], dtype=np.int32), stream=stream)]
+            ),
+        )
```
pentschev, line 84: This test is ok but it only tests exactly 2 ranks, while any other ranks are all no-op. I think we could use a test that would test as many ranks are available too.

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,261 @@
...
+        detail::Chunk::from_packed_data(
+            ordinal_it->second.fetch_add(1, std::memory_order_acq_rel),
+            comm_->rank(),
+            dst,
+            std::move(packed_data)
+        )
+    );
+}
+
+void SparseAlltoall::insert_finished() {
```
pentschev, line 101: We should probably add a check to prevent calling `insert_finished()` more than once.

cpp/tests/streaming/test_sparse_alltoall.cpp
pentschev, line 1: A few tests I think we're missing include:
- `insert()` after `insert_finished()`
- multi-threaded `insert()`
- `extract()` with invalid source rank

cpp/src/streaming/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,64 @@
...
+              // completion of this callback with a task_group.
+              event_.set(ctx_->executor()->get());
+          }
+      )} {}
+
+SparseAlltoall::~SparseAlltoall() noexcept {
+    RAPIDSMPF_EXPECTS_FATAL(
+        event_.is_set(),
+        "~SparseAlltoall: not all notification tasks complete, did you forget to await "
+        "this->wait() or to call this->insert_finished()?"
```
pentschev: Is `this->wait()` a mistake? I don't think we have a `wait()` method.

python/rapidsmpf/rapidsmpf/coll/sparse_alltoall.pyx
pentschev: We should also include `SparseAlltoAll` to https://github.com/rapidsai/rapidsmpf/blob/main/python/rapidsmpf/rapidsmpf/coll/__init__.py .

cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
```diff
@@ -0,0 +1,177 @@
...
+     */
+    [[nodiscard]] std::vector<PackedData> extract(Rank src);
+
+  private:
+    struct SourceState {
+        std::uint64_t expected_count{0};
+        std::uint64_t received_count{0};
+        std::vector<std::unique_ptr<detail::Chunk>> chunks;
+
+        [[nodiscard]] bool ready() const noexcept {
```
nirandaperera, line 154: Nit. 
```suggestion
        [[nodiscard]] bool constexpr ready() const noexcept {
```

cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
```diff
@@ -0,0 +1,177 @@
...
+            return expected_count > 0 && expected_count == received_count;
+        }
+    };
+
+    void send_ready_messages();
+    void receive_metadata_messages();
+    void receive_data_messages();
+    void complete_data_messages();
```
nirandaperera, line 166: It would be nice to add some `@brief` docstrings here for these methods (for future references). Either here, or in the `cpp` file

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,261 @@
...
+      br_{br},
+      srcs_{std::move(srcs)},
+      dsts_{std::move(dsts)},
+      op_id_{op_id},
+      finished_callback_{std::move(finished_callback)} {
+    RAPIDSMPF_EXPECTS(comm_ != nullptr, "the communicator pointer cannot be null");
+    RAPIDSMPF_EXPECTS(br_ != nullptr, "the buffer resource pointer cannot be null");
+    auto const size = comm_->nranks();
+    auto const self = comm_->rank();
+    for (auto src : srcs_) {
```
nirandaperera, line 40: Nit
```suggestion
    source_states.reserve(srcs_.size());
    for (auto src : srcs_) {
```

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,261 @@
...
+            src >= 0 && src < size && src != self, "SparseAlltoall invalid source rank."
+        );
+        RAPIDSMPF_EXPECTS(
+            incoming_by_src_.emplace(src, std::vector<std::unique_ptr<detail::Chunk>>{})
+                .second,
+            "SparseAlltoall source rank list must be unique"
+        );
+        source_states_.emplace(src, SourceState{});
+    }
+    for (auto dst : dsts_) {
```
nirandaperera, line 53: Nit. 
```suggestion
    next_ordinal_per_dst_.reserve(dsts_.size());
    for (auto dst : dsts_) {
```

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,261 @@
...
+    }
+    function_id_ =
+        comm_->progress_thread()->add_function([this]() { return event_loop(); });
+}
+
+SparseAlltoall::~SparseAlltoall() noexcept {
+    RAPIDSMPF_EXPECTS_FATAL(
+        locally_finished_.load(std::memory_order_acquire),
+        "Destroying SparseAlltoall without `insert_finished()`"
+    );
```
nirandaperera, line 73: Marked as `noexcept` but throwing.

cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
```diff
@@ -0,0 +1,177 @@
...
+ */
+class SparseAlltoall {
+  public:
+    /**
+     * @brief Construct a sparse all-to-all collective instance.
+     *
+     * @param comm Communicator for the collective.
+     * @param op_id Collective operation identifier.
+     * @param br Buffer resource used for allocations.
+     * @param srcs Ranks this rank will receive from.
```
nirandaperera, line 43: Question: should the self rank also be included here? Or self transfers are/ should be skipped?

cpp/src/coll/sparse_alltoall.cpp
nirandaperera: Maybe we can check `locally_finished_==false`

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,261 @@
...
+}
+
+void SparseAlltoall::insert_finished() {
+    for (auto dst : dsts_) {
+        outgoing_.insert(
+            detail::Chunk::from_empty(
+                // Number of metadata messages we send to the destination rank, including
+                // this finish message.
+                next_ordinal_per_dst_.at(dst).load(std::memory_order_acquire) + 1,
```
nirandaperera: Nit. bounds are already checked
```suggestion
                next_ordinal_per_dst_[dst].load(std::memory_order_acquire) + 1,
```

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,261 @@
...
+        ordinal_it != next_ordinal_per_dst_.end(),
+        "SparseAlltoall::insert destination is not a valid destination"
+    );
+    outgoing_.insert(
+        detail::Chunk::from_packed_data(
+            ordinal_it->second.fetch_add(1, std::memory_order_acq_rel),
+            comm_->rank(),
+            dst,
+            std::move(packed_data)
+        )
```
nirandaperera, line 97: Edge case. If `packed_data` is empty, `detail::Chunk::from_packed_data` will throw, but by that time ordinal is already incremented. I think this is fine, but maybe it'll be better to either filter out empty packed_data or explicitly check with a `RAPIDSMPF_EXPECTS` statement at the beginning.

cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
```diff
@@ -0,0 +1,177 @@
...
+     * @note If multiple threads are `insert()`ing, you must establish a happens-before
+     * relationship between the completion of all `insert()`s and the final call to
+     * `insert_finished()`.
+     */
+    void insert_finished();
+
+    /**
+     * @brief Wait for local completion.
+     *
+     * @param timeout Optional timeout. Negative values mean no timeout.
```
nirandaperera, line 126: Why not `0`? I think we should consider `<=0` as no timeout, because I think `cv::wait_for` with 0 timeout indicates _polling_. But since we dont return a bool, we cant indicate this. 
So, no timeout if `<=0` else wait for timeout would be a better representation.

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,261 @@
...
+        return chunk->sequence();
+    });
+
+    std::vector<PackedData> result;
+    result.reserve(state.chunks.size());
+    for (auto& chunk : state.chunks) {
+        result.push_back(chunk->release());
+    }
+    state.chunks.clear();
+    return result;
```
nirandaperera, line 154: shouldnt we erase the state_it at the end of extraction? Since it does not rehash, multithreaded extract will not be affected. But it will clearly indicate that the data had been already extracted.
```suggestion
    source_stastes.erase(state_it);
    return result;
```

BTW should we protect/ fail extracting same `src` from multiple threads? Its not very useful, but for completeness' sake?

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,261 @@
...
+                comm_->send(chunk->release_data_buffer(), dst, payload_tag)
+            );
+        }
+    }
+}
+
+void SparseAlltoall::receive_metadata_messages() {
+    Tag const metadata_tag{op_id_, 0};
+    for (auto src : srcs_) {
+        auto& state = source_states_.at(src);
```
nirandaperera: Any particular reason why `source_states_` are accessed indirectly through `srcs_` rather than iterating the map?

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,261 @@
...
+        while (!state.ready()) {
```
nirandaperera, line 174: Woouldnt this `while not ready` loop hog the progress thread until all the messages are received from all sources? I feel like this will be unfair for other concurrent collectives, isnt it?

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,261 @@
...
+                "SparseAlltoall received metadata with unexpected origin"
+            );
+            if (chunk->is_finish()) {
+                RAPIDSMPF_EXPECTS(
+                    state.expected_count == 0,
+                    "SparseAlltoall received duplicate finish control"
+                );
+                state.expected_count = chunk->sequence();
+            } else {
+                incoming_by_src_.at(src).push_back(std::move(chunk));
```
nirandaperera: Nit
```suggestion
                incoming_by_src_.[src].push_back(std::move(chunk));
```

cpp/src/coll/sparse_alltoall.cpp
nirandaperera: Did you mean to use an `if` instead?

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,261 @@
...
+    RAPIDSMPF_EXPECTS(comm_ != nullptr, "the communicator pointer cannot be null");
+    RAPIDSMPF_EXPECTS(br_ != nullptr, "the buffer resource pointer cannot be null");
+    auto const size = comm_->nranks();
+    auto const self = comm_->rank();
+    for (auto src : srcs_) {
+        RAPIDSMPF_EXPECTS(
+            src >= 0 && src < size && src != self, "SparseAlltoall invalid source rank."
+        );
+        RAPIDSMPF_EXPECTS(
+            incoming_by_src_.emplace(src, std::vector<std::unique_ptr<detail::Chunk>>{})
```
nirandaperera: Why cant `incoming_by_src_` a class member of the `SourceState`? I feel like, it is the received metadata queue from a particular source, isnt it?

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,261 @@
...
+                state.chunks.push_back(std::move(chunk));
+            } else {
+                auto buffer = chunk->release_data_buffer();
+                receive_posted_.push_back(std::move(chunk));
+                receive_futures_.push_back(
+                    comm_->recv(src, payload_tag, std::move(buffer))
+                );
+            }
+        }
+        queue.erase(queue.begin(), queue.begin() + processed);
```
nirandaperera, line 218: Wouldn't a `std::deque/queue` better here? Unprocessed chunks will always be moved by `processed` elements in each progress iteration, isnt it?

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,261 @@
...
+    for (auto src : srcs_) {
+        auto& queue = incoming_by_src_.at(src);
+        std::ptrdiff_t processed = 0;
+        for (auto& chunk : queue) {
+            if (!chunk->is_ready()) {
+                break;
+            }
+            processed++;
+            if (chunk->data_size() == 0) {
+                auto& state = source_states_.at(chunk->origin());
```
nirandaperera:
```suggestion
                auto& state = source_states_[chunk->origin()];
```

nirandaperera (dismissed):
Had some comments for the normal impl. I will check the coroutine impl later today.

cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
wence-: Why? It can't possibly be constexpr evaluated, so there is zero point, no?

cpp/src/coll/sparse_alltoall.cpp
wence-: Nope, `std::terminate`ing.

cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
wence-: No, self transfers are explicitly not allowed. I can relax that if that is needed but I don't think it is.

cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
wence-: Because it matches the usage in the other collectives.

cpp/src/coll/sparse_alltoall.cpp
wence-: No, this is correct. As soon as a metadata message is not available we break the loop. But we want to consume as many metadata messages as are available.

cpp/src/coll/sparse_alltoall.cpp
wence-: Maybe, but `std::vector` is a much better container, and this erase-remove idiom is pretty efficient. Especially since the number of messages is likely small.

cpp/src/coll/sparse_alltoall.cpp
wence-: No.

cpp/src/coll/sparse_alltoall.cpp
wence-: I don't want to lock the container to extract, and I don't want one mutex per source. Or indeed build-my-own with an atomic.

cpp/src/coll/sparse_alltoall.cpp
wence-: What do you mean by `empty`? null metadata or data? That is not allowed by the `PackedData` ctor. If you have a `PackedData` it always has non-null metadata and data.

cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
wence-: Done.

cpp/src/coll/sparse_alltoall.cpp
wence-: Good idea, done.

cpp/src/coll/sparse_alltoall.cpp
wence-: DOne.

cpp/src/coll/sparse_alltoall.cpp
wence-: I don't think this is worth in other words.

cpp/src/streaming/coll/sparse_alltoall.cpp
wence-: Yes thanks, fixed.

cpp/tests/streaming/test_sparse_alltoall.cpp
wence-: > multi-threaded insert()

What do you want to test here? It's a little tricky because if I don't impose ordering then the output is timing dependent.

python/rapidsmpf/rapidsmpf/coll/sparse_alltoall.pyx
wence-: Did the latter.

We don't document any thread-safety in the other python docstrings some I'm inclined to dodge for now. I don't think we've even tried with free-threading yet, so I am inclined to not do anything here.

wence-:
Addressed most comments, but have not yet made test updates yet, will do that tomorrow.

cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
nirandaperera: I see. Then it will be best to add a `@note` mentioning it.

cpp/src/coll/sparse_alltoall.cpp
nirandaperera: Yes, you are right. My comment was irrelevant.

nirandaperera: On a different note, should we drop `packed_data` if `packed_data.data->size() == 0`? (this happens for when packing empty tables)

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -0,0 +1,263 @@
...
+    for (auto src : srcs_) {
+        RAPIDSMPF_EXPECTS(
+            src >= 0 && src < size && src != self, "SparseAlltoall invalid source rank."
+        );
+        RAPIDSMPF_EXPECTS(
+            source_states_.emplace(src, SourceState{}).second,
+            "SparseAlltoall source rank list must be unique"
+        );
+    }
+    next_ordinal_per_dst_.reserve(srcs_.size());
```
nirandaperera: nit. 
```suggestion
    next_ordinal_per_dst_.reserve(dsts_.size());
```

nirandaperera (commented):
Only a few nitpicks

cpp/tests/streaming/test_sparse_alltoall.cpp
pentschev: > What do you want to test here? It's a little tricky because if I don't impose ordering then the output is timing dependent.

Essentially the promise that's being made by [`insert` docstrings](https://github.com/rapidsai/rapidsmpf/pull/959/files#diff-d368054a3e660f1dc0b4084538c5389e1a9ce056e466521ca74bbe2514a5427aR96):

`@note Concurrent insertion by multiple threads is supported.`

python/rapidsmpf/rapidsmpf/coll/sparse_alltoall.pyx
pentschev: You're right, I thought of that because of the [C++ multi-threaded docs](https://github.com/rapidsai/rapidsmpf/pull/959#discussion_r3115661723), but there's indeed nothing in the Python API, so let's punt Python multi-threading.

cpp/include/rapidsmpf/coll/sparse_alltoall.hpp
wence-: Done.

cpp/src/coll/sparse_alltoall.cpp
wence-: No, we don't do it in any of the other collectives. Rationale, empty packed data need not have empty metadata, so it has meaning that must be communicated.

cpp/src/coll/sparse_alltoall.cpp
wence-: Done.

cpp/tests/streaming/test_sparse_alltoall.cpp
wence-: Done.

pentschev (approved):
There are some style errors but my concerns have been addressed, thus approving. Thanks Lawrence!

python/rapidsmpf/rapidsmpf/tests/streaming/test_sparse_alltoall.py
wence-: Is it worth it? Those cases are all tested in C++ and this test is just exercising the plumbing

pentschev: I think it's fine as is. Thanks

pentschev:
Ugh, looks like there are some new fixes needed now:

```c++
 │ │ $SRC_DIR/cpp/tests/test_sparse_alltoall.cpp: In member function 'virtual void SparseAlltoallTest_ordered_by_sender_insertion_with_stream_reordering_Test::TestBody()':
 │ │ $SRC_DIR/cpp/tests/test_sparse_alltoall.cpp:310:13: error: invalid initialization of reference of type 'rapidsmpf::BufferResource&' from expression of type 'const rmm::cuda_stream_view'
 │ │   310 |             delayed_stream,
 │ │       |             ^~~~~~~~~~~~~~
 │ │ $SRC_DIR/cpp/tests/test_sparse_alltoall.cpp:58:32: note: in passing argument 4 of 'rapidsmpf::PackedData {anonymous}::make_payload(int, int, rapidsmpf::MemoryType, rapidsmpf::BufferResource&)'
 │ │    58 |     rapidsmpf::BufferResource& br
 │ │       |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
 │ │ $SRC_DIR/cpp/tests/test_sparse_alltoall.cpp:320:13: error: invalid initialization of reference of type 'rapidsmpf::BufferResource&' from expression of type 'const rmm::cuda_stream_view'
 │ │   320 |             fast_stream,
 │ │       |             ^~~~~~~~~~~
 │ │ $SRC_DIR/cpp/tests/test_sparse_alltoall.cpp:58:32: note: in passing argument 4 of 'rapidsmpf::PackedData {anonymous}::make_payload(int, int, rapidsmpf::MemoryType, rapidsmpf::BufferResource&)'
 │ │    58 |     rapidsmpf::BufferResource& br
 │ │       |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
```

wence-:
> Ugh, looks like there are some new fixes needed now:
> 
> ```c++
>  │ │ $SRC_DIR/cpp/tests/test_sparse_alltoall.cpp: In member function 'virtual void SparseAlltoallTest_ordered_by_sender_insertion_with_stream_reordering_Test::TestBody()':
>  │ │ $SRC_DIR/cpp/tests/test_sparse_alltoall.cpp:310:13: error: invalid initialization of reference of type 'rapidsmpf::BufferResource&' from expression of type 'const rmm::cuda_stream_view'
>  │ │   310 |             delayed_stream,
>  │ │       |             ^~~~~~~~~~~~~~
>  │ │ $SRC_DIR/cpp/tests/test_sparse_alltoall.cpp:58:32: note: in passing argument 4 of 'rapidsmpf::PackedData {anonymous}::make_payload(int, int, rapidsmpf::MemoryType, rapidsmpf::BufferResource&)'
>  │ │    58 |     rapidsmpf::BufferResource& br
>  │ │       |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
>  │ │ $SRC_DIR/cpp/tests/test_sparse_alltoall.cpp:320:13: error: invalid initialization of reference of type 'rapidsmpf::BufferResource&' from expression of type 'const rmm::cuda_stream_view'
>  │ │   320 |             fast_stream,
>  │ │       |             ^~~~~~~~~~~
>  │ │ $SRC_DIR/cpp/tests/test_sparse_alltoall.cpp:58:32: note: in passing argument 4 of 'rapidsmpf::PackedData {anonymous}::make_payload(int, int, rapidsmpf::MemoryType, rapidsmpf::BufferResource&)'
>  │ │    58 |     rapidsmpf::BufferResource& br
>  │ │       |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
> ```

My bad, fixed

pentschev:
> My bad, fixed

Did you forget to push?

wence-:
> > My bad, fixed
> 
> Did you forget to push?

No, but I didn't pull the merge in first

PR #960, opened 2026-04-10
Add BufferResource lifetime tracking to Python wrapper classes

vyasr:
## Summary

When Python wrapper objects (e.g. `TableChunk`, `PackedDataChunk`) are created from factory methods, they store a `cuda_stream_view` internally but do not maintain a reference to the owning `BufferResource`. If the `BufferResource` is garbage collected while these objects are still alive, the stream view becomes a dangling reference.

This PR adds an optional `BufferResource br` parameter to all factory methods on the wrapper classes (`TableChunk`, `PackedDataChunk`, `PartitionMapChunk`, `PartitionVectorChunk`, `PackedData`) and stores it as a `cdef BufferResource _br` field. This prevents the `BufferResource` (and thus its stream and memory resource) from being garbage collected while any wrapper object still references it.

Once cudf is updated to pass this argument everywhere (see https://github.com/rapidsai/cudf/pull/22110) I will open a follow-up PR to make this parameter required in rapidsmpf.

### Changes

**Wrapper class factories** (`.pxd`/`.pyx`/`.pyi`):
- `TableChunk.from_handle`, `from_pylibcudf_table`, `from_packed_data`, `from_message`
- `PackedDataChunk.from_handle`, `from_packed_data`, `from_message`
- `PartitionMapChunk.from_handle`, `from_message`
- `PartitionVectorChunk.from_handle`, `from_message`
- `PackedData.from_librapidsmpf`, `from_cudf_packed_columns`, `from_host_bytes`
- `packed_data_vector_to_list` helper

**Integration call sites**:
- `partition_and_pack`, `split_and_pack`, `spill_partitions`, `unspill_partitions`
- `AllGather.wait_and_extract`
- `Shuffler.extract`

**Internal propagation**:
- `TableChunk.make_available` and `copy` forward `self._br`
- `PackedDataChunk.to_packed_data` forwards `self._br`
- `from_packed_data` falls back to source object's `_br` if none provided

All changes are backwards compatible — `br=None` is the default everywhere.

Also applies `@no_gc_clear` to `TableChunk` to prevent the cyclic GC from clearing the `_br` reference prematurely.


python/rapidsmpf/rapidsmpf/memory/packed_data.pxd
```diff
@@ -13,9 +15,12 @@ cdef extern from "<rapidsmpf/memory/packed_data.hpp>" nogil:
 
 cdef class PackedData:
     cdef unique_ptr[cpp_PackedData] c_obj
+    cdef BufferResource _br
```
pentschev, line 20: I would add a comment here or in the `.pyx` file why it's important to store `_br`. This is the type of thing that a year from now may be "cleaned up" if we don't remember the usage context.

pentschev (changes requested):
Looks sensible to me, thanks Vyas for investing so much time investigating this. The only request I have is to add a comment in the code at a minimum why we're doing it, so that we don't accidentally remove it in the future.

mroeschke (approved):
Once we adapt cudf_polars to pass the BufferResource, should we make `br` generally be a required `not None` argument?

pentschev (commented):
Thanks @vyasr !

python/rapidsmpf/rapidsmpf/memory/packed_data.pxd
nirandaperera: @pentschev very true!

pentschev: Ha! I see you were already thinking when you could submit a PR to clean it up! :rofl:

pentschev (approved):

vyasr:
> Once we adapt cudf_polars to pass the BufferResource, should we make `br` generally be a required `not None` argument?

Yes, I should have specified. I plan to make it a required parameter in a follow-up PR. I just wanted to make this change in a non-breaking way so that we didn't block cudf CI in the interim. Once https://github.com/rapidsai/cudf/pull/22110 is merged I'll open the next PR here.

nirandaperera:
@vyasr I agree that keeping buffer resource obj is a solution here. But if we are concerned about the lifetime of the stream pool, shouldnt we just keep that alive, instead of the buffer resource? 
Stream pool is not owned by the BR. Its provided from the caller.

vyasr:
> @vyasr I agree that keeping buffer resource obj is a solution here. But if we are concerned about the lifetime of the stream pool, shouldnt we just keep that alive, instead of the buffer resource? Stream pool is not owned by the BR. Its provided from the caller.

Can you ever use a stream pool that isn't the one that [the BufferResource keeps alive](https://github.com/rapidsai/rapidsmpf/blob/de167acdb45775748a81e92aaf3b4715d15493fd/python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx#L202)? Saving the BufferResource is also better because it will keep the memory resource alive. That lifetime should also be tied to these objects.

nirandaperera:
@vyasr fair enough. 

Now, I'm thinking we should do the same with in the streaming side. For non-actor operations BR would be a good anchor. In the actor/ streaming operations, this would be the Context obj which hold on to BR, coro executor, spillable messages etc. I'm wondering if we should do a similar change to streaming objects like table chunk, message, spillable message, etc where we keep the ctx alive.

PR #961, opened 2026-04-11
Make BufferResource parameters required instead of optional

vyasr:
## Summary

Follow-up to #960 which added `BufferResource` lifetime tracking to Python wrapper classes. This PR makes all of those `br` parameters **required** instead of optional (removes `=None`/`=*` defaults), so callers are forced to pass a `BufferResource` and cannot accidentally omit it.

## Changes

- Removed `=None`/`=*` defaults from `br` parameters in:
  - `PackedData.from_librapidsmpf`, `packed_data_vector_to_list`
  - `PackedDataChunk.from_handle`, `from_packed_data`, `from_message`
  - `PartitionMapChunk` and `PartitionVectorChunk`: `from_handle`, `from_message`
  - `TableChunk.from_handle`, `from_pylibcudf_table`, `from_packed_data`, `from_message`
- Removed `if br is None: br = obj._br` fallback logic in `PackedDataChunk` and `TableChunk`
- `ShufflerAsync` now stores `ctx.br()` as `self._br` and passes it to `packed_data_vector_to_list` in `extract()`
- `AllGather.extract_all` now passes `ctx.br()` to `packed_data_vector_to_list`
- Updated `.pyi` stubs and `Payload` protocol (`PayloadT` TypeVar bound relaxed to allow types with extended `from_message` signatures)
- Updated `test_bloom_filter.py` to pass `br` to `make_table` and `TableChunk.from_pylibcudf_table`

### DIFF
--- a/python/rapidsmpf/rapidsmpf/streaming/chunks/packed_data.pyx
+++ b/python/rapidsmpf/rapidsmpf/streaming/chunks/packed_data.pyx
@@ -48,7 +48,7 @@ cdef class PackedDataChunk:
         return PackedData.from_librapidsmpf(self.release_handle(), self._br)
 
     @staticmethod
-    def from_packed_data(PackedData obj not None, BufferResource br):
+    def from_packed_data(PackedData obj not None, BufferResource br not None):
         """
         Construct a PackedDataChunk from an existing PackedData object.
 
@@ -86,7 +86,7 @@ cdef class PackedDataChunk:
         return ret
 
     @staticmethod
-    def from_message(Message message not None, BufferResource br):
+    def from_message(Message message not None, BufferResource br not None):
         """
         Construct a PackedDataChunk by consuming a Message.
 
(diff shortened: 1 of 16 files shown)


python/rapidsmpf/rapidsmpf/streaming/chunks/packed_data.pyx
```diff
@@ -48,7 +48,7 @@ cdef class PackedDataChunk:
         return PackedData.from_librapidsmpf(self.release_handle(), self._br)
 
     @staticmethod
-    def from_packed_data(PackedData obj not None, BufferResource br=None):
+    def from_packed_data(PackedData obj not None, BufferResource br):
```
mroeschke: Can/should `br` be made `not None` like the first argument in these signatures?

pentschev: Yeah, I think not just here, but in all the pyx files.

madsbk (approved):
Thanks @vyasr, looks good. I took the liberty of removing all the `# noqa: E704` comments, I don’t think they are necessary?

pentschev (approved):
Thanks Vyas!

PR #962, opened 2026-04-12
Increase timeout for test_bootstrap_multiple_clients

pentschev:
PR #901 switched the subprocess in this test from `fork` to `forkserver` and simultaneously reduced the `q.get` timeout from 10 s to 5 s. Unlike `fork`, `forkserver` starts a fresh Python interpreter that must reimport all modules, which is likely slower due to the additional import work. On loaded CI machines this can exceed 5 seconds, causing a spurious `_queue.Empty` failure.

Increase the `q.get` timeout to 30 seconds and the `@gen_test` timeout to 60 seconds to give the subprocess enough headroom.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/tests/test_dask.py
+++ b/python/rapidsmpf/rapidsmpf/tests/test_dask.py
@@ -465,7 +465,7 @@ def test_dask_cudf_join(
             dd.assert_eq(joined, expected, check_index=False)
 
 
-@gen_test(timeout=30)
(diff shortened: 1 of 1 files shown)


pentschev:
Test failure example: https://github.com/rapidsai/rapidsmpf/actions/runs/24300950486/job/70954114704

madsbk (approved):

pentschev:
Thanks Mads.

PR #965, opened 2026-04-13
Replace `cudaMemcpyAsync` with `rapidsmpf::cuda_memcpy_async`

nirandaperera:
Introduces a thin wrapper `rapidsmpf::cuda_memcpy_async` (in `cpp/include/rapidsmpf/memory/cuda_memcpy_async.hpp`) that currently delegates to `cudf::detail::memcpy_async`, which uses `cudaMemcpyBatchAsync` on CUDA 13.0+ with `cudaMemcpySrcAccessOrderStream`. This gives us a single indirection point to change the copy implementation uniformly in the future without touching all call sites.

Changes:

- Add cpp/include/rapidsmpf/memory/cuda_memcpy_async.hpp with the rapidsmpf::cuda_memcpy_async wrapper
- Replace all direct cudaMemcpyAsync calls across src/, tests/, benchmarks/, and the Python Cython extension with rapidsmpf::cuda_memcpy_async (dropping the now-redundant cudaMemcpyKind direction argument, since cudf::detail::memcpy_async infers direction from pointer types)
- Add a pygrep pre-commit hook (use-rapidsmpf-cuda-memcpy-async) to enforce this going forward — mirrored from the equivalent hook in cudf


cpp/benchmarks/utils/random_data.cu
```diff
@@ -78,9 +80,11 @@ void random_fill(rapidsmpf::Buffer& buffer, rmm::device_async_resource_ref mr) {
             );
             buffer.write_access([&](std::byte* buffer_data,
                                     rmm::cuda_stream_view stream) {
-                RAPIDSMPF_CUDA_TRY_ALLOC(cudaMemcpyAsync(
-                    buffer_data, vec.data(), buffer.size, cudaMemcpyDefault, stream
-                ));
+                RAPIDSMPF_CUDA_TRY_ALLOC(
```
bdice: Should this be `TRY` or `TRY_ALLOC`? (it's copying, not allocating)
```suggestion
                RAPIDSMPF_CUDA_TRY(
```

cpp/include/rapidsmpf/memory/buffer.hpp
```diff
@@ -116,11 +116,10 @@ class Buffer {
      * // Snippet: copy data from `src_ptr` into `buffer` on the buffer's stream.
      * buffer.write_access([&](std::byte* buffer_ptr, rmm::cuda_stream_view stream) {
      *   assert(buffer.stream().value() == stream.value());
-     *   RAPIDSMPF_CUDA_TRY_ALLOC(cudaMemcpyAsync(
+     *   RAPIDSMPF_CUDA_TRY_ALLOC(rapidsmpf::cuda_memcpy_async(
```
bdice: Should this be `TRY` or `TRY_ALLOC`? (it's copying, not allocating)

```suggestion
     *   RAPIDSMPF_CUDA_TRY(rapidsmpf::cuda_memcpy_async(
```

bdice (approved):
Approving with a question

nirandaperera:
@bdice thanks. That macro use was an oversight. Should've caught this earlier.

cpp/include/rapidsmpf/memory/cuda_memcpy_async.hpp
```diff
@@ -0,0 +1,37 @@
...
+ * `cudaMemcpyBatchAsync` on CUDA 13.0+ with `cudaMemcpySrcAccessOrderStream`.
+ * The underlying implementation may change in the future; all callers should
+ * use this function so that any such change is applied uniformly.
+ *
+ * @param dst    Destination memory address.
+ * @param src    Source memory address.
+ * @param count  Number of bytes to copy.
+ * @param stream CUDA stream on which the copy is enqueued.
+ * @return cudaError_t CUDA error code.
+ */
```
madsbk, line 43:
```suggestion
/**
 * @brief Asynchronously copies memory between host and/or device buffers.
 *
 * The copy direction is inferred from the pointer types (`cudaMemcpyDefault`).
 * The source buffer must remain valid until the stream executes the copy.
 *
 * This function should be used instead of `cudaMemcpyAsync`, as it provides
 * improved semantics for asynchronous copies, especially from pageable host memory.
 *
 * ## Background
 *
 * The legacy `cudaMemcpyAsync` API accesses non-CUDA-registered host pointers
 * (e.g., allocations from `malloc` or `new`) at the time of the API call,
 * rather than in stream order. This behavior originates from earlier GPU
 * architectures that could not directly access such memory, requiring an
 * immediate CPU-side staging step.
 *
 * Modern systems with HMM/ATS allow GPUs to access these pointers directly.
 * However, the semantics of `cudaMemcpyAsync` cannot be changed without
 * breaking existing code. The batched memcpy APIs (e.g.,
 * `cudaMemcpyBatchAsync`) introduced in CUDA 13.0 allow the caller to specify
 * `cudaMemcpySrcAccessOrderStream`, ensuring that the source is accessed in
 * stream order and enabling true asynchronous copies from pageable host memory.
 *
 * @param dst Destination memory address.
 * @param src Source memory address.
 * @param count Number of bytes to copy.
 * @param stream CUDA stream on which the copy is enqueued.
 * @return cudaError_t CUDA error code.
 */
```

madsbk (approved):
Looks good. I only suggest a bit of background in the docstring.

PR #966, opened 2026-04-13
Use `cudf::pack` with pinned mr in `TableChunk::copy`

nirandaperera:
In [a previous analysis](https://docs.google.com/spreadsheets/d/1yXiB3aFZO8GUD4dAVnwh7o9zjzXKwaSvGYGZphf9jgQ/edit?usp=sharing) we found that using `cudf::pack` with a pinned mem resource performs similar to,
1. packing to device memory and copying. 
2. chunk packing into pinned memory directly. 

Since (1) use unreserved device memory, `TableChunk::copy` is prone to OOM. This PR changes pinned memory to use `cudf::pack` directly. 

Depends on #967


nirandaperera:
This PR depends on #967 because it now requires precise packed sizes for pinned memory.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -298,7 +298,7 @@ class BufferResource {
     );
 
     /**
-     * @brief Move device buffer data into a Buffer.
+     * @brief Move device/ pinned host buffer data into a Buffer.
```
pentschev:
```suggestion
     * @brief Move device or pinned host buffer data into a Buffer.
```

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -310,10 +310,18 @@ class BufferResource {
      * @param data Unique pointer to the device buffer.
      * @param stream CUDA stream associated with the new Buffer. Use or synchronize with
      * this stream when operating on the Buffer.
+     * @param mem_type The memory type of the underlying @p data. This requires to be a
+     * device accessible memory type (ie. MemoryType::DEVICE or MemoryType::PINNED_HOST).
```
pentschev:
```suggestion
     * @param mem_type The memory type of the underlying @p data. A device accessible memory
     * type is required (ie. MemoryType::DEVICE or MemoryType::PINNED_HOST).
```

cpp/src/memory/buffer_resource.cpp
```diff
@@ -196,14 +196,37 @@ std::unique_ptr<Buffer> BufferResource::allocate(
...
     }
-    return std::unique_ptr<Buffer>(new Buffer(std::move(data), MemoryType::DEVICE));
+
+    if (mem_type == MemoryType::DEVICE) {
+        return std::unique_ptr<Buffer>(new Buffer(std::move(data), MemoryType::DEVICE));
+    } else if (mem_type == MemoryType::PINNED_HOST) {
+        RAPIDSMPF_EXPECTS(
+            pinned_mr_ != PinnedMemoryResource::Disabled,
+            "pinned memory resource is not available",
+            std::invalid_argument
```
pentschev: Nit: should this actually be a `std::runtime_error`? The memory type isn't technically invalid, it is an error only from a runtime/system perspective.

cpp/src/streaming/cudf/table_chunk.cpp
```diff
@@ -161,6 +161,10 @@ TableChunk TableChunk::copy(MemoryReservation& reservation) const {
     // 2. The chunk is available and the data is a generic cudf table that is
     //    not already packed. In this case, the table data must first be packed
     //    before copying it to host or pinned memory.
+    //    a. reservaiton in pinned memory - Use cudf::pack to directly copy the table
+    //    data to pinned memory.
+    //    b. reservaiton in host memory - Use cudf::pack to copy the table data to
+    //    intermediate device memory and then copy to host memory.
```
pentschev:
```suggestion
    //    a. reservation in pinned memory - Use cudf::pack to directly copy the table
    //    data to pinned memory.
    //    b. reservation in host memory - Use cudf::pack to copy the table data to
    //    intermediate device memory and then copy to host memory.
```

cpp/src/streaming/cudf/table_chunk.cpp
```diff
@@ -184,9 +188,33 @@ TableChunk TableChunk::copy(MemoryReservation& reservation) const {
...
-        case MemoryType::PINNED_HOST:
-            // Case 2.
+        case MemoryType::PINNED_HOST:  // Case 2a.
+            if (packed_data_ == nullptr) {  // data is in device memory as a table
+                auto stream = this->stream();
+
+                StreamOrderedTiming timing{stream, br->statistics()};
+
+                // use cudf pack with pinned mr. When the pinned mr is warmed up, the
+                // performance
```
pentschev: the performance ...?

pentschev (approved):
Looks good overall, left some requests for typo fixes and phrasing improvements, and a nit.

cpp/src/memory/buffer_resource.cpp
nirandaperera: Yeah, I think that's reasonable.

cpp/src/streaming/cudf/table_chunk.cpp
nirandaperera: I've lost the train of thought! LOL.

pentschev (approved):
Thanks Niranda.

cpp/src/streaming/cudf/table_chunk.cpp
```diff
@@ -184,9 +188,34 @@ TableChunk TableChunk::copy(MemoryReservation& reservation) const {
...
+        case MemoryType::PINNED_HOST:  // Case 2a.
+            if (packed_data_ == nullptr) {  // data is in device memory as a table
+                auto stream = this->stream();
+
+                StreamOrderedTiming timing{stream, br->statistics()};
+
+                // use cudf pack with pinned mr, because it is more device memory
+                // efficient. When the pinned mr is warmed up, the performance is in par
+                // with packing to device memory and copying.
+                auto packed_pinned = cudf::pack(table_view(), stream, br->pinned_mr());
```
wence-:
```suggestion
                auto packed_pinned = cudf::pack(table_view(), stream, br->pinned_mr());
```

Let's write up the documentation about the performance of these approaches properly, rather than littering it in random comments.

cpp/src/streaming/cudf/table_chunk.cpp
wence-, line 178: question: should we fall through to the the single memcpy call in the case that we have data on device but it is in packed form?

cpp/src/memory/buffer_resource.cpp
```diff
@@ -196,14 +196,37 @@ std::unique_ptr<Buffer> BufferResource::allocate(
...
+        );
+
+        auto pinned_host_buffer = std::make_unique<HostBuffer>(
+            HostBuffer::from_rmm_device_buffer(std::move(data), stream, pinned_mr())
+        );
+        return std::unique_ptr<Buffer>(
+            new Buffer(std::move(pinned_host_buffer), stream, MemoryType::PINNED_HOST)
+        );
```
wence-, line 226: We need to think about this going forward. I don't really like the split in the type system between the pinned host buffer and the rmm::device_buffer storage.

Remind me why we need the pinned host buffer to be stored as a HostBuffer inside the Buffer object?

wence- (changes requested):

cpp/src/memory/buffer_resource.cpp
nirandaperera: >I don't really like the split in the type system between the pinned host buffer and the rmm::device_buffer storage.

This was required, because say we pass a pinned MR to a cudf operation, all allocations made by cudf are returned as rmm::device_buffer. So, we need a way to move it into a Host buffer. 

> Remind me why we need the pinned host buffer to be stored as a HostBuffer inside the Buffer object?

PinnedBuffer and HostBuffer were merged together because both impls were very similar. So, now both pinned and host data ptr is passed on to the `HostBuffer`. 
TBH I think we may be able to merge all buffer classes in to a single impl, with the rmm MR changes. Let's see.

cpp/src/streaming/cudf/table_chunk.cpp
nirandaperera: Hmm. That's a good point. @madsbk WDYT?
So, we can have a table chunk created from a table view that is already in packed in device memory.

nirandaperera: Or I think a more relevant one would be, copying a table chunk which has already been made available.

wence-: Yes, those are all statements of "table is on device but in packed form"?

nirandaperera: fixed it here https://github.com/rapidsai/rapidsmpf/pull/966/commits/5fe889b79330f7c8a6329d55f6653575f3ce443f

cpp/src/memory/buffer_resource.cpp
nirandaperera: @wence- this comment line it out-of-place now, because I changed the impl

cpp/include/rapidsmpf/memory/resource_types.hpp
```diff
@@ -16,4 +16,49 @@ using any_device_resource = cuda::mr::any_resource<cuda::mr::device_accessible>;
...
+ *
+ * Queries the resource's `dynamic_accessibility_property` and returns true if
+ * the reported accessibility is host-only or host-and-device.
+ *
+ * @tparam Properties The property pack of the resource reference.
+ * @param mr The memory resource reference to query.
+ * @return True if the resource is host-accessible, false otherwise.
+ */
+template <typename... Properties>
+[[nodiscard]] inline bool is_host_accessible(
```
madsbk:
```suggestion
[[nodiscard]] bool is_host_accessible(
```

cpp/include/rapidsmpf/memory/resource_types.hpp
```diff
@@ -16,4 +16,49 @@ using any_device_resource = cuda::mr::any_resource<cuda::mr::device_accessible>;
...
+ *
+ * Queries the resource's `dynamic_accessibility_property` and returns true if
+ * the reported accessibility is device-only or host-and-device.
+ *
+ * @tparam Properties The property pack of the resource reference.
+ * @param mr The memory resource reference to query.
+ * @return True if the resource is device-accessible, false otherwise.
+ */
+template <typename... Properties>
+[[nodiscard]] inline bool is_device_accessible(
```
madsbk:
```suggestion
[[nodiscard]] bool is_device_accessible(
```

cpp/src/streaming/cudf/table_chunk.cpp
```diff
@@ -152,24 +152,29 @@ bool TableChunk::is_spillable() const {
...
     //
-    // 1. The chunk is available and the reservation specifies device memory.
-    //    In this case, we can directly use cudf to create a deep copy of the
-    //    table by copying the table_view() into device memory.
+    // 1. The chunk is available and not yet packed. The table is copied/packed
+    //    into the reservation-specified memory type using libcudf:
+    //    a. DEVICE       - cudf-copy table_view() into device memory.
+    //    b. PINNED_HOST  - cudf::pack table_view() directly into pinned memory.
+    //    c. HOST         - cudf::pack table_view() into intermediate device
+    //                       memory and then copy to host memory.
```
madsbk:
```suggestion
    //                      memory and then copy to host memory.
```

madsbk (approved):
Looks good, thanks @nirandaperera

cpp/src/memory/buffer_resource.cpp
```diff
@@ -215,6 +216,15 @@ std::unique_ptr<Buffer> BufferResource::move(
         cuda_stream_join(stream, upstream);
         data->set_stream(stream);
     }
+
+    if (is_host_accessible(data->memory_resource())) {
+        auto pinned_host_buffer = std::make_unique<HostBuffer>(
```
pentschev, line 221: The [header](https://github.com/rapidsai/rapidsmpf/pull/966/files#diff-83a954644f6b0574df4bcc42158cfadef144acb428d5e7df001a8faa28c4b6eeR310) says "into a Buffer", not into a "Pinned Host Buffer". Why is it named `pinned_host_buffer` here? I know the name doesn't change anything in practice, but that is inconsistent I think.

pentschev (approved):
Minor question/nit, looks good otherwise, thanks Niranda.

cpp/src/memory/buffer_resource.cpp
nirandaperera: Hmm.. "into a Buffer" happens in L224-226 :-) We can only create a Buffer either by a `rmm::device_buffer` (which forces `DEVICE` mem type) or `HostBuffer`  for `HOST` or `PINNED`. There is some degree of ambiguity here. I have some plans to simplify it later.

cpp/include/rapidsmpf/memory/resource_types.hpp
```diff
@@ -16,4 +16,49 @@ using any_device_resource = cuda::mr::any_resource<cuda::mr::device_accessible>;
...
+ * @return True if the resource is host-accessible, false otherwise.
+ */
+template <typename... Properties>
+[[nodiscard]] bool is_host_accessible(
+    cuda::mr::resource_ref<Properties...> const& mr
+) noexcept {
+    // Unqualified call so ADL finds the hidden-friend `get_property` declared
+    // inside CCCL's type-erasure machinery for `resource_ref`. The qualified
+    // `cuda::mr::get_property` overload is SFINAE-disabled for type-erased
+    // wrappers via `__disable_default_dynamic_accessibility_property`.
```
wence-:
```suggestion
```

cpp/include/rapidsmpf/memory/resource_types.hpp
```diff
@@ -16,4 +16,49 @@ using any_device_resource = cuda::mr::any_resource<cuda::mr::device_accessible>;
...
+    auto const accessibility =
+        get_property(mr, cuda::mr::dynamic_accessibility_property{});
+    return accessibility == cuda::mr::__memory_accessibility::__host
+           || accessibility == cuda::mr::__memory_accessibility::__host_device;
```
wence-, line 36: Is there really no public way to do this?

cpp/include/rapidsmpf/memory/resource_types.hpp
```diff
@@ -16,4 +16,49 @@ using any_device_resource = cuda::mr::any_resource<cuda::mr::device_accessible>;
...
+template <typename... Properties>
+[[nodiscard]] bool is_device_accessible(
+    cuda::mr::resource_ref<Properties...> const& mr
+) noexcept {
+    // See `is_host_accessible` for why the call is unqualified (ADL).
+    auto const accessibility =
+        get_property(mr, cuda::mr::dynamic_accessibility_property{});
+    return accessibility == cuda::mr::__memory_accessibility::__device
+           || accessibility == cuda::mr::__memory_accessibility::__host_device;
+}
```
wence-, line 58: You can avoid the dynamic lookup if you `if constexpr (has_property<mr, device_accessible>)` otherwise fallback?

wence- (approved):
Tiny suggestions

cpp/include/rapidsmpf/memory/resource_types.hpp
nirandaperera: Unfortunately no (at least not that I know of)

cpp/include/rapidsmpf/memory/resource_types.hpp
nirandaperera: I think `get_property` does exactly that https://github.com/NVIDIA/cccl/blob/main/libcudacxx/include/cuda/__memory_resource/get_property.h#L191-L196

PR #967, opened 2026-04-13
[MINOR] Use cudf::packed_size to estimate the memory usage of a table.

### LINKED ISSUE #847: Use `cudf::packed_size()` for reserving memory for packing
When https://github.com/rapidsai/cudf/issues/21321 merges, use that for reserving memory and remove the wiggle-room calculations

nirandaperera:
This change enables `estimated_memory_usage` to return exact packed size. Then we wouldn't need to pad the reservations. 

Closes #847


cpp/src/integrations/cudf/utils.cpp
```diff
@@ -136,15 +137,8 @@ std::size_t estimated_memory_usage(
...
-    return std::transform_reduce(
-        tbl.begin(),
-        tbl.end(),
-        std::size_t{0},
-        std::plus{},
-        [&stream](cudf::column_view const& col) {
-            return estimated_memory_usage(col, stream);
-        }
-    );
+    // cudf::packed_size gives an exact estimate of the memory usage of the table.
```
madsbk: Is it exact or a estimate? :)
If it is exact, let's just remove `estimated_memory_usage` and use it `cudf::packed_size` directly?

nirandaperera:
@madsbk I ended up keeping `estimated_memory_usage` and introduced a new util that inlines `cudf::packed_size`. 
The reason was, we use `estimated_memory_usage` before reserving memory for a cudf table. I think it would be better representative than packed_size for tables. WDYT?

madsbk (approved):

PR #969, opened 2026-04-15
Migrate from pynvml to cuda.core.system

mdboom:
Removes the `nvidia-ml-py` dependency in favor of `cuda-core`.
- Contributes to https://github.com/NVIDIA/cuda-python/issues/1915


madsbk (approved):
Thanks @mdboom

pentschev (approved):

python/rapidsmpf/rapidsmpf/tests/test_ray.py
```diff
@@ -129,10 +129,7 @@ class NonRapidsMPFActor: ...
 
 @toolz.memoize
 def get_gpu_count() -> int:
-    pynvml.nvmlInit()
-    device_count = int(pynvml.nvmlDeviceGetCount())  # casting Any to int
-    pynvml.nvmlShutdown()
-    return device_count
+    return system.get_num_devices()  # type: ignore
```
bdice, line 132: Kind of surprising to me that we have been carrying this dependency for just one function, to get the number of devices.

bdice (approved):

Matt711 (approved):

python/rapidsmpf/rapidsmpf/tests/test_ray.py
mdboom: 100% if you want to do this in some other way, I won't be offended :)

PR #971, opened 2026-04-16
Report failures programmatically in `rrun.bind()`

### LINKED ISSUE #963: [FEA] `rrun.bind()` should report failures programmatically
`rrun.bind()` currently prints warnings to `stderr` when individual binding subsystems (CPU affinity, NUMA memory, network devices) fail, but provides no programmatic way to detect these failures. As a result, callers cannot distinguish between a fully successful bind and a partial failure.

**Current behavior**

* `bind(verbose=False)`: failures are silently ignored
* `bind(verbose=True)`: failures are printed to `stderr`, but the function still returns `None` with no indication of what succeeded or failed

**Desired behavior**

`bind()` should expose per-subsystem success or failure, either via a return value or an exception. This would allow callers to:

1. Fail fast when binding is required (for example, in production multi-node runs where NUMA locality matters)
2. Emit structured diagnostics without parsing `stderr`

**Suggested approach**

Return a structured result:

```python
@dataclasses.dataclass
class BindResult:
    cpu: bool      # CPU affinity applied
    memory: bool   # NUMA memory policy applied
    network: bool  # UCX_NET_DEVICES set
```

Alternatively, raise a `BindError` with per-subsystem details when any requested binding fails, instead of printing to `stderr`.

pentschev:
Instead of writing failures to `std::cerr`, raise an exception in proper programmatic fashion.

Closes #963


cpp/include/rrun/rrun.hpp
```diff
@@ -53,8 +52,10 @@ struct bind_options {
  * When `std::nullopt`, the first GPU in `CUDA_VISIBLE_DEVICES` is used instead.
  * @param options Controls which resource bindings to apply.
  *
- * @throws std::runtime_error if no GPU ID can be determined or the resolved
- *         GPU is not found in the discovered topology.
+ * @throws std::runtime_error if no GPU ID can be determined, topology
+ *         discovery fails, the resolved GPU is not found in the discovered
+ *         topology, or an enabled binding (CPU affinity, NUMA memory policy)
+ *         could not be applied.
```
madsbk: Remove indentation

cpp/include/rrun/rrun.hpp
```diff
@@ -80,8 +81,9 @@ void bind(
  * GPU in `CUDA_VISIBLE_DEVICES` is used instead.
  * @param options Controls which resource bindings to apply.
  *
- * @throws std::runtime_error if no GPU ID can be determined or the resolved
- *         GPU is not found in @p topology.
+ * @throws std::runtime_error if no GPU ID can be determined, the resolved
+ *         GPU is not found in @p topology, or an enabled binding (CPU
+ *         affinity, NUMA memory policy) could not be applied.
```
madsbk: Remove indentation

madsbk (approved):
Looks good, thanks @pentschev

pentschev:
Thanks Mads.

PR #972, opened 2026-04-16
Add functions for resource binding verification to `librrun`

pentschev:
RapidsMPF currently includes a CLI tool `check_resource_binding` for live detection of hardware resource bindings. This functionality can also be useful as a library so that other projects can programmatically retrieve the information, thus move the implementation to `librrun` and adapt the tool to reuse that.

Additionally, make optional use of that verification in `bind()` functions, to strengthen resource binding confidence.


python/rapidsmpf/rapidsmpf/rrun/rrun.pyx
```diff
@@ -43,12 +76,186 @@ cdef _sync_c_env_to_python():
...
+    numa_ok: bool
+    ucx_ok: bool
+    expected_ucx_devices: str
+
+    def all_passed(self) -> bool:
+        """Return ``True`` if all validation checks passed."""
+        return self.cpu_ok and self.numa_ok and self.ucx_ok
+
+
+def check_binding(gpu_id_hint=-1):
```
madsbk: What about using `None` as the _not-available_ value?

cpp/include/rrun/rrun.hpp
```diff
@@ -23,8 +25,96 @@ struct bind_options {
...
+ * Looks up @p gpu_id in @p topology and returns the expected CPU affinity,
+ * memory binding, and network devices.
+ *
+ * @param topology Pre-discovered system topology.
+ * @param gpu_id GPU device index to look up.
+ * @return The expected binding, or `std::nullopt` if @p gpu_id is not found.
+ */
+std::optional<expected_binding> get_expected_binding(
+    cucascade::memory::system_topology_info const& topology, int gpu_id
+);
```
pentschev, line 105: This is currently not exposed to Python because cuCascade has no Python bindings. If we find a proper use case for this we can review this decision and add bindings ourselves.

python/rapidsmpf/rapidsmpf/rrun/rrun.pyx
pentschev: Good idea, I just did a one-to-one translation without much though. Improved this in https://github.com/rapidsai/rapidsmpf/pull/972/commits/eaa51c4effcf1fe3a24a7900508261c3bf52b891.

cpp/include/rrun/rrun.hpp
```diff
@@ -23,8 +25,96 @@ struct bind_options {
...
+ * Queries the current CPU affinity, NUMA memory nodes, UCX network device
+ * configuration, process rank, and GPU information.
+ *
+ * @param gpu_id_hint GPU device index hint. When >= 0 the value is stored
+ * directly; otherwise the GPU ID is read from `CUDA_VISIBLE_DEVICES`.
+ * When a valid GPU ID is available, the PCI bus ID is also queried.
+ *
+ * @return The collected resource binding.
+ */
+resource_binding check_binding(int gpu_id_hint = -1);
```
madsbk, line 91: Is this API only for runs launched via. `rrun`? If so, we should state it somewhere with big letters.

cpp/src/rrun/rrun.cpp
madsbk: If `discovery.discover` can throw, we use restore `CUDA_VISIBLE_DEVICES` on error. Maybe use `ScopedEnvVar` from the test?

cpp/include/rrun/rrun.hpp
pentschev: I'm not sure I understand the question, are you asking if this is currently only used by the `rrun` CLI?

cpp/src/rrun/rrun.cpp
pentschev: Great catch! Fixed in https://github.com/rapidsai/rapidsmpf/pull/972/commits/e2ca8fba3ce9dcf63d5934c2b7b9e54df770a481 .

cpp/include/rrun/rrun.hpp
madsbk: Maybe I’m just mixing things up, but is this API intended for code launched via `rrun`, or can it also be used by workers started through other means, like when using `rrun.bind()`?

pentschev: Yes, that's the idea. One could use this function to retrieve bindings at any point during runtime for validation, reporting, etc., including if `bind()` is not used whatsoever.

madsbk: I think `check_binding()` always falls outside of a `rrun` launch because it calls `rapidsmpf::bootstrap::get_rank()` ?

pentschev: But then in that case `resource_binding.rank == -1` (not available) and there will be no errors on the caller's side.

madsbk: Will `rapidsmpf::bootstrap::get_rank();` not raise?

pentschev: It does and it's already properly considered in https://github.com/rapidsai/rapidsmpf/pull/972/files#diff-8b9a83bbd8a3f3366fe75934ab65c35485f97140d9ead64566b1547e4f6e2a11R357-R360 .

madsbk: I see it now, thanks :)

madsbk (approved):
Thanks @pentschev

pentschev:
Thanks Mads!

PR #974, opened 2026-04-17
Enable adding setup statistics 

nirandaperera:
This PR enables adding setup stats. Main usecase is to record pinned pool creation time. But we can add more stats later if needed

### DIFF
--- a/cpp/include/rapidsmpf/statistics.hpp
+++ b/cpp/include/rapidsmpf/statistics.hpp
@@ -394,6 +394,14 @@ class Statistics {
     Stat get_stat(std::string const& name) const;
 
     /**
+     * @brief Retrieves all statistics by prefix.
+     *
+     * @param prefix The prefix to filter statistics by.
+     * @return A vector of statistics matching the prefix.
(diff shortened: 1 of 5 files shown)


cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -518,6 +526,22 @@ class Statistics {
...
+     * @param duration The duration to record.
+     */
+    void record_setup_stat(std::string prefix, std::size_t nbytes, Duration duration);
+
+    /**
+     * @brief Get the prefixes of all setup statistics.
+     *
+     * @return A reference to the set of setup statistics prefixes.
+     */
+    std::unordered_set<std::string> const& get_setup_stats_prefixes() const;
```
madsbk, line 543: Can we avoid the new prefix semantic and just let the user specify a `mem_type`? I think the current design is general enough to handle this without introducing a special `setup_stats_prefixes_` handle?

nirandaperera:
Closing this in favor of using generic stats in applications

PR #976, opened 2026-04-17
Predefine statistics formatters

madsbk:
Previously, `Statistics` allowed callers to register arbitrary `std::function` formatters via `register_formatter()`. While flexible, this prevented full serialization: lambdas cannot be serialized, so `Statistics::serialize()` dropped formatter information.

This PR replaces dynamic formatters with a closed set of predefined render functions, identified by a new `Statistics::Formatter` enum (`Default`, `Bytes`, `Duration`, `Ratio`, `MemCopy`, `MemAlloc`).

In addition, `Statistics` is now fully picklable. As a result, the `gather_statistics` API is removed. Instead, `cudf-polars` is responsible for pickling and communicating `Statistics` objects directly.


cpp/src/statistics.cpp
```diff
@@ -147,56 +209,37 @@ void Statistics::add_stat(std::string const& name, double value) {
...
-            os << format_nbytes(val);
-            if (count > 1) {
-                os << " | avg " << format_nbytes(val / count);
-            }
-        });
-    }
     add_stat(name, static_cast<double>(nbytes));
+    if (!has_report_entry(name)) {
+        add_report_entry(name, {name}, Formatter::Bytes);
+    }
```
vyasr: Are you ever not wrapping `add_report_entry` in a `has_report_entry` check? Isn't that redundant given the semantics that you've established where add does nothing if there's already something present?

vyasr (approved):
One tiny question, otherwise LGTM.

cpp/src/statistics.cpp
madsbk: It was only to avoid the extra `{name}` construction.

I’ve now addressed this by adding an overload that takes a `std::initializer_list<std::string_view>`.

cpp/src/statistics.cpp
```diff
@@ -147,56 +209,51 @@ void Statistics::add_stat(std::string const& name, double value) {
...
     }
     std::lock_guard<std::mutex> lock(mutex_);
-    formatters_.try_emplace(report_entry_name, stat_names, std::move(formatter));
+    if (report_entries_.contains(report_entry_name)) {
+        return;
+    }
+    report_entries_.try_emplace(
+        report_entry_name,
+        ReportEntry{.stat_names = std::move(stat_names), .formatter = formatter}
+    );
```
madsbk, line 236: This is essentially a duplicate of the `std::initializer_list<std::string_view> stat_names` overload, but I think it’s simpler than introducing a helper function.

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -396,65 +410,56 @@ class Statistics {
...
+    void add_report_entry(
+        std::string const& report_entry_name,
+        std::initializer_list<std::string_view> stat_names,
+        Formatter formatter
+    );
 
+    // clang-format off
     /**
-     * @brief Register a formatter that takes multiple named statistics.
+     * @copydoc add_report_entry(std::string const&,std::initializer_list<std::string_view>, Formatter)
```
pentschev, line 443:
```suggestion
     * @copydoc add_report_entry(std::string const&, std::initializer_list<std::string_view>, Formatter)
```

cpp/src/statistics.cpp
```diff
@@ -16,26 +17,87 @@
...
+        },
+        // Implement `Statistics::Formatter:MemCopy`
+        [](std::ostream& os, std::vector<Statistics::Stat> const& s) {
+            RAPIDSMPF_EXPECTS(
+                s.at(0).count() == s.at(1).count() && s.at(1).count() == s.at(2).count(),
+                "MemCopy formatter expects the record counters to match"
+            );
+            os << format_nbytes(s.at(0).value()) << " | "
+               << format_duration(s.at(1).value()) << " | "
+               << format_nbytes(s.at(0).value() / s.at(1).value()) << "/s"
```
pentschev: Not sure we need or should handle this corner case, but if `s.at(1).value()` is `0.0`, this will result in `inf`/`nan`.

cpp/src/statistics.cpp
```diff
@@ -16,26 +17,87 @@
...
+        },
+        // Implement `Statistics::Formatter:MemAlloc`
+        [](std::ostream& os, std::vector<Statistics::Stat> const& s) {
+            RAPIDSMPF_EXPECTS(
+                s.at(0).count() == s.at(1).count() && s.at(1).count() == s.at(2).count(),
+                "MemAlloc formatter expects the record counters to match"
+            );
+            os << format_nbytes(s.at(0).value()) << " | "
+               << format_duration(s.at(1).value()) << " | "
+               << format_nbytes(s.at(0).value() / s.at(1).value()) << "/s"
```
pentschev, line 84: Ditto.

cpp/src/statistics.cpp
```diff
@@ -16,26 +17,87 @@
...
+               << " | avg-stream-delay "
+               << format_duration(s.at(2).value() / static_cast<double>(s.at(1).count()));
+        },
```
pentschev: `MemCopy` and `MemAlloc` have identical signatures and implementations, the only difference is `MemCopy`/`MemAlloc` in the error message. Do we really need both or could we combine both under a general name (such as `Memory` or `MemoryTransaction`)?

cpp/src/statistics.cpp
```diff
@@ -147,56 +209,51 @@ void Statistics::add_stat(std::string const& name, double value) {
...
-            auto const count = stats[0].count();
-            os << format_duration(val);
-            if (count > 1) {
-                os << " | avg " << format_duration(val / count);
-            }
-        });
-    }
     add_stat(name, seconds.count());
+    add_report_entry(name, {name}, Formatter::Duration);
 }
```
pentschev, line 247: Note: both `add_stat` and `add_report_entry` acquire `mutex_`. This is not a bug but inefficient, as you lock->release->lock->release in both `add_bytes_stat` and `add_report_entry`.

cpp/src/statistics.cpp
```diff
@@ -16,26 +17,87 @@
...
-}  // namespace
+using NamesArray = std::array<Names, MEMORY_TYPES.size()>;
+using Names2DArray = std::array<NamesArray, MEMORY_TYPES.size()>;
+
+// Predefined render functions indexed by `Statistics::Formatter`. Per-entry
+// rendering description lives on the enum in statistics.hpp.
+using FormatterFn = void (*)(std::ostream&, std::vector<Statistics::Stat> const&);
+constexpr std::array<FormatterFn, static_cast<std::size_t>(Statistics::Formatter::_Count)>
+    FORMATTERS = {{
+        // Implement `Statistics::Formatter:Default`
```
pentschev:
```suggestion
        // Implement `Statistics::Formatter::Default`
```

cpp/tests/test_statistics.cpp
```diff
@@ -84,81 +90,51 @@ TEST_F(StatisticsTest, StatMax) {
...
+    stats.add_stat("copy-summary-stream-delay", 0.0001);
+    EXPECT_THAT(stats.report(), ::testing::HasSubstr("copy-summary"));
     EXPECT_THAT(stats.report(), ::testing::HasSubstr("1 MiB"));
-    // The component stats should not appear as individual report entries.
-    EXPECT_THAT(stats.report(), ::testing::Not(::testing::HasSubstr("spill-bytes")));
-    EXPECT_THAT(stats.report(), ::testing::Not(::testing::HasSubstr("spill-time")));
+    // Component stats are consumed by the report entry and don't emit
+    // their own lines.
+    EXPECT_THAT(
+        stats.report(), ::testing::Not(::testing::HasSubstr("copy-summary-bytes:"))
```
pentschev:
```suggestion
        stats.report(), ::testing::Not(::testing::HasSubstr("copy-summary-bytes"))
```

pentschev: Follow the same matching pattern as other checks.

cpp/include/rapidsmpf/statistics.hpp
madsbk: The docs need the full function signature on one line.

cpp/src/statistics.cpp
madsbk: I think `inf` is fine

cpp/src/statistics.cpp
madsbk: Done

cpp/src/statistics.cpp
madsbk: I think this is fine for the sake of simplicity.

cpp/src/statistics.cpp
pentschev: Is there a point in keeping both `MemCopy` and `MemAlloc` `Formatter`s separate instead of the unifying them as well?

pentschev (approved):
LGTM.

cpp/src/statistics.cpp
madsbk: Ha, very good point indeed, I don’t think there is 🙂
Let’s call it `MemoryThroughput`.

pentschev (approved):
Thanks Mads.

PR #983, opened 2026-04-22
Fix `PostBox::spill()` passing stale amount to `std::ranges::lower_bound`

Matt711:
Prevents over spilling.

Each iteration of the spill loop passed the original `amount` to `std::ranges::lower_bound` instead of `amount - total_spilled`. After the first chunk was spilled, the search always found no chunk large enough and fell back to the largest remaining chunk rather than the smallest one covering the remaining need.

xref https://github.com/rapidsai/rapidsmpf/pull/984


Matt711:
/ok to test https://github.com/rapidsai/rapidsmpf/pull/983/commits/013d8577e7cee6d1519abb3c146088169a3e2579

madsbk (approved):
Good catch!

pentschev (approved):

cpp/tests/test_allgather.cpp
```diff
@@ -352,3 +356,39 @@ TEST_F(BaseAllGatherTest, opid_reuse) {
...
+            rapidsmpf::coll::detail::Chunk::INVALID_RANK,
+            rapidsmpf::PackedData{std::move(metadata), std::move(data)}
+        );
+    };
+
+    postbox.insert(make_chunk(20));
+    postbox.insert(make_chunk(80));
+    postbox.insert(make_chunk(90));
+
+    // Correct: 90+20=110. Bug: 90+80=170.
```
wence-:
```suggestion
```

cpp/tests/test_allgather.cpp
```diff
@@ -352,3 +356,39 @@ TEST_F(BaseAllGatherTest, opid_reuse) {
...
 }
+
+// Test that PostBox::spill() tracks the remaining spill need correctly across iterations,
+// rather than passing the original amount to lower_bound every time.
+//
+// With chunks [20, 80, 90] and a request for 100 bytes: the first iteration spills 90
+// (the largest chunk, since none covers 100 alone). The second must search for a chunk
+// >= 10 and pick the 20-byte chunk, totalling 110. The bug passed 100 to lower_bound on
+// every iteration, found nothing >= 100, and fell back to the 80-byte chunk instead,
+// totalling 170.
```
wence-:
```suggestion
// >= 10 and pick the 20-byte chunk, totalling 110.
```

wence- (approved):
Thanks, with minor suggestions for comments. I don't think it makes sense to document in the comments what used to happen.

Matt711:
/ok to test https://github.com/rapidsai/rapidsmpf/pull/983/commits/ff47c68975b3c1f733544d833f5a321d1b1f67e3

PR #985, opened 2026-04-22
Refactor memory resource ownership and type safety

### LINKED ISSUE #978: Use `PinnedMemoryResource` with `shared_resource` rather than `shared_ptr`

nirandaperera:
### `RmmResourceAdaptorImpl` — templated on primary and fallback resources

- Add `PrimaryMR` and `FallbackMR` template parameters (both constrained to
  `cuda::mr::resource_with<device_accessible>`; `FallbackMR` defaults to
  `any_resource<device_accessible>`)
- Move all method implementations into the header to support full template instantiation
- Add `explicit RmmResourceAdaptorImpl(std::in_place_t, Args&&...)` constructor to
  support in-place construction of non-movable types (e.g. `cuda::pinned_memory_pool`)

### `PinnedMemoryResource` — eliminate redundant heap allocation

- Inherit from `cuda::mr::shared_resource<RmmResourceAdaptorImpl<cuda::pinned_memory_pool>>`
  directly, storing the pool inside the shared control block rather than through a
  separate `RmmResourceAdaptor` member
- Remove virtual methods from `HostMemoryResource`, making it a standalone value-semantic type
- Add default-alignment `allocate`/`deallocate` overloads so call sites omitting the
  alignment argument continue to work

### Value-semantic factory methods

- `PinnedMemoryResource::make_if_available` and `from_options` now return
  `std::optional<PinnedMemoryResource>` instead of `std::shared_ptr<PinnedMemoryResource>`
- `PinnedMemoryResource::Disabled` changed from `nullptr` to `std::nullopt_t` — existing
  `== Disabled` comparisons remain valid via `optional::operator==(nullopt_t)`
- `BufferResource` and `Statistics` members/parameters updated from `shared_ptr` to
  `std::optional`
- `BufferResource::host_mr_` reverted to the concrete `HostMemoryResource` type

### Cython bindings

- `PinnedMemoryResource._handle` changed from `shared_ptr[cpp_PinnedMemoryResource]` to
  `optional[cpp_PinnedMemoryResource]`, matching the C++ value semantics throughout
- `buffer_resource.pyx` and `statistics.pyx` updated accordingly

Closes #978


cpp/include/rapidsmpf/memory/pinned_memory_resource.hpp
```diff
@@ -129,11 +120,12 @@ class PinnedMemoryResource final : public HostMemoryResource {
...
      *
-     * @return A shared pointer to the constructed PinnedMemoryResource instance.
+     * @return A `PinnedMemoryResource` if pinned memory is enabled and supported,
+     * otherwise `std::nullopt`.
      */
-    static std::shared_ptr<PinnedMemoryResource> from_options(config::Options options);
+    static std::optional<PinnedMemoryResource> from_options(config::Options options);
 
-    ~PinnedMemoryResource() override;
+    ~PinnedMemoryResource();
```
madsbk: I think "Rule of Zero" applies here, let the compiler generate all five special members.
```suggestion
```

madsbk (approved):
Thanks @nirandaperera, looks good. I only have one change request

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
```diff
@@ -31,19 +36,50 @@ namespace rapidsmpf::detail {
...
+ * resource type (e.g. `cuda::pinned_memory_pool`) to store the resource
+ * directly inside the shared control block, avoiding an extra heap allocation.
+ * @tparam FallbackMR The type of the optional fallback memory resource, used
+ * when the primary resource throws `rmm::out_of_memory`. Defaults to
+ * `cuda::mr::any_resource<cuda::mr::device_accessible>` for type-erased use.
  */
+template <
+    cuda::mr::resource_with<cuda::mr::device_accessible> PrimaryMR,
+    cuda::mr::resource_with<cuda::mr::device_accessible> FallbackMR =
+        cuda::mr::any_resource<cuda::mr::device_accessible>>
```
wence-, line 51: This somehow seems needlessly complicated. Why not just take `any_resource`?

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
```diff
@@ -53,31 +89,77 @@ class RmmResourceAdaptorImpl {
...
+        if (this == &other)
+            return true;
+        // Manual fallback comparison to avoid CCCL 3.2 constraint recursion:
+        // optional::operator== triggers infinite concept checking when the wrapped
+        // type inherits from a CCCL resource_ref.
+        bool fallbacks_equal =
+            (fallback_mr_.has_value() == other.fallback_mr_.has_value())
+            && (!fallback_mr_.has_value() || (*fallback_mr_ == *other.fallback_mr_));
+        return primary_mr_ == other.primary_mr_ && fallbacks_equal;
+    }
```
wence-, line 101: I thought these issues were fixed and we should just be able to do `return this == std::addressof(other)`?

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
```diff
@@ -53,31 +89,77 @@ class RmmResourceAdaptorImpl {
...
 
-    /// @copydoc RmmResourceAdaptor::get_upstream_resource
-    [[nodiscard]] rmm::device_async_resource_ref get_upstream_resource() const noexcept;
+    /**
+     * @brief Returns a reference to the primary upstream resource.
+     * @return Reference to the primary resource.
+     */
+    [[nodiscard]] PrimaryMR const& get_upstream_resource() const noexcept {
+        return primary_mr_;
+    }
```
wence-, line 109: No, come on, return the type-erased thing like before.

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
```diff
@@ -117,7 +260,11 @@ class RmmResourceAdaptorImpl {
      */
     void* allocate_sync(
         std::size_t bytes, std::size_t alignment = rmm::CUDA_ALLOCATION_ALIGNMENT
-    );
+    ) {
+        auto* ptr = allocate(cuda::stream_ref{cudaStream_t{nullptr}}, bytes, alignment);
+        RAPIDSMPF_CUDA_TRY(cudaStreamSynchronize(cudaStream_t{nullptr}));
```
wence-: I think we should use an internal stream (stored in the adaptor) rather than the null stream.

cpp/include/rapidsmpf/memory/pinned_memory_resource.hpp
```diff
@@ -82,31 +81,24 @@ struct PinnedPoolProperties {
 /**
  * @brief Memory resource that provides pinned (page-locked) host memory using a pool.
  *
+ * Inherits from
+ * `cuda::mr::shared_resource<RmmResourceAdaptorImpl<cuda::pinned_memory_pool>>`, which
+ * holds the pool directly inside the shared control block — no extra heap allocation for
+ * the pool itself. Copies share the same underlying pool and memory statistics.
```
wence-, line 87: This aspect seems like it is just not worth it.

wence- (changes requested):

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
nirandaperera: `cuda::mr::pinned_pool` is not copyable or moveable. So, we cant pass it to an `any_resource`. If the adapter impl accepts an `any_resource`, we have to push the pinned pool into a shared_resource. Adapter impl is anyway pushed into a shared_resource. IMO this _shared resource inside another shared resource_ is unnecessary. Templating the resources allows us to construct the pinned pool alongside the adapter impl shared resource.

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
nirandaperera: I only moved the previous impl here. But I think you have a point. 
Consider the following scenario, 
```c++
    rmm::mr::cuda_memory_resource cuda_mr{};

    RmmResourceAdaptor adaptor_a{cuda_mr};
    RmmResourceAdaptor adaptor_b{cuda_mr};

    EXPECT_EQ(adaptor_a, adaptor_b); //1
    EXPECT_EQ(adaptor_a, RmmResourceAdaptor{adaptor_a}); //2
```
Currently this equality holds. If we choose `this == std::addressof(other)`, I think `//1` fails. 

But each adapter has its own set of counters for the same upstream resource. So, should we check the equality based on upstreams or based on adapter impl objects?
@madsbk WDYT?

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
nirandaperera: This is `detail` namespace. Public adapter returns the type erased ref. Otherwise I need to specialize on templates :slightly_frowning_face:

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
bdice: @nirandaperera Do you mean [cuda::pinned_memory_pool](https://nvidia.github.io/cccl/unstable/libcudacxx/api/structcuda_1_1pinned__memory__pool.html)?

nirandaperera: @bdice Yes, I always get these names mixed up :D

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
wence-: What do we want equality to mean? Do we want it to mean that object_a can be used interchangeably with object_b to allocate and deallocate pointers?

Or do we want it to mean the strong thing that the adaptors have the same state. I think the latter.

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
wence-: What do you mean, why can't this be 

```
[[nodiscard]] rmm::device_async_resource_ref get_upstream_resource()
```
?

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
```diff
@@ -53,31 +90,77 @@ class RmmResourceAdaptorImpl {
...
 
-    /// @copydoc RmmResourceAdaptor::get_fallback_resource
-    [[nodiscard]] std::optional<rmm::device_async_resource_ref>
-    get_fallback_resource() const noexcept;
+    /**
+     * @brief Returns a const reference to the optional fallback resource.
+     * @return Const reference to the optional fallback resource, or `std::nullopt` if
+     * no fallback is configured.
+     */
+    [[nodiscard]] std::optional<FallbackMR> const&
```
wence-, line 115: Again, same question here?

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
```diff
@@ -139,13 +289,17 @@ class RmmResourceAdaptorImpl {
...
+    std::optional<FallbackMR> fallback_mr_;
     std::unordered_set<void*> fallback_allocations_;
 
     ScopedMemoryRecord main_record_;
     std::unordered_map<std::thread::id, std::stack<ScopedMemoryRecord>> record_stacks_;
     std::unordered_map<void*, std::thread::id> allocating_threads_;
+
+    rmm::cuda_stream sync_stream_{
+        rmm::cuda_stream::flags::non_blocking
+    };  ///< Custom stream for synchronous allocations and deallocations.
```
wence-:
```suggestion
    };  ///< Stream for synchronous allocations and deallocations.
```

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
wence-: >  IMO this shared resource inside another shared resource is unnecessary. Templating the resources allows us to construct the pinned pool alongside the adapter impl shared resource.

This doesn't make sense to me because now when the `impl` returns `get_upstream_resource()` it can't return an `any_resource` (because it's not holding one). And so that if that upstream is a `cuda::pinned_memory_pool` we now return a const ref to that pinned memory pool. And then the public adaptor when it calls `get_upstream_resource()` must fail because it can't turn an pinned_memory_pool into an any_resource. What am I missing?

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
nirandaperera: @wence- I think I agree with you here. Adapters should be equal iff they have the same shared state. Let's move to `addressof` approach.

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
nirandaperera: `cuda::pinned_memory_pool` cant be implicitly converted to a `device_async_resource_ref`. Only `shared_resource<cuda::pinned_memory_pool>` supports this.

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
nirandaperera: Only `RmmResourceAdapter` uses those methods, `PinnedMemoryResource` doesnt. I will mark the templated get_resource* methods `protected` to mask them from public API. 

public adapter wont be able to accept a `cuda::pinned_memory_pool` itself. If someone wants to do this, they will have to wrap it with a `shared_resource<cuda::pinned_memory_pool>`. 
`PinneMemoryResource` on the other hand, bakes cuda::pinned_memory_pool into the adapter impl in-place. In PinneMemoryResource, we dont advertise `get_upstream_resource()` methods publicly. If anyone want to, they can implicitly convert PinneMemoryResource to device or host resource ref.

nirandaperera: I was wrong about marking methods `protected`'. It doesnt make sense because the adapter inherits `shared_resource<impl>`. But I still think the impl should return the typed refs.

nirandaperera:
@wence- Could you please go through this again? I am inclined to keep the templates in the `detail::AdapterImpl`, because IINM, we dont have a good way to create a resource in-place other than, wrapping it with a shared_resource.

cpp/benchmarks/bench_memory_resources.cpp
```diff
@@ -31,40 +31,70 @@ constexpr std::array<ResourceType, 3> RESOURCE_TYPES{
 };
 
 std::array<std::string, 3> const ResourceTypeStr{
-    "NewDelete", "HostMemoryResource", "PinnedMemoryResource"
+    "SystemMemoryResource", "HostMemoryResource", "PinnedMemoryResource"
```
wence-: Why change this enum label "system" is not the right word here.

cpp/include/rapidsmpf/detail/rmm_resource_adaptor_impl.hpp
wence-: OK, I think I buy this.

cpp/include/rapidsmpf/memory/pinned_memory_resource.hpp
```diff
@@ -82,31 +81,24 @@ struct PinnedPoolProperties {
...
  */
-class PinnedMemoryResource final : public HostMemoryResource {
-  public:
-    /// @brief Sentinel value used to disable pinned host memory.
-    static constexpr auto Disabled = nullptr;
+class PinnedMemoryResource final
+    : public cuda::mr::shared_resource<
+          detail::RmmResourceAdaptorImpl<cuda::pinned_memory_pool>> {
+    using shared_base = cuda::mr::shared_resource<
+        detail::RmmResourceAdaptorImpl<cuda::pinned_memory_pool>>;
```
wence-, line 97: OK, so the only question I now have here is that effectively this bakes that allocation with the pinned pool _always_ uses the `RMMResourceAdaptor`. I suppose we're going to be relying on that, and it's not different to the device MR which also has to have this wrapping?

cpp/src/memory/pinned_memory_resource.cpp
```diff
@@ -95,40 +92,19 @@ std::shared_ptr<PinnedMemoryResource> PinnedMemoryResource::from_options(
...
 std::function<std::int64_t()> PinnedMemoryResource::get_memory_available_cb() const {
     auto const max_pool_size = pool_properties_.max_pool_size.value_or(0);
     if (max_pool_size > 0) {
-        return LimitAvailableMemory{
-            pool_tracker_, safe_cast<std::int64_t>(max_pool_size)
-        };
+        auto const limit = safe_cast<std::int64_t>(max_pool_size);
+        auto tracker = *this;  // shared ownership — same underlying pool
+        return [tracker, limit]() { return limit - tracker.get().current_allocated(); };
     }
```
wence-: What does dereferencing `this` buy us here over `return [this, limit]() {...};`?

wence- (approved):
I have some minor questions/suggestions, but I don't think they are blocking. I am just about convinced that the added complexity of templating here is worth it.

cpp/benchmarks/bench_memory_resources.cpp
nirandaperera: Ah! my bad, I switched this to `rmm::system_memory_resource` earlier, and then reverted.

cpp/src/memory/pinned_memory_resource.cpp
nirandaperera: `this` would simply pass by ptr, isnt it? We want to increment the refcount before passing it in IINM.

cpp/include/rapidsmpf/memory/pinned_memory_resource.hpp
nirandaperera: Yes, I think so. We can provide a _mem-tracking-enabled_ device memory pool inside rapidsmpf that bakes in cuda mempool into the adapter. Do you think we should provide that?

PR #986, opened 2026-04-22
Add `TableChunk.into_packed_data`

Matt711:
We need this API to avoid unecessary copies in cudf-polars when `TableChunk`s to `PackedData` in cudf-polars.

Closes #933


Matt711:
/ok to test https://github.com/rapidsai/rapidsmpf/pull/986/commits/c9fa79846d3191d135c0d5d23f07888164e775cc

cpp/include/rapidsmpf/streaming/cudf/table_chunk.hpp
```diff
@@ -265,6 +265,25 @@ class TableChunk {
...
+     * is required.
+     * @return A unique pointer to the resulting `PackedData`.
+     *
+     * @throws std::invalid_argument If the data is not already packed and
+     * `is_available() == false`.
+     *
+     * @note After this call, this object is in a moved-from state; only reassignment,
+     * movement, or destruction are valid.
+     */
+    [[nodiscard]] std::unique_ptr<PackedData> to_packed_data(BufferResource* br);
```
madsbk: Maybe a better name would be something like `release_as_packed_data`, to make it clear that the table chunk is left in a moved-from state afterwards?

Matt711: That would also be misleading, no? Cause `to_packed_data` will copy if there's no `packed_data_` to release.

madsbk: Fair point, “release” implies handing over something that already exists, so it understates the copy in the table path. Let’s refocus on what actually matters:

1. After this call, the chunk must genuinely be in a *moved-from* state. Right now it isn’t, since `table_` and `owner_` are still alive in the table path. 

2. The API should make the destructive contract visible at the call site, either through naming (something with “release”) or via an rvalue-ref qualifier: `to_packed_data(BufferResource*) &&`, which forces `std::move(chunk).to_packed_data(br)` without committing to a specific naming convention. Maybe a better name would then be `into_packed_data() &&`.

Matt711: That makes sense. I like `into_packed_data`

Matt711: Done.

cpp/src/streaming/cudf/table_chunk.cpp
```diff
@@ -234,6 +234,22 @@ TableChunk TableChunk::copy(MemoryReservation& reservation) const {
     return TableChunk(std::make_unique<PackedData>(std::move(metadata), std::move(data)));
 }
 
+std::unique_ptr<PackedData> TableChunk::into_packed_data(BufferResource* br) {
```
wence-: *question*: Should the caller by required to provide a reservation, given that this might allocate? And/or specify memory types they would like the output to be in?

cpp/include/rapidsmpf/streaming/cudf/table_chunk.hpp
```diff
@@ -265,6 +265,25 @@ class TableChunk {
...
+     * is required.
+     * @return A unique pointer to the resulting `PackedData`.
+     *
+     * @throws std::invalid_argument If the data is not already packed and
+     * `is_available() == false`.
+     *
+     * @note After this call, this object is in a moved-from state; only reassignment,
+     * movement, or destruction are valid.
+     */
+    [[nodiscard]] std::unique_ptr<PackedData> into_packed_data(BufferResource* br);
```
wence-:
```suggestion
    [[nodiscard]] std::unique_ptr<PackedData> into_packed_data(BufferResource* br) &&;
```

cpp/src/streaming/cudf/table_chunk.cpp
wence-:
```suggestion
std::unique_ptr<PackedData> TableChunk::into_packed_data(BufferResource* br) && {
```

wence- (changes requested):

cpp/src/streaming/cudf/table_chunk.cpp
Matt711: I added the same TODO that we have for `TableChunk::copy`. Ie. the allocation is untracked by a memory reservation.

Matt711:
/ok to test https://github.com/rapidsai/rapidsmpf/pull/986/commits/18f1372360b97429cd0310f6dd150cc4c912c4c1

cpp/src/streaming/cudf/table_chunk.cpp
Matt711: I don't think the caller should have to provide anything. Let `into_packed_data` handle the reservation since we know exactly when we need to track the allocation.

python/rapidsmpf/rapidsmpf/streaming/cudf/table_chunk.pyx
```diff
@@ -534,6 +535,42 @@ cdef class TableChunk:
...
+
+        Warnings
+        --------
+        The original table chunk is released and must not be used after this call.
+        """
+        cdef unique_ptr[cpp_PackedData] result
+        cdef cpp_BufferResource* _br = br.ptr()
+        cdef unique_ptr[cpp_TableChunk] handle = self.release_handle()
+        with nogil:
+            result = deref(handle).into_packed_data(_br)
```
wence-:
```suggestion
            result = move(deref(handle)).into_packed_data(_br)
```

python/rapidsmpf/rapidsmpf/streaming/chunks/packed_data.pyx
```diff
@@ -37,7 +37,7 @@ cdef class PackedDataChunk:
         with nogil:
             self._handle.reset()
 
-    def to_packed_data(self):
+    def into_packed_data(self):
```
wence-: I don't think this name should be changed. This function does a slightly different thing, the chunk is always a `PackedData` and this releases it. There is no copy ever.


```suggestion
    def to_packed_data(self):
```

python/rapidsmpf/rapidsmpf/streaming/chunks/packed_data.pyi
```diff
@@ -9,7 +9,7 @@ from rapidsmpf.memory.packed_data import PackedData
 from rapidsmpf.streaming.core.message import Message
 
 class PackedDataChunk:
-    def to_packed_data(self) -> PackedData: ...
+    def into_packed_data(self) -> PackedData: ...
```
wence-:
```suggestion
    def to_packed_data(self) -> PackedData: ...
```

python/rapidsmpf/rapidsmpf/tests/streaming/test_allgather.py
```diff
@@ -76,7 +76,7 @@ def test_allgather_actor(context: Context, comm: Communicator) -> None:
 
     result = unpack_and_concat(
         (
-            PackedDataChunk.from_message(msg, br=context.br()).to_packed_data()
+            PackedDataChunk.from_message(msg, br=context.br()).into_packed_data()
```
wence-:
```suggestion
            PackedDataChunk.from_message(msg, br=context.br()).to_packed_data()
```

python/rapidsmpf/rapidsmpf/tests/streaming/test_allgather.py
```diff
@@ -126,7 +126,7 @@ async def allgather_and_concat(
 ) -> None:
     gather = AllGather(context, comm, op_id)
     while (msg := await ch_in.recv(context)) is not None:
-        chunk = PackedDataChunk.from_message(msg, br=context.br()).to_packed_data()
+        chunk = PackedDataChunk.from_message(msg, br=context.br()).into_packed_data()
```
wence-:
```suggestion
        chunk = PackedDataChunk.from_message(msg, br=context.br()).to_packed_data()
```

wence- (approved):

Matt711:
Thanks @wence-!

PR #987, opened 2026-04-22
Add SPMD multiprocessing pool integration (`rapidsmpf.integrations.mp`)

nirandaperera:
### Summary

Introduces a dependency-free SPMD (Single Program, Multiple Data) execution environment backed by Python `multiprocessing`, analogous to the existing Dask and Ray integrations but without those dependencies. Suitable for local multi-GPU testing and workloads that need guaranteed broadcast semantics.

### Motivation

`multiprocessing.Pool` provides no guarantee that a submitted task executes on *all* workers. This PR uses explicit `multiprocessing.Process` objects with per-worker `Queue` pairs so every `run()` call is truly broadcast — all `N` workers execute the same callable before results are returned to the caller.

#### `python/rapidsmpf/rapidsmpf/integrations/mp.py` *(new)*

- **`MultiprocessingPool`** — client-side API. Spawns `N` persistent worker processes using the `"spawn"` multiprocessing context (required for CUDA safety).
- **Worker lifecycle** (phases inside `_worker_main`): 1. `rrun.bind()` — topology-aware pinning to GPU, CPUs, NUMA nodes, and NICs. 2. `rmpf_worker_local_setup()` — initialises RMM and a per-worker `WorkerContext`. 3. UCXX communicator bootstrap — root-worker exchange of `UCXAddress`, then `barrier`. 4. Persistent event loop — receives `cloudpickle`-serialised tasks, executes them, and returns results or exceptions via result queues. 5. Cleanup — unregisters Python spill callback.
- **`MultiprocessingPool.run(func, *args, **kwargs)`** — broadcasts the callable to all workers, drains *all* result queues before raising, and aggregates per-worker failures into a standard `ExceptionGroup`.
- **`MultiprocessingPool.shutdown()`** — sends a poison-pill sentinel to each worker and joins the processes.
- **Thread safety** — `run()` and `shutdown()` are protected by an instance `threading.Lock`.
- **Oversubscription** — `gpu_devices` defaults to round-robin across available GPUs; more workers than GPUs is explicitly supported.
- **`get_worker_context()`** — public helper (analogous to `rapidsmpf.integrations.single.get_worker_context`) for retrieving the per-process `WorkerContext` from inside a worker task.
- **`sys.path` forwarding** — the parent's `sys.path` (with `""` normalised to `Path.cwd()`) is forwarded to workers so that `cloudpickle`-serialised module-level functions resolve correctly under `pytest --import-mode=importlib`.

#### `python/rapidsmpf/rapidsmpf/tests/test_mp.py` *(new)*

- Module-scoped `mp_pool` fixture parametrised over `nranks ∈ {1, 2, 4}`.
- **`test_mp_pool_ucxx_cluster`** — verifies communicator rank/nranks after bootstrap.
- **`test_mp_pool_worker_context`** — verifies `WorkerContext` is accessible in workers.
- **`test_mp_pool_lambda`** / **`test_mp_pool_closure`** — verifies `cloudpickle` serialisation of lambdas and closures.
- **`test_mp_pool_error_propagation`** — all workers fail; checks `ExceptionGroup` contains `nranks` exceptions.
- **`test_mp_pool_error_propagation_partial`** — even-ranked workers fail; checks `ExceptionGroup` length equals `ceil(nranks / 2)`.
- **`test_mp_pool_multiple_tasks`** — verifies sequential `run()` calls do not interfere.
- **`test_mp_pool_resource_binding`** — verifies `rrun.check_binding()` passes inside workers.
- **`test_mp_pool_shuffle`** — parametrised distributed shuffle via `Shuffler`; row count preserved.
- **`test_mp_pool_allgather_multirun`** — five sequential `AllGather` calls; each rank contributes a unique table, all ranks receive the full concatenated result and verify it in-worker (no data returned to client).
- **`test_mp_pool_oversubscription`** — 2× oversubscription starts correctly.
- Lifecycle / configuration tests: `test_mp_pool_option_statistics`, `test_mp_pool_shutdown_idempotent`, `test_mp_pool_run_after_shutdown`, `test_mp_pool_repr`.

#### Dependency declarations

- **`dependencies.yaml`** — adds `cloudpickle` to `run_rapidsmpf`.
- **`python/rapidsmpf/pyproject.toml`** — adds `cloudpickle` to `[project.dependencies]`.

### Design notes

- `cloudpickle` is used instead of `pickle` to support lambdas, closures, and functions defined in `__main__` or under pytest's `importlib` import mode.
- The UCXX bootstrap follows the same root-worker address-exchange pattern used by the Ray integration.
- Failed `run()` calls fully drain all worker queues before raising, ensuring the pool is reusable after a partial failure.


nirandaperera:
Now that we are ironing out blockers with ray dependency, I dont think this will be very relevant moving forward.

PR #989, opened 2026-04-23
Add `to_dict()` to Python `Statistics`

madsbk:
Adds a `Statistics.to_dict()` method that returns a plain dict snapshot of all tracked statistics, shaped as `{name: {"count", "value", "max"}}`, matching the per-stat layout of `get_stat()`.

Implemented as a thin wrapper over the existing `write_json_string()` path, so the snapshot is atomic (the underlying C++ JSON writer holds the mutex once) and inherits the same name-validation rules. No C++ API changes.


python/rapidsmpf/rapidsmpf/statistics.pyx
```diff
@@ -283,6 +304,55 @@ cdef class Statistics:
...
+            entries.
+        """
+        cdef Statistics ret = Statistics.__new__(Statistics)
+        cdef vector[shared_ptr[cpp_Statistics]] v
+        for item in stats:
+            v.push_back((<Statistics?>item)._handle)
+        with nogil:
+            ret._handle = cpp_merge_statistics(v)
+        return ret
+
```
madsbk, line 355: Untouched, just moved to make the method order more natural

python/rapidsmpf/rapidsmpf/statistics.pyx
```diff
@@ -382,76 +455,21 @@ cdef class Statistics:
         ------
         OSError
             If the file cannot be opened or writing fails.
-        ValueError
-            If any stat name or memory record name contains a double quote, backslash,
-            or ASCII control character (0x00–0x1F).
```
madsbk, line 387: I’ve removed the mention of this error, as it felt misleading. While it’s true that using illegal characters in stat names would trigger it, that’s not something we do in practice and not something users need to worry about.

python/rapidsmpf/rapidsmpf/statistics.pyx
```diff
@@ -241,6 +242,26 @@ cdef class Statistics:
...
+        Report-entry and formatter metadata is not included; use
+        :meth:`report` or :meth:`write_json_string` for those.
+
+        Disabled statistics always return an empty dict.
+
+        Returns
+        -------
+        A ``dict`` mapping each stat name to its ``{"count", "value", "max"}`` dict.
+        """
+        return json.loads(self.write_json_string())["statistics"]
```
madsbk, line 263: This is the only real change to review

pentschev (approved):
LGTM.

PR #991, opened 2026-04-24
Add skill to reproduce CI locally

pentschev:
Introduce a skill that agents can use to assist in [reproducing CI workflows on the local machine](https://docs.rapids.ai/resources/reproducing-ci/).

To use the skill one can simply ask an agent something like:

* I need to debug some problems with the build of rapidsmpf wheels in CI. Please help me run it locally with the rapidsai/ci-wheel:26.06-cuda12.9.1-rockylinux8-py3.11 image.
* I need to debug some problems with the build of rapidsmpf wheels in CI. Please help me reproduce the failed CI job from https://github.com/rapidsai/rapidsmpf/actions/runs/24866449332/job/72832738708?pr=990.

Note that reproducing CI generally requires a GitHub token. Before launching the agent users can `export GH_TOKEN=ghp_...`, so it's capable of using that without asking for it later.

### DIFF
--- a/.agents/skills/reproduce-ci-locally/SKILL.md
+++ b/.agents/skills/reproduce-ci-locally/SKILL.md
@@ -1,0 +1,359 @@
+---
+name: reproduce-ci-locally
+description: >-
+  Reproduce RAPIDS CI builds and tests locally using Docker containers that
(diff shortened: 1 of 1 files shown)


madsbk (approved):
neat

pentschev:
Thanks Mads.

PR #992, opened 2026-04-24
Harden FileBackend key validation and temp file creation

pentschev:
Add input validation to `get_kv_path` to reject keys containing path separators (`/`, `\`), parent directory references (`..`), null bytes, or empty strings. This ensures that key-value paths are always confined to the intended coordination directory on shared filesystems.

Replace the predictable PID-based temporary file naming in `write_file` with `mkstemp()`, which atomically creates a uniquely-named file. This strengthens the atomic write pattern against race conditions on shared NFS mounts where multiple processes may operate concurrently. The write loop now also correctly handles partial writes via POSIX `write()`.

### DIFF
--- a/cpp/src/bootstrap/file_backend.cpp
+++ b/cpp/src/bootstrap/file_backend.cpp
@@ -4,6 +4,7 @@
  */
 
 #include <chrono>
+#include <cstdio>
 #include <cstdlib>
 #include <cstring>
 #include <filesystem>
@@ -224,8 +225,7 @@ void FileBackend::write_file(std::string const& path, std::string_view content) 
         if (written < 0) {
             int err = errno;
             ::close(fd);
-            std::error_code rm_ec;
-            std::filesystem::remove(tmp_path, rm_ec);
+            ::unlink(tmp_path.c_str());
             throw std::runtime_error(
                 "Failed to write to temporary file: " + tmp_path + ": "
                 + std::strerror(err)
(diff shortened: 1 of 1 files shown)


cpp/src/bootstrap/file_backend.cpp
```diff
@@ -193,22 +204,44 @@ bool FileBackend::wait_for_file(std::string const& path, Duration timeout) {
...
+        bytes_left -= static_cast<std::size_t>(written);
+        ptr += written;
     }
-    ofs.write(content.data(), static_cast<std::streamsize>(content.size()));
-    ofs.close();
+    ::close(fd);
 
     // Atomic rename
     std::error_code ec;
     std::filesystem::rename(tmp_path, path, ec);
```
wence-: This should probably be `rename(3)` rather than std::filesystem because the latter is not guaranteed to do the rename with posix semantics.

pentschev: Great catch. Fixed in https://github.com/rapidsai/rapidsmpf/pull/992/commits/90eeb528b9234a600ff64d1fad4d7da53fbc51b8 .

cpp/src/bootstrap/file_backend.cpp
```diff
@@ -193,24 +205,43 @@ bool FileBackend::wait_for_file(std::string const& path, Duration timeout) {
...
-    // Write to temporary file
-    std::ofstream ofs(tmp_path, std::ios::binary | std::ios::trunc);
-    if (!ofs) {
-        throw std::runtime_error("Failed to open temporary file: " + tmp_path);
+    // Write content and close the file descriptor
+    auto bytes_left = content.size();
+    auto const* ptr = content.data();
+    while (bytes_left > 0) {
+        auto written = ::write(fd, ptr, bytes_left);
+        if (written < 0) {
```
madsbk, line 225:
```suggestion
        if (written < 0) {
            if (errno == EINTR)  {
                continue;  // re-try on interrupt
            }
```

madsbk (approved):

cpp/src/bootstrap/file_backend.cpp
pentschev: Good call!

wence- (approved):

PR #993, opened 2026-04-24
[C++] Add OrderScheme partitioning metadata

rjzamora:
**NOTE**: This PR is the C++ component of https://github.com/rapidsai/rapidsmpf/pull/853

Adds mechanism to track ordering within ``ChannelMetadata``. Needed in cudf-polars to keep track of sorted data.

### Proposed API

```c++
struct OrderKey {
    cudf::size_type column_index;  ///< Column to sort on.
    cudf::order order;  ///< ASCENDING or DESCENDING.
    cudf::null_order null_order;  ///< BEFORE or AFTER.

    /**
     * @brief Equality comparison.
     * @return True if all fields are equal.
     */
    bool operator==(OrderKey const&) const = default;
};

struct OrderScheme {
    std::vector<OrderKey> keys;  ///< Sort keys (column, order, null_order per entry).
    std::unique_ptr<TableChunk> boundaries;  ///< N-1 boundary rows for N partitions.
    bool strict_boundaries{false};  ///< Whether unique keys must be mapped to a distinct chunk
```


cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -1,17 +1,22 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights
- * reserved. SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES.
```
wence-:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -31,6 +36,63 @@ struct HashScheme {
...
+ * When `boundaries` is set, its columns must align with `keys`
+ * (same count and compatible dtypes). Mismatched dtypes are a usage error.
+ *
+ * `strict_boundaries`: when true, every row in a chunk belongs to a single partition's
+ * half-open key range (partition keys do not straddle chunk interiors). When false,
+ * a chunk may contain keys spanning multiple partitions (e.g. before a shuffle).
+ */
+struct OrderScheme {
+    std::vector<OrderKey> keys;  ///< Sort keys (column, order, null_order per entry).
+    std::shared_ptr<TableChunk> boundaries;  ///< N-1 boundary rows for N partitions.
```
wence-, line 82: Are boundaries allowed to be null? I think no. But we don't ensure that anywhere.

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,19 +4,73 @@
...
+
+OrderScheme make_order_scheme(
+    std::vector<OrderKey> keys,
+    std::unique_ptr<TableChunk> boundaries,
+    bool strict_boundaries
+) {
+    RAPIDSMPF_EXPECTS(
+        !keys.empty(), "OrderScheme: keys must not be empty", std::invalid_argument
+    );
+    return OrderScheme{
```
wence-: I think the boundaries must be non-null, and must be available.

I think we should also assert that the number of keys matches the number of columns.

So let's implement a custom ctor for `OrderScheme` and do that. Rather than having this factory function.

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,19 +4,73 @@
...
+        [](Message const& msg, MemoryReservation& /*reservation*/) -> Message {
             auto copy = std::make_unique<ChannelMetadata>(msg.get<ChannelMetadata>());
-            return Message{msg.sequence_number(), std::move(copy), {}, msg.copy_cb()};
+            auto copy_cd = get_content_description(*copy);
+            return Message{
+                msg.sequence_number(), std::move(copy), copy_cd, msg.copy_cb()
+            };
         }
     };
 }
```
wence-:
```suggestion
Message to_message(std::uint64_t sequence_number, std::unique_ptr<ChannelMetadata> m) {
    return Message{
        sequence_number,
        std::move(m),
        {},
        [](Message const& msg, MemoryReservation& /*reservation*/) -> Message {
auto copy = std::make_unique<ChannelMetadata>(msg.get<ChannelMetadata>());
            return Message{
                msg.sequence_number(), std::move(copy), {}, msg.copy_cb()
            };
        }
    };
}
```

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,19 +4,73 @@
...
+        return true;
+    }
+    // Both have boundaries: shape only (see doc in header).
+    return boundaries->shape() == other.boundaries->shape();
+}
+
+std::unique_ptr<ChannelMetadata> channel_metadata_from_message(Message msg) {
+    return std::make_unique<ChannelMetadata>(msg.release<ChannelMetadata>());
+}
+
```
wence-:
```suggestion
```

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,19 +4,73 @@
...
+        .strict_boundaries = strict_boundaries,
+    };
+}
+
+bool OrderScheme::operator==(OrderScheme const& other) const {
+    if (keys != other.keys || strict_boundaries != other.strict_boundaries) {
+        return false;
+    }
+    bool this_has = boundaries != nullptr;
+    bool other_has = other.boundaries != nullptr;
```
wence-: As above, I think we _must_ have that the boundaries are non-null

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,19 +4,73 @@
...
+#include <rapidsmpf/error.hpp>
+#include <rapidsmpf/memory/memory_reservation.hpp>
 #include <rapidsmpf/streaming/cudf/channel_metadata.hpp>
 
 namespace rapidsmpf::streaming {
 
+PartitioningSpec PartitioningSpec::from_order(OrderScheme o) {
+    RAPIDSMPF_EXPECTS(
+        !o.keys.empty(), "OrderScheme: keys must not be empty", std::invalid_argument
+    );
```
wence-, line 48: If this is a requirement then the OrderScheme ctor should enforce it.

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -148,6 +221,40 @@ struct ChannelMetadata {
...
+ *
+ * For now this is non-spillable with zero tracked sizes: ORDER boundaries are
+ * expected to stay device-resident on metadata paths. Spill accounting for
+ * embedded boundaries can be added later without changing the Python API.
+ *
+ * @param m Channel metadata to describe.
+ * @return Content description with spillability off and zero-sized content.
+ */
+ContentDescription get_content_description(ChannelMetadata const& m);
+
```
wence-:
```suggestion
```

This is all unnecessary here.

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -31,6 +36,63 @@ struct HashScheme {
...
+ * a chunk may contain keys spanning multiple partitions (e.g. before a shuffle).
+ */
+struct OrderScheme {
+    std::vector<OrderKey> keys;  ///< Sort keys (column, order, null_order per entry).
+    std::shared_ptr<TableChunk> boundaries;  ///< N-1 boundary rows for N partitions.
+    /// See struct-level note on `strict_boundaries` semantics.
+    bool strict_boundaries{false};
+
+    /**
+     * @brief Shallow metadata equality (not semantic boundary value equality).
```
wence-:
```suggestion
     * @brief Shallow metadata equality without comparing boundary values.
```

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -31,6 +36,63 @@ struct HashScheme {
...
+     *
+     * Returns true when `keys` and `strict_boundaries` match, and boundary tables
+     * are consistent in the weak sense: both absent, or both present with the
+     * same shape from `TableChunk::shape()` (not `table_view()`, which requires an
+     * available device table).
+     * Cell values inside `boundaries` are intentionally not compared (that would
+     * require a device comparison API with stream and memory resource). Do not
+     * use `operator==` to assert that two schemes have identical range boundaries.
```
wence-:
```suggestion
     * Returns true when `keys` and `strict_boundaries` match, and boundary tables
     * have the same shape.
     * Cell values inside `boundaries` are intentionally not compared (that would
     * require a device comparison API with stream and memory resource). Do not
     * use `operator==` to assert that two schemes have identical range boundaries.
```

wence- (changes requested):

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
rjzamora: No - ctor checks this now.

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,17 +4,59 @@
...
+           && boundaries->shape() == other.boundaries->shape();
+}
+
 Message to_message(std::uint64_t sequence_number, std::unique_ptr<ChannelMetadata> m) {
     return Message{
         sequence_number,
         std::move(m),
         {},
-        [](Message const& msg, MemoryReservation& /* reservation */) -> Message {
+        [](Message const& msg, MemoryReservation& /*reservation*/) -> Message {
```
rjzamora:
```suggestion
        [](Message const& msg, MemoryReservation& /* reservation */) -> Message {
```

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,11 +4,53 @@
...
+    this->boundaries = std::move(boundaries);
+    this->strict_boundaries = strict_boundaries;
+}
+
+PartitioningSpec PartitioningSpec::from_order(OrderScheme o) {
+    return {.type = Type::ORDER, .hash = std::nullopt, .order = std::move(o)};
+}
+
+bool OrderScheme::operator==(OrderScheme const& other) const {
+    // boundaries is always non-null (enforced by constructor).
```
rjzamora:
```suggestion
```

rjzamora:
Thanks for the review @wence-. I think this is ready for another look

rjzamora:
While integrating the Python API into cudf-polars, I realized it was very convenient to have `replace_keys` and `boundaries_aligned_with` defined at the C++ level. In C++, we track the boundaries using `shared_ptr<TableChunk>`, but we don't really "allow" this at the Python level (without changing a bunch of Cython outside the scope of this PR, we need to copy the boundaries into a `unique_ptr<TableChunk>` for cudf-polars to work with it). For this reason, it's nice to have C++ do the common things we'd like to do in cudf-polars without taking ownership of boundaries (update the key-column mapping and check for alignment).

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -31,6 +36,109 @@ struct HashScheme {
...
+     * @param other The OrderScheme to compare against.
+     * @return True under the shallow rules above.
+     */
+    bool operator==(OrderScheme const& other) const;
+
+    /**
+     * @brief Return a new OrderScheme with updated key column indices, sharing
+     * boundaries.
+     *
+     * The boundary `TableChunk` is shared via `shared_ptr`; no device copy is made.
```
wence-:
```suggestion
```

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -31,6 +36,109 @@ struct HashScheme {
...
+
+    /**
+     * @brief Check whether two schemes share the same boundary values.
+     *
+     * Checks `strict_boundaries` equality and boundary shape, then compares
+     * boundary cell values column-by-column on the device using the LHS stream.
+     * Key column indices and sort directions are intentionally ignored — two
+     * schemes can have compatible partition ranges even when the keys refer to
+     * different column positions (e.g., after a column projection).
+     *
```
wence-: This, if anywhere, should go as a comment.

```suggestion
```

But, can you show an example of when orderschemes with differing sort orders could be compatible? I think if table A is sorted according to some key in increasing order and table B is sorted according to the same logical key but in decreasing order then it cannot be that table A's boundaries are equal to table B's boundaries.

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,11 +4,95 @@
...
+#include <rapidsmpf/memory/memory_reservation.hpp>
 #include <rapidsmpf/streaming/cudf/channel_metadata.hpp>
 
 namespace rapidsmpf::streaming {
 
+OrderScheme::OrderScheme(
+    std::vector<OrderKey> keys,
+    std::shared_ptr<TableChunk> boundaries,
+    bool strict_boundaries
+) {
```
wence-:
```suggestion
) : keys{std::move(keys)}, boundaries{std::move(boundaries)}, strict_boundaries{strict_boundaries} {
```

And then refer to the member slots in the ctor body.

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -31,6 +36,109 @@ struct HashScheme {
...
+     * The boundary `TableChunk` is shared via `shared_ptr`; no device copy is made.
+     * The new key count must match the existing boundary column count.
+     *
+     * @param new_keys Replacement sort keys; size must equal
+     * `boundaries->shape().second`.
+     * @return A new OrderScheme with `new_keys` and the same `boundaries` and
+     *         `strict_boundaries`.
+     * @throws std::invalid_argument if `new_keys` is empty or size mismatches boundaries.
+     */
+    [[nodiscard]] OrderScheme replace_keys(std::vector<OrderKey> new_keys) const;
```
wence-: I think this should be called `with_keys`. Or else modify the scheme in place.

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,11 +4,95 @@
...
+    return OrderScheme(std::move(new_keys), boundaries, strict_boundaries);
+}
+
+bool OrderScheme::boundaries_aligned_with(OrderScheme const& other) const {
+    if (strict_boundaries != other.strict_boundaries)
+        return false;
+    if (boundaries->shape() != other.boundaries->shape())
+        return false;
+    if (boundaries->shape().first == 0)
+        return true;
```
wence-: Please bracketise the conditionals:
```
if (strict_boundaries != other.strict_boundaries || boundaries->shape() != other.boundaries->shape()) {
    return false;
} else if (boundaries->shape().first == 0) {
    return true;
}
```

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,11 +4,95 @@
...
+    auto const lhs = boundaries->table_view();
+    auto const rhs = other.boundaries->table_view();
+    auto const stream = boundaries->stream();
+    // Ensure rhs data is visible on our stream before comparing
+    other.boundaries->stream().synchronize();
```
wence-: Use stream ordering.

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,11 +4,95 @@
...
+
+PartitioningSpec PartitioningSpec::from_order(OrderScheme o) {
+    return {.type = Type::ORDER, .hash = std::nullopt, .order = std::move(o)};
+}
+
+OrderScheme OrderScheme::replace_keys(std::vector<OrderKey> new_keys) const {
+    return OrderScheme(std::move(new_keys), boundaries, strict_boundaries);
+}
+
+bool OrderScheme::boundaries_aligned_with(OrderScheme const& other) const {
```
wence-: This function allocates, therefore needs to take a buffer resource (or at least a memory resource).

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,11 +4,95 @@
...
+    auto const lhs = boundaries->table_view();
+    auto const rhs = other.boundaries->table_view();
+    auto const stream = boundaries->stream();
+    // Ensure rhs data is visible on our stream before comparing
+    other.boundaries->stream().synchronize();
+    for (cudf::size_type i = 0; i < lhs.num_columns(); ++i) {
+        auto eq = cudf::binary_operation(
+            lhs.column(i),
+            rhs.column(i),
+            cudf::binary_operator::EQUAL,
```
wence-: This is incorrect if either column contains nulls, because `null EQUAL null` returns `null`, as does `null EQUAL non-null`. And then `nulls` are ignored in the `reduce`. So you get `[1, None] == [1, 3]` => `True`.


```suggestion
            cudf::binary_operator::NULL_EQUALS,
```

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,11 +4,95 @@
...
+            stream
+        );
+        auto result = cudf::reduce(
+            eq->view(),
+            *cudf::make_all_aggregation<cudf::reduce_aggregation>(),
+            cudf::data_type{cudf::type_id::BOOL8},
+            stream
+        );
+        auto& scalar = static_cast<cudf::numeric_scalar<bool>&>(*result);
+        if (!scalar.is_valid(stream) || !scalar.value(stream)) {
```
wence-:
```suggestion
        if (!scalar.value(stream)) {
```

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -4,11 +4,95 @@
...
+            return false;
+        }
+    }
+    return true;
+}
+
+bool OrderScheme::operator==(OrderScheme const& other) const {
+    return keys == other.keys && strict_boundaries == other.strict_boundaries
+           && boundaries->shape() == other.boundaries->shape();
```
wence-:
```suggestion
    return strict_boundaries == other.strict_boundaries
           && boundaries->shape() == other.boundaries->shape() && keys == other.keys;
```

Comparing the keys is O(nkeys) the other two are O(1), so let's try short-circuiting.

Also, I think this is quite a tight comparison, because it requires that the key column indices used to sort the two order schemes are the same. But trivially a table with columns ["a", "b"] sorted on "a" has the same orderscheme as the same table with its columns permuted. But this equality will say they are different.

cpp/tests/streaming/test_channel_metadata.cpp
```diff
@@ -129,3 +152,83 @@ TEST_F(StreamingChannelMetadata, MessageRoundTrip) {
...
+
+    auto b = make_chunk({100, 200});
+    OrderScheme o1({k0}, b);
+    auto o2 = o1.replace_keys({k5});
+
+    EXPECT_EQ(o2.keys[0].column_index, 5);
+    EXPECT_EQ(o2.keys[0].order, cudf::order::DESCENDING);
+    EXPECT_EQ(o2.strict_boundaries, o1.strict_boundaries);
+    EXPECT_EQ(o2.boundaries->shape(), o1.boundaries->shape());
+    EXPECT_EQ(o2.boundaries.get(), b.get());  // shared_ptr — no device copy
```
wence-:
```suggestion
    EXPECT_EQ(o2.boundaries.get(), b.get());
```

cpp/tests/streaming/test_channel_metadata.cpp
```diff
@@ -129,3 +152,83 @@ TEST_F(StreamingChannelMetadata, MessageRoundTrip) {
...
+    // Same values, different key column indices → still aligned
+    OrderScheme o_shifted({k3}, make_chunk({100, 200}));
+    EXPECT_TRUE(o1.boundaries_aligned_with(o_shifted));
+
+    // Different strict_boundaries → not aligned (short-circuits before device op)
+    OrderScheme o_strict({k0}, make_chunk({100, 200}), /*strict=*/true);
+    EXPECT_FALSE(o1.boundaries_aligned_with(o_strict));
+
+    // Different boundary values → not aligned
+    OrderScheme o_diff({k0}, make_chunk({100, 300}));
```
wence-: Please cull the unnecessary comments throughout.

wence- (changes requested):

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
rjzamora: Okay, sounds good. I had `dataclasses.replace` in the back of my mind, but this isn't python :)

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
rjzamora: No, we definitely need to check the order/null_order somewhere. I'll add it here to `boundaries_aligned_with`

cpp/src/streaming/cudf/channel_metadata.cpp
rjzamora: Yes, I agree the comparison is tight. However, it isn't something we can actually *use* much in practice since we need a proper method to do the full boundary comparison. In practice, `boundaries_aligned_with` will need to do most of the heavy lifting for comparing `OrderScheme`s

rjzamora:
Thanks for the review @wence- - I made some revisions.

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -3,12 +3,114 @@
...
+                return a.order == b.order && a.null_order == b.null_order;
+            }
+        ))
+    {
+        return false;
+    }
+    if (boundaries->shape().first == 0) {
+        return true;
+    }
+    auto const lhs = boundaries->table_view();
```
wence-, line 81: We should wait for https://github.com/rapidsai/cudf/pull/22319 and then use:

```
cuda_stream_join(boundaries->stream(), other.boundaries->stream());
return cudf::tables_equal(boundaries->table_view(), other.boundaries->table_view(), boundaries->stream());
```

(Or approximately that)

rjzamora: Sounds good to me - Is this PR good to go once that PR gets in and the suggested change is made?

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -3,12 +3,114 @@
...
+        );
+        auto& scalar = static_cast<cudf::numeric_scalar<bool>&>(*result);
+        if (!scalar.value(stream)) {
+            return false;
+        }
+    }
+    return true;
+}
+
+bool OrderScheme::operator==(OrderScheme const& other) const {
```
wence-: Let us also implement `bool OrderScheme::operator!=`


```suggestion
bool OrderScheme::operator!=(OrderScheme const& other) const {
    return !(*this == other);
}

bool OrderScheme::operator==(OrderScheme const& other) const {
```

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -31,6 +37,105 @@ struct HashScheme {
...
+struct OrderKey {
+    cudf::size_type column_index;  ///< Column to sort on.
+    cudf::order order;  ///< ASCENDING or DESCENDING.
+    cudf::null_order null_order;  ///< BEFORE or AFTER.
+
+    /**
+     * @brief Equality comparison.
+     * @return True if all fields are equal.
+     */
+    bool operator==(OrderKey const&) const = default;
```
wence-, line 52: Let's also implement `!=`:


```suggestion
    bool operator==(OrderKey const&) const = default;
    /// @brief @return False if not all fields are equal
    bool operator!=(OrderKey const&) const = default;
```

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -3,12 +3,114 @@
...
+            [](OrderKey const& a, OrderKey const& b) {
+                return a.order == b.order && a.null_order == b.null_order;
+            }
+        ))
+    {
+        return false;
+    }
+    if (boundaries->shape().first == 0) {
+        return true;
+    }
```
wence-, line 80: Let's use the shallow equality implementation?

```suggestion
    if (*this != other) {
        return false;
    }
```

wence- (changes requested):

cpp/src/streaming/cudf/channel_metadata.cpp
rjzamora: We need to allow the column indices to be different. The boundaries/order/null_order must match, but we **can** be targeting different columns.

wence-: Ah, ok. But then the converse question is why is shallow equality comparing key indices?

rjzamora: We can do whatever you want for shallow comparison. I'm not using it anywhere, because I need to compare boundaries to make real decisions.

wence-: Ah ok, if we don't use it, let's not offer equality of `OrderScheme`? And just offer this boundaries aligned function? Then we can inline all the stuff here?

rjzamora: Okay, yeah. Works for me.

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -3,12 +3,129 @@
...
+        {
+            auto const& a = *order;
+            auto const& b = *other.order;
+            return a.strict_boundaries == b.strict_boundaries
+                   && a.boundaries->shape() == b.boundaries->shape() && a.keys == b.keys;
+        }
+    }
+    return false;
+}
+
```
wence-: Remind me what we need partitionspec equality for? As with the aligned boundaries this feels like a footgun in the making. For orderschemes, two partitioningspecs can be equal without the schemes actually being equal. Or, vice versa, two orderschemes can have aligned boundaries without the partitioningspecs comparing equal.

rjzamora: Okay, yeah. I think we should just disallow `==` comparison of both `PartitioningSpec` and `ChannelMetadata` objects now that we have this `OrderScheme` complication. I was trying to preserve some of that functionality for simply checking "is this object exactly the same?". However, we **cannot** do that `OrderScheme`, so perhaps it is a footgun to allow `==` comparison at all? The alternative is to raise an error when an `OrderScheme` is attached, but that also seems ugly. I think I'm fine with: "You cannot compare ChannelMetadata with =="

cpp/include/rapidsmpf/streaming/cudf/channel_metadata.hpp
```diff
@@ -31,6 +37,97 @@ struct HashScheme {
...
+    /**
+     * @brief Check whether two schemes share the same boundary values.
+     *
+     * @param other The OrderScheme to compare against.
+     * @param br Buffer resource used for temporary allocations during comparison.
+     * @return True when `strict_boundaries`, shape, and all boundary values match.
+     */
+    [[nodiscard]] bool boundaries_aligned_with(
+        OrderScheme const& other, rapidsmpf::BufferResource const& br
+    ) const;
```
wence-, line 129: nit: This docstring is now misleading. As well as checking boundary values, it also checks for compatibility of sort keys

wence- (approved):
I think this is good now, modulo docstrings. We can change the equality to use the new tables_equal in the libcudf when that comes in.

Issue #996, opened 2026-04-24
Add detailed metrics stating the exact source _and_ destination of communicated data


pentschev:
## Summary

Detailed metrics of communications are important to determine the specific transport layers that are being used, allowing us to identify optimal configurations for compute nodes based on their topology and to prioritize appropriate optimization routes.

### Option 1: Debug information from UCX(X) requests

UCX can provide per-request debug information post-transfers, containing details such as the transport layer used and the bandwidth achieved. This information could be directly plugged into the UCXX `Communicator` and aggregated in statistics for all requests without modifications in call sites.

This is currently blocked as it requires https://github.com/rapidsai/ucxx/pull/437.

### Option 2: Scaffolding in RapidsMPF call sites

Collectives that utilize a `Communicator` may collect statistics on whether it's sending data from host/device, or receiving data to host/device. However, to provide a full picture this is insufficient given the transport layer can be dramatically altered given the remote source/destination memory resource. To provide information that can be used to infer the transport layer used, information of both send and receive memory resources are required.

To realize the condition above, we would require additional scaffolding to retrieve information of either the receiver memory resource type (on the sender side), the sender memory resource type (on the receiver side), or both. This could probably be added as metadata before sending the message and retrieved by the receiver to add ultimately generate statistics on the receiver side. To provide statistics on the sender side this would be more complicated, as it would require an additional message informing of the memory resource type used after the transfer completes. Regardless of the chosen approach, this would require introducing additional complexity at all `Communicator::send()`/`Communicator::recv*()` call sites.

This option would also require knowledge of the system running and UCX configurations, since transport layers are affected by those and are not easily retrievable.

PR #999, opened 2026-04-24
Add coarse comms metrics

### LINKED ISSUE #995: Add coarse comms metrics
Add coarse metrics, stating amount of data communicated from host or from device, and communicated to host or to device. This can be done today with no/little additional scaffolding by simply writing statistics based on the memory resource that is used to send/receive data using a `Communicator`.

nirandaperera:
Depends on https://github.com/rapidsai/rapidsmpf/issues/1008https://github.com/rapidsai/rapidsmpf/issues/1008 

Closes #995 

Coarse comms have been added to,
1. Shuffle 
2. Allgather 
3. Allreduce 
4. Sparse alltoall 

This will add `send-from-<data_type>`, `recv-to-<data_type>` stats for comms operations. 

The communicator now accepts a `Statistics` object which the child classes (MPI & UCXX) uses to record sends and receives. send and recev stats are recorded at the point of _posting_ the comms, not at the point of completing the comms operations.

NOTE:
There is a stat duplication in some tests because communicator stats are global, and tests may decide to create their own stat objects and pass to buffer resource. Stat object accept memory resources in the constructor doesnt help here either. I am planning to fix this in https://github.com/rapidsai/rapidsmpf/issues/979 (its already WIP, need new CCCL for that)


cpp/src/communicator/metadata_payload_exchange/tag.cpp
pentschev, line 1: Looks like you missed the more important `recv()` in `setup_data_receives()`.

cpp/src/coll/allgather.cpp
```diff
@@ -244,6 +248,10 @@ ProgressThread::ProgressState AllGather::event_loop() {
                 break;
             }
             auto chunk = detail::Chunk::deserialize(*msg, br_);
+            // During deserialization, we may have allocated the buffer. So we record the
+            // stats here.
+            statistics_->record_recv(MemoryType::HOST, msg->size());
+            statistics_->record_recv(chunk->memory_type(), chunk->data_size());
```
pentschev: This should be a `record_send`, no? The other `record_recv` should be down below in line 276.

pentschev (changes requested):
It looks like you missed several comms calls, including all of `allreduce.cpp` and `sparse_alltoall.cpp`. Please have another pass.

cpp/src/coll/allgather.cpp
nirandaperera: It's not `send`, is it? 
We are deserializing metadata in L250, and `Chunk::deserialize` allocates buffers for receive buffers. 

> The other record_recv should be down below in line 276.

This is true. That's when we actually take the recv data from the communicator IINM. But during posting the receive, the buffer we provide is exactly the expected size. So, from stats POV, we can record it at either point. 
The only issue would be, if a data receive fails, we will throw, and at that point stats have over-calculated, but they are immaterial if comms fail IMO.

cpp/src/communicator/metadata_payload_exchange/tag.cpp
nirandaperera: @pentschev Maybe I should have explained this in the PR description (I will do it now). Data receive stats are recorded at the point of posting the receive, not at the point of actually receive the data.

nirandaperera:
@pentschev I left out `allreduce` because it did doesnt accept the stats object. But since you brought it up, let me add that as well.

cpp/src/communicator/metadata_payload_exchange/tag.cpp
pentschev: You're right, my bad here, I mixed up the calls.

cpp/src/coll/allgather.cpp
pentschev: Ok, so now based on your other comment about placement of record calls I understand, you're posting both receives (for `recv_from` in line 246 and `recv` in line 276), therefore my comment about line 276 is incorrect as you noted, but my question about `record_send` is probably then valid for the `send` in line 260, no? Note that line 260 is the third call to `send()` in this file, the previous two are both covered but I think the one in line 260 is not.

nirandaperera: Ah. Yes, I see what you mean. I think a finished metadata message is missed in the calculation. Let me check this

nirandaperera: @pentschev good catch. I had under-calculated metadata sends. Thanks. Fixed it now https://github.com/rapidsai/rapidsmpf/pull/999/commits/ece5d526087e08a343dd2b8859432272453654fc

nirandaperera:
@wence- Could you please go through this, since I added stats to allreduce and sparse alltoall.

cpp/include/rapidsmpf/communicator/communicator.hpp
```diff
@@ -640,6 +643,41 @@ constexpr bool COMM_HAVE_MPI = true;
...
+    std::string_view stat_suffix = ""
+) {
+    if constexpr (std::same_as<T, std::vector<std::uint8_t>>) {
+        statistics.record_send(MemoryType::HOST, msg->size(), stat_suffix);
+    } else {
+        statistics.record_send(msg->mem_type(), msg->size, stat_suffix);
+    }
+    return comm.send(std::move(msg), rank, tag);
+}
+
```
pentschev: I'm not quite sure I like this for a few reasons:
1. It only covers send, but none of the `recv*` methods are covered;
2. Adding all `recv*` methods as well will easily bloat this interface substantially;
3. We now have to specify the `Communicator` as argument.

IMO, we would either:
1. Make `Statistics` an integral part of `Communicator`, so that it can be passed to the constructor and all `send`/`recv*` calls will automatically do that; or
2. Move this methods into a separate file to avoid bloating the interface.

I think 1 is a better choice and probably what I'll need to do for #996 , so we should probably go that direction nevertheless but perhaps in a separate PR?

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -159,10 +162,14 @@ void SparseAlltoall::send_ready_messages() {
...
-        fire_and_forget_.push_back(comm_->send(chunk->serialize(), dst, metadata_tag));
+        auto metadata = chunk->serialize();
+        fire_and_forget_.push_back(
+            send_with_stats(*comm_, std::move(metadata), dst, metadata_tag, *statistics_)
+        );
         if (chunk->data_size() > 0) {
+            auto buf = chunk->release_data_buffer();
             fire_and_forget_.push_back(
-                comm_->send(chunk->release_data_buffer(), dst, payload_tag)
+                send_with_stats(*comm_, std::move(buf), dst, payload_tag, *statistics_)
```
pentschev: Why did you decide to create temporary variables (`metadata` and `buf`) that are only used once instead of keeping it like before that the serialization/release are done in-place without the additional variable?

cpp/src/coll/sparse_alltoall.cpp
```diff
@@ -178,6 +185,10 @@ void SparseAlltoall::receive_metadata_messages() {
             }
             state.received_count++;
             auto chunk = detail::Chunk::deserialize(*msg, br_);
+            // During deserialization, we have allocated the recv buffer. So we record the
+            // stats here.
+            statistics_->record_recv(MemoryType::HOST, msg->size());
+            statistics_->record_recv(chunk->memory_type(), chunk->data_size());
```
pentschev: Thinking more about this, your explanation about when to post the receive from https://github.com/rapidsai/rapidsmpf/pull/999/files#r3148360252 was:

> Data receive stats are recorded at the point of posting the receive, not at the point of actually receive the data.

This is not actually true or perhaps not phrased accurately, I think what you meant to say is they are recorded at the point of "posting the _metadata_ receive", which is what is happening here. However, is there a specific reason we need to do that? I think this is a bit flaky for future modifications, if we eventually make changes to the actual data `recv*()` call we're not guaranteed to make changes here as well without careful analysis. Therefore, unless there is a good technical reason to do it this way, it would probably be best to have the `record_recv` next to the actual `recv*()` call.

cpp/src/coll/allreduce.cpp
```diff
@@ -156,9 +158,9 @@ ProgressThread::ProgressState AllReduce::event_loop() {
     // If we don't have a power of two number of ranks, then we have `power_of_two +
     // remainder` ranks. We first take `2 * remainder` ranks, and the even ranks send
     // their contribution to their odd pair. The even ranks then jump to receive a final
-    // contribution, while the rest form a power of two and exchange via the above loop.
-    // Once that is complete, the paired odd ranks send the final answer to their even
-    // counterpart.
+    // contribution, while the rest (ranks >= 2 * remainder or odd) form a power of two
+    // and exchange via the above loop. Once that is complete, the paired odd ranks send
+    // the final answer to their even counterpart.
```
wence-: Why?

cpp/include/rapidsmpf/coll/utils.hpp
```diff
@@ -92,7 +92,7 @@ class Chunk {
      * @brief Return the memory type of the chunk.
      *
      * @return The memory type of the chunk.
-     * @note a finish chunk has memory type host.
+     * @note A finish/ metadata-only chunk has memory type HOST (ie. data_size() == 0).
```
wence-: No.

cpp/include/rapidsmpf/communicator/communicator.hpp
```diff
@@ -640,6 +643,41 @@ constexpr bool COMM_HAVE_MPI = true;
...
+ */
+template <typename T>
+    requires std::same_as<T, std::vector<std::uint8_t>> || std::same_as<T, Buffer>
+[[nodiscard]] inline std::unique_ptr<Communicator::Future> send_with_stats(
+    Communicator& comm,
+    std::unique_ptr<T> msg,
+    Rank rank,
+    Tag tag,
+    Statistics& statistics,
+    std::string_view stat_suffix = ""
```
wence-: The stat_suffix is never used, so I think it is not useful.

wence- (changes requested):
I think this is the wrong approach, because we have to diligently go through every location to remember to record statistics.

Why do we not instead require that a `Communicator` instance is constructed with a `Statistics` object, and then the communicator implementations can add statistics when sending and receiving in their `send` and `recv` implementations.

Then you don't need to touch any of the collectives at all, _and_ you're future proof to any further communication patterns and/or users calling send/recv directly.

nirandaperera:
> I think this is the wrong approach, because we have to diligently go through every location to remember to record statistics.
> 
> Why do we not instead require that a `Communicator` instance is constructed with a `Statistics` object, and then the communicator implementations can add statistics when sending and receiving in their `send` and `recv` implementations.
> 
> Then you don't need to touch any of the collectives at all, _and_ you're future proof to any further communication patterns and/or users calling send/recv directly.

I was actually thinking about this same thing while adding the `send_with_stats` method. I was a bit hesitant because the communicator is purely virtual currently. But let me move this to an adapter/ proxy. Maybe that's a better way.

cpp/include/rapidsmpf/communicator/communicator.hpp
nirandaperera: Because I was thinking whether we need to suffix stats with the OpId/ collective type. Maybe we dont need that?

pentschev:
> I was actually thinking about this same thing while adding the `send_with_stats` method. I was a bit hesitant because the communicator is purely virtual currently. But let me move this to an adapter/ proxy. Maybe that's a better way.

I still don't understand what you mean by that. Why can't you just pass `Statistics` to the `Communicator` constructor, and the implementations handle stats themselves as they see appropriate? We know MPI and UCXX will implement it differently when we have the UCX request info, so it's safe to assume implementations will diverge and thus we don't need any additional layers of indirection here.

cpp/src/coll/allreduce.cpp
nirandaperera: Because this wasnt apparent when I read the comment initially. I thought _rest_ meant just the odd ranks in the initial  2 * remainder ranks.

cpp/include/rapidsmpf/communicator/communicator.hpp
pentschev: Right now, I don't see an immediate use for the OpID or collective type in the statistics, the global amounts should suffice. We can always add this later if we have a specific use case.

nirandaperera:
@pentschev @wence- It was rather an invasive refactor. But I think it was inevitable (and maybe will pay forward for the upcoming changes Peter mentioned).

There is a stat duplication in some tests because communicator stats are global, and tests may decide to create their own stat objects and pass to buffer resource. Stat object accept memory resources in the constructor doesnt help here either. I am planning to fix this in #979 (its already WIP, need new CCCL for that)

madsbk:
I think we need to step back and decide how we want statistics to work: should they be global or local?

I see two approaches:

1. Statistics are tracked globally. There is one statistics object associated with the communicator. Nobody else takes a statistics argument; everyone uses the communicator’s statistics.

2. Statistics are tracked per component. The communicator does **not** own a statistics object. Instead, a statistics object is either passed per send/recv call or manually tracked around the send/recv, which is roughly the current state.

I prefer (1) because it is simpler, and I don’t think we need multiple statistics trackers per communicator in production. That mostly seems useful for testing.

(2) also has merits, since it could allow tracking individual components within an execution. But let’s **not** combine the two. We should choose one model or the other.

In any case, @nirandaperera please submit a PR that only does this refactor and does not also introduce new communication tracking.

wence-:
> Nobody else takes a statistics argument; everyone uses the communicator’s statistics.

We track statistics that are not associated with a communicator at all (particularly, we track when there is no communicator available).

So I think something like:

```
stats = Statistics(...); // or `disabled()`
comm = make_communicator(..., stats);
other_obj = make_obj_that_needs_stats(..., stats);
...
```
This way, the communicator can record send/recv stats, and `other_obj` can record whatever it likes. e.g. today the `BufferResource` has a statistics object because it wants to record local memory copies.

You might say, but the logger that we pass everywhere comes from the communicator and you always need _some_ kind of communicator. But I think that one is the wrong way round as well: it's a historical accident rather than _needing_ a communicator to make a logger.

madsbk:
Right, but it would simplify things a lot if everything used the same communicator-owned statistics object. For example, `BufferResource.statistics` and `Context.statistics` would just point to the communicator’s statistics.

I don’t think it is worth the added complexity to support the case where the communicator uses one `Statistics` instance while the `BufferResource` and all its children use another.

wence-:
I think I agree that our default setup should be that everything uses the same `Statistics`. _But_, I don't want the only way to make a `Statistics` to be by obtaining it from a `Communicator`. I think that would relitigate the mistake we made with the `Logger`.

If we just add the ability to provide a `Statistics` object to the `Communicator` there is no additional complexity in different objects using different `Statistics`. It is then up to the user to post-process and inspect those objects.

In all of our examples and usage we will just have one stats object.

pentschev:
> I think I agree that our default setup should be that everything uses the same `Statistics`. _But_, I don't want the only way to make a `Statistics` to be by obtaining it from a `Communicator`. I think that would relitigate the mistake we made with the `Logger`.
> 
> If we just add the ability to provide a `Statistics` object to the `Communicator` there is no additional complexity in different objects using different `Statistics`. It is then up to the user to post-process and inspect those objects.
> 
> In all of our examples and usage we will just have one stats object.

I agree with this, I'm only confused whether you're saying collectives we provide (e.g., `AllGather` and `AllReduce`) should get the `Statistics` object from the required `Communicator` when unspecified, or if users are _required_ to pass it to their constructors. I think requiring passing the `Statistics` object is better, it leaves no room for confusion when its optional (is this going to use the `Communicator`'s `Statistics` or none at all?) and make all our public interfaces similar where `Statistics` are optionally allowed. This is relevant because in the current state it looks like @nirandaperera may have mixed both, [`AllGather` removes `Statistics`](https://github.com/rapidsai/rapidsmpf/pull/999/files#diff-00a010e89da5c94cc813d2ca86ee79d920980bfbb3df2aff346134293c89ad74) and [`AllReduce` adds `Statistics`](https://github.com/rapidsai/rapidsmpf/pull/999/files#diff-744064b92608e1586c36085380d2bbe5c94e48b2b8e190f0ad4397525d50a04f).

wence-:
Mads and I discussed this a bit, and I think we concluded the following:

- You should explicitly construct statistics `Statistics::from_options(options, mr)` or `std::make_shared<Statistics>(mr)`
- The `Statistics` object should be passed into all "leaf" objects that track statistics (`ProgressThread`, `BufferResource`, ...)
- Non-leaf objects should obtain statistics from one of their leaf arguments: it is implementation-defined which one. If you don't want "partial" statistics you should provide the same stats object to all leaves.

So we prefer that statistics are on if the "leaves" have stats that are on, and otherwise they're off.

Corollary: no `from_options` call should ever construct `Statistics` except the `Statistics::from_options` and they should either be removed or should explicitly accept stats where it is necessary.

This way, the usage would be something like:

```
auto stats = Statistics::from_options(options);
auto thread = std::make_shared<ProgressThread>(stats);
auto comm = new_communicator(options, thread); // communicator gets stats from progress thread
auto br = BufferResource::from_options(mr, ..., stats);
auto ctx = Context::from_options(logger, br, stats);
auto gather = AllGather(comm, br, op_id); // no stats, gets from one of comm or br, just pick (it's implementation defined)
```

pentschev:
I don't think I follow the rationale of picking the statistics from one of the leaves, specifically

> * Non-leaf objects should obtain statistics from one of their leaf arguments: it is implementation-defined which one.

makes everything very confusing. Maybe we won't generally have multiple `Statistics` (is that a fair assessment?), but when we do the implementation deciding which one to pick seems like a bad design as there is no general rule for that. There are probably reasons why you two reached this design, but I can't naively understand the motivation.

wence-:
> Maybe we won't generally have multiple Statistics (is that a fair assessment?), but when we do the implementation deciding which one to pick seems like a bad design as there is no general rule for that. There are probably reasons why you two reached this design, but I can't naively understand the motivation.

The idea is that basically you only have one statistics object. But short of having a global instance that everyone accesses you have to provide it when constructing objects that want to record statistics. So then you ask, do you plumb it in everywhere that explicitly wants statistics, or implicitly. I think both have pros and cons. If you plumb it explicitly, does that mean "I want all the statistics records that emit from this object to go to these stats". Or does it mean "I want the direct statistics from this object to go to these stats". e.g. if I have an `AllGather(comm, br, stats)` do the communication stats from the allgather go to `stats`, or `comm->stats`?

Conversely if I have `AllGather(comm, br)` do the stats from the allgather go to `comm->stats` or `br->stats`? 

My feeling is that since there's ambiguity in both directions, and what we're going to recommend is that you use a single global stats object for everything, that it makes more sense to simplify the constructor signatures of all non-leaf objects.

nirandaperera:
> My feeling is that since there's ambiguity in both directions, and what we're going to recommend is that you use a single global stats object for everything, that it makes more sense to simplify the constructor signatures of all non-leaf objects.

@wence- did you mean a singleton? Wouldn't this be problematic, because we have 2 ways of creating a stats object. 
1. Init stats using ctr in application scope
2.  Using options/ env variables

So, we will need a global getter and setter, isnt it? Then comes the question, what if someone sets a new stats object down the line?

nirandaperera:
@wence- @madsbk @pentschev thanks for the discussion. Sorry for joining late. 

My thoughts are, 
- I like the idea of having one Stats object everywhere. 
- I am +1 for passing stats in the ctr. I understand that this is verbose, but I think this makes intention more clear. Any class that adds stats will accept it in the ctr. We should remove `public auto statistics()` methods (ex. br, communicator). If child classes need to access from parent, we can keep stats obj protected in the parent (ex. Communicator). 

There are some challenges, because currently stats object has a ctr that accepts mem resources. This is a bit out of place, and I am removing it.

wence-:
Before we do another round of refactoring, let's agree on what we want. Because it sounds like we don't.

wence-:
> > My feeling is that since there's ambiguity in both directions, and what we're going to recommend is that you use a single global stats object for everything, that it makes more sense to simplify the constructor signatures of all non-leaf objects.
> 
> @wence- did you mean a singleton? Wouldn't this be problematic, because we have 2 ways of creating a stats object.
> 
> 1. Init stats using ctr in application scope
> 2. Using options/ env variables
> 
> So, we will need a global getter and setter, isnt it? Then comes the question, what if someone sets a new stats object down the line?

I don't mean an actual singleton, but rather "recommended" usage. If you create statistics once and use them everywhere, then everything adds statistics. If you create statistics in an ad-hoc fashion and/or pass them inconsistently then you get inconsistent statistics.

nirandaperera:
Converted to draft since this depends on #1008

PR #1000, opened 2026-04-27
Add socket-based bootstrap coordination backend for `rrun`

pentschev:
Introduces a new `SocketBackend` that replaces filesystem-based coordination with an in-process TCP server hosted by the `rrun` launcher. Before forking any rank processes, `rrun` starts a lightweight coordination server that binds to `127.0.0.1` on an ephemeral port and generates a 256-bit random authentication token. The server address and token are passed to rank processes as environment variables. Each rank connects and authenticates at startup, the server then handles all KV store, barrier, and sync operations entirely in memory, over loopback. This eliminates filesystem artifacts during coordination and prevents unauthorized participation (token required to connect). The socket backend is the new default, while the previous file-based backend remains available via a `--file-backend` flag for backward compatibility for now.

A shared maximum key-length limit (255 bytes, matching POSIX `NAME_MAX`) is now enforced uniformly across all three backends: file, socket, and Slurm/PMIx. The design may also be useful for SSH-based deployment , where remote ranks connect to their own loopback and SSH transparently tunnels the traffic back to the coordinator, with no changes required to the backend code.


cpp/tools/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ add_executable(
   rrun
   "rrun.cpp"
   "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/rrun/rrun.cpp>"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/bootstrap/socket_backend.cpp>"
```
KyleFromNVIDIA: Not sure why the source files are `$<BUILD_INTERFACE:...>`, that genex only matters for `INTERFACE` sources, and these are `PRIVATE`. Nevertheless, we can address that in a follow-up.

KyleFromNVIDIA (approved):
Approved trivial CMake changes with one small comment

cpp/tools/CMakeLists.txt
pentschev: Thanks @KyleFromNVIDIA . I've opened https://github.com/rapidsai/rapidsmpf/pull/1001 to handle all other cases and update this to land correctly once this PR is merged.

cpp/src/bootstrap/socket_backend.cpp
```diff
@@ -0,0 +1,575 @@
...
+
+#include <rapidsmpf/bootstrap/socket_backend.hpp>
+
+// NOTE: Do not use RAPIDSMPF_EXPECTS or RAPIDSMPF_FAIL in this file.
+// Using these macros introduces a CUDA dependency via rapidsmpf/error.hpp.
+// Prefer throwing standard exceptions instead.
+
+namespace rapidsmpf::bootstrap::detail {
+
+static void write_all(int fd, void const* buf, std::size_t n) {
```
madsbk: Use an anonymous namespace instead of `static`

cpp/src/bootstrap/socket_backend.cpp
```diff
@@ -0,0 +1,575 @@
...
+
+    if (wakeup_pipe_[0] >= 0) {
+        ::close(wakeup_pipe_[0]);
+        wakeup_pipe_[0] = -1;
+    }
+
+    std::lock_guard<std::mutex> lk(handler_mutex_);
+    for (auto& t : handler_threads_) {
+        if (t.joinable())
+            t.join();
```
madsbk:
```suggestion
        if (t.joinable()) {
            t.join();
        }
```

cpp/src/bootstrap/socket_backend.cpp
```diff
@@ -0,0 +1,575 @@
...
+    if (key.size() > max_key_size) {
+        throw std::invalid_argument(
+            "Key exceeds maximum length of " + std::to_string(max_key_size)
+            + " bytes: " + key
+        );
+    }
+    send_line("PUT key=" + key + " valuelen=" + std::to_string(value.size()));
+    if (!value.empty())
+        send_bytes(value.data(), value.size());
+    std::string resp = recv_line();
```
madsbk, line 553:
```suggestion
    if (!value.empty()) {
        send_bytes(value.data(), value.size());
    }
```

cpp/src/bootstrap/socket_backend.cpp
pentschev: Done in https://github.com/rapidsai/rapidsmpf/pull/1000/commits/dc056c2d21b07e82718571b3347d3411d96eda10 .

cpp/src/bootstrap/socket_backend.cpp
pentschev: I think the suggestion was only to add curly braces and not to remove `recv_line()`, right? Done in https://github.com/rapidsai/rapidsmpf/pull/1000/commits/fb523399fce1f56c8c5a7898802cbc3718f35e9f .

pentschev (commented):
Thanks @madsbk , your suggestions have been addressed. FYI, I've also pushed two more commits https://github.com/rapidsai/rapidsmpf/pull/1000/commits/c4cb36fb20270b96604e75510e3fbff7b284e991 and https://github.com/rapidsai/rapidsmpf/pull/1000/commits/99a0edeb210c158fc6d516d1998a2a6e110d1686 to improve connection handling which are relatively small.

cpp/include/rapidsmpf/bootstrap/types.hpp
```diff
@@ -16,4 +16,16 @@ using Rank = std::int32_t;
...
 
+/**
+ * @brief Maximum allowed length (in bytes) for a KV coordination key.
+ *
+ * Enforced uniformly by all backend implementations:
+ * - FileBackend: matches the POSIX NAME_MAX filename limit (255 bytes).
+ * - SocketBackend: protocol field width (`%255s`) matches this value.
+ *
+ * Keys must also consist only of printable, non-whitespace ASCII characters
+ * (no `/`, `\`, `..`, or null bytes).
```
wence-: This parenthetical does not make sense with the stipulation on the first line. `/` is certainly printable.

cpp/src/bootstrap/bootstrap.cpp
```diff
@@ -27,8 +28,15 @@ namespace {
...
+    // Socket backend takes highest priority: rrun sets RRUN_SOCKET_ADDR before
+    // forking ranks. Check this before RRUN_COORD_DIR so that new rrun launches
+    // use SocketBackend while old manual setups (RRUN_COORD_DIR only) still work.
+    if (getenv_optional("RRUN_SOCKET_ADDR")) {
+        return BackendType::SOCKET;
+    }
+
+    // Backward compatibility: if RRUN_COORD_DIR is set (but not RRUN_SOCKET_ADDR)
+    // use the file backend.
     if (getenv_optional("RRUN_COORD_DIR")) {
```
wence-, line 40: These comments suggest that there is a deprecation plan. I think we should somehow mark the relevant file-based backend functions as deprecated?

cpp/src/bootstrap/file_backend.cpp
```diff
@@ -135,6 +135,12 @@ std::string FileBackend::get_kv_path(std::string const& key) const {
             "Key contains invalid path characters (e.g., '..', '/', '\\'): " + key
         );
     }
+    if (key.size() > max_key_size) {
```
wence-: Check this first? It's cheaper than the other ones.

cpp/src/bootstrap/socket_backend.cpp
```diff
@@ -0,0 +1,606 @@
...
+            throw std::runtime_error("Connection closed unexpectedly");
+        }
+        ptr += r;
+        n -= static_cast<std::size_t>(r);
+    }
+}
+
+// Read until '\n'; return the line without the newline.
+std::string read_line(int fd) {
+    std::string line;
```
wence-: Use a stringstream/buffer or something rather than `std::string` concat?

cpp/src/bootstrap/socket_backend.cpp
```diff
@@ -0,0 +1,606 @@
...
+    std::array<uint8_t, 32> bytes{};
+    if (::getrandom(bytes.data(), bytes.size(), 0) != static_cast<ssize_t>(bytes.size()))
+    {
+        throw std::runtime_error(
+            "getrandom() failed: " + std::string{std::strerror(errno)}
+        );
+    }
+    static constexpr std::string_view hex = "0123456789abcdef";
+    std::string result;
+    result.reserve(64);
```
wence-:
```suggestion
    result.reserve(bytes.size() * 2);
```
?

cpp/src/bootstrap/socket_backend.cpp
```diff
@@ -0,0 +1,606 @@
...
+            ::close(client_fd);
+            continue;
+        }
+
+        std::lock_guard<std::mutex> lk(handler_mutex_);
+        handler_threads_.emplace_back([this, client_fd]() {
+            handle_connection(client_fd);
+            state_->active_connections.fetch_sub(1, std::memory_order_relaxed);
+        });
+    }
```
wence-, line 277: Do we really need a multi-threaded accept server for exchange of about 5 messages?

cpp/include/rapidsmpf/bootstrap/types.hpp
pentschev: Fixed comment in https://github.com/rapidsai/rapidsmpf/pull/1000/commits/9304660969c2831b4e7d5e54f3c4f8e5f045ed24.

cpp/src/bootstrap/bootstrap.cpp
pentschev: Yes, indeed my plan is to remove this if socket works well. However, we're currently the only consumers of the file-based backend, so deprecating there would only raise warnings for us at build time. I left `--file-backend` as an option in case this proves unstable, if nobody ends up reporting issues in the next ~1 month let's say, we know it's fine to remove it. So my opinion here is we don't need to do anything right now and just monitor for reports, if anything comes we revisit, if we have no reports we can remove `--file-backend` for 26.08. Let me know if you have a strong different opinion.

cpp/src/bootstrap/file_backend.cpp
pentschev: Done in https://github.com/rapidsai/rapidsmpf/pull/1000/commits/e0b8fac053655d48d02169e9dd64fbacbc6137ca.

cpp/src/bootstrap/socket_backend.cpp
pentschev: Switched to use a buffer in https://github.com/rapidsai/rapidsmpf/pull/1000/commits/412b5db25f1de8016df010307b468f0ac4dcd88e, I think a `std::stringstream` would be more expensive than `std::string` here.

cpp/src/bootstrap/socket_backend.cpp
pentschev: Done in https://github.com/rapidsai/rapidsmpf/pull/1000/commits/a1c19eff4d968b2eb28fe6f458f4d0b8b6c5239f.

cpp/src/bootstrap/socket_backend.cpp
pentschev: Yes, because some operations (e.g., barrier) block and we need to allow the server to continue receiving messages from all other ranks. However, I've made this choice for now because we still only support single-node, thus multi-threading is simple and cheap with the amount of ranks we're realistically expecting (up to 8, in most cases), but to support multi-node we would indeed require a better mechanism, as spawning hundreds or thousands of threads doesn't scale. In that case, we would need a better mechanism, like epoll, but that would require large changes. I can make that here if you prefer, but probably punting for now makes more sense. WDYT?

pentschev (commented):
Thanks @wence-, I've addressed/responded all your comments, please have another look.

cpp/src/bootstrap/socket_backend.cpp
wence-: No, ok, that makes sense, thanks.

cpp/include/rapidsmpf/bootstrap/types.hpp
```diff
@@ -16,4 +16,16 @@ using Rank = std::int32_t;
 /// @brief Type alias for Duration type
 using Duration = std::chrono::duration<double>;
 
+/**
+ * @brief Maximum allowed length (in bytes) for a KV coordination key.
+ *
+ * Enforced uniformly by all backend implementations:
+ * - FileBackend: matches the POSIX NAME_MAX filename limit (255 bytes).
```
madsbk, line 23: It would be good checking this

madsbk (commented):
I would be good with a single-process test that spins up a `SocketServer` and exercises the protocol from a client  thread.

cpp/include/rapidsmpf/bootstrap/types.hpp
pentschev: Made the validation a util function that all backends respect in https://github.com/rapidsai/rapidsmpf/pull/1000/commits/7bd42d751a4d6b0d09ac147d52585e20fbec6886, then added tests in https://github.com/rapidsai/rapidsmpf/pull/1000/commits/43ba9b6cef09febc5c82eb666b9e1e486e3500e6. Tests for `SlurmBackend` are intentionally not being added as there's no PMIx server available to test anywhere, we could in the future think of some mock testing or adding a proper PMIx server to our tests, but that's out of the scope of this PR.

pentschev:
> I would be good with a single-process test that spins up a `SocketServer` and exercises the protocol from a client thread.

Added tests in https://github.com/rapidsai/rapidsmpf/pull/1000/commits/8b511e86d640cc6a8d26fa1e2c520be1244ec538.

cpp/src/bootstrap/utils.cpp
```diff
@@ -26,6 +27,24 @@
 
 namespace rapidsmpf::bootstrap {
 
+void validate_key(std::string const& key) {
```
madsbk, line 30: Is this the same as the `validate_key` in `cpp/src/bootstrap/socket_backend.cpp` ?

madsbk (approved):
LGTM, thanks @pentschev

cpp/src/bootstrap/utils.cpp
pentschev: Ugh, yes, that was supposed to be removed from here. It's gone now, thanks for catching.

pentschev:
Thanks all for the reviews!

PR #1001, opened 2026-04-27
Remove unnecessary `$<BUILD_INTERFACE:...>` from CMake targets

pentschev:
The `$<BUILD_INTERFACE:...>` generator expression only has an effect on `PUBLIC` or `INTERFACE` properties, where it controls what downstream consumers see when they consume the target via `find_package()` or `add_subdirectory()`. It was being used in several places where it served no purpose: on `PRIVATE` source files and include directories (which never propagate), and on `RUNTIME_OUTPUT_DIRECTORY` target properties (which are non-propagating build-time-only properties). This PR removes the redundant wrapping, leaving only the one legitimate usage on the `rapidsmpf` library's `PUBLIC` include directory.

### DIFF
--- a/cpp/tools/CMakeLists.txt
+++ b/cpp/tools/CMakeLists.txt
@@ -8,10 +8,10 @@
 add_executable(
   rrun
   "rrun.cpp"
-  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/rrun/rrun.cpp>"
-  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/bootstrap/utils.cpp>"
-  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/system_info.cpp>"
-  "$<$<BOOL:${RAPIDSMPF_HAVE_SLURM}>:$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/bootstrap/slurm_backend.cpp>>"
+  "${PROJECT_SOURCE_DIR}/src/rrun/rrun.cpp"
+  "${PROJECT_SOURCE_DIR}/src/bootstrap/utils.cpp"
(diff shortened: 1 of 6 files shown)


cpp/tools/CMakeLists.txt
```diff
@@ -76,20 +73,16 @@ install(
 
 add_executable(
   check_resource_binding
-  "check_resource_binding.cpp"
-  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/rrun/rrun.cpp>"
-  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/bootstrap/utils.cpp>"
-  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/system_info.cpp>"
+  "check_resource_binding.cpp" "${PROJECT_SOURCE_DIR}/src/rrun/rrun.cpp"
+  "${PROJECT_SOURCE_DIR}/src/bootstrap/utils.cpp" "${PROJECT_SOURCE_DIR}/src/system_info.cpp"
```
KyleFromNVIDIA, line 77: Blegh, looks like cmake-format is being thrown off by the variable substitutions ☹️

KyleFromNVIDIA (approved):

cpp/tools/CMakeLists.txt
pentschev: I don't think it's being thrown off, it's as expected with the line length-based formatting, no?

madsbk (approved):

cpp/tools/CMakeLists.txt
KyleFromNVIDIA: Hmmm, maybe. But then again, I think the old version of this code could have squeezed several source files onto one line.

Doesn't really matter either way, though.

PR #1002, opened 2026-04-27
Enable Ray for aarch64 with conda

pentschev:
Additionally increase pin to >=2.55.1, as there are no older releases supporting aarch64.


bdice (approved):

pentschev:
Thanks @bdice for the approval and for getting aarch64 Ray packages support in conda-forge!

PR #1004, opened 2026-04-30
Skip stack trace if gdb is unavailable

pentschev:
Prevent "gdb not found" errors when timeout is hit in CI if gdb is unavailable. This affects wheels tests in CI, given there's no gdb package in PyPI.


conda/environments/all_cuda-129_arch-x86_64.yaml
```diff
@@ -50,7 +50,7 @@ dependencies:
 - pydata-sphinx-theme>=0.15.4
 - pylibcudf==26.6.*,>=0.0.0a0
 - pytest
-- python>=3.11
+- python=3.13
```
wence-: what happened here?

pentschev: Thanks for catching, I had an unclean local fork and committed this by accident. Fixed now.

wence- (approved):

ci/timeout_with_stack.py
```diff
@@ -109,26 +110,38 @@ def capture_stack_trace(pid: int, stack_type=StackType.C) -> None:
...
+                bt_command,
+                "-ex",
+                "quit",
+            ],
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+    except FileNotFoundError:
+        print(f"Skipping stack trace for process {pid}: gdb not found")
```
wence-: Not sure how we'll get filenotfounderror here given we just checked for gdb above?

wence- (approved):

jameslamb (approved):
Agree with Lawrence's comment about the `FileNotFoundError` bit maybe being unnecessary, but this is fine either way.

pentschev:
Attempting to briefly block merging to address https://github.com/rapidsai/rapidsmpf/pull/1004#discussion_r3167221507 .

PR #1007, opened 2026-04-30
fix(ci): resolve all zizmor findings and add zizmor pre-commit checks

gforsyth:
- **feat: add `zizmor` to pre-commit and add `zizmor` config**
- **fix(ci): add explicit permissions blocks to all workflow jobs**
- **fix(ci): suppress secrets-inherit warnings for shared-workflows**

Part of rapidsai/build-planning#275


jakirkham (approved):
Thanks Gil! 🙏

Issue #1008, opened 2026-04-30
Stats refactor


nirandaperera:
As discussed today in a meeting, we will do the following. 
- Use the same stats object everywhere 
- During object creation, if the args contains an object that contains stats, borrow the stats object from that arg. If there are multiple objects, its up to the class to determine which stats to use (ideally all should point to the same stats instance)

Depends on: 
- #985  
- #1003  

Todo: 
- [ ] Make stats object using a factory method #1009 
- [ ] Use the same stats object everywhere #1009 
- [ ] Pass stats to Communicator 

Post merge:
- #999

PR #1003, opened 2026-04-29
Remove memory resources from Statistics

nirandaperera:
## Remove memory resources from `Statistics` construction

`Statistics` held `RmmResourceAdaptor` and `PinnedMemoryResource` as instance fields, tying resource lifetime to the stats object and preventing a single `Statistics` instance from being used with different resources at report time.

### Solution
- Remove `mr_` / `pinned_mr_` fields and the `Statistics(RmmResourceAdaptor, ...)` constructor; `report()` and `create_memory_recorder()` now accept explicit `std::optional<any_device_resource>` / `std::optional<any_host_device_resource>` parameters — callers supply resources at use time.
- Add `try_pinned_mr()` returning `std::optional<any_host_device_resource>` non-throwingly; correct `pinned_mr()` return type to `host_device_async_resource_ref`.
- Add `memory/resource_types.hpp` with `any_device_resource` / `any_host_device_resource` aliases; update `RAPIDSMPF_MEMORY_PROFILE` macro to require an explicit `mr` argument.

Closes #979 

Depends on https://github.com/rapidsai/rapids-cmake/pull/1008 and #985

PR #1009, opened 2026-05-01
Initial stats cleanup

nirandaperera:
`Statistics` was constructible directly (`make_shared<Statistics>(false)` was not equal to `Statistics::disabled()`), `MemoryRecorder` held a raw `Statistics*` (lifetime hazard noted by an existing TODO), provider `statistics()` accessors returned `shared_ptr` by value (atomic refcount bumps on hot paths), and secondary objects like `AllGather` redundantly carried their own `statistics_` member already reachable through the buffer resource / context.

In this PR, following changes were made
- Make `Statistics` constructible only through factory methods (`create()`, `from_options()`, `disabled()`); the public ctor is removed and `disabled()` is now the canonical singleton for the disabled case.
- Inherit `Statistics` from `std::enable_shared_from_this` so `MemoryRecorder` owns a `shared_ptr<Statistics>` instead of a raw pointer; recorder no-ops when stats are disabled.
- Introduce a `StatisticsProvider` concept and have providers (`BufferResource`, `ProgressThread`, `streaming::Context`) return `std::shared_ptr<Statistics> const&` from `statistics()`; each provider `static_assert`s the concept.
- Drop the redundant `statistics` parameter and `statistics_` member from `coll::AllGather` (and its Python binding); secondary classes now derive their `Statistics` from the provider they already receive.

Related to #1008 

Depends on #1003


cmake/cpm_override_packages.json
wence-, line 1: Committed in error?

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -106,47 +107,27 @@ class Statistics {
...
-     *
-     * @param mr The RMM resource adaptor used for memory profiling.
-     * @param pinned_mr Optional pinned host memory resource for profiling; defaults to
-     * `PinnedMemoryResource::Disabled`.
-     */
-    Statistics(
-        RmmResourceAdaptor mr,
-        std::optional<PinnedMemoryResource> pinned_mr = PinnedMemoryResource::Disabled
-    );
+    static std::shared_ptr<Statistics> create(bool enabled = true);
```
wence-: question: What does having a `create` static method buy us over a constructor?

cpp/include/rapidsmpf/progress_thread.hpp
wence-, line 184: While we're doing this, let's return a `const &` reference to the shared ptr?

nirandaperera:
@wence- this is a draft on top of #1003. Let's merge it first, then I will rebase this.

nirandaperera:
@wence- this is ready for review now

cpp/include/rapidsmpf/statistics.hpp
nirandaperera: factory methods allows us to safely use `std::enable_shared_from_this`, where we can pass a shared_ptr to memory recorder rather than a ptr

cpp/include/rapidsmpf/progress_thread.hpp
nirandaperera: done

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -400,7 +404,7 @@ class BufferResource {
      *
      * @return Shared pointer the Statistics instance.
      */
-    std::shared_ptr<Statistics> statistics();
+    std::shared_ptr<Statistics> const& statistics() const noexcept;
```
pentschev: This is OK as you're going for a hot-path optimization, but the lifetime contract should be explicit. Otherwise returning `shared_ptr` by value is the safer and more idiomatic API.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -105,12 +105,16 @@ class BufferResource {
      *
      * @param mr The RMM resource adaptor.
      * @param options Configuration options.
+     * @param statistics The statistics instance to use (disabled by default). The caller
+     * is responsible for creating and owning this object.
```
pentschev: "and owning this object" doesn't make sense here, `std::shared_ptr<Statistics>` is passed by value.

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -624,13 +631,37 @@ class Statistics {
...
+ * Classes satisfying this concept are *statistics providers* — secondary
+ * classes that receive a provider as a constructor argument should derive their
+ * `Statistics` instance by calling `.statistics()` on it rather than accepting
+ * a separate `std::shared_ptr<Statistics>` argument.
+ *
+ * Returning by const-reference avoids the per-call atomic refcount bump that a
+ * by-value `std::shared_ptr` return would incur on hot paths.
+ *
+ * Each provider asserts this concept via `static_assert` in its own header.
+ * Current providers: `BufferResource`, `ProgressThread`, `streaming::Context`.
```
pentschev:
```suggestion
```

pentschev: Probably best not to document providers here as this will eventually fall out-of-sync.

pentschev (commented):
At a high-level this looks ok. Left some minor suggestions.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
nirandaperera: @pentschev I was addressing @wence- 's [earlier suggestion](https://github.com/rapidsai/rapidsmpf/pull/1009#discussion_r3172566125). 

I am more inclined to use const ref, because I think the `StatisticsProvider` in the scope will likely keep the stats alive.  Ex: 
```c++
void bar(const foo& f){ // foo is a Stats provider
// multiple statistics calls 
f->statistics()->add_bytes_stat(..);
f->statistics()->add_duration_stat(..);
}
```
So, here, unless a temp stats copy is kept in `bar` scope, we will be creating two unnecessary copies. We dont really need a copy anyway, because `f` is keeping stats alive. 
Now, copy will definitely a must, if statistics were a class variable. Ex: `MemoryRecorder`.

pentschev: At a minimum you should make the contract explicit in the docs, it is one of those things that are easy to overlook when you're just glancing at the return type and think "this returns a shared_ptr, so I can manage it's lifetime have a copy while I keep it alive".

nirandaperera: @pentschev After thinking about this a bit more, I reverted stats to return by value.

pentschev (approved):
This looks good to me, thanks Niranda.

cpp/benchmarks/streaming/ndsh/bench_read.cpp
```diff
@@ -413,10 +413,11 @@ int main(int argc, char** argv) {
         std::chrono::duration<double> compute = end - start;
         timings.push_back(pipeline.count());
         timings.push_back(compute.count());
-        comm->logger()->print(ctx->statistics()->report(
+        auto statistics = ctx->statistics();
+        comm->logger()->print(statistics->report(
```
wence-, line 415: Just for future reference, I wish PRs didn't have all this extraneous fluff in them that has no bearing on the actual changes. It makes them much harder to review.

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -624,13 +631,38 @@ class Statistics {
...
+ * classes that receive a provider as a constructor argument should derive their
+ * `Statistics` instance by calling `.statistics()` on it rather than accepting
+ * a separate `std::shared_ptr<Statistics>` argument.
+ *
+ * @note Returning by value provides clear ownership semantics. Callers that invoke
+ * `statistics()` multiple times within the same scope should cache the result
+ * in a local variable to avoid repeated atomic refcount operations on hot paths.
+ *
+ * Each provider asserts this concept via `static_assert` in its own header.
+ * For example, `BufferResource`, `ProgressThread`, `streaming::Context`.
```
wence-: Remove documentation that is implementation detail.
```suggestion
```

cpp/src/statistics.cpp
```diff
@@ -251,15 +256,15 @@ void Statistics::clear() {
...
 )
-    : stats_{stats}, mr_{std::move(mr)}, name_{std::move(name)} {
+    : stats_{std::move(stats)}, mr_{std::move(mr)}, name_{std::move(name)} {
     RAPIDSMPF_EXPECTS(stats_ != nullptr, "the statistics cannot be null");
     mr_->begin_scoped_memory_record();
 }
 
 Statistics::MemoryRecorder::~MemoryRecorder() {
-    if (stats_ == nullptr) {
+    if (stats_ == nullptr || !stats_->enabled()) {
```
wence-: question: What happens if someone makes a memoryrecorder with `stats_->enabled() == true` but then destructs it with `stats_->enabled() == false`?

Also, `stats_ == nullptr` cannot be true, right? Because the `MemoryRecorder` ctor asserts `stats_ != nullptr`.

cpp/src/statistics.cpp
```diff
@@ -162,19 +162,24 @@ Statistics::~Statistics() noexcept {
...
     });
-    return statistics ? std::make_shared<Statistics>(statistics) : Statistics::disabled();
+    return statistics ? Statistics::create() : Statistics::disabled();
 }
 
 std::shared_ptr<Statistics> Statistics::disabled() {
-    static std::shared_ptr<Statistics> ret = std::make_shared<Statistics>(false);
+    static std::shared_ptr<Statistics> ret{new Statistics(false)};
     return ret;
 }
```
wence-, line 176: This should be:

```
std::shared_ptr<Statistics> Statistics::create(bool enabled) {
    return std::shared_ptr<Statistics>(new Statistics(enabled));
}

std::shared_ptr<Statistics> Statistics::from_options(config::Options options) {
    bool const enabled = options.get<bool>(...);
    return Statistics::Create(enabled);
}

std::shared_ptr<Statistics> Statistics::disabled() {
    return Statistics::create(false);
}
```

wence- (changes requested):

cpp/benchmarks/streaming/ndsh/bench_read.cpp
madsbk: And in this case, unnecessarily complicate the code?

cpp/src/statistics.cpp
nirandaperera: I think `disabled()` should be a `static` object or `nullptr`. Otherwise, equality becomes ambiguous. Isnt it?
```c++
auto dis1 = Statistics::disabled();
auto dis2 = Statistics::create(false);
EXPECT_EQ(dis1, dis2); 
```

nirandaperera: I think `nullptr` is not an option, because we expect stats objects to be no-op when disabled.

cpp/src/statistics.cpp
nirandaperera: > Also, `stats_ == nullptr` cannot be true, right? Because the `MemoryRecorder` ctor asserts `stats_ != nullptr`.

This is correct. Let me fix this.

nirandaperera: > question: What happens if someone makes a memoryrecorder with stats_->enabled() == true but then destructs it with stats_->enabled() == false?

I'll double check this. I dont think we've tested this

cpp/benchmarks/streaming/ndsh/bench_read.cpp
nirandaperera: IMO, these changes should've been there in the first place. Since `statistics()` is returning a copy, we should reuse it in the same scope. (this was the case even before this PR)

cpp/src/statistics.cpp
nirandaperera: @wence- this is a valid issue. If we do this, we wouldnt have poped the stack from `mr_`. I fixed it here 0a12bd8e575f2be94109770a1b866ac37b720a0a

cpp/src/statistics.cpp
wence-: Let's step back and decide what we want `Statistics::disabled()` to be. Do we want it as a shorthand for creating a statistics object that is _initially_ disabled. Or do we want it to mean a statistics object you can use that will never be enabled?

The current code (and your changes here) make `statistics::disabled()` mean the same as "statistics that are initially disabled".

nirandaperera: @wence- Yes, I was also thinking about this, and changed the `disabled()` semantics to mean "you wont be able to enable it". Main motivation was, to prevent passing temporary shared_ptrs to default stats arg in constructors. 

If you want to temporarily disable stats, you need to do, 
```
auto stats = Statistics::create();
stats->disable();
...
stats->enable();
...
```

I see several options here, 
1. rename `disable` to `disable_permanently` 
2. Pass an enum to `create` factory, like `enabled | disabled | permanently_disabled`. `permanently_disabled` will return the singleton.

nirandaperera: I did the latter here https://github.com/rapidsai/rapidsmpf/pull/1009/commits/04a3daeeed4528f69fb7991e4f17572f859a6cbd

cpp/benchmarks/streaming/ndsh/bench_read.cpp
madsbk: Fair enough, but to be clear, when you say _copy_ you mean a copy of a shared pointer,  right?

cpp/src/statistics.cpp
wence-: > Main motivation was, to prevent passing temporary shared_ptrs to default stats arg in constructors.

It feels like you're solving an XY problem here. What are we trying to achieve? That you can't get a statistics object out of something that isn't tied to the "main" statistics object? That you can't re-enabled a disabled statistics object? Some other thing?

If you just don't want someone accidentally making another statistics object (because the ctor defaults to something), just remove the default in the ctor.

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -107,41 +109,56 @@ class Statistics {
...
+    enum class Mode : std::uint8_t {
+        /// Tracking is active.
+        Enabled,
+        /// Tracking is off, but may be turned on later via `enable()`.
+        /// Each call to `create(Mode::Disabled)` returns a fresh instance.
+        Disabled,
+        /// Tracking is off and cannot be turned on. All calls to
+        /// `create(Mode::PermanentlyDisabled)` return the same process-wide
+        /// cached singleton, avoiding allocation on hot paths.
+        PermanentlyDisabled,
```
wence-: Where is statistics creation on the "hot path"?

cpp/src/coll/allreduce.cpp
```diff
@@ -59,7 +59,13 @@ AllReduce::AllReduce(
...
     // Note: after this copy, we must check out_buffer's write event before receiving into
     // in_buffer. See StartPreRemainder in the event loop.
-    buffer_copy(Statistics::disabled(), *out_buffer_, *in_buffer_, in_buffer_->size);
+    // TODO: make communicator a statistics provider and pass the statistics instance.
+    buffer_copy(
+        Statistics::create(Statistics::Mode::PermanentlyDisabled),
+        *out_buffer_,
+        *in_buffer_,
+        in_buffer_->size
+    );
```
wence-: Yes, please do that.

cpp/src/streaming/cudf/parquet.cpp
```diff
@@ -105,7 +105,8 @@ class FileCache {
      * @return The cached message, or std::nullopt if the key is not present.
      */
     std::optional<Message> get(std::shared_ptr<Context> ctx, Key const& key) const {
-        auto& stats = *ctx->statistics();
+        auto stats_ptr = ctx->statistics();
+        auto& stats = *stats_ptr;
```
wence-: Completely pointless.

cpp/src/statistics.cpp
```diff
@@ -250,20 +275,27 @@ void Statistics::clear() {
...
     }
+    // Always pop to keep the RMM adaptor's per-thread stack balanced, even if
+    // statistics were disabled after construction (in which case skip publish).
     auto const scope = mr_->end_scoped_memory_record();
-
+    if (!stats_->enabled()) {
+        return;
+    }
     std::lock_guard<std::mutex> lock(stats_->mutex_);
     auto& record = stats_->memory_records_[name_];
```
wence-, line 269: TOCTOU race.

cpp/src/statistics.cpp
```diff
@@ -485,7 +517,9 @@ void Statistics::write_json(std::filesystem::path const& filepath) const {
 
 std::shared_ptr<Statistics> Statistics::copy() const {
     std::lock_guard<std::mutex> lock(mutex_);
-    auto ret = std::make_shared<Statistics>(enabled_.load(std::memory_order_acquire));
+    // Always allocate a fresh instance — `copy()` mutates `stats_` and
+    // `report_entries_` below, so we must not alias the cached singleton.
```
wence-:
```suggestion
```

wence-: What if we're copying the disabled singleton? What then?

cpp/src/statistics.cpp
```diff
@@ -579,7 +613,9 @@ std::shared_ptr<Statistics> Statistics::deserialize(std::span<std::uint8_t const
 
     std::uint8_t enabled{};
     data = read_pod(data, enabled);
-    auto ret = std::make_shared<Statistics>(enabled != 0);
+    // Fresh instance: `ret->stats_` and `ret->report_entries_` are mutated
+    // below, so the cached `Mode::PermanentlyDisabled` singleton is unsafe.
+    auto ret = Statistics::create(enabled != 0 ? Mode::Enabled : Mode::Disabled);
```
wence-:
```suggestion
    auto ret = Statistics::create(enabled != 0 ? Mode::Enabled : Mode::Disabled);

```

cpp/src/statistics.cpp
```diff
@@ -656,7 +692,9 @@ std::shared_ptr<Statistics> Statistics::merge(
 
     bool const any_enabled =
         std::ranges::any_of(snapshots, [](auto const& s) { return s.enabled; });
-    auto ret = std::make_shared<Statistics>(any_enabled);
+    // Fresh instance: the merged data below mutates `ret`'s internal maps,
+    // so the cached `Mode::PermanentlyDisabled` singleton is unsafe here.
+    auto ret = Statistics::create(any_enabled ? Mode::Enabled : Mode::Disabled);
```
wence-:
```suggestion
    auto ret = Statistics::create(any_enabled ? Mode::Enabled : Mode::Disabled);
```

wence- (changes requested):
I think we need to think through what this design, especially around disabled statistics, looks like. Right now it seems very over-complicated and has, from my reading, somewhat inconsistent semantics.

Before writing some more code, let us figure out what it is we actually want to support (it can be different from what we support today!)

cpp/benchmarks/streaming/ndsh/bench_read.cpp
nirandaperera: @madsbk yes 🙂

cpp/src/statistics.cpp
nirandaperera: @wence- 
I was referring to passing `Statistics::disabled()` as the default arg to constructors in other classes (ex. like [this ](https://github.com/rapidsai/rapidsmpf/pull/1009/changes/BASE..04a3daeeed4528f69fb7991e4f17572f859a6cbd#diff-83a954644f6b0574df4bcc42158cfadef144acb428d5e7df001a8faa28c4b6eeL97))

I have seen the following semantics
1. Enabled stats that gets disable during runtime (and vice-versa)
2. Permanently disabled  (ex: stream_ordered_timing.hpp) with a `nullptr`
3. `disabled()` passed as the default value in ctr/ method

So the idea was to bring all of these semantics under a single API.

wence-: OK, but it is possible that rather than hitting the `Statistics` until it fits this shape, we can ban certain approaches because we decide we don't want to support them.

What are the things we actually _do_ want to support?

I think we want to support `disabled` statistics. But it seems there are multiple ways we can state "disabled" today:

- The passed `shared_ptr<Statistics>` is `nullptr`. Consumer that wants to record things must check for nullness
- The passed `shared_ptr<Statistics>` is not `nullptr`, but has `enabled` set to `False`. Most of the time this is fine, but some statistics (especially memory records) are scoped to a range, rather than a point in time. So you can end up recording when you didn't think you would because someone flipped enabled to disabled or vice versa.
- Some other thing?

Additionally, we have ctors that take a `shared_ptr<Statistics>` that defaults to `Statistics::disabled()`, which is a real `Statistics` object with `enabled` set to `False`.

I can't remember why we decided that we needed to be able to runtime selectively enable and disable statistics (and it might even have been me that did that).

But, if we decide we don't need it then I think we can do something like:

- Remove default ctor arguments: always require that someone passes `Statistics`.
- make `Statistics::disabled()` just give you an object (not a singleton) that you can't re-enable

WDYT? What am I missing here?

nirandaperera: > I can't remember why we decided that we needed to be able to runtime selectively enable and disable statistics (and it might even have been me that did that).

I think this is used in benchmarks to disable stats for all iterations except last, and report only results from the last one.  


> Remove default ctor arguments: always require that someone passes Statistics.

This is already WIP https://github.com/rapidsai/rapidsmpf/pull/1017 and depends on this PR. 

> make Statistics::disabled() just give you an object (not a singleton) that you can't re-enable

If we cant reenable it, then we might as well use the same `disabled()` instance everywhere. Isnt it?


Anyways, I think I agree that this 3-modes solution is unnecessarily complicated. Let me simplify things. 
- I will change factories to this suggestion https://github.com/rapidsai/rapidsmpf/pull/1009/changes#r3264727754
- Then `disabled()` will return a new instance every time, which can be enabled at runtime (or better yet, remove it entirely and use `create(false)` )
- After that, #1017 will guarantee that stats obj is always passed as an arg. This will remove any ambiguity. 

WDYT @wence- ?

cpp/src/statistics.cpp
nirandaperera: @wence- 
`enabled_` is not protected by the `mutex_`. https://github.com/rapidsai/rapidsmpf/blob/main/cpp/include/rapidsmpf/statistics.hpp#L157-L166

So, we follow a similar pattern in several places, eg:
https://github.com/rapidsai/rapidsmpf/blob/main/cpp/src/statistics.cpp#L200-L211

Should we change all usages then? We might as well change `enabled_` to a plain `bool` then, and protect it with `mutex_` throughout.

cpp/include/rapidsmpf/statistics.hpp
nirandaperera: this has been reverted

cpp/src/coll/allreduce.cpp
nirandaperera: Will be done in a follow-up.

cpp/src/statistics.cpp
nirandaperera: this has been reverted

cpp/src/statistics.cpp
nirandaperera: this has been reverted

cpp/src/statistics.cpp
wence-: > I think this is used in benchmarks to disable stats for all iterations except last, and report only results from the last one.

Why is `statistics.reset()` not sufficient for this? is it because that doesn't reset memory tracking stats if those are enabled?

> Then disabled() will return a new instance every time, which can be enabled at runtime (or better yet, remove it entirely and use create(false) )

But you think that a singleton is better, and I thought that we should have a disabled thing not be re-enableable?

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -105,12 +105,15 @@ class BufferResource {
...
+     * @param statistics The statistics instance to use (disabled by default).
      *
      * @return A shared pointer to a BufferResource instance configured according to the
      * options.
      */
     static std::shared_ptr<BufferResource> from_options(
-        RmmResourceAdaptor mr, config::Options options
+        RmmResourceAdaptor mr,
+        config::Options options,
+        std::shared_ptr<Statistics> statistics = Statistics::disabled()
```
wence-, line 112: I think we don't want to default a statistics here, but rather require user to specify.

cpp/include/rapidsmpf/streaming/core/context.hpp
```diff
@@ -102,7 +104,8 @@ class Context {
     static std::shared_ptr<Context> from_options(
         RmmResourceAdaptor mr,
         std::shared_ptr<Communicator::Logger> logger,
-        config::Options options
+        config::Options options,
+        std::shared_ptr<Statistics> statistics = Statistics::disabled()
```
wence-, line 108: Again here. no default.

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -559,54 +570,48 @@ class Statistics {
...
         MemoryRecorder& operator=(MemoryRecorder&&) = delete;
 
       private:
-        Statistics* stats_{
-            nullptr
-        };  // TODO: make this shared_ptr using make_shared_from_this
-        std::optional<RmmResourceAdaptor>
-            mr_;  // optional because RmmResourceAdaptor is not default constructible
-        std::string name_;
+        friend class Statistics;
```
wence-: Why does this need to be a friend?

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -559,54 +570,48 @@ class Statistics {
...
+
+        /// Constructs a no-op MemoryRecorder.The dtor short-circuits on
+        /// `!mr_.has_value()` so the null `stats_` is never dereferenced.
```
wence-: Don't document implementation details in docstrings.

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -559,54 +570,48 @@ class Statistics {
...
+        MemoryRecorder() = default;
+
+        /// No-op recorder iff `stats_` is null and `mr_` is `std::nullopt`.
+        std::shared_ptr<Statistics> stats_{nullptr};
+        std::optional<RmmResourceAdaptor> mr_{std::nullopt};
```
wence-: Surely you mean the converse, because this statement implies that it can be an operational recorder if either stats is not null _or_ mr is not nullopt, and I don't think that can work.

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -624,13 +629,35 @@ class Statistics {
...
 };
 
+/**
+ * @brief Satisfied by any type that exposes a `statistics()` method returning
+ *        `std::shared_ptr<Statistics>` by value.
+ *
+ * Classes satisfying this concept are *statistics providers* — secondary
+ * classes that receive a provider as a constructor argument should derive their
+ * `Statistics` instance by calling `.statistics()` on it rather than accepting
+ * a separate `std::shared_ptr<Statistics>` argument.
```
wence-:
```suggestion
 * Classes satisfying this concept are *statistics providers*. Secondary
 * classes that receive a provider as a constructor argument should derive their
 * `Statistics` instance by calling `.statistics()` on it rather than accepting
 * a separate `std::shared_ptr<Statistics>` argument.
```

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -624,13 +629,35 @@ class Statistics {
...
+ *
+ * @note Returning by value provides clear ownership semantics. Callers that invoke
+ * `statistics()` multiple times within the same scope should cache the result
+ * in a local variable to avoid repeated atomic refcount operations on hot paths.
```
wence-, line 650: So I think there was some back and forth here. TBH, I think that the `statistics` method should return `shared_ptr<Statistics> const&`.

That way, the following works "as expected" but everything is efficient.

```
auto br = std::make_shared<BufferResource>(...);

br->statistics()->record(); // doesn't bump refcount

auto stats = br->statistics(); // does bump refcount, because stats is a shared_ptr, not shared_ptr&
```

cpp/src/statistics.cpp
wence-: OK. I think I have read through again and I think the current state is OK. We have `create(enabled/disabled)` and `disabled()` and we end up with no singleton.

Later we will remove the default `disabled()` ctor arguments...

cpp/src/statistics.cpp
wence-: Ah, ok. I think it is fine.

wence- (approved):
I've had another read over this and I think it is in a good place now.

Minor nits to fix in this round before merging: where you introduce new construction arguments that provide a default value for statistics (e.g. from_options) please just require a statistics object (because of planned followup to remove statistics defaults).

Also, I think it is better to return `shared_ptr const&` from statistics providers. The consumer can then either get a `shared_ptr` or just use the method slot directly.

cpp/src/statistics.cpp
nirandaperera: Yes! I was rebasing my previous PRs on top of this, and I felt removing the singleton was a better approach. We need the ability to enable a stat obj that was initially disabled.

cpp/include/rapidsmpf/statistics.hpp
nirandaperera: I was thinking on the same lines [here](https://github.com/rapidsai/rapidsmpf/pull/1009#discussion_r3221367552), but then decided against it, because IMO returning the value forces callers to be mindful of the lifetime aspects (and at minimum its the safest).

cpp/include/rapidsmpf/memory/buffer_resource.hpp
nirandaperera: this is done in #1017

cpp/include/rapidsmpf/streaming/core/context.hpp
nirandaperera: this is done in #1017

cpp/include/rapidsmpf/statistics.hpp
nirandaperera: removed

cpp/include/rapidsmpf/streaming/core/context.hpp
```diff
@@ -83,6 +83,8 @@ class Context {
      * @param logger The logger to use.
      * @param options Configuration options used to initialize the Context and its
      * components.
+     * @param statistics The statistics instance to use (disabled by default). The caller
+     * is responsible for creating and owning this object.
```
pentschev: `statistics` is passed by value, thus the caller is _not_ responsible for owning the object, right?

pentschev (approved):
One minor docstring issue, otherwise LGTM. Thanks Niranda.

PR #1012, opened 2026-05-04
Topology-aware default pool sizing for `PinnedMemoryResource`

nirandaperera:
Replace the hardcoded `pinned_initial_pool_size = 0` / `pinned_max_pool_size = unbounded`
defaults with values scaled to the system's GPU count.

### Changes

- **`get_host_memory_per_gpu()`** — new utility in `system_info` that returns 
`current numa node total host memory/ N GPUs in current numa node`

- **`PinnedMemoryResource`** — two new named constants drive the defaults:
  - `DefaultInitiPoolSizeFactor = "10%"` → initial pool = 10% of per-GPU host memory
  - `DefaultMaxPoolSizeFactor = "80%"`  → max pool = 80% of per-GPU host memory

  Also corrects the documented default from `false` to `true`.

- **Python bindings** — exposes `get_host_memory_per_gpu()` via `system_info.pyx/.pyi`.

- **Tests** — `PinnedResource.from_default_options` verifies pool sizes match the
  factor-scaled values when no options are provided.

- **Benchmark** -- Benchmark initialization time against initial pool size  - [Results](https://docs.google.com/spreadsheets/d/17dghCSjfIEW7dc6in53E0CyUqVn7BHamBAFeAu-T_4c/edit?usp=sharing)

- **Docs** — `configuration.md` updated to reflect the new defaults.


cpp/include/rapidsmpf/memory/pinned_memory_resource.hpp
```diff
@@ -100,6 +100,25 @@ class PinnedMemoryResource final
...
     static constexpr std::nullopt_t Disabled = std::nullopt;
 
+    /**
+     * @brief Fraction of total host memory per GPU used as the initial pinned pool size
+     *        when no explicit `pinned_initial_pool_size` option is provided.
+     *
+     * Applied as: `initial_pool_size = get_host_memory_per_gpu() *
+     * DEFAULT_INIT_POOL_SIZE_FACTOR`.
+     */
+    static constexpr double DEFAULT_INIT_POOL_SIZE_FACTOR = 0.1;
```
madsbk: Let’s be consistent and use a percentage, as we do with the `spill_device_limit` option. 
See `parse_nbytes_or_percent()`, which supports both percentages and absolute byte sizes.

cpp/src/system_info.cpp
```diff
@@ -79,4 +81,24 @@ std::uint64_t get_numa_node_host_memory([[maybe_unused]] int numa_id) noexcept {
...
+        );
+        return discovery;
+    }();
+    return topo.get_topology();
+}
+}  // namespace
+
+std::uint64_t get_host_memory_per_gpu() {
+    auto const num_gpus = get_topology().num_gpus;
+    return get_total_host_memory() / std::max<std::uint64_t>(1, num_gpus);
```
pentschev: Can you double-check how `get_total_host_memory` behaves on devices where GPU memory is also exposed as NUMA nodes? In those cases `free` will also show GPU memory as part of system memory, which may break the assumption `get_total_host_memory` presents only system memory.

IIRC, the conditions for GPU memory to be exposed as system memory are:
1. Coherent system (e.g., Grace-Hopper/Grace-Blackwell)
2. NVIDIA driver in NUMA mode (not CDMM)
3. HMM enabled on the kernel

nirandaperera: Ah! this is a good point.

nirandaperera: @pentschev you were right Peter.

pentschev: Nice that there was a simple solution. I also didn't think of the bigger picture of the problem, we cannot just take the whole host memory and divide equally among GPUs, since their NUMA nodes may have different amount of memory, and looks like your new solution already ensures that.

cpp/src/system_info.cpp
```diff
@@ -79,4 +83,29 @@ std::uint64_t get_numa_node_host_memory([[maybe_unused]] int numa_id) noexcept {
...
+}  // namespace
+
+std::uint64_t get_host_memory_per_gpu() {
+    auto const current_numa_node = get_current_numa_node();
+    auto const& gpus = get_topology().gpus;
+    auto const num_local_gpus = std::ranges::count_if(gpus, [&](auto const& gpu) {
+        return gpu.numa_node == current_numa_node;
+    });
+    return get_numa_node_host_memory(current_numa_node)
+           / std::max<std::uint64_t>(1, static_cast<std::uint64_t>(num_local_gpus));
```
madsbk: Use `safe_cast`

cpp/src/memory/pinned_memory_resource.cpp
```diff
@@ -74,21 +74,27 @@ std::optional<PinnedMemoryResource> PinnedMemoryResource::from_options(
...
     if (pinned_memory && is_pinned_memory_resources_supported()) {
+        auto const host_memory_per_gpu = get_host_memory_per_gpu();
         PinnedPoolProperties pool_properties{
             .initial_pool_size = options.get<size_t>(
                 "pinned_initial_pool_size",
-                [](auto const& s) { return s.empty() ? 0 : parse_nbytes_unsigned(s); }
+                [&](auto const& s) {
+                    return parse_nbytes_or_percent(
+                        s.empty() ? DefaultInitiPoolSizeFactor : s,
+                        static_cast<double>(host_memory_per_gpu)
```
madsbk: Use safe_cast

cpp/src/memory/pinned_memory_resource.cpp
```diff
@@ -74,21 +74,27 @@ std::optional<PinnedMemoryResource> PinnedMemoryResource::from_options(
...
-                "pinned_max_pool_size", [](auto const& s) -> std::optional<size_t> {
-                    auto parsed = parse_optional(s);
-                    if (parsed.has_value() && !parsed->empty()) {
-                        return parse_nbytes_unsigned(*parsed);
-                    }
-                    return std::nullopt;
+                "pinned_max_pool_size", [&](auto const& s) {
+                    return parse_nbytes_or_percent(
+                        s.empty() ? DefaultMaxPoolSizeFactor : s,
+                        static_cast<double>(host_memory_per_gpu)
```
madsbk: use safe_cast

cpp/include/rapidsmpf/system_info.hpp
```diff
@@ -79,5 +79,11 @@ std::vector<int> get_current_numa_nodes() noexcept;
  */
 std::uint64_t get_numa_node_host_memory(int numa_id = get_current_numa_node()) noexcept;
 
+/**
+ * @brief Get the amount of host memory per GPU.
+ *
```
madsbk, line 89: Add something about what _host memory per GPU_ is

python/rapidsmpf/rapidsmpf/utils/system_info.pyx
```diff
@@ -84,3 +87,24 @@ def get_numa_node_host_memory(numa_id = None):
...
+    Get the total host memory divided by the number of GPUs in the system.
+
+    Returns the amount of host memory attributed to each GPU, computed as
+    total host memory divided by the number of GPUs visible to the system.
+    Falls back to total host memory when only one GPU (or none) is present.
+
+    Returns
+    -------
+    int
+        Host memory per GPU in bytes.
```
madsbk:
```suggestion
    Host memory per GPU in bytes.
```

madsbk (approved):
Looks good, I only have minor stuff

cpp/include/rapidsmpf/system_info.hpp
pentschev: The new brief is very long, we should have a `@brief Get the amount of host memory per GPU.` and then a long description below explaining the NUMA node constraint.

pentschev (commented):
Please update the description to reflect recent changes @nirandaperera .

Temporarily blocking to ensure PR and Python `get_host_memory_per_gpu()` are updated.

pentschev:
I think there's some partial GH UI breakage, I cannot comment on the files now for some reason , but `get_host_memory_per_gpu` description in `python/rapidsmpf/rapidsmpf/utils/system_info.pyx` is outdated, @nirandaperera  please update with NUMA info.

pentschev (changes requested):

cpp/benchmarks/bench_memory_resources.cpp
```diff
@@ -18,6 +19,15 @@
 
 using rapidsmpf::safe_cast;
 
+// When the RAPIDSMPF_SMOKE_TEST_MODE env var is set (to any value), each
```
pentschev: This is a bad pattern. One will almost definitely one day attempt `RAPIDSMPF_SMOKE_TEST_MODE=0` and will be confused. A better approach would be adding a CLI argument, like all other `bench_*` binaries we have.

nirandaperera: @pentschev Well, this is a bit tricky. Passing custom args might not work here, because in google benchmarks, `Apply(...)` runs during static initialization. So, to make it work, we have to parse args _before_ registering benchmarks. That means, we need a custom main, parse smoke-test arg, and then move all bench registrations after that. I'd rather keep this simple like this.

cpp/benchmarks/bench_memory_resources.cpp
```diff
@@ -16,15 +16,25 @@
...
+// When the RAPIDSMPF_SMOKE_TEST_MODE env var is set to a truthy value (e.g.
+// "1", "on", "true", "yes"), each benchmark's argument generator emits only a
+// tiny subset of cases so the suite finishes quickly during CI smoke tests.
+// Cached because Apply callbacks invoke this once per registered benchmark.
+//
+// We use an env var rather than a CLI flag because google-benchmark's
+// BENCHMARK(...)->Apply(...) macros run during static initialization, before
+// main() has a chance to parse argv. A CLI-flag approach would require moving
+// every benchmark registration into main() (via benchmark::RegisterBenchmark),
+// which is more invasive. std::getenv works fine during static init.
```
pentschev, line 32: Whelp, this is lame. :disappointed:

pentschev (approved):
Thanks Niranda.

jameslamb (approved):
approving for `ci-codeowners` / `packaging-codeowners`, the changes that pulled in those groups look non-controversial.

PR #1013, opened 2026-05-05
Simplify `HostBuffer` memory ownership

nirandaperera:
`HostBuffer` tracked two redundant deallocation paths (`mr_` + `owned_storage_`), where `mr_` was stored unconditionally but never used when `owned_storage_` was present — and as a non-owning ref it could silently outlive a `PinnedMemoryResource`.

### Solution
- Replace both members with a single `std::function<void(rmm::cuda_stream_view)> deallocate_fn_`, invoked with the live stream at deallocation time to respect `set_stream()`.
- In the RMM-allocated path, wrap `mr` in `cuda::mr::any_resource` before capturing it — deep-copying the resource and incrementing its refcount if it is a `PinnedMemoryResource`.
- Remove the vestigial `mr` parameter from `from_owned_vector` and `from_rmm_device_buffer` (it was always unused for deallocation).
- Remove the leaked `OwnedStorageDeleter` public type alias.


cpp/src/memory/host_buffer.cpp
```diff
@@ -16,45 +16,39 @@ namespace rapidsmpf {
 HostBuffer::HostBuffer(
     std::size_t size, rmm::cuda_stream_view stream, rmm::host_async_resource_ref mr
 )
-    : stream_{stream}, mr_{std::move(mr)} {
+    : stream_{stream} {
     if (size > 0) {
+        auto owned_mr = cuda::mr::any_resource<cuda::mr::host_accessible>{mr};
```
madsbk: I don’t think `owned_mr` actually owns the memory resource. AFAIK, `cuda::mr::any_resource<P>` only owns the `rmm::host_async_resource_ref mr`, which is itself a non-owning reference. So `owned_mr` ends up owning a copy of a non-owning reference, not the underlying memory resource?

wence-: This is a reasonable thought but incorrect. `any_resource(resource_ref)` magically converts the ref into an owning type.

madsbk: Thanks @wence-, good to know! @nirandaperera could you add a comment noting this?

madsbk (commented):
@nirandaperera, it would be good to add a test that constructs a `HostBuffer` from a transient memory resource, for example a `PinnedMemoryResource` that goes out of scope before the buffer is deallocated.

cpp/src/memory/host_buffer.cpp
nirandaperera: Yeah, `any_resource` takes in a copy of the underlying resource in the ref. Idea is to increment the refcount if that resource is shared.

nirandaperera:
@madsbk done

cpp/include/rapidsmpf/memory/host_buffer.hpp
```diff
@@ -222,21 +212,21 @@ class HostBuffer {
      *
      * @param span View of the owned memory.
      * @param stream CUDA stream associated with this buffer.
-     * @param mr Dummy memory resource (not used for deallocation).
-     * @param owned_storage Unique pointer managing the owned storage lifetime.
+     * @param deallocate_fn Callable invoked with the current stream to release the
+     *        underlying memory. It captures all resources needed for deallocation (e.g.
+     *        memory resource, raw pointer, size).
      */
```
madsbk:
```suggestion
     * underlying memory. It captures all resources needed for deallocation (e.g.
     * memory resource, raw pointer, size).
```

cpp/include/rapidsmpf/memory/host_buffer.hpp
```diff
@@ -48,10 +38,15 @@ class HostBuffer {
      *
      * @param size Number of bytes to allocate.
      * @param stream CUDA stream on which allocation and deallocation occur.
-     * @param mr RMM host memory resource used for allocation.
+     * @param mr Host-accessible memory resource used for allocation. Taken by value
+     *        so the buffer shares ownership of the resource (e.g. bumps the refcount
+     *        when constructed from a shared-ownership resource); an implicit
+     *        conversion from `rmm::host_async_resource_ref` is also supported.
```
madsbk:
```suggestion
     * @param mr Host-accessible memory resource used for allocation. Taken by value
     * so the buffer shares ownership of the resource (e.g. bumps the refcount
     * when constructed from a shared-ownership resource); an implicit
     * conversion from `rmm::host_async_resource_ref` is also supported.
```

cpp/include/rapidsmpf/memory/host_buffer.hpp
```diff
@@ -222,21 +212,21 @@ class HostBuffer {
      *
      * @param span View of the owned memory.
      * @param stream CUDA stream associated with this buffer.
-     * @param mr Dummy memory resource (not used for deallocation).
-     * @param owned_storage Unique pointer managing the owned storage lifetime.
+     * @param deallocate_fn Callable invoked with the current stream to release the
+     *        underlying memory. It captures all resources needed for deallocation (e.g.
+     *        memory resource, raw pointer, size).
```
madsbk:
```suggestion
     * @param deallocate_fn Callable invoked with the current stream to release the
     * underlying memory. It captures all resources needed for deallocation (e.g.
     * memory resource, raw pointer, size).
```

madsbk (approved):

PR #1016, opened 2026-05-06
Fix large-message initialization in communication benchmark

pentschev:
`bench_comm` could fail while initializing input buffers for per-peer messages at around 8 GiB and larger, before the benchmark itself ran. This change allows large-message all-to-all cases to run correctly.


cpp/benchmarks/utils/random_data.cu
```diff
@@ -17,21 +20,30 @@
...
     rmm::cuda_stream_view stream,
     rmm::device_async_resource_ref mr
 ) {
     // Fill vector with random data.
-    rmm::device_uvector<std::int32_t> vec(static_cast<std::size_t>(nelem), stream, mr);
+    using index_t = std::int64_t;
+    RAPIDSMPF_EXPECTS(
+        nelem <= static_cast<std::size_t>(std::numeric_limits<index_t>::max()),
+        "random_device_vector size exceeds signed iterator range"
+    );
```
nirandaperera: Nit. Do we consider 32-bit systems? For 64 bit nelem > INT64_MAX seems unlikely isnt it?

cpp/benchmarks/utils/random_data.cu
nirandaperera, line 54: Shouldnt this be `std::size_t` now to reflect the above change? Otherwise we will be clipping values above UINT32_MAX, isnt it?

cpp/benchmarks/utils/random_data.cu
```diff
@@ -72,7 +87,7 @@ void random_fill(rapidsmpf::Buffer& buffer, rmm::device_async_resource_ref mr) {
     case rapidsmpf::MemoryType::DEVICE:
         {
             auto vec = random_device_vector(
-                buffer.size / sizeof(std::int32_t) + sizeof(std::int32_t),
+                (buffer.size + sizeof(std::int32_t) - 1) / sizeof(std::int32_t),
```
nirandaperera: if `buffer` is empty now, we will be passing 0 to `random_device_vector`. I think `random_fill` was trying to prevent that before? I'm not sure.

cpp/benchmarks/utils/random_data.cu
nirandaperera, line 70: Same comment here.

cpp/benchmarks/utils/random_data.cu
madsbk: Always use [`safe_cast()`](https://github.com/rapidsai/rapidsmpf/blob/5b86b15f3acea24d4fad7c380468e4e7c13e3454/cpp/include/rapidsmpf/utils/misc.hpp#L311)

cpp/benchmarks/utils/random_data.cu
wence-: You can't make a column with more than size_type::max rows, so I think the point is moot?

cpp/benchmarks/utils/random_data.cu
pentschev: > Nit. Do we consider 32-bit systems?

No: https://forums.developer.nvidia.com/t/whats-the-last-version-of-the-cuda-toolkit-to-support-32-bit-applications/323106/4

> For 64 bit nelem > INT64_MAX seems unlikely isnt it?

Removed that.

> Always use [`safe_cast()`](https://github.com/rapidsai/rapidsmpf/blob/5b86b15f3acea24d4fad7c380468e4e7c13e3454/cpp/include/rapidsmpf/utils/misc.hpp#L311)

Done.

cpp/benchmarks/utils/random_data.cu
pentschev: Should be fixed now.

cpp/benchmarks/utils/random_data.cu
pentschev: > Shouldnt this be `std::size_t` now to reflect the above change? Otherwise we will be clipping values above UINT32_MAX, isnt it?

Not really, for cuDF columns/rows we should still use `cudf::size_type`, but that shouldn't limit us from generating/filling larger vectors (such as those we need for `bench_comm` that are completely unaware of cuDF).

pentschev: Same as https://github.com/rapidsai/rapidsmpf/pull/1016#discussion_r3203950893 .

wence- (approved):

PR #1017, opened 2026-05-07
Make statistics a mandatory arg for statistics providers

nirandaperera:
**Make `Statistics` a mandatory constructor argument for providers**

Follow-up to #1009. After the initial stats cleanup, `BufferResource`, `ProgressThread`, and `streaming::Context` still defaulted their `statistics` parameter to `Statistics::disabled()`, which made it easy to silently drop stats by forgetting to pass an instance and obscured ownership at the call site (e.g. communicator vs. progress thread vs. buffer resource each potentially holding a different `Statistics`).

In this PR:
- `BufferResource(...)`/`BufferResource::from_options(...)` now require an explicit `Statistics` (passed as the first ctor arg); `ProgressThread(...)` does too. `Communicator` stores a `shared_ptr<ProgressThread>` and forwards `statistics()` to it, so `Single`/`MPI`/`UCXX` drop their own progress-thread members. Tests, examples, and Python bindings are updated; the Python `Communicator` gains a `statistics` property.
- New `MutableStatisticsProvider` concept and `ProgressThread::set_statistics()` (wired through to Python) allow swapping the instance at runtime; the progress loop is paused around the swap.
- `MemoryRecorder` and `StreamOrderedTiming` now hold `Statistics` weakly so a swap can destroy the original even with outstanding scopes; the destructors lock and skip publishing when the instance has expired.
- Benchmarks default to stats enabled (previously `Statistics::disabled()` / `from_options` defaulting to off) and use `clear()` before the final run to keep per-run reporting.

Depends on #1009 

Related to #1008


cpp/benchmarks/bench_comm.cpp
```diff
@@ -317,16 +317,7 @@ int main(int argc, char** argv) {
...
 
     rmm::device_async_resource_ref mr = cudf::get_current_device_resource_ref();
-    BufferResource br{
-        mr,
-        PinnedMemoryResource::Disabled,
-        {},
-        std::chrono::milliseconds{1},
-        std::make_shared<rmm::cuda_stream_pool>(
-            16, rmm::cuda_stream::flags::non_blocking
-        ),
```
pentschev, line 327: Why is it ok to drop all the other values? Same question applies to other benchmarks below.

cpp/src/stream_ordered_timing.cpp
```diff
@@ -79,7 +79,7 @@ void timing_stop_cb(void* data) {
     lock.unlock();
 
     auto statistics = timing.statistics.lock();
-    if (statistics == nullptr) {
+    if (statistics == nullptr || !statistics->enabled()) {
```
pentschev: Can `statistics` still be `nullptr` if we're enforcing it _not_ to be `nullptr` in constructors?

cpp/benchmarks/streaming/bench_streaming_shuffle.cpp
```diff
@@ -460,13 +452,13 @@ int main(int argc, char** argv) {
 
     auto statistics = ctx->statistics();
```
pentschev:
```suggestion
```

pentschev: Unused variable, delete.

cpp/benchmarks/streaming/bench_streaming_shuffle.cpp
```diff
@@ -324,7 +324,8 @@ int main(int argc, char** argv) {
 
     // Initialize configuration options from environment variables.
     rapidsmpf::config::Options options{rapidsmpf::config::get_environment_variables()};
-    auto progress_thread = std::make_shared<rapidsmpf::ProgressThread>();
+    auto stats = rapidsmpf::Statistics::from_options(options);
+    auto progress_thread = std::make_shared<rapidsmpf::ProgressThread>(stats);
```
pentschev: This changes the existing behavior, I think unintentionally, from statistics being enabled by default to disabled by default. As is, users now require specifying `RAPIDSMPF_STATISTICS=1`, which I think is undesirable.

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
```diff
@@ -136,18 +136,18 @@ cdef class BufferResource:
         will be created. Must be an instance of
         ``rmm.pylibrmm.cuda_stream_pool.CudaStreamPool``.
     statistics
-        The statistics instance to use. If None, a disabled statistics instance
-        will be created.
+        The statistics instance to use. Required. Pass
+        ``Statistics.disabled()`` for a no-op recorder.
```
pentschev: Should be moved up to match the parameter ordering.

python/rapidsmpf/rapidsmpf/statistics.pyx
```diff
@@ -186,14 +189,14 @@ cdef class Statistics:
     @classmethod
     def disabled(cls):
         """
-        Returns a disabled (no-op) Statistics instance.
+        Get a shared, no-op Statistics instance.
 
-        Useful when you need to pass a Statistics argument but do not want to
-        collect any data.
+        Returns the same disabled instance on every call. Use this when an API
```
pentschev: I think this is technically incorrect, "the same disabled instance" implies a singleton instance, but C++ implementation `cpp_disabled()` returns a new disabled instance.

cpp/benchmarks/bench_comm.cpp
wence-: These are all the defaults. Previously because the `stats` argument was last we had to specify everything. Now it is not, we don't.

Although I think we should probably be explicit as before anyway.

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -144,6 +144,24 @@ class Statistics : public std::enable_shared_from_this<Statistics> {
...
+     * already mid-execution at reset time may still record one stale stat
+     * (its global-map entry was extracted before `cancel_inflight_timings`
+     * could remove it). Callers that need clean boundaries must quiesce
+     * recorders and synchronise the relevant CUDA streams before calling.
+     *
+     * @param new_options Configuration options whose `"statistics"` field
+     * controls the new enabled state.
+     */
+    void reset_from_options(config::Options new_options);
+
```
wence-: What do you need this for?

cpp/src/communicator/communicator.cpp
```diff
@@ -3,10 +3,24 @@
...
+
+Communicator::Communicator(std::shared_ptr<Statistics> statistics)
+    : statistics_{std::move(statistics)} {
+    RAPIDSMPF_EXPECTS(
+        statistics_ != nullptr,
+        "Communicator statistics cannot be null",
+        std::invalid_argument
+    );
+}
+
```
wence-, line 24: No, let's keep the `Communicator` completely virtual and implement `statistics` as a method on the concrete implementations by just forwarding into progress thread. No need to store another reference to the stats object.

cpp/benchmarks/streaming/bench_streaming_shuffle.cpp
wence-: I think all benchmarks should have enabled statistics. So yes, I think you should `create()` the stats. Rather than `from_options()`ing them

cpp/benchmarks/streaming/bench_streaming_shuffle.cpp
```diff
@@ -411,7 +403,7 @@ int main(int argc, char** argv) {
     for (std::uint64_t i = 0; i < total_num_runs; ++i) {
         // Clear statistics before the last run so only the final run is reported.
         if (i == total_num_runs - 1) {
-            ctx->statistics()->clear();
+            stats->clear();
         }
```
wence-, line 413: Does the `ctx` not advertise a statistics object any more? That seems wrong.

cpp/src/communicator/communicator.cpp
wence-: Or else make the progress thread shared in the base class and implement `statistics()` on the base class by asking the progress thread.

wence-:
Some broader comments:

1. I _still_ think that statistics providers should return a `shared_ptr<Statistics> const&`. That way the caller can either upgrade to an owning `shared_ptr` or just keep a reference. The converse is not possible: if we return a `shared_ptr<Statistics>` there is no way to downgrade it to a reference.
2. Everything that has a `Statistics` object "somewhere" should be a statistics provider.

pentschev:
> * I _still_ think that statistics providers should return a `shared_ptr<Statistics> const&`. That way the caller can either upgrade to an owning `shared_ptr` or just keep a reference. The converse is not possible: if we return a `shared_ptr<Statistics>` there is no way to downgrade it to a reference.

What I don't like about returning a const-ref to a `shared_ptr` is that it's easy to overlook and assume you're getting a copy with potential for use-after-free. One can argue that's true for all const-ref, but a smart pointer falls under a "safe" category in my mental model at least.

wence-:
> > * I _still_ think that statistics providers should return a `shared_ptr<Statistics> const&`. That way the caller can either upgrade to an owning `shared_ptr` or just keep a reference. The converse is not possible: if we return a `shared_ptr<Statistics>` there is no way to downgrade it to a reference.
> 
> What I don't like about returning a const-ref to a `shared_ptr` is that it's easy to overlook and assume you're getting a copy with potential for use-after-free. One can argue that's true for all const-ref, but a smart pointer falls under a "safe" category in my mental model at least.

So there are some usual options, I think:

```c++
struct Foo {
    std::shared_ptr<Statistics> const& statistics();
};

struct Owns {
   std::shared_ptr<Statistics> stats_;
};

Owns takes_ref(std::shared_ptr<Statistics> const& stat) {
   return Owns{stat};
}
...

auto stats = std::make_shared<Statistics>();
Foo foo{stats};

foo.statistics()->some_stats_method(); // This can never be use-after-free because foo is alive

auto owning = takes_ref(foo.statistics()); // transparently upgraded in Owns ctor

auto owning_stat = foo.statistics(); // owning

auto &ref_stat = foo.statistics(); // reference, but obvious
```

pentschev:
> So there are some usual options, I think:
> 
> ```c++
> struct Foo {
>     std::shared_ptr<Statistics> const& statistics();
> };
> 
> struct Owns {
>    std::shared_ptr<Statistics> stats_;
> };
> 
> Owns takes_ref(std::shared_ptr<Statistics> const& stat) {
>    return Owns{stat};
> }
> ...
> 
> auto stats = std::make_shared<Statistics>();
> Foo foo{stats};
> 
> foo.statistics()->some_stats_method(); // This can never be use-after-free because foo is alive
> 
> auto owning = takes_ref(foo.statistics()); // transparently upgraded in Owns ctor
> 
> auto owning_stat = foo.statistics(); // owning
> 
> auto &ref_stat = foo.statistics(); // reference, but obvious
> ```

That looks reasonable to me.

madsbk:
But you only get a reference if you explicitly assign to one, right?

```c++
auto& ref_stat = foo.statistics();  // reference, but explicit
```

Please, let's not introduce a special "reference semantics for smart pointers" concept like:

```c++
Owns takes_ref(std::shared_ptr<Statistics> const& stat) {
    return Owns{stat};
}
```

If you assign to a reference, you are explicitly choosing to take on the associated lifetime risk. That is already true for references in general, and arguably even more so for non-const references. So if someone is not confident about the lifetime guarantees, they simply should not bind to a reference.

wence-:
> But you only get a reference if you explicitly assign to one, right?
> 
> ```c++
> auto& ref_stat = foo.statistics();  // reference, but explicit
> ```

Yes. You have to request a reference explicitly to get a reference

wence-:
Note, I too do not advocate the "takes a reference and upgrades it some time later" model.

nirandaperera:
@wence- @pentschev I am marking this as draft because, I want to revisit `MutableStatisticsProvider`, `set_statistics()` approach.  
https://github.com/rapidsai/rapidsmpf/pull/1017/commits/a6803e664d825f5cb1d94007121cc356ab489fa3

I think this is a more safer approach.

cpp/include/rapidsmpf/statistics.hpp
nirandaperera: @wence- In cudf-polars engines, there is a `_reset_worker(...options)` method. There, we are only resetting buffer resource, context, etc with a new stats object. Then, after calling `_reset_worker`, we would have multiple instances of statistics in the engine ([commuicator, progress thread] vs [buffer resource, context])

wence-: But why do we need to reset the statistics in this setup? When is `_reset_worker` called? Is it only for tests? For every query, something else?

What do we actually need to achieve here? It seems like we want to zero the statistics object that we have around no?

nirandaperera: I was looking into two options. 
1. Introduce `set_stats` for communicator and progress thread and reset the new stats object
2. reset the stats object in-place (clear + enable/disable based on new options)

I think 1. is the safer option.

wence-: That is the _mechanism_ for how to reset stats. I am asking a different question: why do we want to reset stats?

cpp/src/communicator/communicator.cpp
nirandaperera: The reason I took this route is,[ when we enable coarse-grained stats ](https://github.com/rapidsai/rapidsmpf/pull/999/changes#diff-73f7b095035184c1af5cf9c684b44a2706788abe658e5d4617472b446524ab49), we will be calling into stats. So, its best to store a local instance. Either we do it in MPI and UCXX child classes or keep it protected in the parent. I thought latter would be add less code.

cpp/include/rapidsmpf/statistics.hpp
nirandaperera: Because the new options string [here](https://github.com/rapidsai/cudf/blob/main/python/cudf_polars/cudf_polars/engine/dask.py#L297) _could_ have a different config for stats that what it is currently. And if we decide to add more configurations to statistics later on (currently we only have enable/disable config), things can get ugly IMO.

nirandaperera: I think this stems from `StreamingEngine._reset` method. https://github.com/rapidsai/cudf/blob/main/python/cudf_polars/cudf_polars/engine/core.py#L289

cpp/src/communicator/communicator.cpp
wence-: OK, but I think we should therefore consider storing the thing that provides statistics and is shared (the progress thread). Rather than pulling out the statistics.

cpp/include/rapidsmpf/statistics.hpp
wence-: OK, `_reset` is only called in the tests (it is a "hack" to workaround that booting a dask or ray cluster is expensive), and I do not think that we have to tie ourselves in knots to publicly support that api _also_ being able to reset statistics. Note that the docstring of `_reset` _already_ says that it doesn't reset and reinitialize with _all_ the options.

I think we have enough rope with `.enable()` and `.clear()` on the stats object, without introducing a new thing.

nirandaperera: okay. Let's do that.

nirandaperera: I felt `clear()` was insufficient because it only clears the stats, but not the memory records and report entries. So, we can have a scenario where we create a new Buffer resource with new rmm resource adapter, but stats contains old memory records from a previous buffer resource.

wence-: ah, but that problem can be fixed (separately) by clearing memory records

cpp/benchmarks/streaming/bench_streaming_shuffle.cpp
nirandaperera: It does, but its the same as `stats` in this scope

python/rapidsmpf/rapidsmpf/statistics.pyx
nirandaperera: Sorry, I changed the impl at somepoint but forgot to update the docstring

python/rapidsmpf/rapidsmpf/progress_thread.pyx
```diff
@@ -29,14 +31,41 @@ cdef class ProgressThread:
...
+
+        Parameters
+        ----------
+        statistics
+            The new statistics instance. Pass ``Statistics.disabled()`` to opt out of
+            statistics collection.
+
+        Warnings
+        --------
+        Concurrent calls to ``set_statistics`` will result in undefined behavior.
```
pentschev: Why is this warning not documented in the C++ docstring?

cpp/include/rapidsmpf/progress_thread.hpp
```diff
@@ -183,6 +184,24 @@ class ProgressThread {
...
+     * @param statistics The new statistics instance. Pass `Statistics::disabled()` to opt
+     * out of statistics collection.
+     *
+     * @throws std::invalid_argument If @p statistics is null. Validated
+     * before the pause so a null argument does not leave the progress loop
+     * paused.
+     *
+     * @warning Concurrent calls to `set_statistics()` will result in undefined behavior.
+     */
+    void set_statistics(std::shared_ptr<Statistics> statistics);
```
pentschev, line 199: This makes `ProgressThread` a mutable provider, but existing `Statistics` handles are snapshots. A caller can do `auto stats = progress->statistics()`, then another path calls `set_statistics(new_stats)`, and the caller continues recording/reporting against the old object while `ProgressThread` and `Communicator` report the new one. I think this reintroduces the "which stats object owns this?" ambiguity the PR is trying to remove. If runtime reconfiguration is needed, mutating the existing `Statistics` instance is safer than swapping identity. If swapping remains, docs/tests should explicitly state that previously returned handles are not retargeted.

cpp/src/progress_thread.cpp
```diff
@@ -119,6 +119,22 @@ std::shared_ptr<Statistics> ProgressThread::statistics() const noexcept {
...
+        std::invalid_argument
+    );
+    // Pausing ensures the progress loop is not in the middle of dereferencing
+    // `statistics_` while we swap it. Restore the prior running state on exit.
+    bool const was_running = is_running();
+    pause();
+    statistics_ = std::move(statistics);
+    if (was_running) {
+        resume();
+    }
```
pentschev: Calling `statistics()` while this runs is a race. Won't simply adding a mutex to protect `statistics_` or making it atomic suffice to prevent the race and simultaneously prevent anyone from dereferencing it in the progress thread without requiring to pause?

cpp/benchmarks/streaming/bench_streaming_shuffle.cpp
```diff
@@ -364,20 +366,11 @@ int main(int argc, char** argv) {
...
     auto br = std::make_shared<rapidsmpf::BufferResource>(
-        stat_enabled_mr,
-        pinned_mr,
-        std::move(memory_limits),
-        std::nullopt,
-        std::make_shared<rmm::cuda_stream_pool>(
-            16, rmm::cuda_stream::flags::non_blocking
-        ),
-        stats
+        stats, stat_enabled_mr, pinned_mr, std::move(memory_limits)
```
pentschev: Unlike other cases as per https://github.com/rapidsai/rapidsmpf/pull/1017/files#r3317226453, this is changing behavior. `periodic_spill_checking` was set to `set::nullopt`, but the default is `1ms`. Is this what we want here?

nirandaperera: Ah! good catch. I reverted to the original args.

cpp/include/rapidsmpf/progress_thread.hpp
nirandaperera: Fair point. This is how I looked at this. 
Currently, `Statistics` only gets reset when we reset the engine. While running, we initiate async stat recording, like Memory records, stream ordered timings etc. Recording only happens at a later point, and there could be a scenario where the recording happens after a reset. If stat object is the same, we might end up with some stats from before. 
This PR handles it by handing off a weak ptr to async stats recording, and if the ptr has expired, recording is skipped.

cpp/src/progress_thread.cpp
nirandaperera: good call. Let me do that.

python/rapidsmpf/rapidsmpf/progress_thread.pyx
nirandaperera: It is, isnt it? https://github.com/rapidsai/rapidsmpf/pull/1017/changes/BASE..78d8a04e9a906720832b17d43a8192ea0823c691#diff-f9320c48070fa670c43a05334be536be21c0c779e5ab0a8e44d41f978339b0d3R201

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -656,6 +670,28 @@ concept StatisticsProvider = requires(T const& t) {
...
+ *
+ * Extends `StatisticsProvider` with a `set_statistics(std::shared_ptr<Statistics>)`
+ * method that replaces the provider's currently held instance with the given one.
+ * The new instance must not be null; implementations are expected to throw
+ * `std::invalid_argument` (or similar) on a null argument and otherwise install
+ * the new instance unconditionally.
+ *
+ * @warning Concurrent calls to `set_statistics()` are implementation-defined and
+ * likely result in undefined behavior unless the implementation explicitly
+ * documents otherwise. Callers should assume external synchronization is required.
```
pentschev, line 685: Nit: let's not leave this hanging on implementation details and just say this is not thread-safe/concurrent calls are UB.

python/rapidsmpf/rapidsmpf/progress_thread.pyx
pentschev: Yeah, sorry about that, don't know how I missed it. :sweat_smile:

cpp/include/rapidsmpf/progress_thread.hpp
pentschev: Ok, but what will happen to those objects that rely on the `Statistics` to be still alive? Do they have a mechanism to eventually get a new one, or will they just fall out of the `Statistics` loop forever?

nirandaperera: @pentschev If they rely on `Statistics`, it could be a problem. If they want to get the most current statistics, they would have to keep a strong reference to long-living stats provider like `ProgressThread`. 

>will they just fall out of the Statistics loop forever?

Yes. At least in current usages, IMO it will be safe to miss some recordings on the edge cases.

pentschev: What are the edge cases?

> Currently, `Statistics` only gets reset when we reset the engine.

Or more generically, when is the engine expected to reset? I assume be "engine reset" you're referring to cudf-polars' engine, is that right? One case where reset may occur is part of the testing framework, is that what you have in mind? What would other cases be?

nirandaperera: Yes, edge cases I was referring to was, in-flight stats recordings while engine is being reset. 

During tests, I think this is not happening, because we have a long running communicator/ progress thread in the test environment.

.github/workflows/pr.yaml
```diff
@@ -61,6 +61,7 @@ jobs:
           - '!.github/release.yml'
           - '!.pre-commit-config.yaml'
           - '!.shellcheckrc'
+          - '!SECURITY.md'
```
jameslamb, line 64: It's surprising to see these changes from #1064 showing up in the diff again. Those are already out on `main`.

Same with the `cupy` changes throughout packaging files. I suspect something has gone wrong with resolving a merge conflict here.

cpp/include/rapidsmpf/progress_thread.hpp
pentschev: Ok, can you please at a minimum document that previously returned handles are not retargeted?

.github/workflows/pr.yaml
nirandaperera: Oh yes! I think I've messed up bigtime during a merge..

nirandaperera: I fixed the merge mess @jameslamb

jameslamb: Ah great, thanks very much. Commit history looks much better now, and you no longer need `ci-codeowners` / `packaging-codeowners` so I'll unsubscribe from this.

`@` me back if you need my help with anything.

nirandaperera:
closing in favor of https://github.com/rapidsai/rapidsmpf/pull/1082

PR #1018, opened 2026-05-08
Address some performance regressions in shuffle

wence-:
After #927 we lost about 10% performance in the shuffle benchmarks when using cuda async memory.

Recover, from my benchmarking, most of this with a number of updates:

- Switch back to closer to the pre-#927 "wake" scheme. We now wake a waiter once all data is ready to be extracted and all sends have been posted (but not necessarily completed)
- Rather than breaking when we see the first ready buffer per rank, post all receives up to the first non-ready buffer (TODO: confirm this is safe with message ordering, I think it is)
- Apply a circulant shift to the polling for metadata so we don't all look for metadata from rank-0, then rank-1, etc...

In addition, to allow benchmarking just the communication part of the shuffle, add a "discard output without even concatenating it" mode to the shuffle benchmark.

### DIFF
--- a/cpp/benchmarks/bench_shuffle.cpp
+++ b/cpp/benchmarks/bench_shuffle.cpp
@@ -52,7 +52,7 @@ class ArgumentParser {
         }
         try {
             int option;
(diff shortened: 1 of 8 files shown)


wence-:
The waking is not safe yet.

PR #1019, opened 2026-05-08
Allow shuffle extraction before protocol drain

pentschev:
https://github.com/rapidsai/rapidsmpf/pull/927 made `Shuffler::wait()` wait until all internal protocol/send cleanup had drained before returning, which introduced a 10-20% regression in `bench_shuffle` on an 8-GPU node by delaying output extraction and unpack/concat work. For the reported 16 input partitions / 4 output partitions case, mean per-rank elapsed time improved from about 207 ms back down to about 172 ms after restoring overlap. This change makes `wait()` return when local shuffle output is ready to extract, while adding `wait_reusable()` for callers that need the stronger full-drain guarantee before reusing an op_id.


madsbk (commented):
Do we need to update the Python bindings as well?

What about renaming to `wait_drained()`? Or maybe `wait_for_drain()`?

PR #1022, opened 2026-05-12
Fix librapidsmpf wheel builds with current cuDF packages

pentschev:
Current Python wheel builds can fail when a subproject enables C after the top-level configure has already initialized C++ and CUDA. Initialize the language state early enough for that dependency while keeping the main C++ build from depending on C, and use the MPI C++ target for C++ consumers.

The current cuDF wheel dependency set also exposes nvCOMP as a runtime/link dependency without making it reliably available through the installed cuDF package metadata. Resolve and link that dependency explicitly where the affected static dependency chain is used, and keep auditwheel from trying to vendor the external runtime library.


cpp/CMakeLists.txt
```diff
@@ -322,7 +322,7 @@ target_link_libraries(
           $<TARGET_NAME_IF_EXISTS:rapidsmpf_cucascade_internal>
           $<$<NOT:$<TARGET_EXISTS:rapidsmpf_cucascade_internal>>:cuCascade::cucascade>
           $<$<BOOL:${RAPIDSMPF_HAVE_NUMA}>:numa>
-          $<TARGET_NAME_IF_EXISTS:MPI::MPI_C>
+          $<TARGET_NAME_IF_EXISTS:MPI::MPI_CXX>
           $<$<BOOL:${RAPIDSMPF_HAVE_CUPTI}>:CUDA::cupti>
           $<TARGET_NAME_IF_EXISTS:PMIx::PMIx>
```
wence-, line 327: We only rely on the C api of MPI, so I don't understand why this is changing.

python/librapidsmpf/CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ rapids_cuda_init_architectures(librapidsmpf-python)
 project(
   librapidsmpf-python
   VERSION "${RAPIDS_VERSION}"
-  LANGUAGES CXX CUDA
+  LANGUAGES C CXX CUDA
```
pentschev, line 21: This can be dropped once https://github.com/NVIDIA/cuCascade/pull/123 is merged.

cpp/CMakeLists.txt
pentschev: We still only use the MPI C API. The CMake `MPI::MPI_CXX` target is not the removed MPI C++ bindings, that is the separate `MPICXX` component. Per [`FindMPI`](https://cmake.org/cmake/help/latest/module/FindMPI.html), the `CXX` component is the MPI C API usable from C++. Since these are C++ targets and the project no longer enables C, `MPI::MPI_CXX` is the correct imported target.

madsbk (approved):
LGTM

ci/build_wheel_librapidsmpf.sh
```diff
@@ -40,6 +40,7 @@ python -m auditwheel repair \
     --exclude libcudf.so \
     --exclude libkvikio.so \
     --exclude libnvidia-ml.so.1 \
+    --exclude libnvcomp.so.5 \
```
bdice, line 42: Nit: These are not quite alphabetized.

cmake/thirdparty/get_cucascade.cmake
```diff
@@ -5,26 +5,33 @@
...
 #
-# NOTE: We explicitly find kvikio here because:
+# NOTE: We explicitly find kvikio and nvcomp here because:
 #
-# 1. cuCascade depends on cuDF, which depends on kvikio
+# 1. cuCascade depends on cuDF, which depends on kvikio and nvcomp
 # 2. cuCascade runs as a CMake subdirectory/subproject and needs GLOBAL targets to see dependencies
 # 3. CMake's transitive dependency handling can fail with mixed static/shared libraries
-# 4. Without explicit kvikio linking, the linker cannot find libkvikio.so when building tools
+# 4. Without explicit transitive linking, the linker cannot find cuDF's dependent libraries when
```
bdice, line 15: This seems unexpected to me. Maybe @vyasr will know if something broke with recent refactoring.

bdice (approved):
Approving to unblock CI but I’d like @vyasr to confirm the transitive dependencies after the recent cudf CMake changes.

cmake/thirdparty/get_cucascade.cmake
pentschev: This is a consequence of https://github.com/rapidsai/cudf/pull/22341 dependency/export refactor, explicit nvCOMP linking is currently needed for this nested static cuCascade path unless cuDF/RAPIDS CMake exports propagate enough link information for `libcudf.so`’s private DSO dependency.

ci/build_wheel_librapidsmpf.sh
pentschev: Fixed.

cpp/CMakeLists.txt
wence-: Thanks.

cmake/thirdparty/get_cucascade.cmake
vyasr: If the build of cucascade inside rapidsmpf has specific requirements for the nvcomp (or kvikio) target it shouldn't be relying on cudf to provide those targets. It should be using `rapids_cpm_find` or equivalent to ensure that it gets the targets compiled in the way that rapidsmpf requires.

mroeschke: I believe we're seeing the fallout of this change in the cuDF devcontainer CI job where rapidsmpf is being built from source and cmake cannot find nvcomp for cucascade e.g. https://github.com/rapidsai/cudf/actions/runs/25762513526/job/75667189794#step:9:13047

```bash
CMake Error at /home/coder/rapidsmpf/cmake/thirdparty/get_cucascade.cmake:25 (find_package):
  Could not find a package configuration file provided by "nvcomp" with any
  of the following names:

    nvcomp.cps
    nvcompConfig.cmake
    nvcomp-config.cmake

  Add the installation prefix of "nvcomp" to CMAKE_PREFIX_PATH or set
  "nvcomp_DIR" to a directory containing one of the above files.  If "nvcomp"
  provides a separate development package or SDK, be sure it has been
  installed.
Call Stack (most recent call first):
  /home/coder/rapidsmpf/cmake/thirdparty/get_cucascade.cmake:66 (find_and_configure_cucascade)
  CMakeLists.txt:135 (include)

```

python/librapidsmpf/CMakeLists.txt
nirandaperera: Its merged, should we remove this now?

pentschev: It's already removed: https://github.com/rapidsai/rapidsmpf/pull/1024

PR #1023, opened 2026-05-12
Remove py_executor from run_actor_network

### LINKED ISSUE #855: Set default executor in `run_streaming_pipeline`

wence-:
This, despite its name, was only ever used to offload the top-level asyncio.run to a thread so that it could run concurrently with the C++ actors. The multi-threading should belong in the application that chooses to offload work.

To facilitate this, make an awaitable version of when_all to wait for a list of C++ actors. We still need a single thread to run the resulting awaitable tasks because we do so in a fresh event loop which conflicts with any event loop that the cluster manager might have.

- Closes #855

### DIFF
--- a/python/rapidsmpf/rapidsmpf/examples/streaming/basic_example.py
+++ b/python/rapidsmpf/rapidsmpf/examples/streaming/basic_example.py
@@ -4,7 +4,6 @@
 
 from __future__ import annotations
 
-from concurrent.futures import ThreadPoolExecutor
 from typing import TYPE_CHECKING
 
 import cudf
@@ -47,9 +46,6 @@ def main() -> int:
         br=BufferResource(RmmResourceAdaptor(rmm.mr.get_current_device_resource())),
         options=options,
     )
-
-    # Executor for Python actors (asyncio coroutines).
-    py_executor = ThreadPoolExecutor(max_workers=1)
 
     # Create some pylibcudf tables as input to the streaming graph.
     tables = [
@@ -120,12 +116,12 @@ def main() -> int:
 
     # Run all actors. This blocks until every actor has completed.
     run_actor_network(
+        ctx,
         actors=(
             actor1,
             actor2,
             actor3,
         ),
-        py_executor=py_executor,
     )
 
     # Collect and verify results.
--- a/python/rapidsmpf/rapidsmpf/streaming/core/actor.pyi
+++ b/python/rapidsmpf/rapidsmpf/streaming/core/actor.pyi
@@ -4,7 +4,6 @@
 from __future__ import annotations
 
 from collections.abc import Awaitable, Callable, Collection
-from concurrent.futures import ThreadPoolExecutor
 from typing import Concatenate, ParamSpec
 
 from rapidsmpf.streaming.core.channel import Channel
@@ -22,7 +21,5 @@ def define_actor(
     Callable[Concatenate[Context, P], Awaitable[None]],
 ]: ...
 def run_actor_network(
-    *,
-    actors: Collection[CppActor | Awaitable[None]],
-    py_executor: ThreadPoolExecutor | None = None,
+    ctx: Context, *, actors: Collection[CppActor | Awaitable[None]]
 ) -> None: ...
--- a/python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
+++ b/python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
@@ -1,20 +1,64 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES.
 # SPDX-License-Identifier: Apache-2.0
 
+from cpython.object cimport PyObject
+from cpython.ref cimport Py_INCREF
 from cython.operator cimport dereference as deref
-from libcpp.memory cimport unique_ptr
+from libcpp.memory cimport shared_ptr, unique_ptr
 from libcpp.utility cimport move
 
 import asyncio
 import inspect
 from collections.abc import Iterable, Mapping
+from concurrent.futures import ThreadPoolExecutor
 from functools import partial, wraps
+
+from rapidsmpf._detail.exception_handling cimport ex_handler
+from rapidsmpf.owning_wrapper cimport cpp_OwningWrapper
+from rapidsmpf.streaming._detail.libcoro_spawn_task cimport cpp_set_py_future
+from rapidsmpf.streaming.chunks.utils cimport py_deleter
+from rapidsmpf.streaming.core.context cimport Context, cpp_Context
 
 from rapidsmpf.streaming.core.channel import Channel
 from rapidsmpf.streaming.core.context import Context
 
 from rapidsmpf.streaming.core.context cimport Context
 
+
+cdef extern from * nogil:
+    """
+    namespace {
+    coro::task<void> cpp_when_all_task(
+        std::vector<rapidsmpf::streaming::Actor> actors
+    ) {
+        rapidsmpf::streaming::coro_results(co_await coro::when_all(std::move(actors)));
+    }
+
+    void cpp_when_all(
+        std::shared_ptr<rapidsmpf::streaming::Context> ctx,
+        std::vector<rapidsmpf::streaming::Actor> actors,
+        void (*cpp_set_py_future)(void *, const char *),
+        rapidsmpf::OwningWrapper py_future
+    ) {
+        RAPIDSMPF_EXPECTS(
+            ctx->executor()->spawn_detached(
+                cython_libcoro_task_wrapper(
+                    cpp_set_py_future,
+                    std::move(py_future),
+                    cpp_when_all_task(std::move(actors))
+                )
+            ),
+            "libcoro's spawn_detached() failed to spawn task"
+        );
+    }
+    }
+    """
+    void cpp_when_all(
+        shared_ptr[cpp_Context],
+        vector[cpp_Actor],
+        void (*cpp_set_py_future)(void*, const char *),
+        cpp_OwningWrapper py_future
+    ) except +ex_handler
 
 cdef class CppActor:
     """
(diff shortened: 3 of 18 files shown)


wence-:
Marking as do not merge because I don't want to get this in to 26.06, I think.

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyi
```diff
@@ -22,7 +21,5 @@ def define_actor(
     Callable[Concatenate[Context, P], Awaitable[None]],
 ]: ...
 def run_actor_network(
-    *,
-    actors: Collection[CppActor | Awaitable[None]],
-    py_executor: ThreadPoolExecutor | None = None,
+    ctx: Context, *, actors: Collection[CppActor | Awaitable[None]]
```
TomAugspurger, line 24: Given that breaking changes to rapidsmpf now block cudf CI, we should at least consider making this kind of change more gracefully now (deprecate here, remove the usage in cudf-polars, remove entirely here). I'm fine with doing all that in a single release, just spread over multiple PRs.

But, I also don't want to slow things down here too much, and I won't want to have this conversation each time we introduce a breaking change. Maybe we can use this as a chance to write down a policy.

And, I just noticed that for this specific case, we can avoid breaking cudf-polars' CI by removing `py_executor` [here](https://github.com/rapidsai/cudf/blob/4d9b086559e459f900d042f99e5cf872015632e7/python/cudf_polars/cudf_polars/experimental/rapidsmpf/frontend/core.py#L447).

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
```diff
@@ -179,31 +222,66 @@ def define_actor(*, extra_channels=()):
     return partial(decorate_actor, extra_channels)
 
 
-def run_actor_network(*, actors, py_executor = None):
+def sync_wait(fn):
```
madsbk: Add docstring

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
```diff
@@ -225,33 +303,25 @@ def run_actor_network(*, actors, py_executor = None):
...
-            with nogil:
-                cpp_run_actor_network(move(cpp_actors))
-    finally:
-        if len(py_actors) > 0:
-            py_future.result()  # This will raise any unhandled exception.
+    py_actors = [when_all(ctx, cpp_actors), *py_actors]
+    # Need to run in a separate thread in case the cluster runtime already
+    # has an async event loop.
+    executor = ThreadPoolExecutor(max_workers=1)
+    executor.submit(sync_wait, run_py_actors(py_actors)).result()
```
madsbk:
```suggestion
    with ThreadPoolExecutor(max_workers=1) as executor:
        executor.submit(sync_wait, run_py_actors(py_actors)).result()
```

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
```diff
@@ -225,33 +303,25 @@ def run_actor_network(*, actors, py_executor = None):
...
 
-    if len(py_actors) > 0:
-        if py_executor is None:
-            raise ValueError("must provide a py_executor to run Python actors.")
-        py_future = py_executor.submit(asyncio.run, run_py_actors(py_actors))
-
-    try:
-        if cpp_actors.size() > 0:
-            with nogil:
-                cpp_run_actor_network(move(cpp_actors))
```
madsbk, line 254: Remove the `cpp_run_actor_network` from the `.pxd` file as well

madsbk (changes requested):
Thanks @wence-, I think we should include this in this release!

python/rapidsmpf/rapidsmpf/examples/streaming/basic_example.py
```diff
@@ -120,12 +116,12 @@ async def count_num_rows(
 
     # Run all actors. This blocks until every actor has completed.
     run_actor_network(
+        ctx,
         actors=(
             actor1,
             actor2,
             actor3,
         ),
-        py_executor=py_executor,
```
Matt711, line 128: What are the implications for setting the `num_py_executors` config option in cudf-polars after removing this? I don't understand what we're hoping to improve with this change.

FYI: I found this PR whilst trying to understand what `num_py_executors` even is.

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
```diff
@@ -179,31 +222,66 @@ def define_actor(*, extra_channels=()):
     return partial(decorate_actor, extra_channels)
 
 
-def run_actor_network(*, actors, py_executor = None):
+def sync_wait(fn):
+    loop = asyncio.new_event_loop()
+    try:
+        loop.run_until_complete(fn)
+    finally:
+        loop.close()
```
mroeschke: Should we also be calling `loop.run_until_complete(loop.shutdown_asyncgens())` in this finally block? https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.shutdown_asyncgens

Or possibly we can rely on the TaskGroup in `run_py_actors` to close these

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyi
wence-: In some ways, this change is no different from a breaking change in RMM. I have the matching cudf PR up, but that is a reasonable suggestion.

python/rapidsmpf/rapidsmpf/examples/streaming/basic_example.py
wence-: Previously, `num_py_executors` set the size of the thread pool passed in to `py_executor` here. We only ever used one of those threads, independent of the value provided.

The number of python execution threads we use _today_ in cudf-polars is hardware-dependent and selected by the asyncio runtime. I fix that in https://github.com/rapidsai/cudf/pull/22474

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
wence-: Oh, probably yes.

wence-: Used `asyncio.Runner` instead.

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
wence-: Done.

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
wence-: Done.

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
wence-: Done.

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
```diff
@@ -179,31 +223,89 @@ def define_actor(*, extra_channels=()):
...
+    Notes
+    -----
+    This should always be called from a thread we control to ensure no live
+    event loop is running.
+    """
+    with asyncio.Runner() as runner:
+        runner.run(coro)
+
+
+async def when_all(Context ctx not None, list actors):
```
madsbk:
```suggestion
async def when_all(Context ctx not None, list cpp_actors):
```

mroeschke (approved):

nirandaperera (approved):
LGTM. Thanks @wence-

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
```diff
@@ -179,31 +224,89 @@ def define_actor(*, extra_channels=()):
...
+            # This could still fail so we catch and reraise the
+            # cancellation but recording the C++ exception as well.
+            await asyncio.shield(ret)
+        except Exception as cpp_except:
+            raise cancel from cpp_except
+        # Otherwise just reraise the cancellation
+        raise cancel
+
+
+def run_actor_network(Context ctx not None, *, actors):
```
madsbk, line 288: Should we accept (and ignore) a `py_executor` until https://github.com/rapidsai/cudf/pull/22474 has been merged?
```suggestion
def run_actor_network(Context ctx not None, *, actors, py_executor = None):
```

madsbk (approved):
@wence- did you run some numbers?

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
wence-: This doesn't help because the `ctx` argument is also new.

wence-:
> @wence- did you run some numbers?

Yes. With 8 threads in cudf-polars performance on the hardware I ran on was within the noise compared to main.

PR #1028, opened 2026-05-12
Bringing all default options to one place

nirandaperera:
## `[C++/Python]` Centralize option keys and defaults via `OptionDescriptor`

Option keys and default values were scattered as raw string literals across C++ and Python call sites, making it easy for the two languages to drift and for typos to go unnoticed. This PR consolidates every recognised configuration option into a single source of truth in `cpp/include/rapidsmpf/config.hpp` and exposes the same keys and defaults to Python through a new `rapidsmpf.config_defaults` module that pulls the values directly from the C++ header at import time.

- Add `OptionDescriptor{key, default_val}` (non-templated) in `config.hpp`, with **both fields stored as `std::string_view`**. Defaults are written as string literals (e.g. `"false"`, `"16"`, `"1ms"`, `"WARN"`) since options are always parsed from their string representation — the descriptor's default flows through the same factory the call site uses for user-supplied values.
- Replace ad-hoc key/default literals across `statistics.cpp`, `buffer_resource.cpp`, `pinned_memory_resource.cpp`, `communicator.cpp`, `ucxx.cpp`, `coro_executor.cpp`, and `memory_reserve_or_wait.cpp` with the descriptor-based API. Typed defaults (bool / size_t / uint32_t) are parsed via `parse_string<T>` so the canonical string form is the single source of truth.
- Widen `Options::get` to accept `std::string_view` so descriptor keys can be passed directly (existing `std::string` / `const char*` callers still work via implicit conversion).
- Widen `parse_string<T>` to accept `std::string_view` so descriptor defaults flow through unchanged.
- Introduce `rapidsmpf/config_defaults.{pyx,pyi}` exposing one `Final[str]` constant per option (e.g. `BUFFER_RESOURCE_NUM_STREAMS`) plus a read-only `DEFAULTS: Mapping[str, str]` (`MappingProxyType`) sourced from the C++ header via `cdef extern`. A single `RMPF_OPT` macro re-exposes `key.data()` and `default_val.data()` as null-terminated `const char*` aliases — safe because both views are initialised from string literals.
- Migrate Python consumers (`memory/buffer_resource.pyx`, `streaming/core/memory_reserve_or_wait.pyx`, `integrations/core.py`, `tests/test_config.py`) to import from `config_defaults` instead of using raw string keys / hard-coded defaults. Consumers that need typed values parse the string default on-site (e.g. `int(_OPTION_DEFAULTS[BUFFER_RESOURCE_NUM_STREAMS])`, `parse_boolean(_OPTION_DEFAULTS[STREAMING_ALLOW_OVERBOOKING_BY_DEFAULT])`), mirroring the C++ side.
- Promote the previously unnamed streaming option to `streaming::AllowOverbookingByDefaultOption` and drop the `Factor` suffix from the pinned-pool size descriptors.


cpp/include/rapidsmpf/config.hpp
```diff
@@ -380,4 +384,128 @@ std::unordered_map<std::string, std::string> get_environment_variables(
...
+ * @tparam T The C++ type used to represent the option's parsed value.
+ *           For options whose default is a percentage / nbytes / duration
+ *           string, T is `std::string_view` and the default is parsed by the
+ *           same factory the call site uses for user-supplied values.
+ */
+template <typename T>
+struct OptionDescriptor {
+    char const* key;  ///< Null-terminated lookup key passed to `Options::get`.
+    T default_val;  ///< Value used when the option is unset.
+};
```
madsbk, line 410: Since every option has to be able to parse the option from a string, could we simplify by always having the default as a string?
```
struct OptionDescriptor {
    std::string const key;
    std::string const default_val;
};
```

nirandaperera: @madsbk I was thinking about it. I felt like, typed defaults can remove an extra string parse. But I dont have strong opinions about it TBH.

madsbk: I think you should do it :)

nirandaperera: @madsbk Done

cpp/include/rapidsmpf/config.hpp
```diff
@@ -242,10 +246,10 @@ class Options {
      * the same key will result in a `std::bad_any_cast`.
      */
     template <typename T>
-    T const& get(std::string const& key, OptionFactory<T> factory) {
+    T const& get(std::string_view key, OptionFactory<T> factory) {
```
madsbk: Why not just keep  `std::string const& key`?

cpp/include/rapidsmpf/config.hpp
```diff
@@ -380,4 +384,130 @@ std::unordered_map<std::string, std::string> get_environment_variables(
...
+ * always parsed from their string representation at runtime, so the default
+ * is expressed as a string and fed through the same factory the call site
+ * uses for user-supplied values. Descriptors must be initialized from string
+ * literals so that `key.data()` and `default_val.data()` yield
+ * null-terminated `char const*` pointers, which are consumed directly by the
+ * Cython bindings.
+ */
+struct OptionDescriptor {
+    std::string_view key;  ///< Lookup key passed to `Options::get`.
+    std::string_view default_val;  ///< String form of the value used when unset.
```
madsbk: Maybe just use `std::string` to avoid this complication, I don't think it will make any difference performance wise?

cpp/include/rapidsmpf/utils/string.hpp
```diff
@@ -239,8 +239,8 @@ Duration parse_duration(std::string_view text);
  * double d = parse_string<double>("3.14");    // d == 3.14
  */
 template <typename T>
-T parse_string(std::string const& text) {
-    std::stringstream sstream(text);
+T parse_string(std::string_view text) {
+    std::stringstream sstream{std::string{text}};
```
madsbk: Why not `std::string const& text` and avoid a copy?

madsbk (changes requested):
Remove `CMakeFiles/test_sources.dir/test_host_buffer.cpp.o`

cpp/include/rapidsmpf/config.hpp
nirandaperera: wanted to preserve the `constexpr` for `OptionDescriptor`

madsbk (changes requested):
@nirandaperera could you reconsider the string-handling complexity here?

I would like to keep a plain `std::string const& key` in `Options::get`, and drop the `StringHash` / transparent-lookup machinery along with it.

Tracing it back, the whole apparatus seems to exist to serve one decision: making the `OptionDescriptor`s `inline constexpr`. That forces `std::string_view` members, which then forces the heterogeneous-hashing dance in `get()` to avoid a temporary allocation per call.

But nothing in the codebase actually consumes the descriptors in a `constexpr` context. If we switch the descriptors to `inline const` with `std::string` members, the chain collapses:

```cpp
options.get(EnabledOption.key, ...)
```

binds directly to `std::string const&`, with no temporary and no transparent hashing.

The only thing that matters at runtime here is the performance of `Options::get()` itself.

cpp/include/rapidsmpf/utils/string.hpp
madsbk: ping

madsbk (changes requested):
Thanks @nirandaperera, looks good now but please remove `CMakeFiles/test_sources.dir/test_host_buffer.cpp.o` :)

nirandaperera:
> Thanks @nirandaperera, looks good now but please remove `CMakeFiles/test_sources.dir/test_host_buffer.cpp.o` :)

:facepalm:

cpp/include/rapidsmpf/config.hpp
```diff
@@ -380,4 +384,127 @@ std::unordered_map<std::string, std::string> get_environment_variables(
     std::string const& key_regex = "RAPIDSMPF_(.*)"
 );
 
-}  // namespace rapidsmpf::config
+}  // namespace config
```
madsbk: I think everything should live under the `rapidsmpf::config` namespace, and then be accessed through aliases:

```c++
using Log = config::communicator::LogOption;

Log::key
Log::default_val
```

I think that is both more readable and avoids cluttering the `rapidsmpf` namespace. WDYT?

nirandaperera: @madsbk that was my initial approach too. But then I felt it made the callsites too clunky. 
This will require call option descriptors to be prefixed by the namespace/ usage. 
ex: 
`statistics::EnabledOption --> StatisticsEnabledOption`, `pinned_memory::EnabledOption --> PinnedMemoryEnabledOption`

The other aspect in namespaces was, we can group descriptors cleanly.

madsbk: You can still add new namespaces under `rapidsmpf::config`, e.g. `rapidsmpf::config::statistics::EnabledOption`. 
You could even just call it `rapidsmpf::config::statistics::Enabled`.

nirandaperera: Done 
https://github.com/rapidsai/rapidsmpf/pull/1028/commits/03464effc6d3dfcf63f65aa29699aa0eaa285d72

python/rapidsmpf/rapidsmpf/streaming/core/memory_reserve_or_wait.pyx
```diff
@@ -603,7 +608,10 @@ async def reserve_memory(
     """
     if allow_overbooking is None:
         allow_overbooking = ctx.options().get_or_default(
-            "allow_overbooking_by_default", default_value=True
+            STREAMING_ALLOW_OVERBOOKING_BY_DEFAULT,
+            default_value=_parse_boolean(
+                _OPTION_DEFAULTS[STREAMING_ALLOW_OVERBOOKING_BY_DEFAULT]
+            ),
```
madsbk: Parse this as module level to reduce overhead, something like
```python
_ALLOW_OVERBOOK_DEFAULT = _parse_boolean(_OPTION_DEFAULTS[…]
```

cpp/src/communicator/communicator.cpp
```diff
@@ -4,23 +4,24 @@
...
 namespace {
+using config::communicator::LogOption;
+
 Communicator::Logger::LOG_LEVEL level_from_string(std::string const& str) {
-    if (str.empty()) {
-        return Communicator::Logger::LOG_LEVEL::WARN;  // Default log level.
-    }
-    auto trimmed = to_upper(trim(str));
+    auto const value =
+        str.empty() ? std::string{LogOption.default_val} : to_upper(trim(str));
```
madsbk:
```suggestion
        to_upper(trim(str.empty() ? LogOption.default_val : str));
```

madsbk (changes requested):
almost there

cpp/src/communicator/communicator.cpp
nirandaperera: `LogOption.default_val` should already be trimmed and uppercase isnt it?

nirandaperera: Ah! nvm!

python/rapidsmpf/rapidsmpf/streaming/core/memory_reserve_or_wait.pyx
nirandaperera: Templated option descriptors would have helped here TBH. :innocent: anyways. I've updated the PR now.

madsbk (changes requested):
After more consideration, I think we can make this much simpler by using a global const map: https://github.com/rapidsai/rapidsmpf/pull/1062

We lose some compile-time type checking, but the code becomes much simpler, and any option typos should be caught by the test suite anyway.

For context, adding a single option with the current approach touches around seven places: the header descriptor, C++ call site, Cython macro invocation, two `cdef extern` aliases (`_k_*` and `_d_*`), Python `Final[str]` constant, `DEFAULTS` map entry, `__all__` list, and `.pyi` stub.

The Cython bridge alone is about 70 lines of mechanical boilerplate for 12 options.

nirandaperera:
closing in favor of 
https://github.com/rapidsai/rapidsmpf/pull/1062

PR #1030, opened 2026-05-13
Use new cuCascade::cucascade_topology_discovery target

KyleFromNVIDIA:


cmake/thirdparty/get_cucascade.cmake
vyasr, line 1: We should also be able to drop all of the nvcomp and kvikio linking magic in this file now, right?

vyasr (changes requested):

cmake/thirdparty/get_cucascade.cmake
KyleFromNVIDIA: Right, done

vyasr (approved):
LGTM, thanks Kyle!

PR #1035, opened 2026-05-14
More follow-ups to recent cucascade fixes

KyleFromNVIDIA:
- Vendor `FindCUDAToolkit` as of cdacf44fe871fb059159a60117d6ed3164ccfbd9
- Include nvml in dependencies

Also temporarily disable smoke tests due to ongoing driver issues.


dependencies.yaml
```diff
@@ -42,6 +42,7 @@ files:
       - build-cpp
       - build-python
       - build-mpi
+      - build-nvml
```
bdice, line 16: Why is this here in “devcontainers” but not in “all”? Devcontainers should be a subset of all.

Also I think we can just go back to using all, since ray packaging has been fixed up afaik.

KyleFromNVIDIA: Removed `devcontainer` file set

KyleFromNVIDIA: I think removing the `devcontainers` file set will require changing https://github.com/rapidsai/devcontainers. Let's save that for a follow-up. I've added this to `all`.

KyleFromNVIDIA:
https://github.com/rapidsai/devcontainers/pull/699 should fix the pip devcontainer failures

bdice (approved):

bdice (approved):

PR #1039, opened 2026-05-15
Add ``OrderScheme.get_boundaries`` API

rjzamora:
While implementing a prototype to *use* `OrderScheme` to sort a table in cudf_polars, I realized we need a Python method to extract the boundaries table.

~**Note**: I decided it was better to return `tuple[Table, Slice]` than `TableChunk`, because this data is not uniquely owned, and we don't really have a python API for `shared_ptr<TableChunk>`.~


python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -182,6 +184,14 @@ cdef class OrderScheme:
         """Number of boundary rows (N-1 for N partitions)."""
         return self._handle.boundaries.get().shape().first
 
+    def get_boundaries(self) -> tuple[Table, Stream]:
+        """Return the boundary rows and their CUDA stream as a zero-copy view."""
+        cdef const cpp_TableChunk* chunk = self._handle.boundaries.get()
+        cdef Stream stream = Stream._from_cudaStream_t(chunk.stream().value())
+        return Table.from_table_view_of_arbitrary(
+            chunk.table_view(), owner=self, stream=stream
+        ), stream
```
wence-: You could return a `TableChunk` with `exclusive_view=False`. That's pretty close.

rjzamora: Ah, sorry. I made a mistake in my original code, and started overthinking this. I'm returning a `TableChunk` with `exclusive_view=False` now. Thanks.

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -182,6 +184,17 @@ cdef class OrderScheme:
...
+        """Return the boundary rows as a zero-copy TableChunk view."""
+        cdef const cpp_TableChunk* chunk = self._handle.boundaries.get()
+        cdef Stream stream = Stream._from_cudaStream_t(chunk.stream().value())
+        tbl = Table.from_table_view_of_arbitrary(
+            chunk.table_view(), owner=self, stream=stream
+        )
+        return TableChunk.from_pylibcudf_table(
+            tbl, stream, exclusive_view=False, br=br
+        )
+
```
wence-, line 209: As I am looking at this, did the OrderScheme need to keep the `BufferResource` corresponding to the `TableChunk` we created it from alive, I think yes?

rjzamora: Isn't the buffer resource attached to the context? The context should outlive this metadata I think. Am I misunderstanding the question?

madsbk: You are right, I think @wence- was asking what happens if users do not live up to that contract.

I am okay with this PR as-is, since the current contract is that the `BufferResource` outlives this metadata.

That said, I think this highlights a broader design issue we should address separately. We still do not have a clean ownership/lifetime story around `BufferResource` in Python.

I think it is time for me to start working on #641 :)

python/rapidsmpf/rapidsmpf/streaming/cudf/channel_metadata.pyx
```diff
@@ -182,6 +184,17 @@ cdef class OrderScheme:
         """Number of boundary rows (N-1 for N partitions)."""
         return self._handle.boundaries.get().shape().first
 
+    def get_boundaries(self, BufferResource br not None) -> TableChunk:
+        """Return the boundary rows as a zero-copy TableChunk view."""
```
wence-: Please to docstring correctly.

wence- (approved):

PR #1040, opened 2026-05-15
Add `CythonLibcoroTaskWrapperCallback` type alias for the asyncio Future-resolving callback

Matt711:
Because function pointers are confusing to read. And this helps some.


python/rapidsmpf/rapidsmpf/streaming/_detail/libcoro_spawn_task.pxd
```diff
@@ -34,7 +42,7 @@ cdef extern from * nogil:
      * and the Python Future has been notified.
      */
     coro::task<void> cython_libcoro_task_wrapper(
-        void (*cpp_set_py_future)(void*, const char *),
+        PyFutureCallback cpp_set_py_future,
```
Matt711, line 45: Or maybe `AsyncIOFutureCallback`?

madsbk: What about `cython_libcoro_task_wrapper_callback` to avoid the name clash? Since this is in a `.pxd`, it will effectively be copy-pasted into all Cython files importing it.

Alternatively, we could wrap the whole thing in a namespace.

Matt711: > What about cython_libcoro_task_wrapper_callback to avoid the name clash? 

Is there a current name clash or are you thinking of protecting against another `.pxd` using the same name with a different type?

wence-: Let's put the whole thing in an anonymous namespace I think.

Although then I would worry about ODR-use violations...

madsbk: > Is there a current name clash or are you thinking of protecting against another `.pxd` using the same name with a different type?

There is no current name clash. This is mostly about avoiding a future one.

Since this lives under `_detail`, we can always rename it later without a
deprecation period. Still, I think it is good style to use a unique name because this is
declared in a `.pxd` and may be pulled into different generated C++ files.

I do not think an anonymous namespace helps here. A `cdef extern from *` block
is effectively copied into the generated C++ source of each module that imports
it. So if a `.pyx` file also defines or imports a `PyFutureCallback` with a
different type, the names can still collide within that generated translation
unit.

That is why I prefer a more specific name like `cython_libcoro_task_wrapper_callback`.

Matt711: `CythonLibcoroTaskWrapperCallback` or no 😄 ?

madsbk: Heh, yeah I guess it should be `CythonLibcoroTaskWrapperCallback` :)
Since this is only used in one place, a large name is fine.

madsbk (approved):

PR #1043, opened 2026-05-15
Define `AllGather.__enter/exit__` for insert_finished

mroeschke:
In cudf_polars we often have to use a `finally: AllGather.insert_finished()` pattern to ensure that method is called after `AllGather.insert` e.g. https://github.com/rapidsai/cudf/pull/22516.

This PR defines `.__enter/exit__` like we do on `Context` to simplify this convention when we need to call `AllGather.insert`


python/rapidsmpf/rapidsmpf/streaming/coll/allgather.pyx
```diff
@@ -124,6 +124,13 @@ cdef class AllGather:
         with nogil:
             deref(self._handle).insert_finished()
 
+    def __enter__(self):
```
Matt711, line 130: I think we have to be careful in the streaming case because of https://github.com/rapidsai/rapidsmpf/blob/d8e53ff7552b538aae0f70fe990d3d775f2aade3/cpp/include/rapidsmpf/streaming/coll/allgather.hpp#L23-L30

It's the callers responsibility to do this. We probably should only do the "bulk" case in this PR. And keep using ` AllGatherManager.Inserter` in cudf-polars.

Matt711 (changes requested):

python/rapidsmpf/rapidsmpf/streaming/coll/allgather.pyx
mroeschke: Sure we can still use the `AllGatherManager.Inserter` API in cudf_polars, but this would help simplify that implementation too, with this `inserting` could become

```python
@contextmanager
def inserting(self):
    with self.allgather:
        yield self
```

and could move `Inserter.insert` onto `AllGatherManager.insert` if desired

Matt711: Oh ok I think that work, thanks.

Matt711: Yeah I think we can remove the `Inserter` class but we have to keep `inserting` cm, so the caller can control when `insert_finished` is called?

mroeschke: Yeah it's kinda nice to keep an API like `inserting` to signal that "after exiting the cm this calls `insert_finished`"

Matt711 (approved):

python/rapidsmpf/rapidsmpf/tests/streaming/test_allgather.py
```diff
@@ -125,10 +125,10 @@ async def allgather_and_concat(
...
 ) -> None:
     gather = AllGather(context, comm, op_id)
-    while (msg := await ch_in.recv(context)) is not None:
-        chunk = PackedDataChunk.from_message(msg, br=context.br()).to_packed_data()
-        gather.insert(msg.sequence_number, chunk)
-    gather.insert_finished()
+    with gather as ag:
+        while (msg := await ch_in.recv(context)) is not None:
+            chunk = PackedDataChunk.from_message(msg, br=context.br()).to_packed_data()
+            ag.insert(msg.sequence_number, chunk)
```
pentschev: Could we actually test both the context manager and the "raw" use with `insert_finished()`?

python/rapidsmpf/rapidsmpf/tests/test_allgather.py
```diff
@@ -145,15 +145,13 @@ def test_basic_allgather(
...
-
-    # Mark this rank as finished
-    allgather.insert_finished()
+    # Insert data from this rank and mark as finished
+    with allgather as ag:
+        for i in range(n_inserts):
+            packed_data = generate_packed_data(
+                n_elements, gen_offset(i, this_rank), stream, br
+            )
+            ag.insert(i, packed_data)
```
pentschev: Ditto.

pentschev (approved):
Left a suggestion that we keep testing both cases, otherwise LGTM.

wence-:
You need to ban `insert_finished` when being used as a context manager.

mroeschke:
> You need to ban `insert_finished` when being used as a context manager.

Done in da751a3b372e585e6845c23f4e21bf800cd8c23d

python/rapidsmpf/rapidsmpf/tests/test_allgather.py
```diff
@@ -113,16 +115,56 @@ def gen_offset(i: int, r: int) -> int:
...
+    n_elements: int,
+    n_inserts: int,
+) -> AllGather:
+    for i in range(n_inserts):
+        packed_data = generate_packed_data(
+            n_elements, gen_offset(i, allgather.comm.rank), stream, br
+        )
+        allgather.insert(i, packed_data)
+    allgather.insert_finished()
+    return allgather
```
wence-: It seems very sad that we have to define this function that is all but identical twice. Similarly in the define_actor setup

mroeschke: Fair point. In 4fd30048cd1281e41fdc472fcd03b339cba74f0b I used `contextlib.nullcontext` to have the existing logic test both the context manger and existing version.

wence- (approved):

PR #1049, opened 2026-05-18
Remove Dask integration

madsbk:
RapidsMPF no longer provides a Dask shuffle service. For Dask integration with Polars, use the [`DaskEngine`](https://github.com/rapidsai/cudf/blob/cfd60086d0a03df7ee22371bc37bcbc3e7999b5c/python/cudf_polars/cudf_polars/experimental/rapidsmpf/frontend/dask.py#L503) in cudf-polars.

I think this should target the `v26.06` release because we don't want to encourage users to use the legacy Dask integration going forward.


pentschev (commented):
An alternative if we don't want to make this large change still in 26.06 might be to deprecate the integration instead of completely removing it. Ideally we would have made this prior to burndown (I know pre-burndown week was hectic due to build issues, so there's a strong argument that we should still be allowed to move forward with this change).

madsbk:
I think this is a good time to remove it without any deprecation period.

As I see it, RapidsMPF has been in a kind of experimental state so far. I think `v26.06` will be our first *real* release. We still might be in beta, but cudf-polars will now use it by default, so it is starting to get real :)

pentschev:
> I think this is a good time to remove it without any deprecation period.
> 
> As I see it, RapidsMPF has been in a kind of experimental state so far. I think `v26.06` will be our first _real_ release. We still might be in beta, but cudf-polars will now use it by default, so it is starting to get real :)

I don't disagree with this and I think you're right. However, I think we should post internally that we're planning on doing this to confirm there are no objections (both from the removal perspective and timeline), then wait until EOD or tomorrow to merge.

KyleFromNVIDIA (approved):
CI changes for the removal look good. I won't comment on whether the overall removal is a good idea.

rjzamora:
I know we don't plan to use any of this in cudf-polars, but are we deciding that no one should use RapidsMPF with Dask as a "bulk shuffle" service the way people are doing with Ray?

I don't think we want to **maintain** the code you are removing. So, I'm not necessarily protesting. I'm just not feeling super confident. Recall that our rational for removing explicit comms from Dask-CUDA was that people should use RapidsMPF :)

python/rapidsmpf/rapidsmpf/examples/dask.py
```diff
@@ -1,444 +0,0 @@
...
-        RapidsMPF configuration options.
-
-    Returns
-    -------
-    Shuffled Dask-cuDF DataFrame collection.
-
-    Notes
-    -----
-    This API is currently intended for demonstration and
-    testing purposes only.
```
rjzamora, line 224: I suppose we always expected that we may want to remove all this ugly code needed to integrate with RapidsMPF within a task graph.

madsbk:
I strongly believe we do not want to maintain this, so I think we should remove it now.

That said, it might make sense to generalize cudf-polars’ `DaskEngine` and `RayEngine` into something that could form the basis for both cudf-polars and bulk shuffle services for Dask and Ray. However, I think that should be a completely new API design.

bdice (approved):

nirandaperera (approved):

PR #1054, opened 2026-05-20
Pandas 3 test fixes

nirandaperera:
## `[Python]` Fix `test_partition.py` failures under pandas 3.x

After cudf pulled in pandas 3.x as a transitive dependency, empty Python lists in `cudf.DataFrame({"0": [], "1": []})` get inferred as `object` dtype on input but come back as the new `StringDtype(na_value=nan)` after the pylibcudf round-trip, causing `assert_eq` to fail in `test_partition_and_pack_unpack`, `test_split_and_pack_unpack`, and `test_spill_unspill_roundtrip`. This pins the test data to an explicit `int64` dtype so empty inputs aren't ambiguously inferred and the round-trip stays consistent.

- Construct `expect` with `cudf.DataFrame(df, dtype="int64")` in the three affected tests (no-op for the existing integer-valued cases, disambiguates the empty case).

Depends on https://github.com/rapidsai/rapidsmpf/pull/1051


nirandaperera:
Closing as changes have already been applied by #1055

PR #1056, opened 2026-05-21
Avoid cudf (Python) as testing dependency

mroeschke:
Inspired by the breakage in cudf with pandas 3 support (https://github.com/rapidsai/rapidsmpf/pull/1054), I don't think there's a strong reason why tests need to involve cudf Python. This PR refactors Python testing to construct `pylibcudf.Table` inputs directly and make table comparisons using pylibcudf.

Additionally, numpy, cupy, and cudf were runtime dependencies while I do not see them explicitly needed, so removed them and marking this PR as `breaking`


dependencies.yaml
```diff
@@ -422,16 +418,11 @@ dependencies:
       - output_types: [conda, pyproject, requirements]
         packages:
           - cuda-core>=0.7.0
+          - cupy>=13.6.0
```
bdice: The cupy dependency needs to be more like the section that was deleted, for pip:

```yaml
      - output_types: [requirements, pyproject]
        matrices:
          - matrix:
              cuda: "12.*"
            packages:
              - cupy-cuda12x>=13.6.0
          # fallback to CUDA 13 versions if 'cuda' is '13.*' or not provided
          - matrix:
            packages:
              - cupy-cuda13x>=13.6.0
```

I recommend copying from https://github.com/rapidsai/cudf/blob/3938de6cc84239984dd0e4aaca6ba344de0497b5/dependencies.yaml#L1158-L1176

Also we are likely to switch to `>=14.0` and require the `[ctk]` extra soon: https://github.com/rapidsai/build-planning/issues/279

bdice (changes requested):

python/rapidsmpf/rapidsmpf/examples/ray/ray_shuffle_example.py
```diff
@@ -56,26 +52,31 @@ def __init__(
...
-        return cudf.DataFrame(
-            {
-                "a": range(self._num_rows),
-                "b": np.random.randint(0, 1000, self._num_rows),
-                "c": ["cat", "dog"] * (self._num_rows // 2),
-            }
+        return plc.Table(
+            [
+                plc.Column.from_iterable_of_py(
+                    list(range(self._num_rows)), plc.DataType(plc.TypeId.INT64)
```
wence-: A `range` is an "`iterable_of_py`":


```suggestion
                    range(self._num_rows), plc.DataType(plc.TypeId.INT64)
```

python/rapidsmpf/rapidsmpf/examples/ray/ray_shuffle_example.py
```diff
@@ -145,10 +151,10 @@ def run(self) -> None:
                 br=br,
                 stream=stream,
             )
-            assert_eq(
-                pylibcudf_to_cudf_dataframe(partition, column_names=column_names),
+            assert_eq_with_plc(
+                partition,
                 expected[partition_id],
-                sort_rows="a",
+                sort_rows=0,
```
wence-, line 157: Can we just call this `assert_eq`?

python/rapidsmpf/rapidsmpf/testing.py
```diff
@@ -1,64 +1,50 @@
...
 
-from rapidsmpf.utils.cudf import pylibcudf_to_cudf_dataframe
-
 
-def assert_eq(
-    left: Any,
-    right: Any,
+def assert_eq_with_plc(
+    left: pylibcudf.Table,
+    right: pylibcudf.Table,
```
wence-: There seems kind of no need to change this from `assert_eq` as a name. It's an internal testing detail so it doesn't matter that we changed the API.

TomAugspurger:
Seems to be a hang in the test: `tests/test_shuffler.py::test_shuffler_uniform[ucxx-1-None] ` https://github.com/rapidsai/rapidsmpf/actions/runs/26303302761/job/77434732386?pr=1056#step:12:12196

Do you think that could be related to these changes?

mroeschke:
> Seems to be a hang in the test: `tests/test_shuffler.py::test_shuffler_uniform[ucxx-1-None] ` https://github.com/rapidsai/rapidsmpf/actions/runs/26303302761/job/77434732386?pr=1056#step:12:12196
> 
> Do you think that could be related to these changes?

It ended up being a bad conversion from cudf to pylibcudf logic fixed in de862920143497cab65f4810f27c2ecac7583d5f

wence- (approved):

dependencies.yaml
mroeschke: Thanks, I believe I specified this correctly now. Care to re-review @bdice?

bdice (approved):

PR #1061, opened 2026-05-26
Remove custom `MemoryAvailable` functions

madsbk:
As part of the upcoming work for #641, `RmmResourceAdaptor` will be merged into `BufferResource`, which removes most of the motivation for supporting a generic memory-availability callback.

This PR simplifies the API to match how it is actually used. Today, `BufferResource` accepts an arbitrary `std::function<std::int64_t()>` per `MemoryType`, but every real caller simply wraps `LimitAvailableMemory(adaptor, limit)`, which reduces to a straightforward `limit - current_allocated()` computation.

This PR replaces the callback map with a plain integer limit, defaulting to `std::numeric_limits<std::int64_t>::max()` to represent "unlimited".

To further simplify the API, `BufferResource` now constructs the `RmmResourceAdaptor` internally, so users no longer need to interact with `RmmResourceAdaptor` directly.

### DIFF
--- a/cpp/include/rapidsmpf/memory/buffer_resource.hpp
+++ b/cpp/include/rapidsmpf/memory/buffer_resource.hpp
@@ -6,6 +6,8 @@
 #pragma once
 
 #include <array>
+#include <atomic>
+#include <cstdint>
 #include <memory>
 #include <mutex>
 #include <optional>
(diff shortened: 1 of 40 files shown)


cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -146,17 +150,31 @@ class BufferResource {
...
-        return memory_available_.at(mem_type);
-    }
+    [[nodiscard]] std::int64_t memory_available(MemoryType mem_type) const noexcept;
+
+    /**
+     * @brief Updates the memory limit for a given memory type at runtime.
+     *
+     * Allows callers to change the limit after construction — useful for tests
+     * that drive `memory_available()` synthetically and for workloads that
+     * adjust limits between phases.
```
wence-:
```suggestion
```

cpp/tests/test_shuffler.cpp
```diff
@@ -404,24 +403,24 @@ TEST(Shuffler, SpillOnInsertAndExtraction) {
...
 
-    // Create a buffer resource with an available device memory we can control
-    // through the variable `device_memory_available`.
-    std::int64_t device_memory_available{0};
+    // Drive available device memory by adjusting the DEVICE limit at runtime.
+    // BufferResource also wraps `mr` in its own internal RmmResourceAdaptor, so
+    // real allocations affect `memory_available(DEVICE) = limit - allocated`.
+    // To force spilling we use a hugely negative limit; to disable spilling we
+    // use a hugely positive limit. These swamp any contribution from actual
+    // device allocations.
```
wence-: This comment kind of makes no sense? All of the allocations will be observed on the resourceadaptor we pass in, so I don't know why the previous values did not work.

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
```diff
@@ -337,12 +309,29 @@ cdef class BufferResource:
...
+            ret = deref(self._handle).memory_available(mem_type)
         return ret
 
+    def set_memory_limit(self, MemoryType mem_type, int64_t limit):
+        """
+        Set the byte limit for the specified memory type.
+
+        Lets callers drive ``memory_available`` synthetically (useful for tests)
+        and adjust limits between workload phases.
+
```
wence-:
```suggestion
```

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -146,17 +150,31 @@ class BufferResource {
...
+     *
+     * @param mem_type The memory type whose limit is being updated.
+     * @param limit The new byte limit. Negative values are permitted; they make
+     * `memory_available(mem_type)` always negative and so trigger continuous
+     * spilling.
+     */
+    void set_memory_limit(MemoryType mem_type, std::int64_t limit) noexcept;
```
wence-, line 171: nit: Add TOCTOU note to this docstring?

wence- (changes requested):
Small suggestions

cpp/include/rapidsmpf/memory/buffer_resource.hpp
madsbk: Updated to
```c++
    /**
     * @brief Updates the memory limit for a given memory type at runtime.
     *
     * The store is atomic, but readers (e.g. `memory_available()` and `reserve()`)
     * observe the limit and the allocation count independently. A concurrent
     * `set_memory_limit()` call can change the limit between a caller's read of
     * `memory_available()` and a subsequent allocation decision; callers that need
     * a coherent view must serialize updates with higher-level synchronization.
     *
     * @param mem_type The memory type whose limit is being updated.
     * @param limit The new byte limit. Negative values are permitted; they make
     * `memory_available(mem_type)` always negative and so trigger continuous
     * spilling.
     */
```

cpp/tests/test_shuffler.cpp
madsbk: Updated the comment. Previously, `memory_available` came from a synthetic mock that we controlled directly, so `set(1000)` literally meant "1000 bytes available".

Now, the value is computed as `limit - mr.current_allocated()`, and the test has roughly 16 KB of real allocations live. That means `limit = 1000` results in `available = -15000`, so the "disable spilling" branch still triggers spilling.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -408,10 +422,14 @@ class BufferResource {
 
   private:
     std::mutex mutex_;
+    // The internal RmmResourceAdaptor wraps the user's device MR so that
+    // allocations are tracked for the DEVICE memory_available calculation.
+    // Declared before device_mr_ because device_mr_ is initialized from it.
+    RmmResourceAdaptor device_adaptor_;
     cuda::mr::any_resource<cuda::mr::device_accessible> device_mr_;
```
nirandaperera, line 427: We dont really need two objects here. `RmmResourceAdaptor` keeps upstream mr alive. If we need to access the upstream resource, we can call, 
https://github.com/rapidsai/rapidsmpf/blob/main/cpp/include/rapidsmpf/rmm_resource_adaptor.hpp#L70

cpp/src/memory/buffer_resource.cpp
```diff
@@ -21,64 +21,99 @@
...
+        && pinned_mr != PinnedMemoryResource::Disabled)
+    {
+        if (auto const max_pool = pinned_mr->properties().max_pool_size;
+            max_pool.has_value() && *max_pool > 0)
+        {
+            return safe_cast<std::int64_t>(*max_pool);
+        }
     }
-    return memory_available;
+    return std::numeric_limits<std::int64_t>::max();
```
nirandaperera: If `pinned_mr == Disabled()`, this will still return INT64_MAX isnt it? This could be problematic?

nirandaperera (changes requested):

cpp/include/rapidsmpf/memory/buffer_resource.hpp
madsbk: I will remove `RmmResourceAdaptor` in a follow-up PR, so I want my own ref.

cpp/src/memory/buffer_resource.cpp
madsbk: Good catch, fixed

madsbk: actually, let's allow `std::numeric_limits<std::int64_t>::max()` and have `memory_available()` handle the special case.

wence- (approved):

nirandaperera (approved):

PR #1062, opened 2026-05-27
All options in one place, take two

madsbk:
Based on #1028, but using a single plain `std::unordered_map` instead of the namespaced descriptor structs.

We lose some compile-time type checking, but the code becomes much simpler, and any option typos should be caught by the test suite anyway.


cpp/include/rapidsmpf/config.hpp
```diff
@@ -380,4 +384,34 @@ std::unordered_map<std::string, std::string> get_environment_variables(
...
+    {"pinned_max_pool_size", "80%"},
+    {"spill_device_limit", "80%"},
+    {"periodic_spill_check", "1ms"},
+    {"num_streams", "16"},
+    {"num_streaming_threads", "1"},
+    {"memory_reserve_timeout", "100 ms"},
+    {"allow_overbooking_by_default", "true"},
+    {"log", "WARN"},
+    {"ucxx_progress_mode", "thread-blocking"},
+};
```
madsbk: I think we could make this `constexpr` with some hackery, but I really do not think it is worth it.

cpp/src/memory/pinned_memory_resource.cpp
```diff
@@ -73,26 +74,26 @@ std::optional<PinnedMemoryResource> PinnedMemoryResource::make_if_available(
 std::optional<PinnedMemoryResource> PinnedMemoryResource::from_options(
     config::Options options
 ) {
-    bool const pinned_memory = options.get<bool>("pinned_memory", [](auto const& s) {
-        return s.empty() ? EnabledByDefault : parse_string<bool>(s);
+    bool const enabled = options.get<bool>("pinned_memory", [](auto const& s) {
+        return parse_string<bool>(s.empty() ? config::DEFAULTS.at("pinned_memory") : s);
     });
```
wence-: This pattern repeats everywhere and is error prone: we have to spell the name of the field in two places, and always remember to look up the default.

Let's change the contract to be:

```
/**
 * @brief Return an option value.
 *
 * @tparam T Type to convert the string value to.
 * @param name Name of option to look up
 * @param fn function mapping the string value to the option value of type `T`
 * @return The parsed option value
 * @note If the option value is empty, its default value is looked up and used.
 */
 template<typename T, typename F>
    requires (whatever_constraint for F)
 T get(std::string const& name, F fn) {
     return fn(s.empty() ? config::DEFAULTS.at(name) : s);
 }
```

Note that we can only do this if we are changing `Options` to just be a static map of options that rapidsmpf already knows about, and not an extensible map with utility methods for storing/querying things. Which is it?

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
```diff
@@ -651,9 +653,14 @@ def stream_pool_from_options(Options options not None):
...
-    cdef int pool_size = options.get_or_default("num_streams", default_value=16)
+    cdef int pool_size = options.get_or_default(
+        "num_streams",
+        default_value=int(DEFAULTS["num_streams"]),
+    )
     if pool_size < 1:
-        raise ValueError("the `num_streams` options must be greater than 0")
+        raise ValueError(
+            "the `num_streams` options must be greater than 0"
+        )
```
wence-: If we can only look up defaults in the canonical `DEFAULTS` map then `get_or_default` should not provide the ability to specify a default value.

cpp/include/rapidsmpf/config.hpp
```diff
@@ -380,4 +384,34 @@ std::unordered_map<std::string, std::string> get_environment_variables(
...
+ * `from_options` factories.
+ *
+ * The map is the single source of truth. Options are always parsed from
+ * their string representation at runtime, so defaults are expressed as
+ * strings and fed through the same factory the call site uses for
+ * user-supplied values. The Cython bridge re-exports this map to Python
+ * at import time so the two languages cannot drift.
+ *
+ * To add a new option, add an entry here and reference it at the call site
+ * with `config::DEFAULTS.at("<key>")`.
```
wence-: Is the idea that we now support a static set of options in the `Options` object, or rather that we encode options that rapidsmpf knows about natively here?

Aside: please refactor this docstring to make much more sense.

cpp/tests/test_config.cpp
```diff
@@ -77,6 +77,24 @@ TEST(OptionsTest, EnvThrowsIfNoCaptureGroup) {
...
+    EXPECT_EQ(DEFAULTS.at("spill_device_limit"), "80%");
+    EXPECT_EQ(DEFAULTS.at("periodic_spill_check"), "1ms");
+    EXPECT_EQ(DEFAULTS.at("num_streams"), "16");
+    EXPECT_EQ(DEFAULTS.at("num_streaming_threads"), "1");
+    EXPECT_EQ(DEFAULTS.at("memory_reserve_timeout"), "100 ms");
+    EXPECT_EQ(DEFAULTS.at("allow_overbooking_by_default"), "true");
+    EXPECT_EQ(DEFAULTS.at("log"), "WARN");
+    EXPECT_EQ(DEFAULTS.at("ucxx_progress_mode"), "thread-blocking");
+}
+
```
wence-: I don't like this test, because I don't think it really tests anything.

python/rapidsmpf/rapidsmpf/tests/test_config.py
```diff
@@ -596,3 +597,27 @@ def test_context_from_options_can_create_channel() -> None:
...
+    assert stream_pool_from_options(opts).get_pool_size() == 8
+
+
+def test_defaults_sourced_from_cpp() -> None:
+    """The ``DEFAULTS`` mapping is populated from C++ at import time. Spot-
+    check that entries match the canonical C++ defaults."""
+    assert DEFAULTS["statistics"] == "false"
+    assert DEFAULTS["num_streams"] == "16"
+    assert DEFAULTS["log"] == "WARN"
+    assert DEFAULTS["ucxx_progress_mode"] == "thread-blocking"
```
wence-: Again, I don't really like this test. But if we're going to do it, why would you not test all the values?

python/rapidsmpf/rapidsmpf/config_defaults.pyx
```diff
@@ -0,0 +1,39 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-License-Identifier: Apache-2.0
+"""
+String-form default value for every option recognised by ``from_options``
+factories. Sourced from the C++ ``rapidsmpf::config::DEFAULTS`` map at module
+import, so Python and C++ cannot drift.
```
wence-:
```suggestion
factories.
```

python/rapidsmpf/rapidsmpf/config_defaults.pyx
```diff
@@ -0,0 +1,39 @@
...
+
+Example:
+    >>> from rapidsmpf.config_defaults import DEFAULTS
+    >>> DEFAULTS["num_streams"]
+    '16'
+    >>> DEFAULTS["statistics"]
+    'false'
```
wence-: This is, I think, not correctly formatted for an example block.

python/rapidsmpf/rapidsmpf/config_defaults.pyx
```diff
@@ -0,0 +1,39 @@
...
+
+
+cdef _build_defaults():
+    cdef unordered_map[string, string] m = _CPP_DEFAULTS
+    return MappingProxyType({
+        k.decode("utf-8"): v.decode("utf-8") for k, v in m
+    })
+
+
+DEFAULTS: Final[Mapping[str, str]] = _build_defaults()
```
wence-: Can one just inline this as:

```
DEFAULTS: Final[...] = MappingProxyType({k: v for k, v in _CPP_DEFAULTS})
```
?

wence- (changes requested):

python/rapidsmpf/rapidsmpf/tests/test_config.py
madsbk: Removed

cpp/tests/test_config.cpp
madsbk: Removed

python/rapidsmpf/rapidsmpf/config_defaults.pyx
madsbk: now:
```python
"""
String-form default values for config options.

Defaults are stored as strings and parsed through the same factories used
for user-supplied values.

Examples
--------
>>> from rapidsmpf.config_defaults import DEFAULTS
>>> DEFAULTS["num_streams"]
'16'
>>> DEFAULTS["statistics"]
'false'
"""
```

python/rapidsmpf/rapidsmpf/config_defaults.pyx
madsbk: I cannot get this to work because Cython does not track `const` on C++ containers. Iterating `_CPP_DEFAULTS` directly therefore generates an `iterator = const_iterator` assignment that fails to compile.

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
madsbk: We still have `get_or_default`, but it now raises if the option already has a default.

In the case above, we now just use `options.get()`.

cpp/src/memory/pinned_memory_resource.cpp
madsbk: `options.get()` now look up the option defaults.

cpp/include/rapidsmpf/config.hpp
madsbk: We now have:
```c++
/**
 * @brief String-form default values for config options.
 *
 * Defaults are stored as strings and parsed through the same factories used
 * for user-supplied values. `Options::get<T>(key, factory)` consults this
 * map automatically: when the user has not supplied a value, the factory
 * receives the registered default string for `key`.
 *
 * Options are not required to have an entry in this map. If no default is
 * registered for a key, the factory receives the empty string.
 *
 * To add a new option, add an entry here and reference it at the call site
 * via `Options::get<T>("<key>", factory)`.
 */
```

cpp/include/rapidsmpf/memory/pinned_memory_resource.hpp
```diff
@@ -140,13 +118,12 @@ class PinnedMemoryResource final
...
-     * - `pinned_initial_pool_size` (nbytes string): initial pool size; defaults to
-     *   `get_host_memory_per_gpu() * DefaultInitiPoolSizeFactor`.
-     * - `pinned_max_pool_size` (nbytes string or empty): maximum pool size; defaults to
-     *   `get_host_memory_per_gpu() * DefaultMaxPoolSizeFactor`.
+     * Recognized options (see `rapidsmpf::pinned_memory`):
+     * - `pinned_memory` (bool): enables pinned memory.
+     * - `pinned_initial_pool_size` (nbytes string): initial pool size, applied
+     *   as `get_host_memory_per_gpu() * InitialPoolSize`.
+     * - `pinned_max_pool_size` (nbytes string or empty): maximum pool size,
+     *   applied as `get_host_memory_per_gpu() * MaxPoolSize`.
```
wence-: This docstring is now wrong? Or you changed the semantics. I think the "applied, ..." is really "default", no?

Also, the docstring was previously misleading too. I think this should say something like:

> pinned_initial_pool_size (bytes or percentage): initial pool size. If bytes then applied literally, if a percentage then specified as a percentage of `get_host_memory_per_gpu()`.

WDYT?

cpp/include/rapidsmpf/config_defaults.hpp
```diff
@@ -0,0 +1,42 @@
...
+inline const std::unordered_map<std::string, std::string> DEFAULTS{
+    {"statistics", "false"},
+    {"pinned_memory", "false"},
+    {"pinned_initial_pool_size", "0%"},
+    {"pinned_max_pool_size", "80%"},
+    {"spill_device_limit", "80%"},
+    {"periodic_spill_check", "1ms"},
+    {"num_streams", "16"},
+    {"num_streaming_threads", "1"},
+    {"memory_reserve_timeout", "100 ms"},
```
wence-:
```suggestion
    {"memory_reserve_timeout", "100ms"},
```
?

python/rapidsmpf/rapidsmpf/streaming/core/memory_reserve_or_wait.pyx
```diff
@@ -12,6 +12,9 @@ from libcpp.utility cimport move
 from rapidsmpf._detail.exception_handling cimport ex_handler
 from rapidsmpf.config cimport Options, cpp_Options
 from rapidsmpf.memory.buffer cimport MemoryType
+
+from rapidsmpf.utils.string import parse_boolean as _parse_boolean
```
wence-: nit: why rename?

python/rapidsmpf/rapidsmpf/tests/test_config.py
```diff
@@ -596,3 +602,18 @@ def test_context_from_options_can_create_channel() -> None:
         assert ctx is not None
         channel: Channel[Any] = ctx.create_channel()
         assert channel is not None
+
+
+def test_option_keys_round_trip_through_options() -> None:
+    """An Options built with the canonical option keys is read by
+    ``from_options`` as expected (smoke test that the keys are wired the
+    same in C++ and Python).
```
wence-:
```suggestion
    ``from_options`` as expected.
```

wence- (approved):
Tiny docstring suggestions

cpp/include/rapidsmpf/config_defaults.hpp
```diff
@@ -0,0 +1,42 @@
...
+    {"allow_overbooking_by_default", "true"},
+    {"log", "WARN"},
+    {"ucxx_progress_mode", "thread-blocking"},
+};
```
nirandaperera, line 40: I really want the keys to be pulled out as a global param we can refer in the code.  Motivation for #1028  was that I always forget what these key strings were, and I want to refer to a value in a header instead.

nirandaperera (changes requested):
We need to pull the keys out, please! 😇

cpp/include/rapidsmpf/config_defaults.hpp
wence-: How does having:

```
constexpr std::string STATISTICS{"statistics"};
```
help you remember?

nirandaperera: Hehe. Not that per-se, but params would fail hard during compilation and less error prone (ex: typo, case-mismatch, underscore vs dash etc). Also work with intelli-sense

madsbk: I really don’t think it is worth the added complexity, boilerplate, and duplication across both C++ and Python.

Your concern about where to look up options was valid. With this approach, we now have one file, `config_defaults.hpp`, that defines all defaults and serves as the single place to extend them. To add new defaults, you only have to touch this one file; everything else, including the Python bindings, follows from it.

python/rapidsmpf/rapidsmpf/streaming/core/memory_reserve_or_wait.pyx
madsbk: removed

cpp/include/rapidsmpf/memory/pinned_memory_resource.hpp
madsbk: Good catch, you are right on both counts. Updated:
```c++
  /**
   * @brief Construct from configuration options.
   *
   * Recognized options:
   * - "pinned_memory": enable pinned memory.
   * - "pinned_initial_pool_size" (bytes or percentage): initial pool size.
   *   - Byte values (e.g. "1 MiB") are applied literally.
   *   - Percentages (e.g. "10%") are relative to `get_host_memory_per_gpu()`.
   * - "pinned_max_pool_size" (bytes, percentage, or disabled): maximum pool size.
   *   - Byte and percentages uses the same parsing rules as "pinned_initial_pool_size".
   *   - A disabled value (e.g. "off") leaves the pool unbounded.
   *
   * @return A `PinnedMemoryResource` if pinned memory is enabled and supported;
   * otherwise `std::nullopt`.
   */
  static std::optional<PinnedMemoryResource> from_options(config::Options options);
```

nirandaperera (approved):

PR #1066, opened 2026-05-28
Fix rrun tests with fresh Python subprocesses

pentschev:
Switch cuCascade back to `main` and run rrun test bodies in fresh Python interpreters to avoid inherited CUDA/NVML state and `multiprocessing.spawn` module-import failures in CI.


madsbk (approved):

pentschev:
Thanks Mads!

PR #1067, opened 2026-05-28
Merge `RmmResourceAdaptor` into `BufferResource` as a CCCL resource

### LINKED ISSUE #641: Make `BufferResource` usable as an rmm device_memory_resource

madsbk:
Closes #641

Merge `RmmResourceAdaptor` into `BufferResource` so `BufferResource` itself becomes a CCCL `cuda::mr::resource_with<device_accessible>`.

The Python `BufferResource` now subclasses `rmm.pylibrmm.DeviceMemoryResource`. This removes a lifetime foot-gun: previously, callers had to pass `br.device_mr()` (the inner adaptor) to RMM APIs such as `rmm.DeviceBuffer(mr=...)`. That kept the adaptor alive, but not the parent `BufferResource`, so its stream pool could in principle be destroyed while async work on those streams was still in flight.

Now `rmm.DeviceBuffer(mr=br)` is the natural call. It keeps a strong reference to the entire `BufferResource`, ensuring the stream pool stays alive as long as any buffer allocated from it exists.

This will break `cudf-polars` CI, so we should have a PR ready there before merging.


cpp/include/rapidsmpf/memory/detail/buffer_resource_impl.hpp
```diff
@@ -0,0 +1,416 @@
...
+            "calling end_scoped_memory_record() on an empty stack.",
+            std::out_of_range
+        );
+        auto ret = stack.top();
+        stack.pop();
+        if (!stack.empty()) {
+            stack.top().add_subscope(ret);
+        }
+        return ret;
+    }
```
nirandaperera, line 244: I think these can be directly inherited from `RmmResourceAdaptorImpl`. 

Why not extend from it? `BufferResourceImpl: RmmResourceAdapterImpl {...}`

madsbk: I think `RmmResourceAdaptorImpl` should be removed in a later PR.

madsbk:
Closed in favor of #1069.

PR #1069, opened 2026-05-29
Make `BufferResource` shared_ptr-managed, `device_mr()` now keeps BR alive

madsbk:
Fixes the C++ side of #641 by making `BufferResource` shared-pointer-managed.

Lifetime safety is provided by `OwningResourceAdaptor`, a CCCL-compatible adaptor that wraps the device memory resource and carries a `std::weak_ptr<BufferResource>` to its parent.

The adaptor stored inside `BufferResource` itself holds only the weak reference, so there is no reference cycle. When CCCL deep-copies the adaptor, which happens when downstream code promotes the ref into an owning `cuda::mr::any_resource<...>` (for example `rmm::device_buffer::_mr`), the adaptor copy constructor promotes the weak reference to a `std::shared_ptr<BufferResource>`.

As a result, the copy stored inside the owning `any_resource` keeps the entire `BufferResource` alive, including its stream pool, spill manager, and statistics, for as long as the downstream object exists.

If the `BufferResource` has already been destroyed at copy time, the adaptor copy constructor throws `std::bad_weak_ptr` instead of creating a dangling reference.

The caller must still ensure that `BufferResource` outlives any direct use of a bare `device_async_resource_ref`. For example:

```c++
auto mr = br->device_mr();
mr.allocate(...);  // UB if br is already destroyed
```

This matches the existing RMM contract for non-owning resource refs. The new safety guarantees only apply once the ref has been captured into an owning container.

### Context

The original sketch in #641 proposed making `BufferResource` itself a CCCL device memory resource, but that conflates roles. `device_mr()` is the actual device memory resource, while `host_mr()` and `pinned_mr()` expose separate host and pinned resources. What matters is that all three keep the parent `BufferResource` alive.

---

This PR only changes `BufferResource::device_mr()`. Extending the same ownership semantics to `host_mr()` and `pinned_mr()` is tracked separately in #1070.

Python-side lifetime gaps are tracked separately in #1074. Python-created `BufferResource`s still borrow their stream pool, device MR, pinned MR, and statistics from Python-owned objects, so the new C++ ownership model does not yet fully protect those cases.


cpp/src/memory/buffer_resource.cpp
```diff
@@ -94,7 +114,13 @@ void BufferResource::set_memory_limit(MemoryType mem_type, std::int64_t limit) n
...
+cuda::mr::any_resource<cuda::mr::device_accessible> BufferResource::device_mr() {
+    return cuda::mr::any_resource<cuda::mr::device_accessible>{
+        OwningResourceAdaptor{device_adaptor_, shared_from_this()}
+    };
+}
+
+rmm::device_async_resource_ref BufferResource::device_mr_ref() const noexcept {
     return rmm::device_async_resource_ref{
         const_cast<cuda::mr::any_resource<cuda::mr::device_accessible>&>(device_mr_)
     };
```
wence-: I think this change is not what you want. A user of `device_mr_ref()` will produce a `resource_ref` that then gets (internally in the `rmm::device_buffer`s) turned back into an (owning) `any_resource` that correctly keeps `device_mr_` alive. But doesn't have a connection to `BufferResource` (so won't keep the stream, etc...) alive.

I think what you want is that all usages of our `device_mr` to keep `BufferResource` alive. So you probably want something like:
```
rmm::device_async_resource_ref BufferrResource::device_mr() const noexcept {
    return cuda::mr::any_resource<cuda::mr::device_accessible>{OwningResourceAdaptor{device_adaptor_, shared_from_this()}};
}
```

madsbk: That's why I added `device_mr_ref()`: to make it explicit that it is a non-owning reference, together with the corresponding lifetime contract that the `BufferResource` must remain alive for the duration of its use.

We could remove it and force all callers through the owning `device_mr()`, but `cuda::mr::any_resource` (the owning return type) does not bind as a temporary to APIs taking `rmm::device_async_resource_ref`. 

So instead of:
```c++
cudf::func(stream, br->device_mr_ref());  // works today
```

every cudf call site would need to materialize an owning handle first:
```c++
auto mr = br->device_mr();
cudf::func(stream, mr);
```

madsbk: updated to docstring:
```c++
    /**
     * @brief Get a non-owning reference to the device memory resource.
     *
     * Returns an `rmm::device_async_resource_ref` referring to the device memory
     * resource managed by this `BufferResource`. The returned reference does not
     * extend the lifetime of the `BufferResource`.
     *
     * This accessor is intended for transient use, for example when passing a
     * memory resource into a cuDF or RMM API call. It must not be used in storage
     * paths where the memory resource may outlive the owning `BufferResource`.
     * For such cases, use `device_mr()` instead.
     *
     * @return Non-owning reference to the device memory resource.
     */
    [[nodiscard]] rmm::device_async_resource_ref device_mr_ref() const noexcept;
```

cpp/src/integrations/cudf/partition.cpp
```diff
@@ -123,15 +123,15 @@ std::unordered_map<shuffler::PartID, PackedData> split_and_pack(
...
     std::unordered_map<shuffler::PartID, PackedData> ret;
 
     // contiguous split does a deep-copy. Therefore, we need to reserve memory for
     // at least the size of the table.
     auto reservation = br->reserve_device_memory_and_spill(
-        cudf::packed_size(table, stream, br->device_mr()), allow_overbooking
+        cudf::packed_size(table, stream, br->device_mr_ref()), allow_overbooking
     );
-    auto packed = cudf::contiguous_split(table, splits, stream, br->device_mr());
+    auto packed = cudf::contiguous_split(table, splits, stream, br->device_mr_ref());
```
pentschev: I think Lawrence's concern from https://github.com/rapidsai/rapidsmpf/pull/1069/files#r3323828393 is justified. Take this case for example, the allocation of `device_mr_ref` is used as output for `packed`, which is later stored in `PackedData.gpu_data`, therefore `PackedData.gpu_data` doesn't carry a reference to `OwningResourceAdaptor`. The same occurs down in `unpack_and_concat()` in this file, the returned value is created with a non-owning `device_mr_ref`.

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
pentschev, line 1: In `buffer_resource.pyx`, C++ receives a `std::shared_ptr<rmm::cuda_stream_pool>` with a no-op deleter pointing at Python-owned `CudaStreamPool.c_obj` ([lines 31-38](https://github.com/rapidsai/rapidsmpf/blob/58ee92aecd0fa90a352fc8936cb1a97f18568a56/python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx#L31-L38), [174-181](https://github.com/rapidsai/rapidsmpf/blob/58ee92aecd0fa90a352fc8936cb1a97f18568a56/python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx#L174-L181)). The new `device_mr()` back-reference keeps only the C++ `BufferResource` alive, not the Python wrapper that owns `_stream_pool`. So for Python-created `BufferResource`s, the claim that the MR keeps "the entire resource, including the stream pool, alive" is false. This is especially relevant because #641 is a Python lifetime issue. I think the fix needs real ownership of the stream pool on the C++ side, or a `shared_ptr` deleter/control block that keeps the Python `CudaStreamPool` alive safely.

nirandaperera:
@madsbk Let's back up a bit. Prior to https://github.com/rapidsai/rapidsmpf/pull/1061, we could have passed a plain device mr or a `RmmResourceAdapter` wrapped device mr to `BufferResource`. If former was passed, reservations were unlimited, and latter was explicitly passed in `from_options` method. So, there was a notion that, _if I want to limit reservations, I should have a RmmResourceAdapter in my context_.  Callers would have _likely_ used that mr for external operations that needed a memory resource (rmm, cudf etc)
Eg:
```c++
auto cuda_mr ... 
RmmResourceAdapter mr(cuda_mr);
auto br = BufferResource::from_options(mr, ...); 
auto stream_ref = br->stream_pool().get_stream();
...
rmm::device_buffer some_buf(..., mr, stream_ref); // issue here, stream/br could go out of scope without some_buf knowing 
```
But now, RmmResourceAdapter is internal to br. So, then in the call sites, the callers should explicitly take the adapter out from the `br` after construction. 
```c++
auto cuda_mr ... 
auto br = BufferResource::from_options(cuda_mr, ...); 
auto mr = br->device_mr(); // use this everywhere from here on!!!! 
... 
```
It is very easy to miss this last line, and continue using `cuda_mr` everywhere else. So, we are loosing track of allocations outside of rapidsmpf. 


So, my suggestion is, before fixing the lifetime guarantees, we need to accept a `RmmResourceAdapter` to br, instead of an `any_device_resource`.

nirandaperera:
@madsbk Now, regarding ownership guarantees, we anyway need to store a back-reference of buffer resource to the provided device_mr. I agree with managing buffer resource from a `shared_ptr`. What if we attach an `OwningResourceAdaptor` after constructing the shared_ptr. 
eg: 
```c++ 
class BufferResource{
private:
 OwningResourceAdapter<RmmResourceAdapter, shared_ptr<BufferResource>> mr_owner_;

 BufferResource(RmmResourceAdapter mr, ...): mr_owner_{move(mr), nullptr},...{}

public:
shared_ptr create(RmmResourceAdapter mr, ...){
 auto ret = make_shared(move(mr), ...); // ret doesnt have br tied yet! 
 ret->mr_owner_->set_backref(ret); // hacky! but now, we can safely return ref/ any_resource without any compromise. 
 return ret;
}

any_device_resource device_mr() { return mr_owner_;}
device_async_resource_ref device_mr_ref() {return mr_owner_;}
}
```

nirandaperera:
Or else, we can use a `shared_ptr<void>` with a custom deleter to reset the buffer resource shared_ptr. 
```c++
class RmmResourceAdapter{
private:
 shared_ptr<void> owning{}; 

public: 
// might need templates 
void set_owned(shared_ptr<void> needs_owned){
 owning = move(needs_owned);
}
};

class BufferResource{
private:
 RmmResourceAdapter mr_owner_;

 BufferResource(RmmResourceAdapter mr, ...): mr_owner_{move(mr), nullptr},...{}

public:
shared_ptr create(RmmResourceAdapter mr, ...){
 auto ret = make_shared(move(mr), ...); // ret doesnt have br tied yet! 
 ret->mr_owner_->set_owned(std::shared_ptr<void>(nullptr, [ret](void*) {
    }));
 return ret;
}

any_device_resource device_mr() { return mr_owner_;}
device_async_resource_ref device_mr_ref() {return mr_owner_;}
}
```

cpp/src/memory/buffer_resource.cpp
nirandaperera: @madsbk I feel returning different semantics from `device_mr` and `device_mr_ref` could be confusing in the long run. How about taking a slightly different approach?
https://github.com/rapidsai/rapidsmpf/pull/1069#issuecomment-4578023414
https://github.com/rapidsai/rapidsmpf/pull/1069#issuecomment-4578358919

cpp/src/memory/buffer_resource.cpp
```diff
@@ -58,14 +57,23 @@ std::shared_ptr<BufferResource> BufferResource::create(
...
         periodic_spill_check,
         std::move(stream_pool),
         std::move(statistics)
     }};
+    // Install the owning adaptor *after* construction so that
+    // `weak_from_this()` is valid. The adaptor holds only a `weak_ptr` to
+    // `*br` so no refcount cycle. Copies of the adaptor (made by CCCL's
+    // `any_resource(resource_ref)` deep-copy in downstream consumers) promote
+    // that weak ref to a `shared_ptr<BufferResource>`, transferring ownership
+    // to the copy.
```
wence-: transferring is the incorrect verb.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -476,9 +498,14 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
...
     // allocations are tracked for the DEVICE memory_available calculation.
-    // Declared before device_mr_ because device_mr_ is initialized from it.
     RmmResourceAdaptor device_adaptor_;
-    cuda::mr::any_resource<cuda::mr::device_accessible> device_mr_;
+    // Stable storage for the device MR returned by `device_mr()`. Holds a
+    // `weak_ptr<BufferResource>` to *this; copies promote that to a
+    // `shared_ptr<BufferResource>` on the deep-copy path. Installed by
+    // `create()` after construction so that `weak_from_this()` is valid;
+    // wrapped in `optional` so it can be default-constructed empty
+    // (`RmmResourceAdaptor` itself is not default-constructible).
```
wence-: tidy this docstring a bit to remove some of the implementation details.

cpp/include/rapidsmpf/memory/owning_resource_adaptor.hpp
```diff
@@ -147,7 +245,8 @@ class OwningResourceAdaptor
 
   private:
     Resource resource_;
-    BackRef back_ref_;
+    std::weak_ptr<BackRef> weak_;
+    std::shared_ptr<BackRef> strong_;
```
wence-: Nit, explicitly default initialise here: `weak_{nullptr}`/`strong_{nullptr}`

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
madsbk: Very good point, tracking in https://github.com/rapidsai/rapidsmpf/issues/1074

cpp/include/rapidsmpf/memory/owning_resource_adaptor.hpp
```diff
@@ -0,0 +1,252 @@
...
+     * @brief Compare wrapped resource identity.
+     *
+     * Two adaptors compare equal if their wrapped resources compare equal.
+     *
+     * @param other Adaptor to compare against.
+     * @return `true` if both adaptors refer to the same underlying resource.
+     */
+    [[nodiscard]] bool operator==(OwningResourceAdaptor const& other) const noexcept {
+        return resource_ == other.resource_;
+    }
```
wence-, line 252: question: Should one also compare the shared ptrs? Consider providing the same underlying resource to two different buffer resources. This equality definition would lead to:

```
auto mr = ...;
auto br1 = BufferResource::create(mr);
auto br2 = BufferResource::create(mr);

auto ref1 = br1->device_mr();
auto ref2 = br2->device_mr();
cuda::mr::any_resource<...> mr1 = ref1;
cuda::mr::any_resource<...> mr2 = ref2;
mr1 == mr2; // true, but mr1 keeps br1 alive, and mr2 keeps br2 alive.
```

cpp/src/memory/buffer_resource.cpp
```diff
@@ -49,6 +49,37 @@ BufferResource::BufferResource(
...
 
+std::shared_ptr<BufferResource> BufferResource::create(
+    cuda::mr::any_resource<cuda::mr::device_accessible> device_mr,
+    std::optional<PinnedMemoryResource> pinned_mr,
+    std::unordered_map<MemoryType, std::int64_t> memory_limits,
+    std::optional<Duration> periodic_spill_check,
+    std::shared_ptr<rmm::cuda_stream_pool> stream_pool,
+    std::shared_ptr<Statistics> statistics
+) {
+    // Construct via `new` (private ctor visible here as a member function).
```
wence-:
```suggestion
```

cpp/src/streaming/cudf/channel_metadata.cpp
```diff
@@ -57,7 +57,7 @@ OrderScheme OrderScheme::with_keys(std::vector<OrderKey> new_keys) const {
 }
 
 bool OrderScheme::boundaries_aligned_with(
-    OrderScheme const& other, rapidsmpf::BufferResource const& br
+    OrderScheme const& other, rapidsmpf::BufferResource& br
```
wence-, line 60: OK, this is just because `device_mr()` has changed from `const` to non-const.

cpp/src/memory/buffer_resource.cpp
```diff
@@ -94,10 +124,8 @@ void BufferResource::set_memory_limit(MemoryType mem_type, std::int64_t limit) n
     );
 }
 
-rmm::device_async_resource_ref BufferResource::device_mr() const noexcept {
-    return rmm::device_async_resource_ref{
-        const_cast<cuda::mr::any_resource<cuda::mr::device_accessible>&>(device_mr_)
-    };
+rmm::device_async_resource_ref BufferResource::device_mr() {
+    return rmm::device_async_resource_ref{*owning_mr_};
 }
```
wence-: Should still be `noexcept` I think?

cpp/tests/test_buffer_resource.cpp
```diff
@@ -786,3 +788,76 @@ TEST_F(BufferCopyEdgeCases, SameBufferIsDisallowed) {
...
+    // `rmm::device_buffer` stores the resource as an owning `cuda::mr::any_resource`,
+    // which deep-copies the OwningResourceAdaptor. The copied adaptor acquires a
+    // `shared_ptr<BufferResource>`.
+    auto buf = std::make_unique<rmm::device_buffer>(N, stream, br->device_mr());
+
+    // Drop the original shared_ptr. The buffer's internally stored owning memory
+    // resource now holds the only remaining `shared_ptr<BufferResource>`, so the
+    // BR must remain alive.
+    br.reset();
+    EXPECT_FALSE(weak_br.expired()) << "BR freed while buffer still holds it";
```
wence-, line 808:
```suggestion
    // Drop the original shared_ptr. The buffer's internally stored owning memory
    // resource should be keeping the BR alive.
    br.reset();
    EXPECT_FALSE(weak_br.expired()) << "BR freed while buffer still holds it";
```

cpp/tests/test_buffer_resource.cpp
```diff
@@ -786,3 +788,76 @@ TEST_F(BufferCopyEdgeCases, SameBufferIsDisallowed) {
...
+
+    // Destroy the buffer. This releases the owning memory resource and the final
+    // `shared_ptr<BufferResource>`. The stream-ordered deallocation must complete
+    // successfully while the BR is still alive.
+    EXPECT_NO_THROW(buf.reset());
```
wence-:
```suggestion
    EXPECT_NO_THROW(buf.reset());
```

cpp/tests/test_buffer_resource.cpp
```diff
@@ -786,3 +788,76 @@ TEST_F(BufferCopyEdgeCases, SameBufferIsDisallowed) {
...
+
+    // Copy construction promotes the weak back-reference to a
+    // `shared_ptr<BufferResource>`. Since the BR has already been destroyed,
+    // the promotion must fail with `std::bad_weak_ptr`.
+    EXPECT_THROW({ AdaptorT copy{original}; }, std::bad_weak_ptr);
+
+    // Copy assignment uses the same promotion semantics and must likewise
+    // fail when the back-reference has expired.
+    AdaptorT target{adaptor, weak_br};
+    EXPECT_THROW(target = original, std::bad_weak_ptr);
```
wence-, line 837: This is not a very useful test because it doesn't assert invariants in the way we would use the br. How about something like:

```
auto br = BufferResource::create(...);
std::weak_ptr<BufferResource> weak = br;
auto ref = br->device_mr();
EXPECT_FALSE(weak.expired());
br.reset();
EXPECT_TRUE(weak.expired());
EXPECT_THROW(cuda::mr::any_resource<...> = ref, std::bad_weak_ptr);
```
?

cpp/tests/test_buffer_resource.cpp
```diff
@@ -786,3 +788,76 @@ TEST_F(BufferCopyEdgeCases, SameBufferIsDisallowed) {
...
+TEST(BufferResource, OwningAdaptorMemberHasNoStrongCycle) {
+    std::weak_ptr<BufferResource> weak_br;
+    {
+        auto br = BufferResource::create(cudf::get_current_device_resource_ref());
+        weak_br = br;
+        EXPECT_FALSE(weak_br.expired());
+    }
+    EXPECT_TRUE(weak_br.expired()) << "BR leaked - the owning_mr_ member holds a "
+                                      "strong ref to BR, closing a cycle.";
+}
```
wence-: I think this is already covered by the tests above.

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
wence-: This is all a dance because the `CudaStreamPool` wrapper in RMM stores a `unique_ptr`, which is clearly the wrong choice, it should store a `shared_ptr<...>` and then we can just grab a reference here.

But that would be a strict improvement over the current state, so I think we should not hold this change up for that.

wence- (changes requested):
Small changes I think, but otherwise I think this looks good.

madsbk:
> @madsbk Let's back up a bit. Prior to #1061, we could have passed a plain device mr or a `RmmResourceAdapter` wrapped device mr to `BufferResource`. If former was passed, reservations were unlimited, and latter was explicitly passed in `from_options` method. So, there was a notion that, _if I want to limit reservations, I should have a RmmResourceAdapter in my context_. Callers would have _likely_ used that mr for external operations that needed a memory resource (rmm, cudf etc) 
> ...
> It is very easy to miss this last line, and continue using `cuda_mr` everywhere else. So, we are loosing track of allocations outside of rapidsmpf.
> 
> So, my suggestion is, before fixing the lifetime guarantees, we need to accept a `RmmResourceAdapter` to br, instead of an `any_device_resource`.

That's a valid API design concern, but I think it's orthogonal to the lifetime issue this PR is addressing. I've updated the docs to make the tracking semantics explicit: f2663dd478d860a75df7344321cb1198292a9c70.

I'm also not convinced we want to preserve the old semantic. My intuition is that we *do* want to force users through `device_mr()` so allocations become visible to `BufferResource` accounting, limits, and statistics.

Let's discuss that separately.

cpp/src/memory/buffer_resource.cpp
madsbk: changed to:

```c++
    // The adaptor stored inside `BufferResource` itself holds only a `weak_ptr`, avoiding
    // a reference cycle. When downstream code promotes `device_mr()` from a non-owning
    // `resource_ref` to an owning `cuda::mr::any_resource` (for example internally in
    // `rmm::device_buffer`), CCCL deep-copies the adaptor. The copy promotes the weak
    // ref to a `shared_ptr<BufferResource>`, acquiring shared ownership so the
    // `BufferResource` stays alive for as long as the owning resource exists.
```

cpp/include/rapidsmpf/memory/buffer_resource.hpp
madsbk: trimmed to:
```c++
    // Stable storage for the device MR returned by `device_mr()`. See
    // `OwningResourceAdaptor` for the lifetime semantics. Installed by
    // `create()` after construction.
    std::optional<OwningResourceAdaptor<RmmResourceAdaptor, BufferResource>> owning_mr_;
```

cpp/include/rapidsmpf/memory/owning_resource_adaptor.hpp
madsbk: Good catch, done.

cpp/src/streaming/cudf/channel_metadata.cpp
madsbk: Yes

cpp/src/memory/buffer_resource.cpp
madsbk: Yes, fixed

cpp/tests/test_buffer_resource.cpp
madsbk: removed

cpp/tests/test_buffer_resource.cpp
madsbk: We now have:
```c++
TEST(OwningResourceAdaptor, CopyThrowsWhenBackRefExpired) {
    rmm::mr::cuda_memory_resource cuda_mr;
    RmmResourceAdaptor adaptor{
        cuda::mr::any_resource<cuda::mr::device_accessible>{cuda_mr}
    };

    std::weak_ptr<BufferResource> weak_br;
    {
        auto br = BufferResource::create(cudf::get_current_device_resource_ref());
        weak_br = br;
    }
    ASSERT_TRUE(weak_br.expired());

    OwningResourceAdaptor<RmmResourceAdaptor, BufferResource> original{adaptor, weak_br};

    using AdaptorT = OwningResourceAdaptor<RmmResourceAdaptor, BufferResource>;
    EXPECT_THROW({ AdaptorT copy{original}; }, std::bad_weak_ptr);

    AdaptorT target{adaptor, weak_br};
    EXPECT_THROW(target = original, std::bad_weak_ptr);
}
```

madsbk: Tried the BR-centric form first, but it crashes with a double-free. After `br.reset()`, the adaptor itself has already been destroyed, so CCCL's deep copy of the stale ref reads freed memory before the `std::bad_weak_ptr` check can run.

The throw only works when the source adaptor itself still outlives the copy operation, which is what this test exercises. I've clarified that in the `OwningResourceAdaptor` docstring:
```
 * If the owner has already been destroyed when the adaptor is copied, the
 * copy constructor throws `std::bad_weak_ptr` instead of creating a dangling
 * adaptor.
 *
 * This only works while the source adaptor itself is still alive, which is
 * the common case for transient ref/`any_resource` consumers where CCCL
 * performs the deep copy during construction. If both the owner and source
 * adaptor have already been destroyed, the behavior is undefined.
```

wence-: > so CCCL's deep copy of the stale ref reads freed memory before the std::bad_weak_ptr check can run.

We can control this though right, by changing the order of fields in the `OwningResourceAdaptor` so that the first thing that runs in the copy-ctor is the weak_ptr promotion, no?

madsbk: But if the `OwningResourceAdaptor` itself goes out of scope at `br.reset()`, then reading `other.weak_` in the copy ctor is already UB, regardless of member order, right?

But I think this is fine. The contract is that users must promote the ref into an owning `any_resource` if they want the `BufferResource` lifetime to extend beyond the originating scope.

wence-: Oh, I see what you mean, yes.

wence- (approved):

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -115,11 +136,74 @@ class BufferResource {
...
+     *
+     * @par CCCL's lifetime semantic
+     *
+     * The returned `rmm::device_async_resource_ref` is a non-owning
+     * `cuda::mr::resource_ref`, so callers must take care to avoid use-after-free issues.
+     *
+     * When working directly with the returned reference, the caller must ensure that this
+     * `BufferResource` remains alive for the full duration of that use:
+     * @code
+     * auto br = std::make_shared<BufferResource>(...);
```
pentschev: The constructor is now private, examples should be updated to use `create()`.

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
```diff
@@ -138,6 +138,26 @@ cdef class BufferResource:
     statistics
         The statistics instance to use. If None, a disabled statistics instance
         will be created.
+
+    Notes
+    -----
+    Allocation tracking only applies to allocations routed through this
+    ``BufferResource``. The constructor wraps the supplied ``device_mr`` in
+    an internal RMM adaptor that records all allocations and deallocations;
+    that adaptor is exposed via ``BufferResource.device_mr``.
```
pentschev: This appears to overstate the current Python implementation. `BufferResource.device_mr` still returns `self._device_mr`, which is the original `DeviceMemoryResource` passed to the constructor, not the internal tracked adaptor. See the assignment at lines 212-214 and the property at lines 294-303.

Since Python-side lifetime/resource exposure is deferred to #1074, can we avoid documenting `br.device_mr` as the tracked wrapper here? Either this property should actually expose the tracked adaptor, or this Python doc should keep the weaker wording that the C++ BufferResource wraps the MR internally and that allocations made through the original Python MR are not tracked.

Same issue in line 119: tells Python users to use `BufferResource.device_mr` for tracked allocations, but the property currently returns the original unwrapped Python MR. This should be softened unless this PR also changes the property to expose the tracked adaptor.

pentschev (approved):
Left a couple docstrings improvement suggestions, otherwise LGTM.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
madsbk: fixed

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
madsbk: Very good point. It has been fixed in 19f50ef15eaf78c9c460a6bcb2882a184deff40d.
@pentschev and @wence- can you take a look before we merge?

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
```diff
@@ -138,6 +161,19 @@ cdef class BufferResource:
...
+    Allocation tracking only applies to allocations routed through this
+    ``BufferResource``.
+
+    To ensure allocations are included in memory-limit accounting and
+    statistics, use ``BufferResource.device_mr`` for all CUDA allocations
+    associated with this resource.
+
+    Allocations performed through other memory resources, including the
+    original resource passed to the constructor or allocations made outside
+    ``BufferResource``, are not tracked by this class.
```
pentschev, line 176: Nit: The comment in lines 228-229 (can't comment directly there as the lines are not part of the diff) looks stale after `device_mr` was changed to return an `OwningDeviceMemoryResource` from the C++ `BufferResource::device_mr()`. `_device_mr` is no longer used by the Python `device_mr` property, it is kept to preserve the lifetime of the original Python MR passed into `make_any_device_resource()`.

Maybe update to something like:

```
# Keep the original Python memory resources alive while the C++ BufferResource
# holds resources derived from them.
```

pentschev (approved):
One final nit, otherwise LGTM. Thanks Mads!

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
madsbk: updated to:
```python
        # Keep the original Python memory resources alive while the C++
        # BufferResource holds resources derived from them. These anchors are
        # likely redundant: `make_any_device_resource(...)` deep-copies the
        # device MR into a self-sufficient owning any_resource, and
        # `cpp_PinnedMemoryResource` is copied by value into the C++ BR. Kept
        # defensively for now.
        # TODO: drop these once verified against pool/upstream-adaptor MRs.
        #       https://github.com/rapidsai/rapidsmpf/issues/1074
        self._device_mr = device_mr
        self._pinned_mr = pinned_mr
```

pentschev (approved):

cpp/include/rapidsmpf/memory/owning_resource_adaptor.hpp
```diff
@@ -0,0 +1,260 @@
...
+ * adaptor have already been destroyed, the behavior is undefined.
+ *
+ * This does not make direct use of a bare `resource_ref` safe after the owner
+ * has been destroyed. A call such as `ref.allocate(...)` still dispatches
+ * through the `resource_ref` into the original resource object, so the owner
+ * must outlive all direct uses of the non-owning ref.
+ */
+template <typename Resource, typename BackRef>
+class OwningResourceAdaptor
+    : public cuda::forward_property<OwningResourceAdaptor<Resource, BackRef>, Resource> {
```
nirandaperera, line 83: @madsbk I feel like making this implement `Resource` concept is an overkill. How about converting this to a mixin, `BackRefMixin`? 
```c++
template<typename BackRef>
class BackRefMixin{

  void set_backref(std::weak_ptr<BackRef> backref){...}
// might need the equality, copy constructors,.  overload
  private:
    std::weak_ptr<BackRef> weak_{nullptr};
    std::shared_ptr<BackRef> strong_{nullptr};
}
```
That way, we can mix it in with `RmmResourceAdaptor` or create a simple class, 
`RmmResourceAdapterWithBR: RmmResourceAdaptor, BackRefMixin`

This will remove the Resource boilerplate

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -422,12 +506,24 @@ class BufferResource {
...
+        std::optional<Duration> periodic_spill_check,
+        std::shared_ptr<rmm::cuda_stream_pool> stream_pool,
+        std::shared_ptr<Statistics> statistics
+    );
+
     std::mutex mutex_;
     // The internal RmmResourceAdaptor wraps the user's device MR so that
     // allocations are tracked for the DEVICE memory_available calculation.
-    // Declared before device_mr_ because device_mr_ is initialized from it.
     RmmResourceAdaptor device_adaptor_;
```
nirandaperera: We dont really need to keep this in the `BufferResource`, isnt it?

nirandaperera (approved):
I wanted to float the idea of a mixin here. But dont want to block because @vyasr is waiting for this. 
So, approving.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
madsbk: good catch, removed

cpp/include/rapidsmpf/memory/owning_resource_adaptor.hpp
madsbk: Interesting idea, but I think it becomes cumbersome once we need to handle other resources like the pinned and host MRs. You would likely end up needing a separate `XWithBR` glue class for each adaptor/resource combination.

I also like that the current adaptor is generic enough that we could potentially upstream it to RMM at some point, whereas a mixin-based approach feels a bit more tied to our specific `BufferResource` use case.

nirandaperera: On the other hand, a mixin can be used for any class. We could easily add that to Pinned mr and host mr, and well as stream pool (will need a rapidsmpf stream pool in the end). Let me sketch something up, now that this PR is merged.

nirandaperera: @madsbk This is what I had in mind https://github.com/rapidsai/rapidsmpf/pull/1078 
I feel its more generic.

PR #1073, opened 2026-05-29
Use rapids_cpm_libcoro from rapids-cmake

vyasr:
Depends on https://github.com/rapidsai/rapids-cmake/pull/1035

Replaces the manual libcoro CPM integration in `cmake/thirdparty/get_libcoro.cmake` with the new `rapids_cpm_libcoro()` wrapper from rapids-cmake. This change is important so that we can ensure libcoro version alignment between rapidsmpf and cudf.


pentschev (approved):
LGTM. Thanks Vyas!

PR #1075, opened 2026-06-01
Fix AddressSanitizer build and test runs

madsbk:
Several issues prevented `./build.sh --asan` and the test binaries from running cleanly under AddressSanitizer. This PR fixes those issues so an ASan run of `gtests/single_tests` exits successfully with no LSan findings.

Run with something like:
```bash
LSAN_OPTIONS=suppressions=$PWD/cpp/asan_suppressions.txt \
    ASAN_OPTIONS=protect_shadow_gap=0 \
    cpp/build/gtests/single_tests
```


cpp/src/config.cpp
```diff
@@ -10,7 +10,17 @@
 #include <cstring>
 #include <limits>
 #include <mutex>
+// GCC bug 109442: libstdc++'s <regex> headers raise spurious -Wmaybe-uninitialized
+// warnings, which become errors under -Werror.
+// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=109442
```
pentschev, line 16: It seems this is fixed in GCC 15. Could we also check for the version in the macro, so we can some day eventually remove this?

cpp/tests/streaming/test_lineariser.cpp
```diff
@@ -68,6 +68,10 @@ class StreamingLineariser : public BaseStreamingFixture {
 };
 
 TEST_F(StreamingLineariser, ManyProducers) {
+#ifdef RAPIDSMPF_ASAN
+    GTEST_SKIP() << "Skipped under AddressSanitizer: extremely slow and prone "
+                    "to stack overflow via libcoro ring_buffer mutual resume.";
```
pentschev: Is that a bug in libcoro that should be fixed or just really a test too large for ASAN?

asan.supp
pentschev, line 1: Is this file stale? It tells users to use `cpp/asan_suppressions.txt`. Can this be removed in favor of the new one?

cpp/asan_suppressions.txt
```diff
@@ -0,0 +1,23 @@
...
+#       cpp/build/gtests/single_tests
+#
+# Append `:print_suppressions=0` to LSAN_OPTIONS to silence the
+# "Suppressions used:" footer.
+
+# cuDF process-lifetime singletons. Note: event_for_thread is defined in an
+# anonymous namespace, which gets mangled into the symbolicated name between
+# `cudf::detail::` and `event_for_thread`, so we match the leaf name only
+# since LSan suppression matching is plain substring.
+leak:event_for_thread
```
pentschev, line 18: This seems to also reinforce that `asap.supp` is stale with this being more up-to-date?

cpp/asan_suppressions.txt
pentschev, line 1: Can we update https://github.com/rapidsai/rapidsmpf/blob/main/docs/source/getting-started.md#addresssanitizer-enabled-build or add a new section reflecting this?

asan.supp
madsbk: yes, removed

cpp/asan_suppressions.txt
madsbk: removed

cpp/tests/streaming/test_lineariser.cpp
madsbk: I is not really a bug, more an limitation, Updated the comment:
```c++
    // libcoro's coro::ring_buffer resumes the peer coroutine inline on the C
    // stack from produce/consume. At -O2 the compiler tail-call-optimizes
    // those resumes, so the stack stays flat in release. ASan builds disable
    // those optimizations and inflate frame sizes, so the chain accumulates
    // real stack on every produce/consume round-trip and a workload this
    // large reliably overflows.
    GTEST_SKIP() << "Skipped under AddressSanitizer: libcoro ring_buffer "
                    "mutual-resume chain overflows the stack when frames are "
                    "not tail-call-optimized.";
```

cpp/src/config.cpp
madsbk: done

cpp/asan_suppressions.txt
madsbk: done

pentschev (approved):
Thanks Mads!

madsbk:
thanks @pentschev

Issue #1070, opened 2026-05-29
Make `BufferResource::host_mr()` and `pinned_mr()` return owning MRs


madsbk:
### Summary

PR #1069 changed `BufferResource::device_mr()` to return an owning `cuda::mr::any_resource<device_accessible>` backed by `OwningResourceAdaptor<RmmResourceAdaptor, std::shared_ptr<BufferResource>>`. This ensures that any object storing the returned MR also keeps the owning `BufferResource` alive, including associated runtime state such as the stream pool.

To keep the PR scope manageable, `host_mr()`, `pinned_mr()`, and `try_pinned_mr()` were left unchanged and still return non-owning resources. This issue tracks extending the same ownership semantics to those accessors for consistency and lifetime safety.

`HostMemoryResource` and `PinnedMemoryResource` both depend on the owning `BufferResource`'s stream pool for stream-ordered deallocation. As a result, the same lifetime hazard that motivated the `device_mr()` fix also exists for host and pinned allocations: a buffer allocated through one of these memory resources can outlive its owning `BufferResource` and later access destroyed runtime state during destruction.

In practice, this is currently unlikely to occur:

* `host_mr()` has no external call sites in the codebase.
* `pinned_mr()` has only two call sites.
* `try_pinned_mr()` has six call sites, all in benign `Statistics::report`-style read-only paths.

So this issue is primarily about closing a latent lifetime gap and making the API behavior symmetric across all MR accessors, rather than fixing a known user-visible bug.

### Proposed change

Make `host_mr()`, `pinned_mr()`, and `try_pinned_mr()` return owning memory resource handles that keep the parent `BufferResource` alive for as long as the returned MR exists, matching the semantics of `device_mr()`.

This will likely reuse `OwningResourceAdaptor` in the same way as `device_mr()`.

PR #1078, opened 2026-06-01
Replace `OwningResourceAdaptor` with a BackRef mixin

nirandaperera:
Replaces the `OwningResourceAdaptor<Resource, BackRef>` wrapper with a slim `BackRefMixin<BackRef>` that `RmmResourceAdaptor` inherits directly. `OwningResourceAdaptor` wrapped `RmmResourceAdaptor` only to carry a `weak_ptr`/`shared_ptr<BackRef>` pair and promote weak→strong on copy. Its `allocate`/`deallocate`/property forwarding was redundant since `RmmResourceAdaptor` already satisfies the CCCL resource concept.

## Back-reference contract

`BackRefMixin` enforces a strict lifetime contract: a back-reference **must** be installed via `set_backref()` before an instance is copied. Copying an uninstalled (or installed-but-expired) instance throws `std::bad_weak_ptr`. This makes accidental lifetime mistakes loud instead of silently producing a copy that no longer keeps its owner alive.

To make the contract impossible to violate by mistake, **`RmmResourceAdaptor`'s primary-resource constructor is now private**. Only `BufferResource` (a `friend`) can construct one, and it installs the back-reference before the adaptor becomes observable. External callers obtain an adaptor exclusively via `BufferResource::device_mr_adaptor()`, so every adaptor a caller can reach is guaranteed to be back-referenced and safely copyable.

### C++

- **New** `cpp/include/rapidsmpf/memory/back_ref_mixin.hpp`: templated `BackRefMixin<BackRef>` with default-empty `weak_`/`strong_`, `set_backref()`, copy ctor/assignment that promote weak→strong (throwing `std::bad_weak_ptr` when uninstalled or expired), and owner-based `operator==`. The promote logic is inlined into the copy ctor/assignment (no helper).
- **Deleted** `cpp/include/rapidsmpf/memory/owning_resource_adaptor.hpp`.
- `RmmResourceAdaptor` now publicly inherits `BackRefMixin<BufferResource>`. Its primary-resource constructor is **private**; `BufferResource` is a `friend` and is the sole producer of new instances. `operator==` checks both shared state and back-reference identity.
- `HostMemoryResource` and `PinnedMemoryResource` do **not** inherit the mixin (deferred); they remain freely copyable. Their `operator==` is pool-identity based (pinned) / always-equal (stateless host).
- `BufferResource`'s `owning_mr_` is a plain `RmmResourceAdaptor` member (no `std::optional`). `device_mr` is threaded through the private constructor so `owning_mr_` is initialized in the member-initializer list; `BufferResource::create()` installs the back-ref afterward via `owning_mr_.set_backref(weak_from_this())` (device only).
- **New** `BufferResource::device_mr_adaptor()`: exposes the internal device `RmmResourceAdaptor` directly (e.g. to query `get_main_record()` / `current_allocated()`), so callers no longer construct an `RmmResourceAdaptor` externally.
- `streaming::Context::from_options(...)` first parameter changed from `RmmResourceAdaptor mr` to `any_device_resource mr` (a plain device MR), since external callers can no longer construct an `RmmResourceAdaptor`. The `Context` hands the device MR to its internal `BufferResource`, which wraps it for tracking.
- Call sites (tests + benchmarks) updated to pass a vanilla memory resource to `BufferResource::create()` / `Context::from_options()` and to read stats through `device_mr_adaptor()` instead of creating and wrapping an adaptor themselves.

### Python bindings

- `RmmResourceAdaptor` (Cython) now extends `DeviceMemoryResource` instead of `UpstreamResourceAdaptor`. Its `__init__` **raises `TypeError`** to mirror the C++ private-constructor policy; instances are produced internally via a new `cdef _from_cpp(const cpp_RmmResourceAdaptor&)` factory that copies a back-ref'd C++ adaptor. The `get_upstream` accessor is removed.
- **New** `BufferResource.device_mr_adaptor()` (`cpdef`): returns a back-ref'd `RmmResourceAdaptor` whose copies keep the `BufferResource` alive. This is the only way to obtain an adaptor from Python.
- `streaming.core.context.Context.from_options(...)` now expects `mr: DeviceMemoryResource` instead of `RmmResourceAdaptor` (wrapped internally), matching the C++ change.
- `Statistics.memory_profiling` docs/examples updated to obtain the adaptor via `BufferResource(...).device_mr_adaptor()`.
- Tests/conftests updated: stop pre-wrapping device MRs in `RmmResourceAdaptor`; pass plain `rmm.mr.*` resources to `BufferResource`/`Context.from_options` and read stats via `br.device_mr_adaptor()`.

## Latent side effect

`Statistics::create_memory_recorder` calls `cuda::mr::resource_cast<RmmResourceAdaptor>(&mr)` on the resource passed in. Previously this cast failed for resources obtained from `BufferResource::device_mr()` (the contained type was `OwningResourceAdaptor<…>`), silently returning a no-op `MemoryRecorder`. After this PR the contained type is `RmmResourceAdaptor`, so the cast succeeds and memory profiling becomes active for BR-allocated memory. No tests were relying on the previous no-op behavior.


madsbk:
Thanks @nirandaperera, I think this is a good idea. My main motivation for a generic wrapper like `OwningResourceAdaptor` was that we could potentially upstream it to RMM. I suspect other libraries built on top of RMM will need something similar now that we have the owning `any_resource` type. But a fully generic wrapper does not compose well with `resource_cast`, which is a real issue.

Before we go deeper into this PR, could you also implement, or at least sketch, the same pattern for `host_mr` and `pinned_mr` (#1070)? It would be useful to see the full picture before we settle on the final design.

nirandaperera:
I already added the Pinned MR and Host MR changes in this PR. Is that what you asked @madsbk ? Or did you want to sketch the owning wrapper impl for them?

madsbk:
> Or did you want to sketch the owning wrapper impl for them?

Yes, but on second thought, I think the trade-off is already pretty clear. It really comes down to one thing:

* A templated wrapper like `OwningResourceAdaptor` is nice because it avoids modifying the existing resource types.
* A mixin is nice because it composes naturally with `cuda::mr::resource_cast`.

And I agree that the mixin approach is probably preferable overall.

nirandaperera:
> > Or did you want to sketch the owning wrapper impl for them?
> 
> Yes, but on second thought, I think the trade-off is already pretty clear. It really comes down to one thing:
> 
> * A templated wrapper like `OwningResourceAdaptor` is nice because it avoids modifying the existing resource types.
> * A mixin is nice because it composes naturally with `cuda::mr::resource_cast`.
> 
> And I agree that the mixin approach is probably preferable overall.

I agree. Even I didnt realize this resource_cast bug when I reviewed `OwningResourceAdapter` PR. Claude is nifty in that sense ;-) LOL

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
```diff
@@ -0,0 +1,175 @@
...
+ * The mixin is the building block previously provided by the now-removed
+ * `OwningResourceAdaptor`: it captures the "back-reference acquires shared
+ * ownership when CCCL deep-copies a resource into an owning
+ * `cuda::mr::any_resource`" lifetime trick without the surrounding
+ * allocate/deallocate/`forward_property` wrapper.
+ *
+ * @par Lifetime semantics
+ *
+ * - A default-constructed mixin holds an empty `weak_` and empty `strong_`.
+ *   Copying such an instance is a no-op for the back-ref (no throw, no
```
wence-: Please rewrite the docstrings/comments to avoid talking about "how it used to be".

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
```diff
@@ -0,0 +1,175 @@
...
+    /**
+     * @brief Copy constructor that promotes the back-reference to shared
+     * ownership.
+     *
+     * - If @p other already holds a strong back-reference, it is shared.
+     * - Otherwise, if @p other has an installed weak back-reference, it is
+     *   promoted via `std::shared_ptr<BackRef>{other.weak_}`.
+     * - If @p other has no installed back-reference (default state), this is
+     *   a no-op for the back-ref state.
+     *
```
wence-: Please describe semantics with implementation.

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
```diff
@@ -0,0 +1,175 @@
...
+        }
+        return *this;
+    }
+
+    /// @brief Move constructor.
+    BackRefMixin(BackRefMixin&&) noexcept = default;
+
+    /**
+     * @brief Move assignment operator.
+     * @return Reference to this mixin.
```
wence-, line 71:
```suggestion
     * @return Reference to this.
```

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
```diff
@@ -0,0 +1,175 @@
...
+     *
+     * Two mixins compare equal when their weak references refer to the same
+     * control block under `std::weak_ptr::owner_before` (or when both are
+     * empty). The strong reference does not participate: a mixin that has
+     * promoted its weak ref via copy compares equal to its source.
+     *
+     * This replicates the back-ref half of the equality previously exposed by
+     * `OwningResourceAdaptor::operator==`. Combine with the host type's own
+     * identity check to get full equality.
+     *
```
wence-: remove

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
```diff
@@ -0,0 +1,175 @@
...
+
+    ~BackRefMixin() = default;
+
+    /**
+     * @brief Owner-based equality of the back-reference.
+     *
+     * Two mixins compare equal when their weak references refer to the same
+     * control block under `std::weak_ptr::owner_before` (or when both are
+     * empty). The strong reference does not participate: a mixin that has
+     * promoted its weak ref via copy compares equal to its source.
```
wence-: Please describe semantics not implementation.

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
```diff
@@ -0,0 +1,175 @@
...
+     *
+     * This replicates the back-ref half of the equality previously exposed by
+     * `OwningResourceAdaptor::operator==`. Combine with the host type's own
+     * identity check to get full equality.
```
wence-: What does this cryptic second half of the docstring mean?

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
```diff
@@ -0,0 +1,175 @@
...
+     * Returns an empty `shared_ptr` when @p other has no installed
+     * back-reference (so default-constructed mixins copy without throwing).
+     */
+    static std::shared_ptr<BackRef> promote(BackRefMixin const& other) {
+        if (other.strong_) {
+            return other.strong_;
+        }
+        // Distinguish "never installed" (default-constructed weak_ptr) from
+        // "installed but expired". An installed weak_ptr has a non-empty
+        // owner; default-constructed ones owner-compare equal to each other.
```
wence-: question: Should it be the case that one ever attempts to `promote` a default-constructed backref? I _think_ no?

I would prefer somehow if this state were not permitted at the type level.

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
```diff
@@ -0,0 +1,175 @@
...
+class BufferResource;
+
+/**
+ * @brief Convenience alias: `BackRefMixin` instantiated for `BufferResource`.
+ *
+ * Inherit from this alias to give a type the standard back-reference lifetime
+ * contract `BufferResource::create()` relies on. See `BackRefMixin` for the
+ * full semantics.
+ */
+using WithBufferResourceBackRef = BackRefMixin<BufferResource>;
```
wence-: This seems unnecessary

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -516,10 +535,15 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
...
-    // `create()` after construction.
-    std::optional<OwningResourceAdaptor<RmmResourceAdaptor, BufferResource>> owning_mr_;
+    // Stable storage for the resources returned by `device_mr()`, `host_mr()`,
+    // `pinned_mr()`, and `try_pinned_mr()`. Each carries a
+    // `WithBufferResourceBackRef`; `create()` installs the back-reference on all
+    // of them via `set_backref(weak_from_this())` after construction. Copies
+    // of these resources (e.g. inside a `cuda::mr::any_resource` that CCCL
+    // promotes from one of our `resource_ref` accessors) then promote that
+    // weak ref to a `shared_ptr<BufferResource>` and keep this resource
+    // alive. See `BackRefMixin` for the lifetime semantics.
```
wence-: the previous comment was better because it doesn't replicate implementation detail in a comment.

cpp/include/rapidsmpf/memory/host_memory_resource.hpp
```diff
@@ -28,8 +29,18 @@ namespace rapidsmpf {
...
+ *
+ * The resource also carries an optional back-reference to a `BufferResource`
+ * via `WithBufferResourceBackRef`. When unset (the default, used for standalone
+ * instances in tests/benchmarks/Python bindings), the back-ref machinery is
+ * a no-op. `BufferResource::create()` installs a weak back-reference via
+ * `set_backref()` so that any copy of the resource (for example the one
+ * CCCL makes when promoting `BufferResource::host_mr()` from a non-owning
+ * `cuda::mr::resource_ref` to an owning `cuda::mr::any_resource`) promotes
+ * that weak reference to a `shared_ptr`, keeping the `BufferResource` alive
+ * for as long as the copy lives.
```
wence-: Again, what is this implementation stuff doing in a docstring?

cpp/include/rapidsmpf/memory/host_memory_resource.hpp
```diff
@@ -111,23 +122,23 @@ class HostMemoryResource {
     /**
      * @brief Compares this resource to another resource.
      *
-     * All instances are stateless and interchangeable, so this always returns
-     * true.
+     * `HostMemoryResource` is otherwise stateless, so equality is fully
+     * delegated to the `WithBufferResourceBackRef` base: two instances compare
+     * equal when their installed back-references are owner-equivalent, and
+     * standalone instances (no `set_backref()` ever called) all compare equal
+     * to each other.
```
wence-: This can be shorter I think.

cpp/src/memory/buffer_resource.cpp
```diff
@@ -56,25 +57,30 @@ std::shared_ptr<BufferResource> BufferResource::create(
...
-    // ref to a `shared_ptr<BufferResource>`, acquiring shared ownership so the
+    // Install the back-reference on every internal resource *after*
+    // construction so that `weak_from_this()` is valid. Each resource holds
+    // only a `weak_ptr`, avoiding a reference cycle. When downstream code
+    // promotes a non-owning `resource_ref` returned by `device_mr()`,
+    // `host_mr()`, `pinned_mr()`, or `try_pinned_mr()` to an owning
+    // `cuda::mr::any_resource` (for example internally in `rmm::device_buffer`
+    // or any cudf API that stores the MR by owning value), CCCL deep-copies
+    // the resource. The copy promotes the weak ref to a
+    // `shared_ptr<BufferResource>`, acquiring shared ownership so the
```
wence-: It seems like this is needless churn of the comment.

cpp/tests/test_buffer_resource.cpp
```diff
@@ -815,11 +816,52 @@ TEST(BufferResource, DeviceMrKeepsBufferResourceAlive) {
...
+// BR's owning resource must not affect the other BR's lifetime.
+//
+// Note on equality: `any_resource::operator==` dispatches to the wrapped
+// `RmmResourceAdaptor::operator==`, which compares shared-impl identity
+// first. Each `BufferResource::create()` constructs a brand-new
+// `cuda::mr::shared_resource<Impl>` with its own heap-allocated impl, so the
+// two `any_resource`s here compare unequal even though they share an upstream
+// resource. This matches the previous `OwningResourceAdaptor::operator==`
+// behavior. The substantive invariant exercised below is the per-instance
+// lifetime isolation.
```
wence-: Why?

wence- (changes requested):
The core idea looks good, please apply some sanity to all of the docstrings and comments

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
nirandaperera: I thought about this. I think the issue there would be, then, any class that uses this mix-in needs to call `set_backref(...)` before being copied. Eg: this test case will fail. 
https://github.com/rapidsai/rapidsmpf/pull/1078/changes#diff-2a413e2040efcd78454a48108f8a881e3ea4cb8f0d103d180f58804674fc2431R882-R895

I felt it was too restrictive.

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
nirandaperera: just syntactic sugar. Wanted to avoid the fwd declaration in all memory resource headers.

nirandaperera:
@madsbk @wence- I updated the PR with better docstrings

cpp/include/rapidsmpf/memory/host_memory_resource.hpp
```diff
@@ -111,23 +117,22 @@ class HostMemoryResource {
     /**
      * @brief Compares this resource to another resource.
      *
-     * All instances are stateless and interchangeable, so this always returns
-     * true.
+     * `HostMemoryResource` is otherwise stateless, so equality reduces to
+     * the back-reference contract from `WithBufferResourceBackRef`: two
+     * instances are equal iff they reference the same owning
+     * `BufferResource`, or are both standalone.
```
madsbk:
```suggestion
```

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
madsbk: If it is only used by tests, I think we should raise if you try to copy or promote an uninstalled BackRef.

nirandaperera: @madsbk If we do that, the following will fail 
```c++ 
    RmmResourceAdaptor mr{rmm::mr::cuda_memory_resource{}};
    auto mr_copy = mr;
```
IMO RmmResourceAdapter's buffer resource back ref should be optional

madsbk: > IMO `RmmResourceAdaptor`'s buffer-resource backref should be optional.

Why? `RmmResourceAdaptor` is something only RapidsMPF uses.

I would rather keep the semantics reasonably simple. I don't think we need to support patterns like:

```c++
RmmResourceAdaptor mr{rmm::mr::cuda_memory_resource{}};
auto mr_copy = mr;
```

In fact, if we want that kind of composability, I think we should keep the `OwningResourceAdaptor` instead :)

nirandaperera: Okay. fair enough. Let me change the promotion logic.

nirandaperera: @madsbk With this change, I think `RmmResourceAdapter` will be useless outside of rapidsmpf. We might have to move it to a `detail` namespace later on.

madsbk: Sounds very good :)

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -192,11 +191,25 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
...
+     * reference is valid for as long as this `BufferResource` is alive.
+     */
+    [[nodiscard]] RmmResourceAdaptor const& device_mr_adaptor() const noexcept;
+
     /**
      * @brief Get the RMM host memory resource.
      *
      * @return Reference to the RMM resource used for host allocations.
      */
+    // TODO: returned ref will not keep the BufferResource alive
```
madsbk: Something like:
```
/**
 * @brief Get the RMM host memory resource.
 *
 * The returned reference does not by itself maintain the lifetime of this
 * BufferResource. See device_mr() for details.
 *
 * @return Reference to the RMM resource used for host allocations.
 */
```

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -205,6 +218,7 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
      * @throws std::invalid_argument if no pinned memory resource is available.
      * @return Reference to the RMM resource used for pinned host allocations.
      */
+    // TODO: returned ref will not keep the BufferResource alive
```
madsbk: Something like:
```
   /**
    * @brief Get the RMM pinned host memory resource.
    *
    * The returned reference does not by itself maintain the lifetime of this
    * BufferResource. See device_mr() for details.    
    *
    * @throws std::invalid_argument if no pinned memory resource is available.
    * @return Reference to the RMM resource used for pinned host allocations.
    */
```

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -213,6 +227,7 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
      * @return The pinned host memory resource as an `any_resource`, or `std::nullopt` if
      * pinned host memory is not available.
      */
+    // TODO: returned ref will not keep the BufferResource alive
```
madsbk: Something like:
```
/**
 * @brief Get the pinned host memory resource if available.
 *
 * The returned reference does not by itself maintain the lifetime of this
 * BufferResource. See device_mr() for details. 
 * 
 * @return The pinned host memory resource as an `any_resource`, or `std::nullopt` if
 * pinned host memory is not available.
 */
```

jameslamb (commented):
Seems like this PR is still under active development and discussion. Please `@` me for a `ci-codeowners` approval once you're ready for that.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
nirandaperera: @madsbk I removed the Host and Pinned MR backref, because after making `set_backref` mandatory, it required a lot changes in the callsites. So, I will immediately start work on it after this merges.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
nirandaperera: Same here

cpp/include/rapidsmpf/memory/buffer_resource.hpp
nirandaperera: Same here

cpp/include/rapidsmpf/memory/buffer_resource.hpp
madsbk: Alright, move the the TODO instead the docstring then

nirandaperera: Done

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
```diff
@@ -0,0 +1,145 @@
...
+ * external @p BackRef instance. Each instance is in one of two states:
+ *
+ * - **Uninstalled** (default-constructed): the instance is not bound to any
+ *   owner. Copying an uninstalled instance throws `std::bad_weak_ptr`; a
+ *   back-reference must be installed via `set_backref()` before copying.
+ * - **Installed** (after `set_backref()`): the instance is bound to a
+ *   specific owner. Each copy of an installed instance acquires shared
+ *   ownership of that owner for the lifetime of the copy. If the owner has
+ *   been destroyed before the copy is made, copying throws
+ *   `std::bad_weak_ptr`.
```
wence-: Please fix this description. How can an instance be bound to a specific owner and then the instance obtains via copy shared ownership of that owner?

How is the owner, who is the ownee?

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * Move operations transfer state without re-acquiring ownership. Equality
+ * is owner-based: two instances compare equal iff they reference the same
+ * owner, or are both uninstalled.
```
wence-: Why do I care about equality at all?

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
```diff
@@ -0,0 +1,145 @@
...
+     * After construction, *this* references the same owner as @p other and
+     * keeps it alive for as long as *this* lives.
+     *
+     * @param other Mixin to copy from.
+     * @throws std::bad_weak_ptr if @p other is uninstalled, or is installed
+     * but its owner has been destroyed.
+     */
+    BackRefMixin(BackRefMixin const& other)
+        // Reuse `other`'s strong ref when present; otherwise lock its weak ref,
+        // which throws `std::bad_weak_ptr` if `other` is uninstalled or expired.
```
wence-:
```suggestion
```

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
```diff
@@ -0,0 +1,145 @@
...
+     *
+     * @param other Mixin to copy from.
+     * @return Reference to this.
+     * @throws std::bad_weak_ptr if @p other is uninstalled, or is installed
+     * but its owner has been destroyed.
+     */
+    BackRefMixin& operator=(BackRefMixin const& other) {
+        if (this != &other) {
+            // Promote first so a throw leaves *this unchanged. Locking an
+            // uninstalled or expired weak ref throws `std::bad_weak_ptr`.
```
wence-:
```suggestion
            // Promote first so a throw leaves *this unchanged.
```

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
```diff
@@ -0,0 +1,145 @@
...
+class BufferResource;
+
+/**
+ * @brief Convenience alias: `BackRefMixin` instantiated for `BufferResource`.
+ *
+ * Inherit from this alias to give a type the back-reference lifetime
+ * contract that `BufferResource::create()` installs on its internal
+ * resources. See `BackRefMixin` for the contract.
+ */
+using WithBufferResourceBackRef = BackRefMixin<BufferResource>;
```
wence-: Still seems unnecessary.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -503,10 +516,11 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
...
 
     std::mutex mutex_;
-    // Stable storage for the device MR returned by `device_mr()`. See
-    // `OwningResourceAdaptor` for the lifetime semantics. Installed by
-    // `create()` after construction.
-    std::optional<OwningResourceAdaptor<RmmResourceAdaptor, BufferResource>> owning_mr_;
+    // The device adaptor is installed with a back-reference to this instance
+    // during the `create()` factory method, so the `BufferResource` stays alive
+    // for as long as copies of the device adaptor are alive. The pinned and host
+    // resources do not (yet) carry a back-reference.
```
wence-:
```suggestion
```

cpp/include/rapidsmpf/rmm_resource_adaptor.hpp
```diff
@@ -25,10 +26,24 @@ namespace rapidsmpf {
...
+ * is copied (copying an uninstalled adaptor throws `std::bad_weak_ptr`).
+ * Adaptors installed by `BufferResource::create()` keep their owning
+ * `BufferResource` alive for as long as any copy of the adaptor lives.
+ *
+ * @note A standalone adaptor (constructed directly, without a `BufferResource`)
+ * has no back-reference and therefore **must not be copied**. Triggering
+ * operations include type-erasing the adaptor into an owning
+ * `cuda::mr::any_resource`, constructing an `rmm::device_buffer` with it as
+ * the memory resource, and passing it by value to `Statistics::MemoryRecorder`.
+ * Use `BufferResource::device_mr_adaptor()` to obtain a copyable adaptor.
```
wence-: This is terrible.

cpp/tests/test_buffer_resource.cpp
```diff
@@ -827,11 +826,48 @@ TEST(OwningResourceAdaptor, CopyThrowsWhenBackRefExpired) {
...
+    RmmResourceAdaptor target{any_device_resource{cuda_mr}};
     EXPECT_THROW(target = original, std::bad_weak_ptr);
 }
+
+TEST(RmmResourceAdaptor, CopyWithoutBackRefThrows) {
+    // A `set_backref()` must be installed before an adaptor is copied. Copying
+    // an uninstalled adaptor throws `std::bad_weak_ptr`.
+    rmm::mr::cuda_memory_resource cuda_mr;
+
+    RmmResourceAdaptor original{any_device_resource{cuda_mr}};
```
wence-: I think this is terrible as a design. This adaptor is completely fine and copyable because its MR is copyable and it keeps the MR alive.

The problem we're trying to solve is that we want the adaptor _if it comes from a `BufferResource`_ to keep the `BufferResource` alive. But that shouldn't preclude using the adaptor when it _does not_ come from a `BufferResource`.

cpp/tests/test_buffer_resource.cpp
```diff
@@ -827,11 +826,48 @@ TEST(OwningResourceAdaptor, CopyThrowsWhenBackRefExpired) {
...
+    }
+
+    auto const& records = stats->get_memory_records();
+    ASSERT_EQ(records.count("br-scope"), 1u);
+    auto const& rec = records.at("br-scope");
+    EXPECT_EQ(rec.num_calls, 1u);
+    EXPECT_EQ(rec.global_peak, static_cast<std::int64_t>(kAllocBytes));
+    EXPECT_EQ(rec.scoped.peak(), static_cast<std::int64_t>(kAllocBytes));
+    EXPECT_EQ(rec.scoped.total(), static_cast<std::int64_t>(kAllocBytes));
+}
```
wence-: What is this test actually testing? The comment is _meaningless_.

wence- (changes requested):

wence-:
Please also apply some work to fix the PR description.

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
nirandaperera: Ah my bad! semantics changed since I made it mandatory to set a backref before copying, So, uninstalled state is no longer valid.

cpp/tests/test_buffer_resource.cpp
nirandaperera: Related to this comment as well, https://github.com/rapidsai/rapidsmpf/pull/1078/changes#r3415079826

This was based on @madsbk 's suggestion https://github.com/rapidsai/rapidsmpf/pull/1078#discussion_r3378338624

We decided to disable `RmmResourceAdapter` instantiation outside of rapidsmpf.

cpp/tests/test_buffer_resource.cpp
nirandaperera: `br->device_mr()` returns a type-erased device resource ref. `Statistics` casts this ref to a `RmmResourceAdapter`. If casting fails this will be a no-op. This checks whether this casting machinery works as expected.

wence-: Thanks, this makes sense. Can we update to state this? Ie that the type erased thing should be cast correctly and hence the statistics recorded?

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
nirandaperera: Two back_ref objects will be equal if their weak ptrs point to the same control block. We cant simply use `this == addressof(other)` here.

wence-: OK, but what do I need equality for? What goes wrong if (for example) we delete `operator==`

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -192,6 +191,19 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
...
+    /**
+     * @brief Access the internal device memory resource adaptor.
+     *
+     * `BufferResource` wraps the device memory resource in an internal
+     * `RmmResourceAdaptor` for allocation tracking. This exposes that adaptor
+     * directly, e.g. to query allocation statistics via `get_main_record()` or
+     * `current_allocated()`.
+     *
+     * @return Reference to the internal device `RmmResourceAdaptor`. The
+     * reference is valid for as long as this `BufferResource` is alive.
```
wence-: nit: "internal" in this docstring suggests that this perhaps ought to be private.

Perhaps "access the concrete device memory resource adaptor".

I think also note here that one should use `device_mr` for actual allocations etc...?

cpp/include/rapidsmpf/rmm_resource_adaptor.hpp
```diff
@@ -25,10 +27,19 @@ namespace rapidsmpf {
...
  * This class is copyable and shares ownership of its internal state via
  * `cuda::mr::shared_resource`.
+ *
+ * The primary-resource constructor is private: instances can only be created by
+ * `BufferResource`, which installs the back-reference before the adaptor becomes
+ * observable. Obtain one via `BufferResource::device_mr_adaptor()` (copies of
+ * which are valid). This guarantees the `BackRefMixin<BufferResource>` lifetime
+ * contract always holds — every adaptor a caller can reach already has a
+ * back-reference installed, so copies keep their owning `BufferResource` alive
+ * for as long as any copy lives.
```
wence-: Perhaps simplify this (and its implementation detail) to say something like:

> It is not possible to construct an `RmmResourceAdaptor` directly. Instead obtain one by creating a `BufferResource` with the memory resource you wish to use for allocations and then obtain the adaptor with `device_mr_adaptor()`

?

cpp/include/rapidsmpf/rmm_resource_adaptor.hpp
```diff
@@ -39,27 +50,21 @@ class RmmResourceAdaptor
...
-
     ~RmmResourceAdaptor() = default;
 
     /**
      * @brief Equality comparison.
      *
-     * Two adaptors are equal if and only if they share the same underlying shared state.
+     * Two adaptors are equal iff they share the same underlying shared
+     * state **and** reference the same owning `BufferResource` (or are
+     * both standalone).
```
wence-: If we're making things private to `BufferResource` can we ever have standalone adaptors?

wence- (approved):
Some minor documentation suggestions

cpp/include/rapidsmpf/memory/back_ref_mixin.hpp
nirandaperera: I think for current usages, we dont need it, because rmm resource adapter defines its own equality operator. But I wanted to preserve the `default` equlity operator behavior. 
Eg: 
```c++
class Foo: BackRefMixin<int>{
int a; 
bool operator==(const Foo&) const = default;
}
```
I think for this to work, we need BackRefMixin to define an equality opertor.

cpp/include/rapidsmpf/rmm_resource_adaptor.hpp
nirandaperera: You are right. This simplifies to `get() == other.get()`

nirandaperera:
@jameslamb This is ready

madsbk (approved):

jameslamb (approved):
Believe this no longer needs `ci-codeowners` / `packaging-codeowners` but approving for you anyway.

I skimmed the changes and don't see any issues, deferring to the much more thorough and better-informed reviews given by other reviewers.

PR #1079, opened 2026-06-02
Improve rrun NUMA binding

pentschev:
Make rrun NUMA memory binding use process memory-policy syscalls, gracefully skip unavailable memory binding, tighten NUMA validation and make GPU PCI lookup use unfiltered topology.


madsbk (approved):
Looks good

pentschev:
Thanks for the review, Mads!

PR #1080, opened 2026-06-02
Use cuCascade topology-only build mode

vyasr:
This PR switches rapidsmpf to use cuCascade's new `CUCASCADE_TOPOLOGY_ONLY=ON` mode, which builds only the topology discovery library without requiring cudf/rmm dependencies at cuCascade configure time.

This breaks the circular dependency where:
- rapidsmpf depends on cuCascade (topology discovery)
- cuCascade (full build) depends on cudf
- cudf depends on rapidsmpf

With topology-only mode, cuCascade only needs CUDAToolkit (for NVML), making the dependency chain linear.

Depends on:
- https://github.com/NVIDIA/cuCascade/pull/141

### DIFF
--- a/cmake/thirdparty/get_cucascade.cmake
+++ b/cmake/thirdparty/get_cucascade.cmake
@@ -11,16 +11,17 @@
 function(find_and_configure_cucascade)
   rapids_cpm_find(
     cuCascade 0.1.0
-    GLOBAL_TARGETS cuCascade::cucascade cuCascade::cucascade_topology_discovery
+    GLOBAL_TARGETS cuCascade::cucascade_topology_discovery
     CPM_ARGS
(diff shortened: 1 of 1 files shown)


madsbk (approved):

vyasr:
Converting to draft. This was a workaround that I thought we might need as part of https://github.com/rapidsai/cudf/pull/22747, but at least for the moment we seem to be avoiding any issues in the jobs that do involve refetching cuCascade. I'll keep this in draft for now just in case an edge case pops up while we're moving through the rapidsmpf->cudf transition.

vyasr:
I think we are going to need this to be able to merge https://github.com/rapidsai/cudf/pull/22786 safely.

bdice (approved):

vyasr:
Closing, this change is folded into https://github.com/rapidsai/rapidsmpf/pull/1083.

PR #1083, opened 2026-06-05
Remove cudf components

vyasr:
This PR removes all the parts of rapidsmpf that rely on libcudf from rapidsmpf. Those parts have all been moved to the cudf_streaming library in the cudf repository.

Tests and benchmarks remain (at least for now), with benchmarks having been moved to `librapidsmpf-tests` package to avoid introducing dependencies to `libcudf_streaming` in the library package.


ci/build_wheel_librapidsmpf.sh
```diff
@@ -37,7 +37,6 @@ export SITE_PACKAGES
 ./ci/build_wheel.sh "${package_name}" "${package_dir}"
 
 python -m auditwheel repair \
-    --exclude libcudf.so \
     --exclude libkvikio.so \
     --exclude libnvcomp.so.5 \
```
pentschev: I think kvikio and nvcomp were here for cuDF, so they should probably be removed too.

cmake/thirdparty/get_cucascade.cmake
```diff
@@ -11,16 +11,17 @@
 function(find_and_configure_cucascade)
   rapids_cpm_find(
     cuCascade 0.1.0
-    GLOBAL_TARGETS cuCascade::cucascade cuCascade::cucascade_topology_discovery
+    GLOBAL_TARGETS cuCascade::cucascade_topology_discovery
     CPM_ARGS
     GIT_REPOSITORY https://github.com/NVIDIA/cuCascade.git
-    GIT_TAG main
+    GIT_TAG 6a8cf0f7c545c601cec4b41a0da22ad6a17eeb7c
```
pentschev: Why is this pinning to a specific commit and not `main`?

cpp/benchmarks/CMakeLists.txt
```diff
@@ -5,34 +5,6 @@
...
-target_link_libraries(
-  bench_shuffle PRIVATE rapidsmpf::rapidsmpf ucxx::ucxx $<TARGET_NAME_IF_EXISTS:MPI::MPI_CXX>
-                        $<TARGET_NAME_IF_EXISTS:conda_env> maybe_asan bench_utils
-)
-install(
-  TARGETS bench_shuffle
-  COMPONENT benchmarking
-  DESTINATION bin/benchmarks/librapidsmpf
-  EXCLUDE_FROM_ALL
-)
```
pentschev, line 34: Why are we removing benchmarks? I think they're not necessarily part of the move to `cudf_streaming`, in particular those that are not part of the `streaming` submodule, are they?

cpp/tests/test_shuffler.cpp
pentschev, line 1: I'm not sure we want to remove `Shuffler` tests entirely. We probably want to find a middle-ground here, since removing this file completely will leave `Shuffler` pretty much untested. Maybe for benchmarks tests we should still link to cuDF since they're not packaged and keep them mostly "as is"?

pentschev (commented):
Thanks @vyasr . I did an initial pass and left a few comments. In particular I'm a bit concerned about removing certain benchmarks/examples and more importantly tests that we had for "core" components both in C++ and Python. I think some of them were not necessarily intended to be moved to `cudf_streaming`, but maybe I'm missing some context. For tests in particular, I think we should still keep core components being tests against cuDF, as they provide important coverage that we would lose if they're not handled here. Would be also good to hear @madsbk 's opinion (and Lawrence too, but I'm not tagging him since he's on PTO).

vyasr:
@pentschev the short answer to your questions is that I think it is possible for tests and benchmarks to still link to cudf, we'll just have to disable those in devcontainer builds by default where that would result in the circular dependency persisting. For conda and wheel builds it should be OK, I think...

cpp/benchmarks/CMakeLists.txt
nirandaperera: I was thinking the same. Cant benchmarks depend on `cudf_streaming`?

cpp/include/rapidsmpf/memory/cuda_memcpy_async.hpp
```diff
@@ -44,7 +45,7 @@ namespace rapidsmpf {
 [[nodiscard]] inline cudaError_t cuda_memcpy_async(
     void* dst, void const* src, std::size_t count, rmm::cuda_stream_view stream
 ) {
-    return cudf::detail::memcpy_async(dst, src, count, stream);
+    return cudaMemcpyAsync(dst, src, count, cudaMemcpyDefault, stream.value());
```
nirandaperera: Oh no. this is what we want to skip 😅 We want to use `cudaMemcpyBatchAsync` instead. 
This will fail a precommit hook.

cpp/tests/test_allgather_packed_data.cpp
```diff
@@ -0,0 +1,36 @@
...
+    rapidsmpf::coll::AllGather allgather{comm, 0, br.get()};
+
+    auto packed_data = generate_packed_data(4, 0, stream, *br);
+    allgather.insert(0, std::move(packed_data));
+    allgather.insert_finished();
+
+    auto results = allgather.wait_and_extract(
+        rapidsmpf::coll::AllGather::Ordered::NO, std::chrono::seconds{30}
+    );
+    EXPECT_FALSE(results.empty());
```
nirandaperera: it would be nice to verify the contents here as well (I think it should be `comm.nranks() x [0..3]`)

cpp/tests/test_shuffler.cpp
nirandaperera: I agree.

nirandaperera (dismissed):
I agree with @pentschev take here. It would be nice to  link `cudf_streaming` to tests and benchmarks, and leave them as is.

cpp/include/rapidsmpf/memory/cuda_memcpy_async.hpp
vyasr: Well we definitely cannot `cudf::detail::memcpy_async` here. If we need to use `cudaMemcpyBatchAsync` then we will have to vendor the cudf function. We absolutely cannot use cudf detail functions here.

nirandaperera: Yes, I agree. This is rather a straigthforward impl. Let me send a patch

cpp/benchmarks/bench_pack.cpp
pentschev, line 1: Possibly this and `bench_partition.cpp` could move to `cudf_streaming`, they're primarily cuDF tests with no collectives, I don't think it has to be in this PR though. WDYT @nirandaperera ?

cpp/scripts/ndsh.py
pentschev, line 1: Seems like the change in permissions is wanted since execution is removed from this file that contains a shebang.

vyasr: Reverted this change.

cmake/Modules/ConfigureCUDA.cmake
```diff
@@ -31,7 +31,7 @@ list(
   RAPIDSMPF_CUDA_FLAGS
   -Xcompiler=-Wall,-Werror,-Wextra,-Wsign-conversion,-Wno-unknown-pragmas,-Wno-missing-field-initializers,-Wno-error=deprecated-declarations
 )
-# This warning needs to be suppressed because some parts of cudf instantiate templated CCCL
+# This warning needs to be suppressed because some parts of rapidsmpf instantiate templated CCCL
 # functions in contexts where the resulting instantiations would have internal linkage (e.g. in
 # anonymous namespaces). In such contexts, the visibility attribute on the template is ignored, and
 # the compiler issues a warning. This is not a problem and will be fixed in future versions of CCCL.
```
bdice, line 37: Follow up: This suppression has been here since @madsbk's "initial commit" 2 years ago. We should re-evaluate if it is still needed and remove it if not.

conda/recipes/librapidsmpf/recipe.yaml
```diff
@@ -134,7 +132,6 @@ outputs:
           - ${{ stdlib("c") }}
         by_name:
           - cuda-cupti
-          - libcudf
```
bdice, line 137: We still need to ignore `libcudf` (and possibly other dependencies? Please double-check the list below). This is now being included from `cache-host`, because `libcudf` [has a run-export on itself](https://github.com/rapidsai/cudf/blob/363920c83694ee88f2af12568241250d81983144/conda/recipes/libcudf/recipe.yaml#L154-L155). https://github.com/rapidsai/rapidsmpf/actions/runs/27112901660/job/80014151119?pr=1083#step:11:1401

```
 │ │ Finalized run dependencies (librapidsmpf-26.08.00a74-cuda13_260608_c925c61f):
 │ │ ╭──────────────────┬──────────────────────────────────────────────────────╮
 │ │ │ Name             ┆ Spec                                                 │
 │ │ ╞══════════════════╪══════════════════════════════════════════════════════╡
 │ │ │ Run dependencies ┆                                                      │
 │ │ │ cuda-cupti       ┆ *                                                    │
 │ │ │ cuda-version     ┆ >=13,<14.0a0 (PC)                                    │
 │ │ │ libcudf          ┆ >=26.8.0a615,<26.9.0a0 (RE of [cache-host: libcudf]) │
 │ │ │ libnuma          ┆ >=2.0.18,<3.0a0 (RE of [host: libnuma])              │
 │ │ │                  ┆ >=2.0.18,<3.0a0 (RE of [cache-host: libnuma])        │
 │ │ │ libpmix          ┆ >=5.0,<6.0                                           │
 │ │ │                  ┆ >=5.0.8,<6.0a0 (RE of [host: libpmix-devel])         │
 │ │ │                  ┆ >=5.0.8,<6.0a0 (RE of [cache-host: libpmix-devel])   │
 │ │ │ librmm           ┆ 26.8.*                                               │
 │ │ │ libucxx          ┆ 0.51.*                                               │
 │ │ │                  ┆ >=0.51.0a25,<0.52.0a0 (RE of [host: libucxx])        │
 │ │ │                  ┆ >=0.51.0a25,<0.52.0a0 (RE of [cache-host: libucxx])  │
 │ │ │ openmpi          ┆ >=5.0                                                │
 │ │ ╰──────────────────┴──────────────────────────────────────────────────────╯
```

Only the `librapidsmpf-tests` package should have `libcudf` / `libcudf-streaming` dependencies. https://github.com/rapidsai/rapidsmpf/actions/runs/27112901660/job/80014151119?pr=1083#step:11:1980

```
 │ │ Finalized run dependencies (librapidsmpf-tests-26.08.00a74-cuda13_260608_c925c61f):
 │ │ ╭───────────────────┬──────────────────────────────────────────────────────╮
 │ │ │ Name              ┆ Spec                                                 │
 │ │ ╞═══════════════════╪══════════════════════════════════════════════════════╡
 │ │ │ Run dependencies  ┆                                                      │
 │ │ │ cuda-cupti        ┆ *                                                    │
 │ │ │ cuda-version      ┆ >=13,<14.0a0 (PC)                                    │
 │ │ │ libcudf           ┆ 26.8.*                                               │
 │ │ │                   ┆ >=26.8.0a615,<26.9.0a0 (RE of [host: libcudf])       │
 │ │ │                   ┆ >=26.8.0a615,<26.9.0a0 (RE of [cache-host: libcudf]) │
 │ │ │ libcudf-streaming ┆ 26.8.*                                               │
 │ │ │ libnuma           ┆ >=2.0.18,<3.0a0 (RE of [host: libnuma])              │
 │ │ │                   ┆ >=2.0.18,<3.0a0 (RE of [cache-host: libnuma])        │
 │ │ │ libpmix           ┆ >=5.0.8,<6.0a0 (RE of [cache-host: libpmix-devel])   │
 │ │ │ librmm            ┆ 26.8.*                                               │
 │ │ │ libucxx           ┆ 0.51.*                                               │
 │ │ │                   ┆ >=0.51.0a25,<0.52.0a0 (RE of [cache-host: libucxx])  │
 │ │ │ openmpi           ┆ >=5.0                                                │
 │ │ ╰───────────────────┴──────────────────────────────────────────────────────╯
```

conda/recipes/librapidsmpf/recipe.yaml
```diff
@@ -149,6 +146,7 @@ outputs:
       script:
         content: |
           cmake --install cpp/build --component testing
+          cmake --install cpp/build --component=benchmarking
```
bdice, line 151: This has an `=` but the line above doesn't. Let's be consistent.

conda/recipes/librapidsmpf/recipe.yaml
```diff
@@ -77,8 +77,9 @@ cache:
       - cuda-version =${{ cuda_version }}
       - cuda-cupti-dev
       - cuda-nvml-dev
-      - libcudf =${{ minor_version }}
       - libnuma
+      - libcudf-streaming =${{ minor_version }}
+      - libcudf =${{ minor_version }}
```
bdice: We should alphabetize these (libcudf before libcudf-streaming). Please apply throughout.

cpp/benchmarks/streaming/ndsh/bench_read.cpp
```diff
@@ -75,8 +75,8 @@ rapidsmpf::streaming::Actor consume_channel_parallel(
             if (msg.empty()) {
                 break;
             }
-            if (msg.holds<rapidsmpf::streaming::TableChunk>()) {
-                auto chunk = co_await msg.release<rapidsmpf::streaming::TableChunk>()
+            if (msg.holds<cudf_streaming::streaming::TableChunk>()) {
```
bdice, line 78: It's a little odd to have "streaming" twice in the name, maybe this can be refactored later.

cpp/benchmarks/CMakeLists.txt
bdice: If I'm reading this right, it's still there, but moved down below into a conditional block `if(BUILD_CUDF_TESTS)`.

cpp/.clang-format
```diff
@@ -58,7 +58,7 @@ IncludeCategories:
     Priority: 30 # CUDA includes
   - Regex: '^<(thrust|cub|cuda)/'
     Priority: 40 # CCCL includes
-  - Regex: '^<(cudf.*|rmm|cugraph|cuml|raft|kvikio|cucascade)'
+  - Regex: '^<(rmm|cugraph|cuml|raft|kvikio|cucascade)'
```
bdice: This change seems unexpected to me. Why remove cudf here? We still have a few optional includes, in the tests/benchmarks. We try to keep this file mostly-consistent across RAPIDS, which is why it mentions other libraries too.

cpp/CMakeLists.txt
```diff
@@ -69,6 +70,7 @@ message(STATUS "  BUILD_SHARED_LIBS       : ${BUILD_SHARED_LIBS}")
 message(STATUS "  RAPIDSMPF_CLANG_TIDY    : ${RAPIDSMPF_CLANG_TIDY}")
 message(STATUS "  RAPIDSMPF_ASAN          : ${RAPIDSMPF_ASAN}")
 message(STATUS "  RAPIDSMPF_VERBOSE_INFO  : ${RAPIDSMPF_VERBOSE_INFO}")
+message(STATUS "  BUILD_CUDF_TESTS          : ${BUILD_CUDF_TESTS}")
```
bdice: Spacing.
```suggestion
message(STATUS "  RAPIDSMPF_VERBOSE_INFO  : ${RAPIDSMPF_VERBOSE_INFO}")
message(STATUS "  BUILD_CUDF_TESTS        : ${BUILD_CUDF_TESTS}")
```

dependencies.yaml
```diff
@@ -676,3 +620,33 @@ dependencies:
...
+              cuda: "13.*"
+              cuda_suffixed: "true"
+            packages:
+              - pylibcudf-cu13==26.8.*,>=0.0.0a0
+          - matrix:
+            packages:
+              - *pylibcudf_unsuffixed
+  depends_on_cudf_streaming:
+    common:
+      - output_types: conda
```
bdice, line 650: Are there wheels for cudf-streaming yet? I thought I approved some code for that but maybe it wasn't merged at the time this was written. Please check.

cmake/Modules/ConfigureCUDA.cmake
vyasr: Added this task to https://github.com/rapidsai/cudf/issues/22771.

conda/recipes/librapidsmpf/recipe.yaml
vyasr: Good catch.

conda/recipes/librapidsmpf/recipe.yaml
vyasr: Fixed — using `--component=testing` now for consistency.

conda/recipes/librapidsmpf/recipe.yaml
vyasr: Fixed — `libcudf` before `libcudf-streaming` throughout the recipe.

cpp/benchmarks/streaming/ndsh/bench_read.cpp
vyasr: Renaming the namespaces is tracked in https://github.com/rapidsai/cudf/issues/22771.

cpp/benchmarks/CMakeLists.txt
vyasr: Correct.

cpp/.clang-format
vyasr: You're right, tests and benchmarks still `#include <cudf/...>`. Added `cudf` back to the RAPIDS includes regex.

cpp/CMakeLists.txt
vyasr: Fixed, thanks.

dependencies.yaml
vyasr: Not yet. We should have had them after https://github.com/rapidsai/cudf/pull/22760, but there was an oversight in that PR's changes to `build.yaml` — it didn't properly specify that Python artifacts of cudf_streaming use the stable ABI, which causes failures like [this one](https://github.com/rapidsai/cudf/actions/runs/27000103965/job/79681708134). The circularity of the rapidsmpf↔cudf dependency tree (partly alleviated by this branch making cudf a test-only dep of rapidsmpf) meant we couldn't merge a fix there until this rapidsmpf branch lands. Once this merges we'll get that fixed and enable the wheels.

bdice (approved):
Thank you. Approving packaging and CI and CMake.

pentschev (approved):

pentschev:
It seems Niranda's concerns (like mine) have been addressed, so I'm going ahead and merging it to speedup unblocking.

PR #1084, opened 2026-06-07
Cudf-free shuffler tests

nirandaperera:
## Summary

The shuffler round-trip tests depended on cudf for input generation and result checking (`random_table_with_index`, `partition_and_pack`/`unpack_and_concat`, `CUDF_TEST_EXPECT_TABLES_EQUIVALENT`, `cudf::test::BaseFixture`), which made them heavy and coupled to the cudf integration. This replaces that with a self-contained, conservation-preserving data model built directly from `PackedData`, so the shuffler tests no longer depend on cudf and run as part of the default (non-`BUILD_CUDF_TESTS`) build.

### C++ utilities
- Add `ceil_div` and a `chunk_indices` view to `utils/misc.hpp`. `chunk_indices` tiles `[0, count)` into exactly `num_chunks` contiguous (possibly empty) ranges whose sizes sum to `count`.
- `ceil_div` is computed as `x / y + (x % y != 0)` rather than the naive `(x + y - 1) / y`, which overflows for large unsigned values and is UB on signed overflow near the type maximum.
- Add compile-time edge-case coverage for `ceil_div` in `test_misc.cpp` (`static_assert`s for exact/remainder division, zero numerator, denominator of one, and values at `uint64`/`int64` maxima).

### C++ shuffler tests (now cudf-free)
- Rework the shuffler tests (`test_shuffler.cpp`, `streaming/test_shuffler.cpp`) around `make_partition_data` (produces routed chunks) and `validate_partition_data` (verifies received chunks), dropping all `cudf`/`cudf_test` usage and switching to `rmm::` stream/MR helpers and `::testing::Test`.
- Generalize the test helpers `generate_packed_data`/`validate_packed_data` in `tests/utils.hpp` over the element type and build device buffers via the buffer resource + `cuda_memcpy_async` instead of cudf.
- De-couple `streaming/test_allgather.cpp` from `cudf_streaming/integrations/partition.hpp`.
- `cpp/tests/CMakeLists.txt`: move `test_shuffler.cpp`, `streaming/test_shuffler.cpp`, and `streaming/test_allgather.cpp` out of the `BUILD_CUDF_TESTS` block into the default build; the cudf block now only carries `test_partition.cpp`, `test_shuffler_many_streams.cpp`, and `streaming/test_leaf_actor.cpp`.

### Python (now cudf-free)
- Add a `PackedData.from_device_buffer(gpu_data, metadata, stream, br)` factory (plus `.pyi` stub) so packed data can be built from an rmm `DeviceBuffer` without cudf.
- Add shared, cudf-free shuffle test helpers to `rapidsmpf/testing.py`: `chunk_indices`, `generate_packed_data`, `validate_packed_data`, `make_partition_data`, and `validate_partition_data`.
- Rewrite the Python shuffler tests (`tests/test_shuffler.py`, `tests/streaming/test_shuffler.py`) around those helpers and the `Shuffler`/`ShufflerAsync` object interfaces, removing the `cudf` / `cudf_streaming` / `pylibcudf` dependencies.

Depends on https://github.com/rapidsai/rapidsmpf/pull/1087

### DIFF
--- a/cpp/include/rapidsmpf/utils/misc.hpp
+++ b/cpp/include/rapidsmpf/utils/misc.hpp
@@ -4,8 +4,10 @@
  */
 #pragma once
 
+#include <algorithm>
 #include <chrono>
 #include <cmath>
+#include <concepts>
 #include <cstdlib>
 #include <functional>
 #include <memory>
@@ -191,6 +193,49 @@ bool is_running_under_valgrind();
 template <typename T>
 constexpr T safe_div(T x, T y) {
     return (y == 0) ? 0 : x / y;
+}
+
+/**
+ * @brief Computes the ceiling of the division of two integers.
+ *
+ * Returns the smallest integer not less than `x / y`. Both operands must be
+ * non-negative and the denominator must be non-zero.
+ *
+ * @tparam T An integral type.
+ * @param x The numerator (must be non-negative).
+ * @param y The denominator (must be positive).
+ * @return T The ceiling of `x / y`.
+ */
+template <std::integral T>
+constexpr T ceil_div(T x, T y) {
+    return (x + y - 1) / y;
+}
+
+/**
+ * @brief Splits the index range `[0, count)` into exactly `num_chunks` contiguous chunks.
+ *
+ * Each chunk is a half-open `[begin, end)` index pair. Chunks are front-loaded with
+ * size `ceil(count / num_chunks)`; the last non-empty chunk may be smaller and, when
+ * `count < num_chunks`, the trailing chunks are empty (`begin == end`). The chunks
+ * exactly tile `[0, count)`, so their sizes sum to `count`.
+ *
+ * Unlike `std::ranges::chunk_view` (C++23), this always yields exactly `num_chunks`
+ * chunks, injecting empty trailing chunks as needed.
+ *
+ * @param count The number of elements to split.
+ * @param num_chunks The number of chunks to produce (must be positive).
+ * @return A lazy view of `num_chunks` `std::pair<std::size_t, std::size_t>` `[begin,
+ * end)` index pairs.
+ */
+[[nodiscard]] inline auto chunk_indices(std::size_t count, std::size_t num_chunks) {
+    std::size_t const chunk_size = ceil_div(count, num_chunks);
+    return std::views::iota(std::size_t{0}, num_chunks)
+           | std::views::transform([count, chunk_size](std::size_t k) {
+                 return std::pair<std::size_t, std::size_t>{
+                     std::min(k * chunk_size, count),
+                     std::min((k + 1) * chunk_size, count)
+                 };
+             });
 }
 
 // Macro to concatenate two tokens x and y.
--- a/cpp/include/rapidsmpf/utils/misc.hpp
+++ b/cpp/include/rapidsmpf/utils/misc.hpp
@@ -199,7 +199,8 @@ constexpr T safe_div(T x, T y) {
  * @brief Computes the ceiling of the division of two integers.
  *
  * Returns the smallest integer not less than `x / y`. Both operands must be
- * non-negative and the denominator must be non-zero.
+ * non-negative and the denominator must be non-zero. Computed as `x / y + (x % y != 0)`
+ * to avoid the overflow (and signed UB)
  *
  * @tparam T An integral type.
  * @param x The numerator (must be non-negative).
@@ -208,7 +209,7 @@ constexpr T safe_div(T x, T y) {
  */
 template <std::integral T>
 constexpr T ceil_div(T x, T y) {
-    return (x + y - 1) / y;
+    return x / y + (x % y != 0 ? T{1} : T{0});
 }
 
 /**
--- a/cpp/tests/streaming/test_shuffler.cpp
+++ b/cpp/tests/streaming/test_shuffler.cpp
@@ -13,6 +13,7 @@
 #include <rapidsmpf/cuda_stream.hpp>
 #include <rapidsmpf/memory/buffer.hpp>
 #include <rapidsmpf/shuffler/shuffler.hpp>
+#include <rapidsmpf/streaming/chunks/partition.hpp>
 #include <rapidsmpf/streaming/coll/shuffler.hpp>
 #include <rapidsmpf/streaming/core/actor.hpp>
 #include <rapidsmpf/streaming/core/context.hpp>
@@ -98,25 +99,21 @@ TEST_P(StreamingShuffler, basic_shuffler) {
     const int64_t base = static_cast<int64_t>(comm->rank()) * num_rows;
     std::vector<Message> input_chunks;  // Message contains a PartitionMapChunk
     for (size_t chunk_idx = 0; chunk_idx < num_chunks; ++chunk_idx) {
-        ContentDescription cd{};
         std::unordered_map<shuffler::PartID, PackedData> chunks;
         chunks.reserve(num_partitions);
         for (size_t j = 0; j < num_partitions; ++j) {
             auto [start, end] = pieces[chunk_idx * num_partitions + j];
             // end > start is guaranteed.
-            auto [it, _] = chunks.emplace(
+            chunks.emplace(
                 static_cast<shuffler::PartID>(j),
                 generate_packed_data<int64_t>(
                     end - start, base + static_cast<int64_t>(start), stream, *br
                 )
             );
-            cd.content_size(it->second.data->mem_type()) += it->second.data->size;
         }
-        input_chunks.emplace_back(Message(
-            chunk_idx,
-            std::make_unique<PartitionMapChunk>(std::move(chunks)),
-            std::move(cd)
-        ));
+        input_chunks.emplace_back(
+            to_message(chunk_idx, std::make_unique<PartitionMapChunk>(std::move(chunks)))
+        );
     }
     EXPECT_EQ(input_chunks.size(), num_chunks);
 
@@ -150,7 +147,7 @@ TEST_P(StreamingShuffler, basic_shuffler) {
 
         auto p_vec = chunk.release<PartitionVectorChunk>();
         // for each local pid, it should receive num_chunks * nranks chunks.
-        EXPECT_EQ(p_vec.data.size(), num_chunks * n_ranks);
+        ASSERT_EQ(p_vec.data.size(), num_chunks * n_ranks);
 
         // since values are offset by rank, if we sort packed data by their first element,
         // then it will be in rank & chunk-index order.
--- a/cpp/tests/streaming/test_shuffler.cpp
+++ b/cpp/tests/streaming/test_shuffler.cpp
@@ -3,18 +3,16 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <algorithm>
+#include <ranges>
+
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
-
-#include <cudf/copying.hpp>
-#include <cudf_streaming/integrations/partition.hpp>
-#include <cudf_streaming/streaming/partition.hpp>
-#include <cudf_streaming/streaming/table_chunk.hpp>
-#include <cudf_test/table_utilities.hpp>
 
 #include <rapidsmpf/communicator/single.hpp>
 #include <rapidsmpf/cuda_stream.hpp>
 #include <rapidsmpf/memory/buffer.hpp>
+#include <rapidsmpf/shuffler/shuffler.hpp>
 #include <rapidsmpf/streaming/coll/shuffler.hpp>
 #include <rapidsmpf/streaming/core/actor.hpp>
 #include <rapidsmpf/streaming/core/context.hpp>
@@ -55,13 +53,12 @@ TEST_F(BaseStreamingShuffle, zero_owned_partitions_completes) {
 class StreamingShuffler : public BaseStreamingShuffle,
                           public ::testing::WithParamInterface<int> {
   public:
-    const unsigned int num_partitions = 10;
-    const unsigned int num_rows = 1000;
-    const unsigned int num_chunks = 5;
-    const unsigned int chunk_size = num_rows / num_chunks;
-    const std::int64_t seed = 42;
-    const cudf::hash_id hash_function = cudf::hash_id::HASH_MURMUR3;
-    const OpID op_id = 0;
+    static constexpr size_t num_partitions = 10;
+    static constexpr size_t num_rows = 1000;
+    static constexpr size_t num_chunks = 5;
+    static constexpr OpID op_id = 0;
+
+    shuffler::Shuffler::PartitionOwner partition_owner = shuffler::Shuffler::round_robin;
 
     void SetUp() override {
         BaseStreamingShuffle::SetUpWithThreads(GetParam());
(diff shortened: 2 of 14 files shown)


cpp/include/rapidsmpf/memory/spill.hpp
pentschev, line 1: AFAICT, this is not public API anymore, although you still use it for cudf Python integration which seems appropriate to be removed. Move this to somewhere under `cpp/tests`.

cpp/tests/utils.hpp
```diff
@@ -184,58 +185,70 @@ template <std::integral T = std::int64_t>
...
+    auto data = br.make_buffer(stream, std::move(reservation));
+
+    data->write_access([d_ptr = metadata->data(), m_size = metadata->size()](
+                           std::byte* ptr, rmm::cuda_stream_view op_stream
+                       ) {
+        RAPIDSMPF_CUDA_TRY(rapidsmpf::cuda_memcpy_async(ptr, d_ptr, m_size, op_stream));
+    });
 
-    return {std::move(metadata), br.move(std::move(data), stream)};
+    return {std::move(metadata), std::move(data)};
```
pentschev, line 229: Why is `stream` removed here? `write_access` does not sync the stream, so I think it's still needed for the consumer.

nirandaperera: @pentschev before, we were creating a `rmm::device_buffer` and "moving" it into buffer resource. IMO the better way here should have been make a reservation and then make buffer. This is done in L213 and it sets `stream` for the buffer. `op_stream == stream` in this case.

cpp/include/rapidsmpf/memory/spill.hpp
nirandaperera: I think Spill/UnSpillPartitions should reside in rapidsmpf. It has no dependency to cudf, so I moved it to a separate header.

cpp/include/rapidsmpf/utils/misc.hpp
```diff
@@ -193,6 +195,49 @@ constexpr T safe_div(T x, T y) {
...
+ * non-negative and the denominator must be non-zero.
+ *
+ * @tparam T An integral type.
+ * @param x The numerator (must be non-negative).
+ * @param y The denominator (must be positive).
+ * @return T The ceiling of `x / y`.
+ */
+template <std::integral T>
+constexpr T ceil_div(T x, T y) {
+    return (x + y - 1) / y;
```
pentschev:
```suggestion
    return x / y + (x % y != 0);
```

pentschev: This is now public API, but the original overflows for large unsigned values and is UB for signed overflow, let's thus improve that. Add a test for edge cases too.

cpp/tests/test_shuffler.cpp
```diff
@@ -285,112 +302,83 @@ INSTANTIATE_TEST_SUITE_P(
...
             shuffler,
             total_num_partitions,
-            [&](auto&& packed_chunks) { insert_fn(shuffler, std::move(packed_chunks)); },
-            [&]() { insert_finished_fn(shuffler); },
             100'000,  // total_num_rows
-            t_id,  // seed
-            cudf::hash_id::HASH_MURMUR3,
             stream,
             br.get()
         ));
```
pentschev, line 400: The original cuDF test used `t_id` as the seed, the new one does not pass a distinct base into `test_shuffler`, so cross-talk between concurrent shufflers can pass unnoticed. Please make data different for each thread.

cpp/tests/streaming/test_shuffler.cpp
```diff
@@ -175,12 +78,111 @@ INSTANTIATE_TEST_SUITE_P(
...
+    // output chunk.
+    EXPECT_EQ(local_pids.size(), output_chunks.size());
+    const size_t n_ranks = static_cast<size_t>(comm->nranks());
+    for (auto& chunk : output_chunks) {
+        auto pid = chunk.sequence_number();
+        std::erase_if(local_pids, [pid](auto& p) { return p == pid; });
+
+        auto p_vec = chunk.release<PartitionVectorChunk>();
+        // for each local pid, it should receive num_chunks * nranks chunks.
+        EXPECT_EQ(p_vec.data.size(), num_chunks * n_ranks);
```
pentschev: This should be `ASSERT_EQ`. If the actor drops or duplicates chunks, the test continues and indexes `p_vec.data[r * num_chunks + c_idx]` below, which can go out of bounds instead of failing cleanly at the size mismatch.

cpp/tests/streaming/test_shuffler.cpp
```diff
@@ -175,12 +78,111 @@ INSTANTIATE_TEST_SUITE_P(
...
+                    end - start, base + static_cast<int64_t>(start), stream, *br
+                )
+            );
+            cd.content_size(it->second.data->mem_type()) += it->second.data->size;
+        }
+        input_chunks.emplace_back(Message(
+            chunk_idx,
+            std::make_unique<PartitionMapChunk>(std::move(chunks)),
+            std::move(cd)
+        ));
```
pentschev: Could this use `to_message(chunk_idx, std::make_unique<PartitionMapChunk>(std::move(chunks)))` instead of constructing `Message` manually? The manual `ContentDescription cd{}` leaves the message non-spillable and omits the copy callback, so this no longer exercises the normal `PartitionMapChunk` streaming message path through `SpillableMessages`.

pentschev (changes requested):

cpp/tests/streaming/test_shuffler.cpp
nirandaperera: Oh yes. Good idea!

cpp/include/rapidsmpf/utils/misc.hpp
nirandaperera: Done

cpp/tests/test_shuffler.cpp
```diff
@@ -130,116 +130,135 @@ MemoryLimitsMap get_memory_limits_map(rapidsmpf::MemoryType priorities) {
...
+// are empty). Sub-region (local_pidx, split_idx) is piece k = local_pidx*P + split_idx
+// and is routed to destination partition split_idx; input region local_pidx is the union
+// of its P sub-regions. The pieces exactly tile [0,N), so the total shuffled data == N
+// regardless of rank/partition counts (conservation). A per-shuffle `base` offset is
+// added to every value so distinct shuffles carry distinct data.
+
+// Produces the non-empty sub-regions of one owned input region `local_pidx`, keyed by
+// destination partition. Since local_partitions() across ranks partition [0,P), every
+// input region is produced exactly once, so rows are not replicated.
+std::unordered_map<rapidsmpf::shuffler::PartID, rapidsmpf::PackedData>
```
madsbk, line 173: Use a normal docstring format explaining all `@param`

cpp/tests/test_shuffler.cpp
```diff
@@ -130,116 +130,135 @@ MemoryLimitsMap get_memory_limits_map(rapidsmpf::MemoryType priorities) {
...
+                )
+            );
+        }
+    }
+    return chunks;
+}
+
+// Verifies that the `received` chunks for partition `j` match the non-empty sub-regions
+// expected for it.
+void validate_partition_data(
```
madsbk, line 218: Use a normal docstring format explaining all `@param`

python/rapidsmpf/rapidsmpf/testing.py
```diff
@@ -6,13 +6,23 @@
 
 from typing import TYPE_CHECKING
 
+import numpy as np
 import pylibcudf
```
madsbk, line 10: Remove?

pentschev: Not ready for that yet. This PR adds new tests, removal of cudf dependencies need to be handled separately as a follow-up to https://github.com/rapidsai/cudf/pull/22814 in RapidsMPF.

python/rapidsmpf/rapidsmpf/testing.py
```diff
@@ -63,3 +73,178 @@ def assert_eq(
...
+        CUDA stream used for the device allocation.
+    br
+        Buffer resource used for memory management.
+
+    Returns
+    -------
+    A ``PackedData`` containing the integer sequence.
+    """
+    data = np.arange(offset, offset + n_elements, dtype=_DTYPE).tobytes()
+    gpu_data = rmm.DeviceBuffer.to_device(data, stream=stream)
```
pentschev: Please allocate this test payload through `br.device_mr` (or a small BR-backed helper) before wrapping it in `PackedData`. `to_device()` uses the current/default RMM resource and `from_device_buffer()` only adopts the allocation, so the replacement tests bypass `BufferResource` tracking that the old `partition_and_pack(..., br=br)` path covered.

python/rapidsmpf/rapidsmpf/tests/streaming/test_shuffler.py
```diff
@@ -3,290 +3,95 @@
...
-            num_partitions,
-            partition_assignment=PartitionAssignment.CONTIGUOUS,
+    """
+    End-to-end correctness of the async streaming shuffler.
+
+    Each rank inserts the input regions it owns and, after shuffling, every local
+    partition is validated against the conserved data model.
+    """
+    br = context.br()
+    shuffler = ShufflerAsync(context, comm, 0, total_num_partitions)
```
pentschev, line 55: This replacement loses Python coverage of the streaming actor wrapper. The old cudf-based test drove the channel pipeline through `rapidsmpf.streaming.coll.shuffler.shuffler()`, while this now instantiates `ShufflerAsync` directly and never exercises the wrapper or its `PartitionAssignment.CONTIGUOUS` branch. Please add a cudf-free actor/channel test, or expose a generic `PartitionMapChunk` factory so the new `PackedData` model can drive `shuffler()`, ideally covering both assignment modes.

pentschev (changes requested):

python/rapidsmpf/rapidsmpf/tests/streaming/test_shuffler.py
nirandaperera: Good call!

python/rapidsmpf/rapidsmpf/testing.py
nirandaperera: Done

python/rapidsmpf/rapidsmpf/tests/streaming/test_shuffler.py
nirandaperera: I ported the tests over.

madsbk (approved):
Thanks @nirandaperera

cpp/tests/test_shuffler.cpp
```diff
@@ -130,116 +141,181 @@ MemoryLimitsMap get_memory_limits_map(rapidsmpf::MemoryType priorities) {
...
+/// Conservation-preserving data model shared by the shuffler round-trip tests.
+///
+/// The index range `[0, total_num_rows)` is split into `total_num_partitions^2`
+/// contiguous sub-regions via `chunk_indices` (front-loaded, so trailing sub-regions
+/// are empty when `N < P*P`). Sub-region `(local_pidx, split_idx)` is piece
+/// `k = local_pidx * P + split_idx` and is routed to destination partition
+/// `split_idx`. The pieces exactly tile `[0, N)`, so the total shuffled row
+/// count equals `N` regardless of rank or partition counts (conservation). A
+/// per-shuffle `base` offset is added to every value so distinct shuffles carry
+/// distinct data.
```
pentschev, line 153: Is this supposed to be part of `make_partition_data` docs? Currently looks like a strange placing not really connected to anything.

python/rapidsmpf/rapidsmpf/streaming/chunks/partition.pyx
```diff
@@ -27,6 +95,62 @@ cdef class PartitionMapChunk:
...
+
+        Returns
+        -------
+        A new PartitionMapChunk owning the given packed data.
+
+        Raises
+        ------
+        ValueError
+            If any of the provided PackedData objects is empty.
+        """
```
pentschev, line 120: The stub advertises `Mapping`/`Sequence` for these new helpers, but the Cython signatures require concrete `dict`/`list`. That means callers accepted by the type hints fail at runtime, for example a `MappingProxyType` or tuple. Since these helpers are now public in the `.pyi`, please either loosen the Cython parameters to accept the advertised protocols or narrow the stub to `dict[int, PackedData]` / `list[PackedData]`. The same applies to `from_packed_data_list`.

python/rapidsmpf/rapidsmpf/tests/streaming/test_shuffler.py
```diff
@@ -155,8 +137,8 @@ async def generate_inputs(
 async def do_shuffle(
     context: Context,
     comm: Communicator,
-    ch_in: Channel[TableChunk],
-    ch_out: Channel[TableChunk],
+    ch_in: Channel[PartitionMapChunk],
+    ch_out: Channel[PartitionVectorChunk],
```
pentschev: I can't add the comment on the exact lines, but assumes this extends to line 146:

This replacement still bypasses the exported streaming actor wrapper. The removed cudf-based test drove `rapidsmpf.streaming.coll.shuffler.shuffler()` in the channel pipeline ([see that in 26.06 for example](https://github.com/rapidsai/rapidsmpf/blob/78b66f1a8ab6068cd38d73429e3935c3cb97c90f/python/rapidsmpf/rapidsmpf/tests/streaming/test_shuffler.py#L98-L108)), which exercised `streaming/coll/shuffler.pyx` including the `cpp_shuffler` call and `PartitionAssignment` mapping. This new actor instantiates `ShufflerAsync` directly, so a regression in the public `shuffler()` wrapper would no longer be caught. Now that `PartitionMapChunk.from_packed_data_map()` exists, can this test use the actual `shuffler()` actor between `ch_in` and `ch_shuffled`?

pentschev (changes requested):

pentschev:
I have opened #1092 and #1093 adding cuDF-free coverage for the remaining missing tests. Both require changes from this PR (specifically, `rapidsmpf.testing.generate_packed_data` and `rapidsmpf.testing.validate_packed_data`). _I think_ after this PR #1084 plus the aforementioned PRs are merged we would have covered everything missing and could then remove all remaining cuDF pieces without functionality coverage loss.

cpp/tests/test_shuffler.cpp
nirandaperera: I was trying to explain test approach here, especially the `P^2` splits because I felt that a new reader might find it a little confusing.

nirandaperera: Since this relates to make_partition_data, validate_partition_data, test_shuffler methods I thought this is the best place to put it, more as a detail rather than a doc string.

python/rapidsmpf/rapidsmpf/tests/streaming/test_shuffler.py
nirandaperera: Yes you are right. I was more focused on porting the previous tests as is, and missed this detail. I removed the `do_shuffle` actor alltogether, because we can strictly replace it from `shuffle` actor as you said.

python/rapidsmpf/rapidsmpf/streaming/chunks/partition.pyx
nirandaperera: done

python/rapidsmpf/rapidsmpf/tests/streaming/test_shuffler.py
```diff
@@ -222,16 +168,27 @@ def test_shuffler_runtime_obeys_contiguous_assignment(
...
-    expected_local = list(
-        range(
-            rank * num_partitions // nranks,
-            (rank + 1) * num_partitions // nranks,
-        )
-    )
-    assert set(received_pids) == set(expected_local)
-    assert len(received_pids) == len(expected_local)
+    # Single rank, so every partition is local to this rank.
+    assert set(received_pids) == set(range(num_partitions))
```
pentschev, line 172:
```suggestion
    assert set(received_pids) == set(range(num_partitions))
    assert len(received_pids) == num_partitions
```

pentschev: Checks for duplicates, like previously.

pentschev (approved):
Thanks @nirandaperera , one final minor request, otherwise LGTM.

nirandaperera:
Thanks @pentschev

cpp/tests/test_shuffler.cpp
pentschev: It's a strange placement, but I don't have a better idea where to put it either. Let's leave it there for now.

PR #1099, opened 2026-06-12
Hide CUPTI types behind CuptiMonitor PImpl

pentschev:
List of changes:
- move `CuptiMonitor` implementation details into a private PImpl
- remove `cuda_runtime.h` and `cupti.h` from the public `rapidsmpf/cupti.hpp` header
- add `rapidsmpf::CuptiCallbackId` as the public callback counter key wrapper
- update Cython CUPTI bindings to avoid importing `<cupti.h>`
- add a compile-only public header regression test that does not inherit CUDA/CUPTI include paths

### DIFF
--- a/cpp/include/rapidsmpf/cupti.hpp
+++ b/cpp/include/rapidsmpf/cupti.hpp
@@ -1,22 +1,19 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 #pragma once
 
 #ifdef RAPIDSMPF_HAVE_CUPTI
 
-#include <atomic>
 #include <chrono>
 #include <cstddef>
-#include <mutex>
+#include <cstdint>
+#include <functional>
+#include <memory>
 #include <string>
-#include <thread>
 #include <unordered_map>
 #include <vector>
-
-#include <cuda_runtime.h>
-#include <cupti.h>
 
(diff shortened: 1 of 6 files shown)


cpp/tests/test_cupti_public_api.cpp
pentschev, line 1: Note this test doesn't execute, it is only meant to fail during compile time.

cpp/include/rapidsmpf/cupti.hpp
```diff
@@ -30,6 +27,45 @@ struct MemoryDataPoint {
     std::size_t used_memory;  ///< Used GPU memory in bytes
 };
 
+/**
+ * @brief Public wrapper for a CUPTI callback identifier.
+ *
+ * Keeps CUPTI types out of the public RapidsMPF API while preserving the
+ * numeric callback identifier for callers that inspect callback counters.
+ */
+class CuptiCallbackId {
```
vyasr, line 36: Minor nit: This name is only two characters different from the real cupti name, which might lead to a bit of confusion when reading. It's namespaced in rapidsmpf though so maybe it's fine.

vyasr (approved):
LGTM, thanks!

cpp/include/rapidsmpf/cupti.hpp
pentschev: Yeah, I unfortunately don't have a much better idea on how to disambiguate that doesn't just look weird (like naming it `rapidsmpf::CallbackId`, being _too_ generic). Since they're different _types_, I think that's ok at least for now. I'm gonna merge this but happy to revisit next week if we have a better concrete idea.

pentschev: Ugh, you're not part of @rapidsai/rapidsmpf-cpp-codeowners . :disappointed:

bdice (approved):

bdice (approved):

PR #1102, opened 2026-06-15
Fix CudaEvent move assignment

fallintoplace:
## Summary

Fix `CudaEvent::operator=(CudaEvent&&)` so it takes ownership of the source CUDA event handle before clearing the moved-from object.

The move assignment operator kept the existing empty-destination guard, but after the completion-cache removal it only nulled `other.event_`. That left the destination empty and dropped the source handle.

This also adds a focused regression test that moves into an emptied `CudaEvent` and checks that the handle transfers to the destination and is cleared from the source.

## Validation

- `git diff --check`

### DIFF
--- a/cpp/src/cuda_event.cpp
+++ b/cpp/src/cuda_event.cpp
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <utility>
+
 #include <rapidsmpf/cuda_event.hpp>
 
 namespace rapidsmpf {
@@ -36,7 +38,7 @@ CudaEvent& CudaEvent::operator=(CudaEvent&& other) {
             "cannot move into an already-initialized CudaEvent",
             std::invalid_argument
         );
-        other.event_ = nullptr;
+        event_ = std::exchange(other.event_, nullptr);
     }
     return *this;
 }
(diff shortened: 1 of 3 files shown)


pentschev:
@fallintoplace thanks for contributing. It seems there are some style errors could you fix them? I saw you opened a few more PRs, and generally throughout RAPIDS you should be able to get them all fixed with:

```
pre-commit install .  # After this, `git commit` should also do that for you on _new_ changes
pre-commit run -a
```

cpp/src/cuda_event.cpp
```diff
@@ -36,7 +38,7 @@ CudaEvent& CudaEvent::operator=(CudaEvent&& other) {
             "cannot move into an already-initialized CudaEvent",
             std::invalid_argument
         );
```
pentschev, line 40: I know this is not about your change, but since you're here would you mind documenting this behavior as well? We document in [cuda_event.hpp](https://github.com/rapidsai/rapidsmpf/blob/20af487cb6720cabc3c016eae75bc2d2e2ec4a77/cpp/include/rapidsmpf/cuda_event.hpp#L80-L86) the move assignment generically, but the implementation enforces that the destination must already be empty and throws otherwise. Ideally, the public docs should mention the precondition and `std::invalid_argument`.

pentschev (approved):
This looks good to me, thanks for contributing, @fallintoplace !

PR #1103, opened 2026-06-15
Wait for UCXX listener addresses before distribution

fallintoplace:
## Summary

Fix UCXX root barrier setup so listener-address distribution only starts after the root has both the endpoint and listener address for every rank.

In the `HostPortPair` path, endpoint registration and listener-address registration are separate: `listener_callback()` records the endpoint when a rank connects, and `RegisterListenerAddress` can be handled later. Previously the root barrier waited only for `rank_to_endpoint_` to reach `nranks()`, so `distribute_listener_addresses()` could run with a partial `rank_to_listener_address_`, mark distribution complete, and skip late addresses.

This change makes the root wait for both maps to reach `nranks()`, documents the stronger precondition, and checks the invariants under the existing map mutexes before setting `listener_addresses_distributed_`.

## Validation

- `git diff --check`
- `pre-commit run --files cpp/src/communicator/ucxx.cpp`

### DIFF
--- a/cpp/src/communicator/ucxx.cpp
+++ b/cpp/src/communicator/ucxx.cpp
@@ -445,8 +445,7 @@ class SharedResources {
 
     void barrier() {
         // The root needs to have endpoints to all other ranks to continue.
-        while (rank_ == 0 && rank_endpoint_count() != safe_cast<std::size_t>(nranks()))
-        {
+        while (rank_ == 0 && rank_endpoint_count() != safe_cast<std::size_t>(nranks())) {
             progress_worker();
         }
 
--- a/cpp/src/communicator/ucxx.cpp
+++ b/cpp/src/communicator/ucxx.cpp
@@ -361,9 +361,11 @@ class SharedResources {
         futures_.push_back(std::move(future));
     }
 
-    [[nodiscard]] std::size_t rank_endpoint_count() {
-        std::lock_guard<std::mutex> lock(endpoints_mutex_);
-        return rank_to_endpoint_.size();
+    [[nodiscard]] bool all_barrier_maps_registered() {
+        std::scoped_lock lock(endpoints_mutex_, listener_mutex_);
+        auto const expected_size = safe_cast<std::size_t>(nranks());
+        return rank_to_endpoint_.size() == expected_size
+               && rank_to_listener_address_.size() == expected_size;
     }
 
     [[nodiscard]] RankToEndpointMap rank_to_endpoint_snapshot() {
@@ -380,7 +382,8 @@ class SharedResources {
      * worker may no longer be progressed when a peer attempts lazy endpoint
      * creation.
      *
-     * Must only be called on the root rank, after all ranks have connected.
+     * Must only be called on the root rank, after all ranks have connected and
+     * registered listener addresses.
      */
     void distribute_listener_addresses() {
         RankToEndpointMap rank_to_endpoint;
@@ -390,9 +393,19 @@ class SharedResources {
             if (listener_addresses_distributed_) {
                 return;
             }
-            listener_addresses_distributed_ = true;
+            auto const expected_size = safe_cast<std::size_t>(nranks());
+            RAPIDSMPF_EXPECTS(
+                rank_to_endpoint_.size() == expected_size,
+                "cannot distribute listener addresses before all endpoints are registered"
+            );
+            RAPIDSMPF_EXPECTS(
+                rank_to_listener_address_.size() == expected_size,
+                "cannot distribute listener addresses before all listener addresses are "
+                "registered"
+            );
             rank_to_endpoint = rank_to_endpoint_;
             rank_to_listener_address = rank_to_listener_address_;
+            listener_addresses_distributed_ = true;
         }
 
         // Pack all listener addresses (except the destination's own) into a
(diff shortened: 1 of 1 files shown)


cpp/src/communicator/ucxx.cpp
```diff
@@ -373,25 +383,32 @@ class SharedResources {
...
-        if (listener_addresses_distributed_) {
-            return;
+        RankToEndpointMap rank_to_endpoint;
+        RankToListenerAddressMap rank_to_listener_address;
+        {
+            std::scoped_lock lock(endpoints_mutex_, listener_mutex_);
+            if (listener_addresses_distributed_) {
+                return;
+            }
+            listener_addresses_distributed_ = true;
```
pentschev: I think we need to wait for the listener-address map to be complete before setting this flag and snapshotting the maps. The root barrier currently only waits for `rank_to_endpoint_` to reach `nranks()`, but endpoint registration and listener-address registration are separate for the `HostPortPair` path: `listener_callback()` registers the endpoint when the rank connects, and the `RegisterListenerAddress` control message is handled later. That means this can snapshot a partial `rank_to_listener_address_`, mark distribution as done, and never send the missing addresses.

Could we wait for both `rank_to_endpoint_.size()` and `rank_to_listener_address_.size()` to reach `nranks()` before distributing, and also assert both invariants here while holding both mutexes before setting `listener_addresses_distributed_ = true`? I’d also update the doc comment from "after all ranks have connected" to "after all ranks have connected and registered listener addresses.".

pentschev (changes requested):

cpp/src/communicator/ucxx.cpp
fallintoplace: Thank you for your attention, could you please check again?

cpp/src/communicator/ucxx.cpp
```diff
@@ -427,17 +457,17 @@ class SharedResources {
     }
 
     void barrier() {
-        // The root needs to have endpoints to all other ranks to continue.
-        while (rank_ == 0 && rank_to_endpoint_.size() != safe_cast<std::size_t>(nranks()))
-        {
+        // The root needs endpoints and listener addresses for all ranks to continue.
+        while (rank_ == 0 && !all_barrier_maps_registered()) {
             progress_worker();
         }
```
wence-, line 454: This change is necessary I think, and fixes a race bug.

cpp/src/communicator/ucxx.cpp
```diff
@@ -427,17 +457,17 @@ class SharedResources {
...
 
         if (rank_ == 0) {
+            auto const rank_to_endpoint = rank_to_endpoint_snapshot();
```
wence-: I think this one is not however. For the following reason: we can only reach this line once `all_barrier_maps_registered()` returns true. But, by definition that call can only return true once all other threads have finished modifying the `rank_to_endpoint_` map having registered the relevant endpoints. IOW, as soon as `all_barrier_maps_registered()` returns true, we are guaranteed that there are going to be no more modifications of `rank_to_endpoint_` and hence it is safe to read without the lock because there will be no data races.

Do you agree?

cpp/src/communicator/ucxx.cpp
```diff
@@ -370,28 +382,46 @@ class SharedResources {
...
      */
     void distribute_listener_addresses() {
-        if (listener_addresses_distributed_) {
-            return;
+        RankToEndpointMap rank_to_endpoint;
+        RankToListenerAddressMap rank_to_listener_address;
+        {
+            std::scoped_lock lock(endpoints_mutex_, listener_mutex_);
+            if (listener_addresses_distributed_) {
+                return;
```
wence-, line 387: Here, again, the precondition for this function, I think requires that all modifications of `rank_to_endpoint_` and `rank_to_listener_address_` have been committed ,and so there is no possibility of a data race. What am I missing?

wence- (changes requested):

cpp/src/communicator/ucxx.cpp
fallintoplace: Thank you for your review, I’m happy to simplify this and remove the snapshots later today.

wence-: No I think there is some misunderstanding here. I contend that these changes weren't necessary (for the reasons in https://github.com/rapidsai/rapidsmpf/pull/1103#discussion_r3420220002) but I am not sure. So I asked for a second opinion given that you'd observed this potential race.

It might be that there is a race and my understanding of how the code fits together is wrong, so I would like to understand one way (or the other) to improve overall understanding

fallintoplace: Yes, I agree that after `all_barrier_maps_registered()` returns true, the later direct reads are safe by the protocol, so I removed the snapshots.

Looking at the history, this seems to come from #934 / 33a8511: that change moved listener-address lookup from on-demand `QueryListenerAddress` to pre-distribution during the first barrier, but the root barrier readiness check stayed as the older endpoint-only check.

The ordering hole is in the `HostPortPair` path: root records the endpoint in `listener_callback()`, but the listener address is sent later via `RegisterListenerAddress` after the non-root receives its assigned rank. So the old barrier could see all endpoints, distribute with a partial listener-address map, and mark distribution done.

I think the necessary part is just waiting for both maps, plus checking those invariants before setting `listener_addresses_distributed_`.

PR #1104, opened 2026-06-15
Fix Buffer exclusive access ordering

fallintoplace:
## Summary

Fix `Buffer::exclusive_data_access()` so it acquires the Buffer lock before checking whether the latest stream-ordered write is complete.

Previously the readiness check happened first through `is_latest_write_done()`, then the exclusive lock was acquired. A concurrent `write_access()` could pass its lock check and enqueue/record a newer write in between, allowing exclusive access based on a stale completed event.

The new ordering acquires `lock_` first, checks `latest_write_event_` while the lock blocks other Buffer API access, and releases the lock again if the readiness check fails or throws.

This also adds a regression test that blocks the buffer stream with a host callback, verifies `exclusive_data_access()` rejects the pending write, and confirms the failed attempt does not leave the buffer locked.

## Validation

- `git diff --check`

### DIFF
--- a/cpp/src/memory/buffer.cpp
+++ b/cpp/src/memory/buffer.cpp
@@ -75,8 +75,6 @@ std::byte const* Buffer::data() const {
 }
 
 std::byte* Buffer::exclusive_data_access() {
-    RAPIDSMPF_EXPECTS(is_latest_write_done(), "the latest write isn't done");
-
     bool expected = false;
     RAPIDSMPF_EXPECTS(
         lock_.compare_exchange_strong(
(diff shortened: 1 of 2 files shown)


nirandaperera:
I am not sure about this change. `write_access, is_latest_write_done, exclusive_data_access` methods are not there for concurrency. Docs clearly mention this, 
https://github.com/rapidsai/rapidsmpf/blob/main/cpp/include/rapidsmpf/memory/buffer.hpp#L152-L157

IMO this change gives a false-sense of thread-safety, but in reality `Buffer` is not thread-safe for concurrent modification.

fallintoplace:
Yeah, I think you're right. I was treating the Buffer lock as if it serialized concurrent write/exclusive access, but the docs explicitly avoid that contract, and this change still would not make concurrent modification safe in all interleavings. I'll withdraw this PR rather than imply stronger thread-safety than Buffer provides.

pentschev:
It seems that everyone agrees (including myself) this change is unnecessary and will not change the existing behavior in the intended way, and that no further action should be taken here. Therefore, I'm closing this. Please reopen if there's anything else that needs to be done.

Issue #1074, opened 2026-06-01
Python `BufferResource` lifetime management


madsbk:
## Problem

With #1069, the C++ `BufferResource` lifetime is now maintained at the C++ level. Any downstream consumer holding a deep-copied `any_resource<>` from `br.device_mr` keeps the underlying C++ `BufferResource` alive.

However, this is still insufficient for Python-created `BufferResource`s. The C++ object only *borrows* its stream pool, device MR, pinned MR, and statistics objects. Those resources are actually owned by Python and kept alive through `self._<thing>` attributes on the Python wrapper.

If the Python `BufferResource` wrapper is garbage collected while a downstream object still holds `device_mr`, the C++ `BufferResource` survives, but its borrowed dependencies are destroyed underneath it. The next allocation can therefore trigger a use-after-free.

This is the lifetime issue tracked in #641.

### 1. Stream pool: move ownership into C++

The C++ constructor already supports owning the stream pool through a normal
`shared_ptr` deleter, for example the default:

```c++
std::make_shared<rmm::cuda_stream_pool>(16, ...)
```

from:

```text
cpp/include/rapidsmpf/memory/buffer_resource.hpp:91
```

We should therefore remove the Python-side borrowed stream pool entirely.

Replace the Python constructor argument:

```python
stream_pool: CudaStreamPool
```

with:

```python
stream_pool_size: int
```

and expose streams through:

```python
BufferResource.get_stream()
```

implemented as:

```python
Stream._from_cudaStream_t(view.value(), owner=self)
```

The `owner=self` reference ensures the Python `BufferResource` wrapper stays alive
for the lifetime of any handed-out stream.

### 2. Device MR, pinned MR, and statistics: Python-aware deleter

These resources are user-supplied, so ownership must remain on the Python side.

Wrap each borrowed resource in a `std::shared_ptr` whose deleter holds a strong
Python reference:

* `Py_INCREF` when constructing the control block
* `Py_DECREF` under the GIL when destroying it

The resulting C++ control block then transitively pins the Python wrapper for as
long as any downstream object still holds the associated resource.

Apply this pattern to all three handles passed into the C++ `BufferResource`:

* device MR
* pinned MR
* statistics

---

For some more context, see @pentschev's comment: https://github.com/rapidsai/rapidsmpf/pull/1069#discussion_r3324916912

wence-:
> Device MR, pinned MR, and statistics: Python-aware deleter

[...]

PR #1105, opened 2026-06-17
Share stream pool better between C++ and Python

wence-:
Now that RMM's `CudaStreamPool` stores the C++ object as a `shared_ptr` we can use it directly when constructing the `BufferResource`.

Partially addresses #1074.


cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -475,7 +475,7 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
      *
      * @return Reference to the underlying CUDA stream pool.
```
pentschev: Stale doc, please update since it's not a reference anymore.

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
```diff
@@ -310,8 +275,16 @@ cdef class BufferResource:
...
-        return &deref(self._handle).stream_pool()
+    @property
+    def stream_pool(self):
+        """
+        The stream pool associated with this buffer resource.
+
+        Returns
+        -------
+        An rmm CudaStreamPool.
+        """
```
pentschev, line 286: This removes the `cdef stream_pool(self)` implementation, but it's still declared in: https://github.com/rapidsai/rapidsmpf/blob/da2d09ff56bc93079d5c7c6230579733de168f5b/python/rapidsmpf/rapidsmpf/memory/buffer_resource.pxd#L61

The above should be removed. The return type of `stream_pool()` method declaration should also be updated:

https://github.com/rapidsai/rapidsmpf/blob/da2d09ff56bc93079d5c7c6230579733de168f5b/python/rapidsmpf/rapidsmpf/memory/buffer_resource.pxd#L47

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
```diff
@@ -310,8 +275,16 @@ cdef class BufferResource:
...
-    cdef const cuda_stream_pool* stream_pool(self):
-        return &deref(self._handle).stream_pool()
+    @property
+    def stream_pool(self):
+        """
+        The stream pool associated with this buffer resource.
+
+        Returns
+        -------
+        An rmm CudaStreamPool.
```
pentschev:
```suggestion
        An RMM CudaStreamPool.
```

pentschev: Nit: probably capitalize, since this looks more like describing the library, rather than the module's name?

pentschev (changes requested):

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
wence-: Thanks, not sure how I screwed that up.

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
wence-: Done.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
wence-: Done.

wence-:
Needs new RMM packages to appear.

pentschev (approved):
LGTM.

madsbk (approved):
Nice

PR #1106, opened 2026-06-22
 Add a strict Buffer Resource back-reference to Host and Pinned memory resources

### LINKED ISSUE #1070: Make `BufferResource::host_mr()` and `pinned_mr()` return owning MRs

nirandaperera:
`HostMemoryResource` and `PinnedMemoryResource` are now **`BufferResource`-only** resources that carry a strict `BackRefMixin<BufferResource>` back-reference. When a copy of one of these resources is made (e.g. when CCCL promotes `host_mr()` / `pinned_mr()` into an owning `cuda::mr::any_resource` inside a `HostBuffer`), the copy promotes the stored `weak_ptr` to a `shared_ptr`, keeping the owning `BufferResource` alive for as long as any derived buffer lives. This closes a lifetime gap where a `Buffer` could outlive the `BufferResource` that produced its memory resource.

The mixin is *strict*: copying an instance without an installed back-reference throws `std::bad_weak_ptr`. To make that safe, these resources can no longer be constructed as standalone objects — they are created only by `BufferResource`, which installs the back-reference immediately after construction. This mirrors how `RmmResourceAdaptor` already works on this branch.

## C++ changes

- **`HostMemoryResource`**: inherits `BackRefMixin<BufferResource>`; constructor is private with `friend class BufferResource`. Equality remains stateless (`always true`) since instances are interchangeable and the back-reference is installed exactly once.
- **`PinnedMemoryResource`**: inherits `BackRefMixin<BufferResource>` as a second base; constructor is private (`friend class BufferResource`). The `make_if_available` / `from_options` factories were **removed** so no back-reference-less instance can ever escape. Equality compares the shared pool state only.
- **`PinnedPoolProperties`**: gained a `numa_id` field (defaults to the calling thread's NUMA node) so all pinned configuration flows as one struct. Added a free helper `pinned_pool_properties_from_options(...)`.
- **`PinnedMemoryDisabled`**: new `constexpr std::nullopt_t` sentinel used to disable pinned host memory (replaces the old `PinnedMemoryResource::Disabled`).
- **`BufferResource::create()` / `from_options()`**: now take `std::optional<PinnedPoolProperties> pinned_pool_properties` (default `PinnedMemoryDisabled`) instead of a pre-built resource. The pinned resource is constructed internally and the back-reference is installed on `owning_mr_`, `host_mr_`, and `pinned_mr_` before `create()` returns.
- **`try_pinned_mr()`**: now returns `std::optional<PinnedMemoryResource>` (a back-referenced handle) so Python/Statistics get a concrete, lifetime-safe handle.

## Python changes

- **`PinnedMemoryResource`**: now an opaque, non-constructible handle (`__init__` raises `TypeError`); obtained via `BufferResource.pinned_mr`.
- **`PinnedPoolProperties`**: new `@dataclass` (`initial_pool_size`, `max_pool_size`, `numa_id`) used to configure pinned memory.
- **`BufferResource(...)`**: replaces the old `pinned_mr=` argument with `pinned_pool_properties: PinnedPoolProperties | None = None` (`None` disables pinned host memory). `from_options` derives this from the config options.
- **`Statistics.report(pinned_mr=...)`**: unchanged signature; the handle is now sourced from `BufferResource.pinned_mr`.

## Tests

- Added `HostMrKeepsBufferResourceAlive` and `PinnedMrKeepsBufferResourceAlive` regression tests mirroring `DeviceMrKeepsBufferResourceAlive`.
- Routed all C++ tests/benchmarks through a `BufferResource` (no standalone Host/Pinned MR construction); reworked `test_host_buffer`, `test_config`, `test_memory_resources`, and `bench_memory_resources`.
- Updated Python `test_config.py` to configure pinned memory via `BufferResource` and assert on `BufferResource.pinned_mr`.

## Breaking changes

- C++: `HostMemoryResource` / `PinnedMemoryResource` can no longer be constructed directly; `PinnedMemoryResource::make_if_available` / `from_options` and the `Disabled` sentinel are removed. `BufferResource::create()` takes `PinnedPoolProperties` instead of a `PinnedMemoryResource`.
- Python: `PinnedMemoryResource(...)` is no longer constructible; the `BufferResource(pinned_mr=...)` argument is replaced by `pinned_pool_properties=`.

Depends on #1078

Closes #1070


python/rapidsmpf/rapidsmpf/memory/pinned_memory_resource.pyx
```diff
@@ -36,42 +48,73 @@ cpdef bool_t is_pinned_memory_resources_supported():
...
+    :attr:`~rapidsmpf.memory.buffer_resource.BufferResource.pinned_mr`.
+
+    The returned handle holds shared ownership of its owning ``BufferResource``,
+    so it (and any copy of it) keeps the ``BufferResource`` alive.
+    """
+    def __init__(self, *args, **kwargs):
+        raise TypeError(
+            "PinnedMemoryResource cannot be constructed directly; configure pinned "
+            "memory on a BufferResource (e.g. "
+            "`BufferResource(mr, pinned_pool=PinnedPoolProperties())`) and obtain it "
```
madsbk:
```suggestion
            "`BufferResource(mr, pinned_pool_properties=PinnedPoolProperties())`) and obtain it "
```

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -217,6 +219,12 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
      * @brief Get the RMM host memory resource.
      *
      * @return Reference to the RMM resource used for host allocations.
+     *
+     * @note The returned non-owning `resource_ref` does not by itself keep this
+     * `BufferResource` alive, but the underlying `HostMemoryResource` carries a
+     * back-reference: promoting the ref to an owning `cuda::mr::any_resource` (as
+     * RMM/CCCL containers do internally) keeps this `BufferResource` alive for the
+     * lifetime of that owning copy.
```
madsbk: I find this a bit confusing. The key point is that the returned reference is
non-owning. To keep the `BufferResource` alive, the caller must upgrade it to an
owning reference, right?

What about referring to the `@par CCCL's lifetime semantics` section in the
`device_mr()` docstring and only describing any differences here?

Something like:

```suggestion
     *
     * @note Lifetime semantics are identical to `device_mr()`. See its
     * `@par CCCL lifetime semantics` section for details. In brief, the returned
     * `resource_ref` is non-owning. Promote it to a
     * `cuda::mr::any_resource<cuda::mr::host_accessible>` to extend the
     * `BufferResource` lifetime.
```

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -225,16 +233,24 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
      *
      * @throws std::invalid_argument if no pinned memory resource is available.
      * @return Reference to the RMM resource used for pinned host allocations.
+     *
+     * @note The returned non-owning `resource_ref` does not by itself keep this
+     * `BufferResource` alive, but the underlying `PinnedMemoryResource` carries a
+     * back-reference: promoting the ref to an owning `cuda::mr::any_resource` (as
+     * RMM/CCCL containers do internally) keeps this `BufferResource` alive for the
+     * lifetime of that owning copy.
```
madsbk: The same here

cpp/include/rapidsmpf/memory/host_memory_resource.hpp
```diff
@@ -28,10 +31,14 @@ namespace rapidsmpf {
  * region. THP can improve device-host memory transfer performance for large
  * buffers. The hint is applied via `madvise(MADV_HUGEPAGE)` and may be ignored
  * by the kernel depending on system configuration or resource availability.
+ *
+ * @note Instances are constructed only by `BufferResource`, which installs a
+ * back-reference (via `BackRefMixin<BufferResource>`) so that any copy of the
+ * resource keeps its owning `BufferResource` alive. There is no public
+ * constructor; obtain the resource through a `BufferResource`.
```
madsbk:
```suggestion
```

cpp/include/rapidsmpf/memory/pinned_memory_resource.hpp
```diff
@@ -88,50 +126,21 @@ struct PinnedPoolProperties {
  * This resource allocates and deallocates pinned host memory asynchronously through
  * CUDA streams. It offers higher bandwidth and lower latency for device transfers
  * compared to regular pageable host memory.
+ *
+ * @note Instances are constructed only by `BufferResource`, which installs a
+ * back-reference (via `BackRefMixin<BufferResource>`) so that any copy of the
+ * resource keeps its owning `BufferResource` alive. There is no public
+ * constructor; obtain the resource through a `BufferResource`. Use the free
+ * function `is_pinned_memory_resources_supported()` to query availability.
```
madsbk:
```suggestion
```

madsbk (commented):
Overall looks good

cpp/benchmarks/bench_memory_resources.cpp
```diff
@@ -96,6 +99,22 @@ class NewDelete {
     friend void get_property(NewDelete const&, cuda::mr::host_accessible) noexcept {}
 };
 
+// Build a lightweight `BufferResource` that owns a pinned pool with the given
+// properties. Host and pinned memory resources are constructible only by a
+// `BufferResource`. The dedicated spill-check thread is disabled and a
+// single-stream pool is used to keep overhead minimal for micro-benchmarks.
```
wence-: nit: This name kind of doesn't make sense because the buffer resource is not pinned.

Also, what is "lightweight" about this buffer resource?

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -225,16 +233,24 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
...
 
     /**
      * @brief Get the pinned host memory resource if available.
      *
-     * @return The pinned host memory resource as an `any_resource`, or `std::nullopt` if
-     * pinned host memory is not available.
+     * @return A back-referenced copy of the pinned host memory resource, or
+     * `std::nullopt` if pinned host memory is not available. The returned handle
+     * keeps this `BufferResource` alive for as long as the handle (or any copy of
+     * it) lives.
```
wence-: *nit*: Back-referenced is an implementation detail and back-referenced to what?

Why not just:

```
The `PinnedMemoryResource` is available, or `std::nullopt` if pinned host memory is not available. The returned handle keeps this `BufferResource` alive as long as the handle (or any copy) exists.
```

cpp/include/rapidsmpf/memory/host_memory_resource.hpp
```diff
@@ -112,7 +119,9 @@ class HostMemoryResource {
      * @brief Compares this resource to another resource.
      *
      * All instances are stateless and interchangeable, so this always returns
-     * true.
+     * true. Instances can only be created by `BufferResource` (which installs
+     * the back-reference exactly once), so the back-reference does not affect
+     * equality.
```
wence-: *nit*: This implementation detail doesn't have any effect on the observable behaviour of the equality comparison, so we can remove it.


```suggestion
     * true.
```

cpp/include/rapidsmpf/memory/pinned_memory_resource.hpp
```diff
@@ -64,6 +68,16 @@ inline bool is_pinned_memory_resources_supported() {
     return supported;
 }
 
+/**
+ * @brief Sentinel used to disable pinned host memory.
+ *
+ * Pass this in place of a `PinnedPoolProperties` (e.g. to `BufferResource::create()`)
+ * to disable pinned host memory allocations. It is simply `std::nullopt` typed as
+ * `std::nullopt_t`, so it implicitly converts to an empty
+ * `std::optional<PinnedPoolProperties>`.
```
wence-:
```suggestion
 * Pass this in place of a `PinnedPoolProperties` (e.g. to `BufferResource::create()`)
 * to disable pinned host memory allocations.
```

cpp/include/rapidsmpf/memory/pinned_memory_resource.hpp
```diff
@@ -172,6 +181,10 @@ class PinnedMemoryResource final
     /**
      * @brief Equality comparison.
      *
+     * Instances can only be created by `BufferResource` (which installs the
+     * back-reference exactly once), so equality is determined solely by the
+     * underlying shared state.
+     *
```
wence-:
```suggestion
```

cpp/include/rapidsmpf/memory/pinned_memory_resource.hpp
```diff
@@ -217,16 +230,17 @@ class PinnedMemoryResource final
     /**
      * @brief Construct a pinned (page-locked) host memory resource.
      *
-     * Private — use `make_if_available` or `from_options` to obtain an instance.
+     * Private — only `BufferResource` constructs instances (it installs the
+     * back-reference immediately after construction).
      *
```
wence-:
```suggestion
     * Private: use `BufferResource` to construct instances.
     *
```

cpp/src/memory/buffer_resource.cpp
```diff
@@ -50,12 +50,21 @@ BufferResource::BufferResource(
...
+    std::optional<PinnedPoolProperties> pinned_pool_properties,
     std::unordered_map<MemoryType, std::int64_t> memory_limits,
     std::optional<Duration> periodic_spill_check,
     std::shared_ptr<rmm::cuda_stream_pool> stream_pool,
     std::shared_ptr<Statistics> statistics
 ) {
+    // Construct the pinned resource in place (`BufferResource` is a friend of
+    // `PinnedMemoryResource`), so no observable, back-reference-less instance is
+    // ever minted. Only create it when pinned memory is both requested and
+    // supported on this system.
```
wence-:
```suggestion
```

cpp/src/memory/pinned_memory_resource.cpp
```diff
@@ -41,57 +41,47 @@ cuda::memory_pool_properties get_memory_pool_properties(
...
               is_pinned_memory_resources_supported(),
               "Pinned host memory is not supported on this system. "
               "CUDA " RAPIDSMPF_PINNED_MEM_RES_MIN_CUDA_VERSION_STR
               " is one of the requirements, but additional platform or driver "
-              "constraints may apply. If needed, use `PinnedMemoryResource::Disabled` "
-              "to disable pinned host memory, noting that this may significantly "
-              "degrade spilling performance.",
+              "constraints may apply. If needed, disable pinned host memory (pass "
+              "`std::nullopt` for the pinned pool to `BufferResource`), noting that "
+              "this may significantly degrade spilling performance.",
```
wence-: `PinnedMemoryResource::Disabled` still exists, right? Or something similar?

Also, this suggestion is still wrong because we can't pass a pinned pool to `BufferResource` any more, only pool properties.

cpp/tests/test_memory_resources.cpp
```diff
@@ -11,28 +11,40 @@
...
 #include <rapidsmpf/memory/pinned_memory_resource.hpp>
 #include <rapidsmpf/memory/resource_types.hpp>
 
 namespace {
 
+// `HostMemoryResource` and `PinnedMemoryResource` are constructible only by a
+// `BufferResource`. The `any_resource` copies returned below each carry a
+// back-reference that keeps the (otherwise local) `BufferResource` alive, so no
+// external anchor is needed.
 std::vector<cuda::mr::any_resource<cuda::mr::host_accessible>> make_host_resources() {
```
wence-: All of these comments about back references everywhere should be removed. The docstring of the relevant method is where we should talk about ownership.

These kind of inline explanations only ever go out of date

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
```diff
@@ -329,12 +357,20 @@ cdef class BufferResource:
...
         The pinned host memory resource, or None if pinned host allocations
         are disabled.
         """
-        return self._pinned_mr
+        cdef optional[cpp_PinnedMemoryResource] opt
+        with nogil:
+            opt = deref(self._handle).try_pinned_mr()
+        if not opt.has_value():
+            return None
+        return PinnedMemoryResource._from_cpp(opt.value())
```
wence-: Everywhere else in rapidsmpf this method is called `from_handle`.

wence- (dismissed):
Code changes broadly look good, but please clean up the docstrings and comments throughout.

In general, please try and elide as much implementation detail in comments and, if it is necessary, have it only in one place rather than everywhere. Docstrings and comments should generally talk about observable behaviour and semantics, not implementation specifics.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
nirandaperera: Yes, that is correct.

cpp/src/memory/pinned_memory_resource.cpp
nirandaperera: `PinnedMemoryResource::Disabled` is replaced by `PinnedMemoryDisabled`. Let me fix the exception string.

cpp/benchmarks/bench_memory_resources.cpp
nirandaperera: I'll fix this comment

madsbk (approved):
Looks good. The only thing we should consider is the enablement semantics. Currently, setting `pinned_pool_properties` to a non-`None` value does not guarantee that pinned memory is enabled, since that still depends on system support. 

It would be helpful to have an option that means: enable pinned memory, or raise an error if it is unavailable.

Alternatively, this option could be implemented at the cudf-polars level. After creating the memory resource, cudf-polars could verify that pinned memory was actually enabled and raise an error otherwise.

nirandaperera:
> Looks good. The only thing we should consider is the enablement semantics. Currently, setting `pinned_pool_properties` to a non-`None` value does not guarantee that pinned memory is enabled, since that still depends on system support.
> 
> It would be helpful to have an option that means: enable pinned memory, or raise an error if it is unavailable.
> 
> Alternatively, this option could be implemented at the cudf-polars level. After creating the memory resource, cudf-polars could verify that pinned memory was actually enabled and raise an error otherwise.

I made `BufferResource::create` throw, if `PinnedProperties` provided and pinned memory is not supported

cpp/tests/test_memory_resources.cpp
nirandaperera: removed

PR #1113, opened 2026-06-26
Reject out-of-range reservation sizes in `reserve_memory`

madsbk:
## Summary

The streaming memory reservation path could fail with an opaque C++ cast error from deep inside `BufferResource::reserve`:

```
RuntimeError: RapidsMPF cast error at:
  cpp/src/memory/buffer_resource.cpp:165, value out of range (value=18446744062495230272)
```

Values larger than `INT64_MAX` cannot represent real allocation requests. In practice, they almost always result from an unsigned (`size_t`) subtraction that underflowed below zero and wrapped around. The existing `safe_cast<std::int64_t>` correctly rejects these values, but the error surfaces in C++ with no Python context, making it difficult to identify the offending caller.

This PR adds a shared `check_reservation_size` helper and calls it from every Python API that accepts a reservation size, including `reserve_memory`, `MemoryReserveOrWait`, and `BufferResource`. 

Behavior is otherwise unchanged. The new check rejects exactly the same values that the downstream `safe_cast<std::int64_t>` already rejected, but with a much more actionable diagnostic.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
+++ b/python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
@@ -14,6 +14,8 @@ from libcpp.vector cimport vector
 
 from rmm.pylibrmm import CudaStreamFlags
 
+from rapidsmpf.utils.memory import check_reservation_size
+
 from rmm.librmm.memory_resource cimport (any_resource, device_accessible,
                                          device_async_resource_ref,
                                          make_any_device_resource)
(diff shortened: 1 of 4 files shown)


wence- (approved):
OK, though I wish we could tie together the C++ and python backtraces better.

TomAugspurger (approved):
Makes sense to me. My only worry about putting it here is that we need to remember to call it from everywhere that makes a reservation, but I'm not aware of a way to avoid that.

madsbk:
> OK, though I wish we could tie together the C++ and python backtraces better.

When we move to C++23, we might be able to use `std::stacktrace::current()` for a better UX.

PR #1118, opened 2026-07-02
Add send/receive payload tracking for non-shuffle collectives

wence-:
Now that the communicator carries a Statistics object, we can plumb through basic communication volume statistics for the remaining collectives.

Additionally, consistently take the statistics object from the communicator's progress thread when recording statistics for an object with a communicator.


python/rapidsmpf/rapidsmpf/statistics.pxd
```diff
@@ -60,6 +60,10 @@ cdef extern from "<rapidsmpf/statistics.hpp>" nogil:
 cdef class Statistics:
     cdef shared_ptr[cpp_Statistics] _handle
 
+    @staticmethod
+    cdef Statistics from_handle(shared_ptr[cpp_Statistics] handle)
```
wence-, line 66: This is so I can get statistics out of the progress thread without having to store the matching python class on construction.

cpp/src/coll/allgather.cpp
```diff
@@ -223,6 +224,11 @@ ProgressThread::ProgressState AllGather::event_loop() {
                 // of insertions from that rank.
                 mark_finish(chunk->sequence());
             } else {
+                if (chunk->data_size() > 0) {
+                    statistics->add_bytes_stat(
+                        "allgather-payload-send", chunk->data_size()
```
wence-, line 229: question (for all the collectives): Do we want to stratify the send/receive data by `op_id` (so that we can measure per collective). Or are we OK with this level of coarse-grained data and more detailed stuff should come via tracing?

TomAugspurger: > this level of coarse-grained data and more detailed stuff should come via tracing?

That's my preference.

cpp/src/shuffler/shuffler.cpp
```diff
@@ -89,7 +89,7 @@ class Shuffler::Progress {
         RAPIDSMPF_NVTX_SCOPED_RANGE_VERBOSE("Shuffler.Progress", p_iters++);
         auto const t0_event_loop = Clock::now();
 
-        auto& stats = *shuffler_.statistics_;
+        auto const& stats = shuffler_.comm_->progress_thread()->statistics();
```
pentschev, line 90: Now that stats is the progress thread’s statistics object, both this function and `ProgressThread::event_loop()` record `event-loop-total` into the same accumulator. This adds two differently scoped samples per shuffler iteration. Could we rename the shuffler timer (for example, `shuffler-event-loop-total`) or remove one of the recordings, and update the documentation accordingly?

cpp/src/shuffler/shuffler.cpp
```diff
@@ -328,10 +327,6 @@ void Shuffler::insert(detail::Chunk&& chunk) {
     }
     if (dst_rank == comm_->rank()) {
         // this is a local chunk, so we can move it straight to received.
-        if (chunk.is_data_buffer_set()) {
-            statistics_->add_bytes_stat("shuffle-payload-send", chunk.data_size());
-            statistics_->add_bytes_stat("shuffle-payload-recv", chunk.data_size());
-        }
         insert_into_received(std::move(chunk));
```
pentschev, line 317: Removing the counters from the local path changes `shuffle-payload-{send,recv}` to exclude self-transfers, while https://github.com/rapidsai/rapidsmpf/blob/main/docs/source/statistics.md explicitly says they are included. Since the new test expects `nranks() - 1`, could we update those descriptions to document the new semantics?

cpp/src/coll/allgather.cpp
```diff
@@ -223,6 +224,11 @@ ProgressThread::ProgressState AllGather::event_loop() {
...
+                    );
```
pentschev, line 230: Could we add the new `allgather-payload-{send,recv}` (and `allreduce-payload-{send,recv}`, and `sparsealltoall-payload-{send,recv}) entries to https://github.com/rapidsai/rapidsmpf/blob/main/docs/source/statistics.md?

cpp/src/coll/allgather.cpp
pentschev: I also think the coarse-grained data is enough for now. We can always change that if we have a need for it.

nirandaperera: But I thought the general comment in [this PR](https://github.com/rapidsai/rapidsmpf/pull/999) was to remove per-operation stats and collect only total send/ recv stats. 🤔  (I am sorry it got pushed back due to several refactors)

nirandaperera: I think I have a better idea about the reasoning behind this PR after attending the cudf meeting today. 

So, maybe this should supersede https://github.com/rapidsai/rapidsmpf/pull/999?

nirandaperera: Also shouldnt we suffix the stats key with `op_id`? There could be several collective operations in a single stat context, isnt it?

cpp/tests/environment.hpp
```diff
@@ -4,10 +4,40 @@
...
 
+#include <utility>
+
 #include <gtest/gtest.h>
 #include <mpi.h>
 
 #include <rapidsmpf/communicator/communicator.hpp>
+#include <rapidsmpf/statistics.hpp>
+
+class ClearedStatistics {
```
nirandaperera, line 20: Let's add a simple docstring here. 
```suggestion
/// @brief RAII struct that enables stats and restores the original state (enabled/ disabled) of stats upon destruction. 
class ClearedStatistics {
```

cpp/tests/test_shuffler.cpp
```diff
@@ -364,6 +364,53 @@ TEST_P(MemoryLimits_NumPartition, round_trip) {
...
+        chunks.emplace(
+            pid, generate_packed_data(n_elements, offset, rmm::cuda_stream_default, *br)
+        );
+    }
+    shuffler.insert(std::move(chunks));
+    shuffler.insert_finished();
+    shuffler.wait(std::chrono::seconds{30});
+
+    auto const local_pid =
+        rapidsmpf::safe_cast<rapidsmpf::shuffler::PartID>(comm->rank());
```
nirandaperera, line 391: shall we make `total_num_partitions` arbitrary and use `shuffler.local_partitions()` to get local_pids?

nirandaperera (commented):
LGTM. Thanks @wence-

cpp/src/shuffler/shuffler.cpp
wence-: thanks, will do

cpp/src/coll/allgather.cpp
wence-: yes

python/rapidsmpf/rapidsmpf/statistics.pyx
```diff
@@ -164,6 +164,12 @@ cdef class Statistics:
         with nogil:
             self._handle = cpp_create(enable)
 
+    @staticmethod
+    cdef Statistics from_handle(shared_ptr[cpp_Statistics] handle):
+        cdef Statistics obj = Statistics.__new__()
```
madsbk:
```suggestion
        cdef Statistics obj = Statistics.__new__(Statistics)
```

I guess we never test this path?

madsbk (changes requested):
Looks good beside a missing argument

cpp/src/coll/allgather.cpp
wence-: As per Tom's comment, and I think I agree, more granular data should be via tracing.

cpp/src/coll/allgather.cpp
wence-: done.

cpp/src/shuffler/shuffler.cpp
wence-: Done.

python/rapidsmpf/rapidsmpf/statistics.pyx
wence-: Thanks, added a test and fixed.

madsbk (approved):

cpp/src/shuffler/shuffler.cpp
pentschev: This seems to still be true.

wence-: Done. I also removed some eventloop timing stats from the shuffle. Again, I think if we want detail here we should do it via tracing.

pentschev: Seems reasonable, thanks.

pentschev (approved):

PR #1131, opened 2026-07-16
Ensure spawned libcoro tasks complete when cancelling

wence-:
If we are awaiting a libcoro task and are cancelled on the Python side, we currently just cancel the Python future leading to an asyncio warning when the C++ callback eventually fires to set a result. Furthermore, we might not end up waking C++ waiters if, for example, Python cancellation fails to successfully shut down a channel.

To mitigate against this for most awaitables that bridge from Python into C++, introduce new utilities to shutdown channels, and await a C++ spawned task, handling cancellation, and optionally running some cleanup before raising the cancellation error.

### DIFF
--- a/python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
+++ b/python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
@@ -225,6 +225,27 @@ def define_actor(*, extra_channels=()):
     return partial(decorate_actor, extra_channels)
 
 
+async def run_and_publish_task(coro, task_ready):
+    """
+    Run a coroutine and expose its event loop and task to another thread.
+
+    Parameters
+    ----------
+    coro
+        Coroutine to run.
+    task_ready
+        Rendezvous point shared with calling thread. Receives the running
+        event loop and cancellable task before ``coro`` starts, allowing
+        the caller to cancel the task from outside the event loop.
+
+    Returns
+    -------
+    Result returned by ``coro``.
+    """
+    task_ready.set_result((asyncio.get_running_loop(), asyncio.current_task()))
+    return await coro
+
+
 def sync_wait(coro, task_ready):
     """
     Run, and wait for completion of, a coroutine.
@@ -240,10 +261,7 @@ def sync_wait(coro, task_ready):
     """
     try:
         with asyncio.Runner() as runner:
-            loop = runner.get_loop()
-            task = loop.create_task(coro)
-            task_ready.set_result((loop, task))
-            return runner.run(task)
+            return runner.run(run_and_publish_task(coro, task_ready))
     except BaseException as error:
         if not task_ready.done():
             task_ready.set_exception(error)
--- a/python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
+++ b/python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 from cpython.object cimport PyObject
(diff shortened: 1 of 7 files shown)


TomAugspurger (approved):

python/rapidsmpf/rapidsmpf/streaming/core/cancellation.pyi
```diff
@@ -0,0 +1,10 @@
+import asyncio
```
madsbk, line 4:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.
# SPDX-License-Identifier: Apache-2.0
import asyncio
```

python/rapidsmpf/rapidsmpf/streaming/core/cancellation.pyx
```diff
@@ -0,0 +1,77 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
madsbk:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
```diff
@@ -224,7 +225,26 @@ def define_actor(*, extra_channels=()):
...
+    coro
+        Thing to await, must be a coroutine.
+    task_ready
+        concurrent.futures.Future representing the task we're monitoring the status of.
+
+    Returns
+    -------
+    Result of awaiting the coroutine.
+    """
+    task_ready.set_result((asyncio.get_running_loop(), task_ready))
```
madsbk: We want to set the running task not the Future itself, right?
```suggestion
    task_ready.set_result((asyncio.get_running_loop(), asyncio.current_task()))
```

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
```diff
@@ -224,7 +225,26 @@ def define_actor(*, extra_channels=()):
...
+    ----------
+    coro
+        Thing to await, must be a coroutine.
+    task_ready
+        concurrent.futures.Future representing the task we're monitoring the status of.
+
+    Returns
+    -------
+    Result of awaiting the coroutine.
+    """
```
madsbk, line 244: What about?
```suggestion
    """
    Run a coroutine and expose its event loop and task to another thread.

    Parameters
    ----------
    coro
        Coroutine to run.
    task_ready
        Rendezvous point shared with the calling thread. Receives the running
        event loop and cancellable task before ``coro`` starts, allowing the
        caller to cancel the task from outside the event loop.

    Returns
    -------
    Result returned by ``coro``.
    """
```

madsbk (changes requested):
Overall looks very good

python/rapidsmpf/rapidsmpf/streaming/core/actor.pyx
wence-: Done.

madsbk (approved):

python/rapidsmpf/rapidsmpf/streaming/core/cancellation.pyx
```diff
@@ -0,0 +1,77 @@
...
+    Await a C++ future handling cancellation.
+
+    Parameters
+    ----------
+    future
+        Awaitable future returning None bridging into libcoro.
+    on_cancel
+        Optional callback to run if cancellation is raised while awaiting
+        the future. Must be a zero-argument function that returns an
+        awaitable.
```
vyasr, line 47: Probably should document that callback will be run before the future, no? Otherwise on_cancel could deadlock if it's set up to be ordered-after future finishes.

vyasr (approved):
One suggestion, otherwise LGTM.

python/rapidsmpf/rapidsmpf/streaming/core/cancellation.pyx
wence-: It runs after the future is submitted

jameslamb (approved):

PR #1133, opened 2026-07-17
Ensure spawned libcoro tasks complete when cancelling

wence-:
If we are awaiting a libcoro task and are cancelled on the Python side, we currently just cancel the Python future leading to an asyncio warning when the C++ callback eventually fires to set a result. Furthermore, we might not end up waking C++ waiters if, for example, Python cancellation fails to successfully shut down a channel.

To mitigate against this for most awaitables that bridge from Python into C++, introduce new utilities to shutdown channels, and await a C++ spawned task, handling cancellation, and optionally running some cleanup before raising the cancellation error.

Backport of #1131 onto 26.08

### DIFF
--- a/python/rapidsmpf/rapidsmpf/streaming/core/cancellation.pyi
+++ b/python/rapidsmpf/rapidsmpf/streaming/core/cancellation.pyi
@@ -9,5 +9,7 @@ from rapidsmpf.streaming.core.context import Context
 
 async def shutdown_channels(ctx: Context, *chs: Channel) -> None: ...
 async def await_cpp_future(
-    future: asyncio.Future[None], *, on_cancel: Callable[[], Awaitable[None]] | None
+    future: asyncio.Future[None],
+    *,
+    on_cancel: Callable[[], Awaitable[None]] | None = None,
 ) -> None: ...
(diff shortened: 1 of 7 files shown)


python/rapidsmpf/rapidsmpf/streaming/core/cancellation.pyi
```diff
@@ -0,0 +1,13 @@
...
+
+import asyncio
+from collections.abc import Awaitable, Callable
+
+from rapidsmpf.streaming.core.channel import Channel
+from rapidsmpf.streaming.core.context import Context
+
+async def shutdown_channels(ctx: Context, *chs: Channel) -> None: ...
+async def await_cpp_future(
+    future: asyncio.Future[None], *, on_cancel: Callable[[], Awaitable[None]] | None
```
madsbk:
```suggestion
    future: asyncio.Future[None], *, on_cancel: Callable[[], Awaitable[None]] | None = None
```

madsbk: We should also fix this in `main`

madsbk (approved):

python/rapidsmpf/rapidsmpf/streaming/core/cancellation.pyi
wence-: If we merge it here the automerger should make the forward merge work.

TomAugspurger (approved):
Looks good once the precommit formatting issue is fixed.

bdice (approved):
Approving backport.

PR #1015, opened 2026-05-06
Add support for Ray in Python 3.14 and wheels tests

pentschev:
There are Ray packages for Python 3.14 https://github.com/conda-forge/ray-packages-feedstock/pull/277 and thus it's now possible to enable it for all Python version. Tentatively enable also for wheels tests.

Closes https://github.com/rapidsai/rapidsmpf/issues/897

Issue #897, opened 2026-03-03
fix(ray): re-enable Ray testing on Python 3.14 once wheels + conda packages are available


gforsyth:
In #881 I am adding support for Python 3.14, as part of rapidsai/build-planning#205

`ray` does not yet have wheels or conda packages for Python 3.14.  While `ray` is used in the test suite and is supported by `rapidsmpf`, it is not itself a hard dependency of `rapidsmpf`, so we are ok with skipping a few tests to get partial Python 3.14 support out there.

Tracking link for Python 3.14 support in `ray`: https://github.com/ray-project/ray/issues/56434

Also there are currently two feedstock dependencies of `ray` that also haven't been built for Python 3.14 yet:

  * https://github.com/conda-forge/mujoco-feedstock/pull/116
  * https://github.com/conda-forge/pygame-feedstock/pull/26

PR #1136, opened 2026-07-20
Fix docs build on 26.10

wence-:
Ray packages now exist for python 3.14 so we can remove the TODO note and do what it tells us.


jameslamb (approved):
Ah great!

jameslamb (changes requested):
The issue in the comment (#897) is already closed, guess this reference was just missed in #1015

I double-checked and there's one more reference to update in `build.yaml`. I'll push a commit for that.

jameslamb (approved):
`build.yaml` needed this same change. I've pushed that.

That won't at all affect PR CI, so admin-merging this to get PR CI unblocked.

PR #1141, opened 2026-07-24
test: isolate library wheel smoke test

bdice:
## Description

Install the library wheel in a clean virtual environment and call its `load_library()` entry point before installing the high-level wheel or test extras. This exposes missing library-wheel runtime dependencies that the full test environment can otherwise mask.

xref: https://github.com/rapidsai/build-planning/issues/307


wence- (approved):

wence-:
@bdice I marked this as ready for review because I think it makes sense.

jameslamb (approved):

PR #1142, opened 2026-07-25
Adding reservation aware resource adapter

nirandaperera:
This PR adds `ReservationAwareResourceAdaptor`, whose reservations are themselves RMM memory resources: hand one to cudf and every allocation made through it is charged against the reservation, turning "reserve before allocate" into something the type system and the allocator enforce rather than a convention.

- `MemoryReservation` is a `shared_resource`, so it can be passed anywhere a `rmm::device_async_resource_ref` is accepted. Allocating draws down its balance and throws `rmm::out_of_memory` when the grant is exhausted.
- The adaptor extends `RmmResourceAdaptor`'s allocation tracking with a runtime-adjustable limit and a reserved-bytes counter (follows the limit logic in `BufferResource`)
- **NOTE:** Unspent balance is refunded only when the last reference dies. So, callers need to be cautious when over reserving!.

### Usage

- General usage
```cpp
    ReservationAwareResourceAdaptor adaptor{rmm::mr::cuda_memory_resource{}, 1MB};
    auto res = adaptor.reserve(2KB, AllowOverbooking::NO);

    {
        rmm::device_buffer buf1{1KB, stream, res};
        rmm::device_buffer buf2{1KB, stream, res};
        EXPECT_THROW((rmm::device_buffer buf3{1KB, stream, res}, rmm::out_of_memory); // res is exhausted
    }
   rmm::device_buffer buf4{2KB, stream, res}; // works again because buffers are released
```

- Create another reservation from existing reservation. 
```cpp
  void foo(MemoryReservation res, ...){
    ... 
    auto res2 = res.adapter().reserve(...); 
     ...
  }
```

[Notes](https://docs.google.com/document/d/1on97Sl0l5LQlOLX3SImrW9nTHm4Qtaw1-YBikyudq1I/edit?usp=sharing)


cpp/include/rapidsmpf/detail/reservation_aware_resource_adaptor_impl.hpp
```diff
@@ -0,0 +1,350 @@
...
+    void deallocate(
+        cuda::stream_ref stream,
+        void* ptr,
+        std::size_t bytes,
+        std::size_t alignment = rmm::CUDA_ALLOCATION_ALIGNMENT
+    ) noexcept {
+        auto const padded_bytes =
+            safe_cast<std::int64_t>(rmm::align_up(bytes, alignment));
+        balance_.fetch_add(padded_bytes, std::memory_order_acq_rel);
+        adaptor_->total_reserved_.fetch_add(padded_bytes, std::memory_order_acq_rel);
```
bdice: I would use a method to do this, rather than making `total_reserved_` a public member.

cpp/include/rapidsmpf/detail/reservation_aware_resource_adaptor_impl.hpp
```diff
@@ -0,0 +1,350 @@
...
+     * @param ptr Pointer to the memory to deallocate.
+     * @param bytes Number of bytes to deallocate.
+     * @param alignment Alignment of the original allocation.
+     */
+    void deallocate_sync(
+        void* ptr,
+        std::size_t bytes,
+        std::size_t alignment = rmm::CUDA_ALLOCATION_ALIGNMENT
+    ) noexcept {
+        deallocate(adaptor_->sync_stream_, ptr, bytes, alignment);
```
bdice, line 303: For correctness, this must sync too.

cpp/include/rapidsmpf/detail/reservation_aware_resource_adaptor_impl.hpp
nirandaperera: Its private, but `MemoryReservationImpl<>` has been friended in  `ReservationAwareResourceAdaptorImpl`.

nirandaperera:
@bdice thank you for the review. 
BTW I'd like to know your thoughts about the general direction of this idea, _reserve before allocate_. I think the worse case scenario would be, if there are `N` allocations, we could be creating `N` separate `shared_resource` instances for each reservation.

cpp/include/rapidsmpf/detail/reservation_aware_resource_adaptor_impl.hpp
```diff
@@ -0,0 +1,354 @@
...
+            // Availability *after* the reservation would be made. Negative means the
+            // request overbooks the limit.
+            std::int64_t const headroom =
+                limit() - this->current_allocated() - (reserved + want);
+            std::size_t const overbooking =
+                headroom < 0 ? safe_cast<std::size_t>(-headroom) : 0;
+            if (overbooking > 0 && !allow_overbooking) {
+                return {0, overbooking};
+            }
+            if (total_reserved_.compare_exchange_weak(
```
wence-: `current_allocated` is "expensive" since it has to acquire a mutex. So let's make this `compare_exchange_strong` and lose the while loop, no?

nirandaperera: Maybe not. I think the `while` loop serves 2 purposes. 
- spurious failures
- `total_reserved_` reserved could also be modified between L108-119.
So, the latter might still be in effect. I think we can rework the logic to avoid this. Let me try

nirandaperera: I changed this to a more _pessimistic_ approach on `total_reserved_`. We claim upfront, and if overbooking is not allowed, we rollback.

PR #1147, opened 2026-07-31
Add test for `bind()` and CUDA initialization.

TomAugspurger:
https://github.com/NVIDIA/cuCascade/commit/8a3b42f02b69e27dd05402c215d4369ff97e4f1c
changed cucascade's topology discovery in a way that ultimately broke multi-GPU
cudf-polars.

Some part of rapidsmpf's ``bind`` is sensitive to whether or not CUDA is initialized
during topology discovery. This PR adds a test to ensure we catch it sooner
if this happens again.

https://github.com/rapidsai/rapidsmpf/pull/1145 pinned release/26.08 to a known-good commit.
https://github.com/rapidsai/rapidsmpf/pull/1146 forward-merged that to `main`. But we
don't want that pin on `main`: we'd prefer to catch regressions and benefit from improvements
immediately. So the second commit reverts that pin.

### DIFF
--- a/cmake/thirdparty/get_cucascade.cmake
+++ b/cmake/thirdparty/get_cucascade.cmake
@@ -14,7 +14,7 @@ function(find_and_configure_cucascade)
     GLOBAL_TARGETS cuCascade::cucascade_topology_discovery
     CPM_ARGS
     GIT_REPOSITORY https://github.com/NVIDIA/cuCascade.git
-    GIT_TAG d515bb0536b8766bae61ec60a530df394467af64
+    GIT_TAG main
     OPTIONS "CUCASCADE_BUILD_TESTS OFF"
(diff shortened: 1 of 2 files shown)


TomAugspurger:
/ok to test https://github.com/rapidsai/rapidsmpf/pull/1147/commits/69075eca59f4318a9a8a5201f2339aec6aabcc66

TomAugspurger:
I expect the new test to fail, since we should be building rapidsmpf against a cucascade commit with the problematic rmm usage in topology discovery.

TomAugspurger:
/ok to test https://github.com/rapidsai/rapidsmpf/pull/1147/commits/d4bf9ee875690c94c4c64cd0781f0de81372e865

PR #1149, opened 2026-08-05
Add Python bindings for `rapidsmpf::Buffer`

Matt711:
Adds Python bindings for `rapidsmpf::Buffer` and exposes `BufferResource::make_buffer`.

Needed by https://github.com/rapidsai/cudf/pull/23317


Matt711:
/ok to test https://github.com/rapidsai/rapidsmpf/pull/1149/commits/a402174a7b296a6e3d2c1ce879d9d81a1a4f5512

python/rapidsmpf/rapidsmpf/memory/buffer.pyx
```diff
@@ -1,2 +1,49 @@
...
+        """Size of the buffer in bytes."""
+        return deref(self._handle).size
+
+    @property
+    def mem_type(self):
+        """Memory type of this buffer."""
+        return deref(self._handle).mem_type()
+
+    def __getbuffer__(self, Py_buffer* view, int flags):
+        if deref(self._handle).mem_type() != MemoryType.PINNED_HOST:
```
nirandaperera: Technically this can be for `HOST` type as well, isnt it?

python/rapidsmpf/rapidsmpf/tests/test_buffer.py
```diff
@@ -0,0 +1,79 @@
...
+def pinned_br() -> BufferResource:
+    from rapidsmpf.memory.pinned_memory_resource import PinnedPoolProperties
+
+    pool_size = 4 * 1024 * 1024
+    mr = rmm.mr.CudaMemoryResource()
+    return BufferResource(
+        mr,
+        pinned_pool_properties=PinnedPoolProperties(initial_pool_size=pool_size),
+        memory_limits={MemoryType.PINNED_HOST: pool_size},
+    )
```
nirandaperera, line 33: Not all CI runners can make a pinned pool. Best to skip if not https://github.com/rapidsai/rapidsmpf/blob/main/python/rapidsmpf/rapidsmpf/memory/pinned_memory_resource.pyi#L8

python/rapidsmpf/rapidsmpf/tests/test_buffer.py
```diff
@@ -0,0 +1,79 @@
...
+def test_make_buffer_device() -> None:
+    size = 1024
+    mr = rmm.mr.CudaMemoryResource()
+    br = BufferResource(mr, memory_limits={MemoryType.DEVICE: size * 4})
+    stream = Stream()
+    reservation, _ = br.reserve(MemoryType.DEVICE, size, allow_overbooking=False)
+    buf = br.make_buffer(size, stream, reservation)
+    assert isinstance(buf, Buffer)
+    assert buf.size == size
+    assert buf.mem_type == MemoryType.DEVICE
```
nirandaperera: Maybe do the same for all 3 types?

python/rapidsmpf/rapidsmpf/memory/buffer.pyx
Matt711: Yes

Matt711: Done

python/rapidsmpf/rapidsmpf/memory/buffer.pyx
```diff
@@ -1,2 +1,52 @@
...
+    def size(self):
+        """Size of the buffer in bytes."""
+        return deref(self._handle).size
+
+    @property
+    def mem_type(self):
+        """Memory type of this buffer."""
+        return deref(self._handle).mem_type()
+
+    def __getbuffer__(self, Py_buffer* view, int flags):
```
madsbk: I'm not sure this is safe as-is. Even though this is host memory, `Buffer` is still stream-ordered, `data()` doesn't check `is_latest_write_done()`, so `__getbuffer__` can hand out a `memoryview` while an async write is still in flight on the buffer's stream (e.g. a pending `HostBuffer` allocation or an in-progress `write_access()`).

At minimum I think we need two things:
1. Raise if the latest write isn't done yet i.e. check `is_latest_write_done()` in `__getbuffer__` and throw if it's `False`. It's already a public method, so this shouldn't need much new surface.
2. The exported view should be read-only. Right now `readonly=False` is passed to `PyBuffer_FillInfo` by deliberately casting away `data()`'s `const`, which lets arbitrary Python code mutate the buffer with no synchronization against concurrent stream-ordered writes on the C++ side. Even with fix (1), a write happening later on the buffer's stream could race with Python code still holding the view.

Actually, stepping back. I wonder if the buffer protocol is the right fit here at all, given it can't express "done with the stream" back to `Buffer`. Fix (1) only checks the write is done *at acquisition time*; nothing stops a new `write_access()` from starting while the `memoryview` is still alive, since `__getbuffer__`/`__releasebuffer__` don't take a lock. Might be better to instead expose `exclusive_data_access()`/`unlock()` directly (e.g. as a context manager) so the lock is held for the actual lifetime of Python's access, rather than trying to fake that guarantee through the buffer protocol.

madsbk (changes requested):

python/rapidsmpf/rapidsmpf/memory/buffer.pyx
Matt711: Switched to a cm in https://github.com/rapidsai/rapidsmpf/pull/1149/commits/6bcd2fec2570fc2b2b5ce54ca9083039cff04db4

python/rapidsmpf/rapidsmpf/memory/buffer.pyx
```diff
@@ -1,2 +1,98 @@
...
+        Returns
+        -------
+        BufferHostView
+            A context manager that yields a writable ``memoryview`` of the buffer.
+
+        Raises
+        ------
+        TypeError
+            If the buffer is not a host buffer (``HOST`` or ``PINNED_HOST``).
+        std::logic_error
```
madsbk:
```suggestion
        RuntimeError
```

python/rapidsmpf/rapidsmpf/memory/buffer.pyx
```diff
@@ -1,2 +1,98 @@
...
+
+        Acquires an exclusive lock on the buffer for the duration of the ``with``
+        block, preventing concurrent stream-ordered operations on the C++ side.
+        The lock is released (and the returned ``memoryview`` must not be used)
+        once the block exits.
+
+        Returns
+        -------
+        BufferHostView
+            A context manager that yields a writable ``memoryview`` of the buffer.
```
madsbk:
```suggestion
        A context manager that yields a writable ``memoryview`` of the buffer.
```

python/rapidsmpf/rapidsmpf/memory/buffer.pyx
```diff
@@ -1,2 +1,98 @@
...
+from cpython.buffer cimport PyBUF_WRITE
+from cpython.memoryview cimport PyMemoryView_FromMemory
+from cython.operator cimport dereference as deref
+from libcpp.utility cimport move
+from rmm.pylibrmm.stream cimport Stream
+
+from rapidsmpf.memory.buffer_resource cimport BufferResource
+
+
+cdef class BufferHostView:
```
madsbk, line 13: Add a docstring

madsbk (approved):
Very good

PR #1151, opened 2026-08-13
Configure UCXX wakeup support based on progress mode

pentschev:
Configure UCXX contexts so wakeup support is enabled only for blocking progress modes and disabled for polling modes.

Polling modes do not use wakeup-based progress, so enabling this feature is unnecessary and can result in incorrect UCXX configuration (e.g., disabling transports such as SRD that do not support wakeup). This change aligns the context features with the selected progress strategy.


cpp/src/communicator/ucxx_internal.hpp
pentschev, line 1: Only added as a separate header so this new function is testable.

wence- (approved):

pentschev:
Thanks @wence- !

PR #1158, opened 2026-08-19
X-ORG-1157: Publish API docs to docs.nvidia.com

### LINKED ISSUE #1157: [X-ORG] Publish docs to docs.nvidia.com via new shared workflow
As part of publishing API docs to docs.nvidia.com, we need to integrate a [new shared workflow](https://github.com/rapidsai/shared-workflows/blob/main/.github/workflows/publish-api-docs.yaml) into the build workflow.

This integration will take the docs publish by the docs-build job and push them to the docs.nvidia.com S3 bucket. It does not automatically make those docs go live - that needs to be handled out-of-band by the docs team themselves.

XREF: https://github.com/rapidsai/build-infra/issues/380

josephine-wolf-oberholtzer:
Closes #1157

In order to publish docs to docs.nvidia.com, we need to:
- Integrate a [new shared workflow](https://github.com/rapidsai/shared-workflows/blob/main/.github/workflows/publish-api-docs.yaml) for pushing to the docs.nvidia.com S3 bucket
- Activate the version switcher drop-down in Sphinx, using a `versions.json` file located on docs.nvidia.com

N.B.:
- This _does not replace_ publishing docs to docs.rapidsai.org. This is _in addition to_.
- Virtually identical changes have already been made for cudf, cugraph, cuvs, nvforest, rmm, etc.
- Unlike the initial batch of repos, I'm setting the `version-map` parameter via `vars` rather than hard-coding. I'll migrate the initial batch of repos to use this same pattern in the near future so we can coordinate versions independent of code changes.


docs/source/conf.py
```diff
@@ -63,9 +63,14 @@
             "type": "fontawesome",
         },
     ],
-    "show_toc_level": 2,
     "navbar_align": "right",
+    "navbar_center": "navbar-nav, version-switcher, navbar-external-links",
     "navigation_with_keys": True,
+    "show_toc_level": 2,
+    "switcher": {
+        "json_url": "https://docs.nvidia.com/cucim/rapidsmpf.json",
```
wence-: This URL looks wrong?

wence- (changes requested):

docs/source/conf.py
josephine-wolf-oberholtzer: Yeah, it's wrong. Fixed!

bdice (approved):

PR #1161, opened 2026-08-24
ci: Make GB300 tests optional for successful completion

wence-:
Due to nvbug 6565551, multi-process tests on GB300 can fail when the virtualisation breaks. Since this is nothing we can recover from in user space, there's not much we can do.

To avoid this blocking CI, let's make the GB300 test runs optional.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -39,6 +39,26 @@ jobs:
       branch: ${{ inputs.branch }}
       container-options: "--cap-add CAP_SYS_PTRACE --shm-size=8g --ulimit=nofile=1000000:1000000"
       date: ${{ inputs.date }}
+      matrix_filter: map(select(.GPU != "gb300"))
+      script: ci/test_cpp.sh
+      sha: ${{ inputs.sha }}
+  conda-cpp-tests-gb300:
+    if: inputs.build_type != 'pull-request'
+    permissions:
+      actions: read
+      contents: read
+      id-token: write
+      packages: read
+      pull-requests: read
+    secrets: inherit # zizmor: ignore[secrets-inherit]
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-tests.yaml@main
+    with:
+      build_type: ${{ inputs.build_type }}
+      branch: ${{ inputs.branch }}
+      container-options: "--cap-add CAP_SYS_PTRACE --shm-size=8g --ulimit=nofile=1000000:1000000"
+      continue-on-error: true
+      date: ${{ inputs.date }}
+      matrix_filter: map(select(.GPU == "gb300"))
       script: ci/test_cpp.sh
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -39,6 +39,26 @@ jobs:
       branch: ${{ inputs.branch }}
       container-options: "--cap-add CAP_SYS_PTRACE --shm-size=8g --ulimit=nofile=1000000:1000000"
       date: ${{ inputs.date }}
+      matrix_filter: map(select(.GPU != "gb300"))
+      script: ci/test_cpp.sh
+      sha: ${{ inputs.sha }}
+  conda-cpp-tests-gb300:
+    if: inputs.build_type != 'pull-request'
```
pentschev, line 50: Why disabling it for pull requests?

.github/workflows/test.yaml
```diff
@@ -72,6 +92,27 @@ jobs:
       branch: ${{ inputs.branch }}
       container-options: "--cap-add CAP_SYS_PTRACE --shm-size=8g --ulimit=nofile=1000000:1000000"
       date: ${{ inputs.date }}
+      matrix_filter: map(select(.GPU != "gb300"))
+      run_codecov: false
+      script: ci/test_python.sh
+      sha: ${{ inputs.sha }}
+  conda-python-tests-gb300:
+    if: inputs.build_type != 'pull-request'
```
pentschev, line 108: Ditto.

pentschev (approved):
Overall looks good to me, I'm just wondering why we disable it for PRs. Not sure if it's possible, but perhaps having a maximum number of failures (like no more than 7 consecutive nightlies failures) would be probably be good too, so as to ensure we don't miss a valid problem.

.github/workflows/test.yaml
bdice: We only have enough GB300 capacity to run it in nightlies. PRs shouldn't use GB300 nodes at this time.

.github/workflows/test.yaml
```diff
@@ -39,6 +39,26 @@ jobs:
       branch: ${{ inputs.branch }}
       container-options: "--cap-add CAP_SYS_PTRACE --shm-size=8g --ulimit=nofile=1000000:1000000"
       date: ${{ inputs.date }}
+      matrix_filter: map(select(.GPU != "gb300"))
+      script: ci/test_cpp.sh
+      sha: ${{ inputs.sha }}
+  conda-cpp-tests-gb300:
```
bdice, line 45: We need an issue linked here documenting the problems we've observed, so that we know when to revert this. Same for the Python job.

bdice (approved):

.github/workflows/test.yaml
wence-: Done. I just reference the nvbug

PR #1162, opened 2026-08-24
Make `spill_to_make_headroom()` account for reservations

madsbk:
`spill_to_make_headroom(n)` now spills until `n` bytes can actually be reserved. It used to spill until `n` bytes were merely unallocated, which ignored outstanding reservations, so it could report success while a following `reserve(n)` still failed.

`Shuffler::insert` and the periodic spill thread now start spilling once allocations plus reservations hit the limit instead of waiting for allocations alone to get there.

### Breaking change: `memory_reserved()` removed

This drops `BufferResource::memory_reserved()` from both the C++ and Python APIs. No deprecation period, since nothing uses it. The only callers are our own tests, and a sweep of the RAPIDS projects checked out locally turned up nothing.


cpp/tests/test_shuffler.cpp
```diff
@@ -587,6 +587,45 @@ TEST(Shuffler, SpillOnInsertAndExtraction) {
...
+    EXPECT_EQ(comm->nranks(), 1);
+    rapidsmpf::shuffler::Shuffler shuffler(comm, /* op_id = */ 0, 1, br.get());
+
+    // Reserve twice the limit. Availability stays positive, but the memory available
+    // for reservation goes negative.
+    auto [reservation, overbooking] = br->reserve(
+        rapidsmpf::MemoryType::DEVICE,
+        2 * static_cast<std::size_t>(k_no_spill_limit),
+        rapidsmpf::AllowOverbooking::YES
+    );
```
pentschev, line 614: Could we either assert for successful reservation and nonzero overbooking, or alternatively mark this `[[maybe_unused]]`?

pentschev (approved):
Minor test sanity improvement, otherwise LGTM. Thanks Mads.

cpp/tests/test_shuffler.cpp
madsbk: Good call, went with the asserts.

madsbk:
Thanks @pentschev

PR #1163, opened 2026-08-25
Squash more HtoD copy stream sync issues

wence-:
We previously went round and did an audit of the HtoD copies and inserted stream syncs, but missed a few. These are all the rest I found.

Mostly mechanical, with one notable change. Since `buffer_copy` knows the memory type of both the source and destination I chose to explicitly insert the stream synchronisation in the copy when we know the source is in pageable host memory. This is morally the same as the `stream_wait` call we already have.

### DIFF
--- a/cpp/tests/test_buffer.cpp
+++ b/cpp/tests/test_buffer.cpp
@@ -97,6 +97,7 @@ TEST_P(BufferRebindStreamTest, RebindStreamAndCopy) {
     auto rmm_buffer = std::make_unique<rmm::device_buffer>(
         random_data.data(), buffer_size, stream1, br->device_mr()
     );
+    stream1.synchronize();
 
     auto [reserve1, overbooking1] =
         br->reserve(mem_type, buffer_size, AllowOverbooking::YES);
--- a/cpp/include/rapidsmpf/memory/buffer.hpp
+++ b/cpp/include/rapidsmpf/memory/buffer.hpp
@@ -376,6 +376,9 @@ class Buffer {
 /**
  * @brief Asynchronously copy data between buffers.
  *
+ * @note The copy is stream-ordered on @p dst's stream, correct cross-stream ordering
+ * between @p src's stream and @p dst's stream is provided automatically.
+ *
  * Copies @p size bytes from @p src, starting at @p src_offset, into @p dst at
  * @p dst_offset.
  *
--- a/cpp/tests/streaming/test_fanout.cpp
+++ b/cpp/tests/streaming/test_fanout.cpp
@@ -12,6 +12,7 @@
 #include <coro/coro.hpp>
 
 #include <rapidsmpf/memory/buffer.hpp>
+#include <rapidsmpf/memory/memory_type.hpp>
 #include <rapidsmpf/memory/pinned_memory_resource.hpp>
 #include <rapidsmpf/streaming/core/actor.hpp>
 #include <rapidsmpf/streaming/core/coro_utils.hpp>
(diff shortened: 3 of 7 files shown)


cpp/src/memory/buffer.cpp
```diff
@@ -168,9 +169,16 @@ void buffer_copy(
             dst_data + dst_offset, src.data() + src_offset, size, stream
         ));
     });
-    // after the dst.write_access(), its last_write_event is recorded on dst.stream(). So,
-    // we need the src.stream() to wait for that event.
-    dst.latest_write_event().stream_wait(src.stream());
+    if (src.mem_type() == MemoryType::HOST) {
+        // If source memory is host, its destruction will cannot be stream ordered after
+        // the copy we just enqueued, so we need to make sure the copy has completed.
```
pentschev:
```suggestion
        // If source memory is host, its destruction cannot be stream ordered after
        // the copy we just enqueued, so we need to make sure the copy has completed.
```

cpp/src/memory/buffer.cpp
```diff
@@ -168,9 +169,16 @@ void buffer_copy(
...
+        dst.latest_write_event().host_wait();
+    } else {
+        // Otherwise, destruction will be stream ordered. After the dst.write_access(),
+        // its last_write_event is recorded on dst.stream(). So, we need the src.stream()
+        // to wait for that event.
+        dst.latest_write_event().stream_wait(src.stream());
+    }
```
pentschev: Is the intent that a pageable-host source is immediately safe for host-side reuse when `buffer_copy()` returns? If so, could we document that `HOST`-source copies are intentionally synchronous, since this differs from the general asynchronous `buffer_copy` contract? Otherwise, the previous `stream_wait(src.stream())` preserves source lifetime and lets callers synchronize `src.stream()` only when they actually need host access.

wence-: Yes, this is intentional.

The problem is this that for pageable source, the `stream_wait(src.stream())` doesn't ensure that any destruction of `src` is stream-ordered after this copy.

wence-: I decided to make the sync happen inside buffer copy otherwise all callers of buffer_copy would have to have code like the following:

```
buffer_copy(..., dst, src, stream);
if (src.mem_type() == MemoryType::HOST) {
    dst.latest_write_event().host_wait();
    // or stream.synchronize();
}
```
And it's easy to forget.

wence-: Added docstring note.

pentschev: > Yes, this is intentional.
> 
> The problem is this that for pageable source, the `stream_wait(src.stream())` doesn't ensure that any destruction of `src` is stream-ordered after this copy.

When you say destruction, do you literally mean deallocation or any modification of the buffer? If it's deallocation I think syncing the stream is also covered in https://github.com/rapidsai/rapidsmpf/blob/b9e86061631b8f5189607e6d6c539d8fce384b67/cpp/src/memory/host_memory_resource.cpp#L51-L55.

I'm not sure what's the best solution here, but I'm fine either way, at least the docstring update that you have done already was necessary so I'm ok with that.

pentschev (approved):

cpp/src/memory/buffer.cpp
wence-: Ah, in which case we don't need this synchronisation.

wence-: Removed.

pentschev (approved):
Thanks Lawrence!

cpp/tests/test_buffer.cpp
```diff
@@ -97,6 +97,7 @@ TEST_P(BufferRebindStreamTest, RebindStreamAndCopy) {
     auto rmm_buffer = std::make_unique<rmm::device_buffer>(
         random_data.data(), buffer_size, stream1, br->device_mr()
     );
+    stream1.synchronize();
```
nirandaperera, line 100: I think we can drop this, isnt it? random_data stays alive until the end of the test case.

nirandaperera (commented):
Just a small comment

cpp/tests/test_buffer.cpp
wence-: One has to think a lot to realise the `stream2.synchronize()` at the end of this test also ensures that this code has run. I would rather just sync here.

PR #1164, opened 2026-08-25
Make `MemoryReserveOrWait` spill

madsbk:
`MemoryReserveOrWait` now spills. When no pending request fits the available memory, the wait loop frees enough for the highest-priority one and re-polls, so a waiter is unblocked by spilling instead of sitting until its timeout expires. Spilling from the poll path is non-blocking, so it never queues behind an in-flight spill or parks the streaming executor.

Before this it never spilled. It waited for the available memory to rise while nothing in the mechanism caused it to rise. The background spiller could not cover for it either. It runs with a headroom of zero, and a request parked in `reserve_or_wait` holds no reservation, only the intent to take one, so it is invisible to that check. #1162 made the spiller account for outstanding reservations, which does not help here for the same reason: a pending request is not yet a reservation.

The result was circular. Under pressure `reserve_or_wait` timed out, the `reserve_or_wait_or_overbook` fallback allocated past the limit, and only then did available memory go negative and the periodic thread begin spilling. Overbooking was what armed the spiller, so memory was spilled after the limit was exceeded rather than to avoid exceeding it.

### Breaking changes

`MemoryReserveOrWait`'s constructor takes a mandatory `std::shared_ptr<Logger>` as its second argument, so it can warn when a spill function throws. This is source-breaking for anyone constructing one directly. All in-repo call sites are updated, libcudf_streaming never constructs one, and the Python signature is unchanged because the binding already had the `Context` to take the logger from.

`spill_to_make_headroom()` keeps its signature but changes behavior. Under concurrency it can now return 0 where it previously returned a nonzero amount, because a caller that waited for an in-progress spill sees the memory that spill freed instead of spilling again for a shortfall that no longer exists.

### DIFF
--- a/cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
+++ b/cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
@@ -10,6 +10,7 @@
 
 #include <coro/task.hpp>
 
+#include <rapidsmpf/communicator/logger.hpp>
 #include <rapidsmpf/config.hpp>
 #include <rapidsmpf/memory/buffer_resource.hpp>
 #include <rapidsmpf/streaming/core/actor.hpp>
@@ -55,12 +56,14 @@ class MemoryReserveOrWait {
      * still cannot be satisfied.
      *
      * @param options Configuration options.
+     * @param logger Shared pointer to a logger.
      * @param mem_type The memory type for which reservations are requested.
      * @param executor Shared pointer to a coroutine executor.
      * @param br Buffer resource for memory allocation.*
      */
     MemoryReserveOrWait(
         config::Options options,
+        std::shared_ptr<Logger> logger,
         MemoryType mem_type,
         std::shared_ptr<CoroThreadPoolExecutor> executor,
         std::shared_ptr<BufferResource> br
@@ -281,12 +284,14 @@ class MemoryReserveOrWait {
     mutable std::mutex mutex_;
     std::uint64_t sequence_counter{0};
     MemoryType const mem_type_;
+    std::shared_ptr<Logger> logger_;
     std::shared_ptr<CoroThreadPoolExecutor> executor_;
     std::shared_ptr<BufferResource> br_;
     Duration const timeout_;
     std::set<Request> reservation_requests_;
     std::atomic<std::uint64_t> periodic_memory_check_counter_{0};
     std::optional<coro::task<void>> periodic_memory_check_task_;
+    bool periodic_task_running_{false};
 };
 
 /**
--- a/cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
+++ b/cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
@@ -98,13 +98,15 @@ class MemoryReserveOrWait {
      * highest-priority one needs, and again just before progress is forced. Both are
      * limited to `MemoryType::DEVICE`.
      *
-     * The timeout does not apply specifically to this request. Instead, it is used
-     * as a global progress guarantee: if no pending reservation request can be
-     * satisfied within the timeout, `MemoryReserveOrWait` forces progress by
-     * selecting the smallest pending request and attempting to reserve memory for
-     * it. The forced reservation attempt may result in an empty `MemoryReservation`
-     * if the selected request still cannot be satisfied, for example when nothing
-     * is spillable.
+     * The timeout does not apply specifically to this request. Instead, it bounds
+     * when a final recovery attempt begins, not when it completes: if no pending
+     * reservation request can be satisfied within the timeout, `MemoryReserveOrWait`
+     * forces progress by selecting the smallest pending request, spilling to make
+     * room for it, and attempting to reserve memory. That spill waits for any
+     * in-flight spill to finish, so the call can return later than the timeout. The
+     * forced reservation attempt may result in an empty `MemoryReservation` if the
+     * selected request still cannot be satisfied, for example when nothing is
+     * spillable.
      *
      * When multiple reservation requests are eligible, `MemoryReserveOrWait` uses
      * @p net_memory_delta as a heuristic to prefer requests that are expected to
--- a/cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
+++ b/cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
@@ -30,6 +30,11 @@ class Context;
  * memory with backpressure. Callers submit reservation requests via
  * `reserve_or_wait()`, which suspends until enough memory is available or
  * progress must be forced.
+ *
+ * While requests are pending and none of them fit into the available memory,
+ * spilling is triggered on the caller's behalf. This applies to `MemoryType::DEVICE`
+ * only, since `SpillManager` measures headroom against device memory. Requests of
+ * other memory types wait without spilling.
  */
 class MemoryReserveOrWait {
   public:
@@ -54,6 +59,8 @@ class MemoryReserveOrWait {
      * progress by selecting the smallest pending request and attempting to reserve
      * memory for it. This attempt may result in an empty reservation if the request
      * still cannot be satisfied.
+     *
+     * Spilling is attempted before that timeout is reached, see the class docs.
      *
      * @param options Configuration options.
      * @param logger Shared pointer to a logger.
@@ -87,12 +94,17 @@ class MemoryReserveOrWait {
      * (including other pending requests) makes progress within the configured
      * timeout.
      *
+     * While no pending request fits, spilling is attempted to free the memory the
+     * highest-priority one needs, and again just before progress is forced. Both are
+     * limited to `MemoryType::DEVICE`.
+     *
      * The timeout does not apply specifically to this request. Instead, it is used
      * as a global progress guarantee: if no pending reservation request can be
      * satisfied within the timeout, `MemoryReserveOrWait` forces progress by
      * selecting the smallest pending request and attempting to reserve memory for
      * it. The forced reservation attempt may result in an empty `MemoryReservation`
-     * if the selected request still cannot be satisfied.
+     * if the selected request still cannot be satisfied, for example when nothing
+     * is spillable.
      *
      * When multiple reservation requests are eligible, `MemoryReserveOrWait` uses
      * @p net_memory_delta as a heuristic to prefer requests that are expected to
--- a/cpp/src/memory/spill_manager.cpp
+++ b/cpp/src/memory/spill_manager.cpp
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <mutex>
+#include <optional>
 #include <utility>
 
 #include <rapidsmpf/memory/buffer_resource.hpp>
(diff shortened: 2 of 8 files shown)


cpp/src/memory/spill_manager.cpp
```diff
@@ -65,10 +67,8 @@ void SpillManager::remove_spill_function(SpillFunctionID fid) {
     }
 }
 
-std::size_t SpillManager::spill(std::size_t amount) {
-    RAPIDSMPF_NVTX_FUNC_RANGE();
+std::size_t SpillManager::spill_impl(std::size_t amount) {
```
nirandaperera: Nit. Lets rename `*_impl` to `*_unsafe`. It wasnt immediately apparent to me why the lock was removed here.

cpp/src/streaming/core/memory_reserve_or_wait.cpp
```diff
@@ -198,13 +203,68 @@ coro::task<void> MemoryReserveOrWait::periodic_memory_check() {
...
+            if (self != nullptr) {
+                std::lock_guard lock(self->mutex_);
+                self->periodic_task_running_ = false;
+            }
+        }
+
+        void dismiss() noexcept {
+            self = nullptr;
+        }
+    } running_flag_guard{.self = this};
```
nirandaperera: Let's not do this. Personally, its not very readable. Lets break up the declaration and init.

cpp/tests/streaming/test_memory_reserve_or_wait.cpp
```diff
@@ -40,6 +41,61 @@ class StreamingMemoryReserveOrWait
     std::int64_t get_mem_avail() {
         return br->memory_available(rapidsmpf::MemoryType::DEVICE);
     }
+
+    // Pins the memory available for reservation to `available` bytes and registers a
+    // spill function that frees exactly what it is asked for, recording each request.
+    // Pinning keeps the buffer resource's periodic spill thread out of the way, it only
+    // acts once the available memory goes negative.
+    class SpillRecorder {
```
nirandaperera: Let's pull this out from `StreamingMemoryReserveOrWait` and move to L332? It doesnt seemed to be accessing any private members.

cpp/tests/streaming/test_memory_reserve_or_wait.cpp
```diff
@@ -268,6 +330,88 @@ TEST_P(StreamingMemoryReserveOrWait, NoDeadlockWhenSpawningWithStaleHandle) {
...
+
+    // The available memory covers the request on its own, but an outstanding
+    // reservation consumes all of it, so only spilling can unblock the waiter.
+    SpillRecorder spills{br.get(), /* available = */ 10};
+    auto [outstanding, _] = br->reserve(MemoryType::DEVICE, 10, AllowOverbooking::NO);
+    ASSERT_EQ(outstanding.size(), 10);
+
+    std::vector<Actor> actors;
+    actors.push_back(waiter(mrow, 10, 0, 10));
+    run_actor_network(std::move(actors));
```
nirandaperera, line 359:
```suggestion
    run_actor_network({waiter(mrow, 10, 0, 10)});
```

cpp/tests/streaming/test_memory_reserve_or_wait.cpp
```diff
@@ -40,6 +41,61 @@ class StreamingMemoryReserveOrWait
...
+                  )
+                  + available
+              } {
+            br_->set_memory_limit(rapidsmpf::MemoryType::DEVICE, limit_);
+            fid_ = br_->spill_manager().add_spill_function(
+                [this](std::size_t amount) -> std::size_t {
+                    std::lock_guard lock(mutex_);
+                    amounts_.push_back(amount);
+                    limit_ += rapidsmpf::safe_cast<std::int64_t>(amount);
+                    br_->set_memory_limit(rapidsmpf::MemoryType::DEVICE, limit_);
```
nirandaperera: this keeps on incrementing the device limit everytime a spill function is called. Wouldnt that be problematic?

cpp/tests/streaming/test_memory_reserve_or_wait.cpp
```diff
@@ -268,6 +330,88 @@ TEST_P(StreamingMemoryReserveOrWait, NoDeadlockWhenSpawningWithStaleHandle) {
...
+    SpillManager::SpillFunction func = [&](std::size_t) -> std::size_t {
+        spill_called = true;
+        throw std::runtime_error("spill function failed");
+    };
+    auto fid = br->spill_manager().add_spill_function(func, /* priority = */ 1);
+
+    // The waiter completes via the timeout instead of hanging on its queue.
+    std::vector<Actor> actors;
+    actors.push_back(waiter(mrow, 10, 0, 0));
+    run_actor_network(std::move(actors));
```
nirandaperera, line 394: same nit

cpp/tests/streaming/test_memory_reserve_or_wait.cpp
```diff
@@ -268,6 +330,88 @@ TEST_P(StreamingMemoryReserveOrWait, NoDeadlockWhenSpawningWithStaleHandle) {
...
+        MemoryType::DEVICE,
+        ctx->executor(),
+        ctx->br()
+    };
+
+    SpillRecorder spills{br.get(), /* available = */ 1024};
+
+    std::vector<Actor> actors;
+    actors.push_back(waiter(mrow, 10, 0, 10));
+    run_actor_network(std::move(actors));
```
nirandaperera, line 413: same

cpp/tests/streaming/test_memory_reserve_or_wait.cpp
madsbk: Not a correctness problem for these tests, and it matches the existing idiom in `test_spill_manager.cpp`. But an unbounded mock can never fail: a regression that spills repeatedly would keep being granted memory, and the test would still pass.

`SpillRecorder` now has a finite pool. It frees what is requested until the spillable memory is exhausted, then returns 0.

cpp/tests/streaming/test_memory_reserve_or_wait.cpp
madsbk: done

cpp/src/streaming/core/memory_reserve_or_wait.cpp
madsbk: done

cpp/src/memory/spill_manager.cpp
madsbk: done

cpp/tests/streaming/test_memory_reserve_or_wait.cpp
madsbk: This doesn't compile. `Actor` is `coro::task<void>`, which is move-only, and `std::initializer_list` exposes its elements as `const`, so `std::vector<Actor>` can't be constructed from one.

I also tried a variadic overload of `run_actor_network()`, but that breaks five existing sites because a bare overloaded function name can no longer be resolved as a callable there. That's a breaking change, so if we want it, it should go in a separate PR.

cpp/src/streaming/core/memory_reserve_or_wait.cpp
```diff
@@ -271,6 +348,9 @@ coro::task<void> MemoryReserveOrWait::periodic_memory_check() {
         Request request = reservation_requests_.extract(it).value();
         lock.unlock();
 
+        // Last chance before handing back a zero-size reservation.
+        make_headroom(request.size, /* blocking = */ true);
```
pentschev, line 351: Could we keep this final spill attempt non-blocking too? `spill_to_make_headroom()` waits on the spill-manager mutex, so this periodic reservation task can park an executor worker behind an in-flight spill. Using `try_spill_to_make_headroom()` here and proceeding with the existing zero-size fallback if the lock is unavailable would preserve progress without blocking the streaming executor.

cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
```diff
@@ -55,12 +56,14 @@ class MemoryReserveOrWait {
      * still cannot be satisfied.
      *
      * @param options Configuration options.
+     * @param logger Shared pointer to a logger.
      * @param mem_type The memory type for which reservations are requested.
      * @param executor Shared pointer to a coroutine executor.
      * @param br Buffer resource for memory allocation.*
      */
     MemoryReserveOrWait(
```
pentschev, line 71: Could we update the docstring? The implementation now attempts a non-blocking device spill before `memory_reserve_timeout` when no request fits, but the C++ and Python docs only describe the timeout-based fallback. Documenting the pre-timeout, device-only spill behavior (and the fallback when it cannot spill) would match the new implementation.

cpp/src/streaming/core/memory_reserve_or_wait.cpp
```diff
@@ -198,13 +203,70 @@ coro::task<void> MemoryReserveOrWait::periodic_memory_check() {
...
+            }
+            return br_->spill_manager().try_spill_to_make_headroom(target).has_value();
+        } catch (...) {
+            if (!std::exchange(spill_failure_logged, true)) {
+                try {
+                    logger_->warn(
+                        "a spill function threw while making headroom for ",
+                        format_nbytes(headroom),
+                        ", falling back to the reservation timeout"
+                    );
```
pentschev, line 228: Could the warning avoid saying "falling back to the reservation timeout" here? `make_headroom()` is also invoked after the timeout at line 352, where that wording is no longer accurate. A neutral message such as "continuing with the reservation fallback" would describe both call paths correctly.

cpp/src/streaming/core/memory_reserve_or_wait.cpp
madsbk: I think it is best to block here. At this point we are about to hand back a zero-size reservation, so the caller either throws or accepts overbooking. 

If a spill is in flight it is freeing exactly the memory this request needs, so I think waiting for it is acceptable even if it means blocking one of the coroutine threads. 

For a workload running close to the limit, that wait may be the difference between staying under it and an OOM. It runs at most once per timeout window, so the stall is bounded.

cpp/include/rapidsmpf/streaming/core/memory_reserve_or_wait.hpp
madsbk: done

cpp/src/streaming/core/memory_reserve_or_wait.cpp
madsbk: done

cpp/src/streaming/core/memory_reserve_or_wait.cpp
pentschev: Thanks, I agree. One documentation detail: this final call can wait longer than `memory_reserve_timeout`, because it blocks on `SpillManager::mutex_` while another spill callback is running. Could we adjust the public docs’ "global progress guarantee" wording to clarify that the timeout starts the final recovery attempt, which may wait for an in-flight spill, rather than acting as an absolute bound on completion?

pentschev (approved):
Left one last suggestion about docs, otherwise LGTM. Thanks Mads.

cpp/src/streaming/core/memory_reserve_or_wait.cpp
madsbk: what about: 40917ca977ebe9a01ce24b4584730a0cc462bdfd

pentschev: Looks great, thanks!

pentschev (approved):

wence- (approved):

PR #1175, opened 2026-08-27
Migrate cuda stream views to cuda::stream_ref

vyasr:
## Summary
- migrate public and internal stream parameters from `rmm::cuda_stream_view` to `cuda::stream_ref`
- use `<cuda/stream>` and `stream_ref::sync()`
- pass raw CUDA stream handles with `.get()` at CUDA runtime API boundaries
- update Cython bindings to expose `cuda::stream_ref`

### DIFF
--- a/cpp/benchmarks/bench_memory_resources.cpp
+++ b/cpp/benchmarks/bench_memory_resources.cpp
@@ -9,8 +9,9 @@
 
 #include <benchmark/benchmark.h>
 
+#include <cuda/stream>
+
 #include <rmm/cuda_stream_pool.hpp>
-#include <rmm/cuda_stream_view.hpp>
 #include <rmm/device_buffer.hpp>
 #include <rmm/mr/cuda_memory_resource.hpp>
 #include <rmm/mr/per_device_resource.hpp>
@@ -73,7 +74,7 @@ class NewDelete {
     }
 
     void* allocate(
-        rmm::cuda_stream_view,
+        cuda::stream_ref,
         std::size_t size,
         std::size_t alignment = rmm::CUDA_ALLOCATION_ALIGNMENT
     ) {
@@ -81,7 +82,7 @@ class NewDelete {
     }
 
     void deallocate(
-        rmm::cuda_stream_view,
+        cuda::stream_ref,
         void* ptr,
         std::size_t,
         std::size_t alignment = rmm::CUDA_ALLOCATION_ALIGNMENT
@@ -142,7 +143,7 @@ cuda::mr::any_resource<cuda::mr::host_accessible> create_host_memory_resource(
 }
 
 void BM_Allocate(benchmark::State& state) {
-    rmm::cuda_stream_view stream = rmm::cuda_stream_default;
+    cuda::stream_ref stream{cudaStreamLegacy};
     auto const allocation_size = static_cast<std::size_t>(state.range(0));
     auto const resource_type = static_cast<ResourceType>(state.range(1));
 
@@ -157,11 +158,11 @@ void BM_Allocate(benchmark::State& state) {
     for (auto _ : state) {
         void* ptr = mr.allocate(stream, allocation_size, rmm::CUDA_ALLOCATION_ALIGNMENT);
         benchmark::DoNotOptimize(ptr);
-        stream.synchronize();
+        stream.sync();
 
         state.PauseTiming();
         mr.deallocate(stream, ptr, allocation_size, rmm::CUDA_ALLOCATION_ALIGNMENT);
-        stream.synchronize();
+        stream.sync();
         state.ResumeTiming();
     }
 
@@ -175,7 +176,7 @@ void BM_Allocate(benchmark::State& state) {
 }
 
 void BM_Deallocate(benchmark::State& state) {
-    rmm::cuda_stream_view stream = rmm::cuda_stream_default;
+    cuda::stream_ref stream{cudaStreamLegacy};
     auto const allocation_size = safe_cast<std::size_t>(state.range(0));
     auto const resource_type = static_cast<ResourceType>(state.range(1));
 
@@ -190,11 +191,11 @@ void BM_Deallocate(benchmark::State& state) {
     for (auto _ : state) {
         state.PauseTiming();
         void* ptr = mr.allocate(stream, allocation_size, rmm::CUDA_ALLOCATION_ALIGNMENT);
-        stream.synchronize();
+        stream.sync();
         state.ResumeTiming();
 
         mr.deallocate(stream, ptr, allocation_size, rmm::CUDA_ALLOCATION_ALIGNMENT);
-        stream.synchronize();
+        stream.sync();
     }
 
     state.SetBytesProcessed(
@@ -207,7 +208,7 @@ void BM_Deallocate(benchmark::State& state) {
 }
 
 void BM_DeviceToHostCopyInclAlloc(benchmark::State& state) {
-    rmm::cuda_stream_view stream = rmm::cuda_stream_default;
+    cuda::stream_ref stream{cudaStreamLegacy};
     auto const transfer_size = static_cast<std::size_t>(state.range(0));
     auto const resource_type = static_cast<ResourceType>(state.range(1));
 
@@ -224,8 +225,8 @@ void BM_DeviceToHostCopyInclAlloc(benchmark::State& state) {
     // Allocate device memory
     auto src = rmm::device_buffer(transfer_size, stream, *device_mr);
     // Initialize src to avoid optimization removal
-    RAPIDSMPF_CUDA_TRY(cudaMemsetAsync(src.data(), 0xAB, transfer_size, stream));
-    stream.synchronize();
+    RAPIDSMPF_CUDA_TRY(cudaMemsetAsync(src.data(), 0xAB, transfer_size, stream.get()));
+    stream.sync();
 
     for (auto _ : state) {
         void* dst =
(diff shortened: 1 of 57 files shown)


cpp/benchmarks/bench_memory_resources.cpp
```diff
@@ -274,7 +275,7 @@ void bench_copy(
 }
 
 void BM_DeviceToHostCopy(benchmark::State& state) {
-    rmm::cuda_stream_view stream = rmm::cuda_stream_default;
+    cuda::stream_ref stream = cuda::stream_ref{};
```
bdice: This is deprecated, I think.

python/rapidsmpf/rapidsmpf/memory/packed_data.pyx
```diff
@@ -201,7 +201,7 @@ cdef class PackedData:
         cdef const uint8_t* meta_ptr = NULL
         if meta_size > 0:
             meta_ptr = <const uint8_t*>&metadata[0]
-        cdef cuda_stream_view sv = stream.view()
+        cdef stream_ref sv = stream.view().value()
```
pentschev: I think we need to explicitly construct `stream_ref` here, e.g. `stream_ref(stream.view().value())` here. Cython does not apply the C++ converting constructor during assignment, I think this is the cause of [build failures](https://github.com/rapidsai/rapidsmpf/actions/runs/33034559195/job/98394338096).

cpp/benchmarks/bench_memory_resources.cpp
vyasr: Fixed in 610a2d6e: all default stream refs now explicitly use `cudaStreamLegacy`, avoiding the deprecated default constructor.

python/rapidsmpf/rapidsmpf/memory/packed_data.pyx
vyasr: Fixed in 610a2d6e by explicitly constructing `stream_ref(stream.view().value())`. This addresses the reported Cython build failure.

cpp/examples/example_cupti_monitor.cpp
```diff
@@ -56,7 +58,9 @@ int main() {
                       << " MB on GPU using rmm::device_buffer...\n";
             try {
                 // Allocate device memory using rmm::device_buffer
-                rmm::device_buffer buf(allocation_size, rmm::cuda_stream_default);
+                rmm::device_buffer buf(
+                    allocation_size, cuda::stream_ref{cudaStream_t{nullptr}}
```
nirandaperera: shall we add a `get_default_stream()` util?

cpp/include/rapidsmpf/memory/buffer.hpp
```diff
@@ -368,7 +368,7 @@ class Buffer {
   private:
     MemoryType const mem_type_;
     std::variant<DeviceBufferT, HostBufferT> storage_;
-    rmm::cuda_stream_view stream_;
+    cuda::stream_ref stream_{cudaStream_t{nullptr}};
```
nirandaperera: I think we have, `cudaStream_t{nullptr}` and `cudaStreamLegacy`. IMO we should use one, to be more consistent.

cpp/src/coll/utils.cpp
```diff
@@ -146,7 +148,8 @@ std::unique_ptr<Chunk> Chunk::deserialize(
         Chunk::INVALID_RANK,
         std::move(metadata),
         br->make_buffer(
-            br->stream_pool()->get_stream(), br->reserve_or_fail(data_size, MEMORY_TYPES)
+            cuda::stream_ref{br->stream_pool()->get_stream().value()},
```
nirandaperera: this is implicit, isnt it? So, cant we leave it as is?
```suggestion
            br->stream_pool()->get_stream(),
```

nirandaperera (commented):
just some comments about consistency

cpp/include/rapidsmpf/memory/buffer.hpp
vyasr: Agreed, let's use `cudaStreamLegacy`.

cpp/examples/example_cupti_monitor.cpp
vyasr: I think if we use `cudaStreamLegacy` everywhere instead of `cudaStream_t{nullptr}` that's brief enough to not need another helper function. WDYT?

cpp/include/rapidsmpf/memory/buffer.hpp
vyasr: Addressed in 01524f69: `Buffer::stream_` now consistently initializes with `cudaStreamLegacy`.

vyasr: Follow-up in f9111f93: applied `cudaStreamLegacy` consistently to every default `cuda::stream_ref` construction introduced by this branch (including test fixtures and members).

cpp/src/coll/utils.cpp
vyasr: Addressed in f9111f93: restored the direct `br->stream_pool()->get_stream()` call; the C++ conversion is implicit here.

cpp/include/rapidsmpf/memory/buffer.hpp
vyasr: Done.

cpp/src/coll/utils.cpp
vyasr: You're right. I changed this back.

bdice (approved):
I pushed a few minor fixes to allow implicit conversions from `rmm::cuda_stream_view` to `cuda::stream_ref` since later changes in RMM will actually return `cuda::stream_ref` and thus remove the implicit conversion. I also removed some unused `cimport`s from Cython.

cpp/include/rapidsmpf/cuda_stream.hpp
```diff
@@ -1,12 +1,14 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 #pragma once
 
 #include <memory>
 #include <ranges>
```
igorpeshansky: Looks like `cuda/stream` does not provide iterators by default, so we also need `#include <iterator>` here, or we get build failures like [this](https://github.com/NVIDIA/cudf/actions/runs/33158248511/job/98807177960?pr=23862#step:11:4135).

I wonder why the `rapidsmpf` CI didn't catch it…

cpp/include/rapidsmpf/cuda_stream.hpp
```diff
@@ -89,9 +91,7 @@ void cuda_stream_join(
  * @see cuda_stream_join(Range1 const&, Range2 const&, CudaEvent*)
  */
 inline void cuda_stream_join(
-    rmm::cuda_stream_view downstream,
-    rmm::cuda_stream_view upstream,
-    CudaEvent* event = nullptr
+    cuda::stream_ref downstream, cuda::stream_ref upstream, CudaEvent* event = nullptr
```
igorpeshansky, line 94: [Opened #1178]

Downstream consumers (e.g., `libcudf_streaming`) still pass in `rmm::cuda_stream_view` arguments to `cuda_stream_join`, which now attempts to instantiate the template ranged version on line 36 with something that isn't a range (`rmm::cuda_stream_view`) and results in build failures like [this](https://github.com/NVIDIA/cudf/actions/runs/33158248511/job/98807177960?pr=23862#step:11:4135):

```
cuda_stream.hpp: In instantiation of 'void rapidsmpf::cuda_stream_join(const Range1&, const Range2&, CudaEvent*)
  [with Range1 = rmm::cuda_stream_view; Range2 = rmm::cuda_stream_view]':
cpp/libcudf_streaming/src/bloom_filter.cpp:81:32: required from here
cuda_stream.hpp:41:13: error: 'begin' was not declared in this scope
```

Previously, `rmm::cuda_stream_view` matched the old signature exactly and was preferred over the template. This new scalar overload requires an implicit `rmm::cuda_stream_view`→`cuda::stream_ref` conversion, while the unconstrained template version wins by being an exact match. We need to constrain the template parameters to be ranges to prevent such matches and redirect scalar consumers to this overload, e.g.:

```cpp
template <std::ranges::input_range Range1, std::ranges::input_range Range2>
void cuda_stream_join(
    Range1 const& downstreams, Range2 const& upstreams, CudaEvent* event = nullptr
) { … }
```

or

```cpp
template <typename Range1, typename Range2>
  requires std::ranges::input_range<Range1> && std::ranges::input_range<Range2>
void cuda_stream_join(
    Range1 const& downstreams, Range2 const& upstreams, CudaEvent* event = nullptr
) { … }
```

> [!NOTE]
> Catching this required a consumer that still passes `rmm::cuda_stream_view` into `cuda_stream_join`, and all in-repo callers were migrated as part of this PR, so the CI didn't surface it.
>
> We could potentially add a test that would flag a regression of this (e.g., the snippet below; just needs to compile, not run):
>
> ```c++
> CudaEvent event;
> rmm::cuda_stream stream;
> cuda_stream_join(stream.view(), stream.view(), &event);
> cuda_stream_join(cudaStreamLegacy, cudaStreamLegacy, &event);
> cuda_stream_join(cuda::stream_ref{cudaStreamLegacy}, stream.view(), &event);
> ```
>
> Ironically, [`cpp/src/memory/buffer_resource.cpp` lines 284-288](https://github.com/rapidsai/rapidsmpf/blob/77e368d22348f35f9635d63f845b19dfc4027a37/cpp/src/memory/buffer_resource.cpp#L284-L288) would have triggered the exact same failure **in this repo** if not for the intermediate local with the new type…

> [!TIP]
> [Optional] This would still trigger an inscrutable error if a consumer accidentally passes an iterable whose elements aren't streams (e.g., `std::vector<int>`). We could guard against that as well by adding more constraints, e.g., `std::convertible_to<std::ranges::range_reference_t<Range1>, cuda::stream_ref> && std::convertible_to<std::ranges::range_reference_t<Range2>, cuda::stream_ref>`…

PR #1179, opened 2026-08-28
Add range constraints on cuda_stream_join

### LINKED ISSUE #1178: cuda_stream_join instantiation errors with rmm::cuda_stream_view
Downstream consumers (e.g., `libcudf_streaming`) still pass in `rmm::cuda_stream_view` arguments to `cuda_stream_join`, which now attempts to instantiate the template ranged version on [line 36](https://github.com/rapidsai/rapidsmpf/blob/77e368d22348f35f9635d63f845b19dfc4027a37/cpp/include/rapidsmpf/cuda_stream.hpp#L36) with something that isn't a range (`rmm::cuda_stream_view`) and results in build failures like [this](https://github.com/NVIDIA/cudf/actions/runs/33158248511/job/98807177960?pr=23862#step:11:4135):

```
cuda_stream.hpp: In instantiation of 'void rapidsmpf::cuda_stream_join(const Range1&, const Range2&, CudaEvent*)
  [with Range1 = rmm::cuda_stream_view; Range2 = rmm::cuda_stream_view]':
cpp/libcudf_streaming/src/bloom_filter.cpp:81:32: required from here
cuda_stream.hpp:41:13: error: 'begin' was not declared in this scope
```

Previously, `rmm::cuda_stream_view` matched the old signature exactly and was preferred over the template. This new scalar overload requires an implicit `rmm::cuda_stream_view`→`cuda::stream_ref` conversion, while the unconstrained template version wins by being an exact match. We need to constrain the template parameters to be ranges to prevent such matches and redirect scalar consumers to this overload, e.g.:

```cpp
template <std::ranges::input_range Range1, std::ranges::input_range Range2>
void cuda_stream_join(
    Range1 const& downstreams, Range2 const& upstreams, CudaEvent* event = nullptr
) { … }
```

or

```cpp
template <typename Range1, typename Range2>
  requires std::ranges::input_range<Range1> && std::ranges::input_range<Range2>
void cuda_stream_join(
    Range1 const& downstreams, Range2 const& upstreams, CudaEvent* event = nullptr
) { … }
```

> [!NOTE]
> Catching this required a consumer that still passes `rmm::cuda_stream_view` into `cuda_stream_join`, and all in-repo callers were migrated as part of this PR, so the CI didn't surface it.
>
> We could potentially add a test that would flag a regression of this (e.g., the snippet below; just needs to compile, not run):
>
> ```c++
> CudaEvent event;
> rmm::cuda_stream stream;
> cuda_stream_join(stream.view(), stream.view(), &event);
> cuda_stream_join(cudaStreamLegacy, cudaStreamLegacy, &event);
> cuda_stream_join(cuda::stream_ref{cudaStreamLegacy}, stream.view(), &event);
> ```
>
> Ironically, [`cpp/src/memory/buffer_resource.cpp` lines 284-288](https://github.com/rapidsai/rapidsmpf/blob/77e368d22348f35f9635d63f845b19dfc4027a37/cpp/src/memory/buffer_resource.cpp#L284-L288) would have triggered the exact same failure **in this repo** if not for the intermediate local with the new type…

> [!TIP]
> [Optional] This would still trigger an inscrutable error if a consumer accidentally passes an iterable whose elements aren't streams (e.g., `std::vector<int>`). We could guard against that as well by adding more constraints, e.g., `std::convertible_to<std::ranges::range_reference_t<Range1>, cuda::stream_ref> && std::convertible_to<std::ranges::range_reference_t<Range2>, cuda::stream_ref>`…

_Originally posted by @igorpeshansky in https://github.com/rapidsai/rapidsmpf/pull/1175#discussion_r3883678991_

igorpeshansky:
Constrains the template version of `cuda_stream_join` to ensure that both type arguments are ranges of `cuda::stream_ref`. Also adds a compile-only test case.

A follow-on to #1175.

Fixes #1178.

### DIFF
--- a/cpp/include/rapidsmpf/cuda_stream.hpp
+++ b/cpp/include/rapidsmpf/cuda_stream.hpp
@@ -15,9 +15,17 @@
 namespace rapidsmpf {
 
 namespace detail {
+
+/**
+ * @brief Internal concept for identifying ranges of a given type.
+ *
+ * @tparam R Range type.
+ * @tparam T Expected content type.
+ */
 template <typename R, typename T>
 concept input_range_of = std::ranges::input_range<R>
                          && std::convertible_to<std::ranges::range_reference_t<R>, T>;
+
 }  // namespace detail
 
 /**
--- a/cpp/include/rapidsmpf/cuda_stream.hpp
+++ b/cpp/include/rapidsmpf/cuda_stream.hpp
@@ -4,6 +4,7 @@
  */
 #pragma once
 
+#include <concepts>
 #include <memory>
 #include <ranges>
 
@@ -13,6 +14,11 @@
 
 namespace rapidsmpf {
 
+namespace detail {
+template <typename R, typename T>
+concept input_range_of = std::ranges::input_range<R>
+                         && std::convertible_to<std::ranges::range_reference_t<R>, T>;
+}  // namespace detail
 
 /**
  * @brief Make downstream CUDA streams wait on upstream CUDA streams.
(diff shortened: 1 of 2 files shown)


bdice (approved):

nirandaperera (approved):
LGTM

PR #1180, opened 2026-08-28
Expose stream pool streams as cuda::stream_ref

vyasr:
## Summary
- wrap the RMM-backed stream pool behind an MPF `StreamPool` API
- return `cuda::stream_ref` from stream-pool accessors, including indexed access
- retain Python compatibility by converting supplied RMM pools at the binding boundary

### DIFF
--- a/cpp/include/rapidsmpf/memory/buffer_resource.hpp
+++ b/cpp/include/rapidsmpf/memory/buffer_resource.hpp
@@ -7,6 +7,7 @@
 
 #include <array>
 #include <atomic>
+#include <cstddef>
 #include <cstdint>
 #include <memory>
 #include <mutex>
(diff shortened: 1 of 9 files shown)


cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -45,6 +46,43 @@ enum class AllowOverbooking : bool {
...
 };
 
+/**
+ * @brief Pool of non-blocking CUDA streams.
+ *
+ * The pool is backed by RMM until CUDA Core provides an owning stream-pool
+ * implementation. Its public interface returns CUDA Core stream references so
+ * callers do not need to bridge between stream abstractions.
+ */
+class StreamPool {
```
bdice, line 56: I keep reading these as template parameters... rapidmpf's PascalCase and rmm/cudf's snake_case are so tricky.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -45,6 +46,43 @@ enum class AllowOverbooking : bool {
...
+        : pool_{std::make_shared<rmm::cuda_stream_pool>(
+              num_streams, rmm::cuda_stream::flags::non_blocking
+          )} {}
+
+    explicit StreamPool(std::shared_ptr<rmm::cuda_stream_pool> pool)
+        : pool_{std::move(pool)} {
+        RAPIDSMPF_EXPECTS(pool_ != nullptr, "the stream pool pointer cannot be NULL");
+    }
+
+    [[nodiscard]] static std::shared_ptr<StreamPool> from_rmm(
```
bdice: Can we just use the constructor above and drop this factory? Or does Cython dislike that?

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
```diff
@@ -230,12 +230,14 @@ cdef class BufferResource:
                 _props.numa_id = <int>pinned_pool_properties.numa_id
             cpp_pinned_pool = _props
         with nogil:
+            # TODO: Replace this RMM pool with a cuda-python stream pool once one is
+            # available.
```
vyasr:
```suggestion
            # TODO: Replace this RMM pool with a cuda-python stream pool once a suitable
            # one is available with all the necessary CCCL interop.
```

cpp/include/rapidsmpf/memory/buffer_resource.hpp
vyasr: Addressed in 8dea24c5: Cython can use `make_shared[cpp_StreamPool](stream_pool.c_obj)`, so the factory was unnecessary and has been removed.

bdice (approved):

bdice (approved):
Great, [8dea24c](https://github.com/rapidsai/rapidsmpf/pull/1180/commits/8dea24c5875a9b37b103d09eafe60f66f3d9460f) is what I was looking for.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
vyasr: Yeah we probably should switch rapidsmpf to match the more dominant convention in RAPIDS, but problem for another day...

bdice (approved):

PR #1181, opened 2026-08-31
Migrate stream APIs from rmm::cuda_stream_view to cuda::stream_ref

bdice:
## Summary

Track the coordinated migration of stream APIs and call sites from `rmm::cuda_stream_view` to CCCL's `cuda::stream_ref`. This propagates `cuda::stream_ref` through RMM containers and memory resources, RAFT resource and handle APIs, downstream C++ interfaces, Python/Cython bindings, benchmarks, tests, and documentation.

This updates RapidsMPF C++ and Python/Cython stream handling to preserve references returned by RMM stream pools and extract raw handles only where required.

Depends on rapidsai/rmm#2372.

Tracked in rapidsai/build-planning#318.

## Migrations

- Pass `cuda::stream_ref` through stream pools, resource accessors, conditionals, and downstream APIs without converting to `rmm::cuda_stream_view`
- Use `cuda::stream_ref` constructions for default/legacy/per-thread streams
  - `rmm::cuda_stream_default` ➡️ `cuda::stream_ref{cudaStream_t{cudaStreamDefault}}`
  - `rmm::cuda_stream_legacy` ➡️ `cuda::stream_ref{cudaStreamLegacy}`
  - `rmm::cuda_stream_per_thread` ➡️ `cuda::stream_ref{cudaStreamPerThread}`
- Use `.get()` when calling an API that requires a raw `cudaStream_t`, including CUDA runtime, library, CUB, and legacy API boundaries (previously `rmm::cuda_stream_view` used `value()`)
- Use `.sync()` when synchronizing a `cuda::stream_ref` (previously `rmm::cuda_stream_view` used `synchronize()`)
- Update Cython declarations and call sites to pass stream references directly where supported


nirandaperera (approved):
LGTM

pentschev (commented):
Could we update the declaration [`void wait()`](https://github.com/rapidsai/rapidsmpf/blob/728a7192708fef5d69dfc13393b535de25d86003/python/rapidsmpf/rapidsmpf/_detail/cuda_stream_ref.pxd#L12) declaration to `void sync()` as part of the migration? I don't think it's currently used anywhere and probably exposed only for completeness, regardless, I think it makes sense to update it (or remove it completely) rather than having a stale one declaration.

bdice:
Rebased onto preparation PR #1199 and validated against the rebased RMM stack. Local `build-rapidsmpf -j0` completed successfully, and the CUDA stream-join tests passed. CI is rerunning at `11a1656d`.

bdice:
CI follow-up: all four devcontainer failures use the pre-migration RMM Cython declaration where `Stream.view()` returns `cuda_stream_view`, so Cython rejects forwarding it as `stream_ref`. The rebased stack includes RMM #2372, where the declaration is migrated; the full build and both `CudaStreamJoinCppOnly.*` tests pass locally. These failures are expected until the RMM migration is available to CI.

wence-:
> Could we update the declaration [`void wait()`](https://github.com/rapidsai/rapidsmpf/blob/728a7192708fef5d69dfc13393b535de25d86003/python/rapidsmpf/rapidsmpf/_detail/cuda_stream_ref.pxd#L12) declaration to `void sync()` as part of the migration? I don't think it's currently used anywhere and probably exposed only for completeness, regardless, I think it makes sense to update it (or remove it completely) rather than having a stale one declaration.

Done.

wence- (approved):

wence-:
Needs coordiantion with rapidsai/rmm#2372

PR #1182, opened 2026-08-31
Bind `MemoryReservation::split()` to Python

madsbk:
Exposes the `split()` added in #1177 as `MemoryReservation.split(size)`.

Together with the existing `opaque_memory_usage` context manager, callers can scope part of a reservation to the allocation it covers:
```python
with opaque_memory_usage(reservation.split(reorder_bytes)):
    reordered, offsets = plc.partitioning.hash_partition(...)
# the reorder's bytes are released here
```
instead of computing the remainder and releasing it by hand. This lets cudf-polars' shuffle actors drop a helper and its bookkeeping at two call sites.

Also fixes the `opaque_memory_usage` docstring example, which passed an argument the function doesn't take.


pentschev (approved):
LGTM.

madsbk:
Thanks @pentschev

Issue #1170, opened 2026-08-26
Disk-spilling shuffler


madsbk:
Shuffle extraction is often the OOM point. It acts as a barrier that waits on all data before continuing, so at that point a rank holds its entire local share of the shuffle. In RapidsMPF the normal answer is to spill device memory to host, but on very large datasets host memory is limiting too, and there is no tier below it.

We propose an option for the existing `Shuffler` to spill shuffled data to disk, and to unspill it back to host or device only when the user extracts it.

### Proposal

Add a `DiskBuffer`, a file-backed handle to a single spilled chunk. Disk becomes a tier the shuffler can move a chunk into, and it stays entirely outside the `MemoryType` taxonomy.

Concretely, `rapidsmpf::shuffler::detail::Chunk` gains a disk-resident state, `ReceivedChunks::spill` gets a disk fallback instead of a throw, and `Shuffler::extract` rehydrates. Nothing in `BufferResource`, `Buffer`, `MemoryType`, or the general spill machinery changes.

#### Host limit

The `Shuffler` takes a host limit, a byte count or percentage mirroring the existing `spill_device_limit`. Device pressure spills device to host exactly as today, and crossing the host limit is what triggers host to disk. Disk is therefore entered on a threshold the user sets, rather than only once host allocation has already failed.

### Why not `MemoryType::DISK`

A general `MemoryType::DISK` is something we may well want later, but it is a much bigger change than the enum makes it look, and most of the cost lands downstream.

Within RapidsMPF, `Buffer` cannot represent disk. Its storage is a `std::variant` of a device and a host buffer (`buffer.hpp:370`), and `data()`, `write_access()`, and `exclusive_data_access()` all return a raw pointer that a file-backed buffer does not have. Every cross-tier move funnels through `buffer_copy()` (`buffer.cpp:134`), one `cuda_memcpy_async` between two pointers against a single reservation, so disk cannot join that funnel without turning it into a tier dispatch that affects every `BufferResource` user. On top of that, once `DISK` is in `MEMORY_TYPES` it becomes reachable by accident from the sites that pass the whole array as a preference list.

Downstream is the larger cost. `cudf_streaming` has its own exhaustive switch over `rapidsmpf::MemoryType` with a `default: RAPIDSMPF_FAIL`, reached on the spill path, and it sizes an array by `rapidsmpf::MEMORY_TYPES.size()`.

madsbk:
[...]

PR #1186, opened 2026-08-31
Add a KvikIO-backed `DiskResource` for blocking disk spill I/O

### LINKED ISSUE #1183: Add a Disk Memory Resource

nirandaperera:
This adds a public, blocking `rapidsmpf::disk::DiskResource` on KvikIO `CompatMode::AUTO` (GDS when available) and keeps disk outside `MemoryType` / `Buffer` / `BufferResource`.

- `write` / `read` / `flush` take a path, pointer, size, device flag, and optional file offset; callers must sync device pointers before I/O.
- Spill location comes from config option `disk_spill_dir` (`RAPIDSMPF_DISK_SPILL_DIR`); empty uses the system temp directory.
- C++ tests cover host, pinned, and device `Buffer` round-trips, unaligned offsets, flush, and the spill-dir option.
- `libkvikio` is a public dependency of `librapidsmpf` (CMake, conda, DFG / pyproject).

Closes https://github.com/rapidsai/rapidsmpf/issues/1183
Related to #1170


jameslamb (commented):
`@` me when this is ready for a full review (I still see plenty of failing CI here and it's behind `main`, unsure if it's ready yet).

Quick note... if this is merged, please also update RAPIDS dependency graph to reflect that, following these examples: https://github.com/NVIDIA/cuvs/pull/2257#pullrequestreview-5042632463

cpp/src/disk/disk_resource.cpp
```diff
@@ -0,0 +1,133 @@
...
+        "close after fdatasync failed: " + std::string{std::strerror(errno)},
+        std::runtime_error
+    );
+}
+
+std::filesystem::path default_spill_directory(config::Options options) {
+    return options.get<std::filesystem::path>(
+        "disk_spill_dir", [](std::string const& value) {
+            if (value.empty()) {
+                return std::filesystem::temp_directory_path();
```
wence-, line 126: Often this will be a ramdisk, which seems like a bad default.

cpp/include/rapidsmpf/disk/disk_resource.hpp
```diff
@@ -0,0 +1,135 @@
...
+    [[nodiscard]] std::unique_ptr<DiskFuture> read(
+        std::filesystem::path const& path,
+        void* data,
+        std::size_t size,
+        MemoryType mem_type,
+        std::size_t file_offset = 0
+    );
+
+    /**
+     * @brief Durably synchronize file data to storage (fdatasync).
```
pentschev, line 116:
```suggestion
     * @brief Durably synchronize file data to storage.
```

pentschev: Let's not add implementation details on the docstrings.

cpp/include/rapidsmpf/disk/disk_resource.hpp
```diff
@@ -0,0 +1,135 @@
...
+     *
+     * Not used on the default spill path; exposed for benchmark durability cases.
+     *
+     * @param path File path.
+     */
+    void flush(std::filesystem::path const& path);
+};
+
+/**
+ * @brief Spill directory from `disk_spill_dir` (`RAPIDSMPF_DISK_SPILL_DIR`).
```
pentschev, line 126:
```suggestion
 * @brief Spill directory from `disk_spill_dir`.
```

pentschev: I don't think we need to list all possible ways to set a configuration, users should refer to the docs.

cpp/src/disk/disk_resource.cpp
```diff
@@ -0,0 +1,133 @@
...
+std::unique_ptr<DiskFuture> DiskResource::write(
+    std::filesystem::path const& path,
+    void const* data,
+    std::size_t size,
+    [[maybe_unused]] MemoryType mem_type,
+    std::size_t file_offset
+) {
+    auto file = std::make_unique<kvikio::FileHandle>(
+        path.string(), "w+", kvikio::FileHandle::m644, kvikio::CompatMode::AUTO
+    );
```
pentschev, line 70:
```suggestion
  auto const flags = std::filesystem::exists(path) ? "r+" : "w+";
  auto file = std::make_unique<kvikio::FileHandle>(
      path.string(), flags, kvikio::FileHandle::m644, kvikio::CompatMode::AUTO);
```

pentschev: As written, `"w+"` enables `O_TRUNC`, so every `write()` first truncates the path and a later `pwrite()` cannot retain bytes written by an earlier call. KvikIO supports `"r+"` for read/write access without truncation, so selecting `"r+"` when the spill file already exists (and `"w+"` only for initial creation) would preserve offset writes. I think the proposal above fixes nonzero offset writes.

Please also add a test that writes two distinct ranges to one path and verifies both remain intact.

pentschev (changes requested):

cpp/src/disk/disk_resource.cpp
pentschev: Why would this be a ramdisk? Wouldn't a ramdisk effectively be a proxy for spill-to-host?

pentschev: Ah nevermind, I misunderstood your statement. `temp_directory_path()` will often be a ramdisk, you're not saying users would often be a ramdisk. I agree with your statement.

nirandaperera:
superseded by #1195

pentschev:
@nirandaperera really? You just close it, say it's superseded by #1195 and ignore reviews/change requests that were made here? Please address them in #1195.

pentschev:
> @nirandaperera really? You just close it, say it's superseded by #1195 and ignore reviews/change requests that were made here? Please address them in #1195.

Sorry Niranda, I didn't mean to be rude, but please do address the reviews in #1195 . :slightly_smiling_face:

cpp/src/disk/disk_resource.cpp
nirandaperera: Ah good catch. thanks

cpp/src/disk/disk_resource.cpp
nirandaperera: This is no longer valid. disk_spill_dir is optional and disk resource is explicitly opt-in

PR #1190, opened 2026-09-01
Avoid spilling during memory reservation admission

pentschev:
## Summary

Avoid spilling queued streaming data from `MemoryReserveOrWait` when a device-memory reservation cannot be admitted immediately, partially reverting changes from #1164 .

The reservation flow continues to use reservation-aware availability, request ordering, smallest-net-delta selection, and the existing progress timeout implemented in #1164 . When no request fits, it now waits for a reservation release. At the timeout, it preserves the existing bounded-progress behavior: callers that allow overbooking receive a zero-size reservation and decide how to proceed, while non-overbooking callers retain their established failure path.

The change adds regression coverage that verifies neither ordinary polling nor the progress-timeout path invokes a spill function.

## Motivation and observations

In an isolated SF1000 TPC-H Q1 in an AWS g7e.8xlarge instance, admission-triggered spilling was associated with substantially higher device-memory residency and slower execution. Spilling queued messages can cause their rematerialization to overlap with incoming Parquet reads, increasing peak residency under this high-concurrency workload.

| Configuration | Q1 wall time | Peak observed device use | Result |
|---|---:|---:|---|
| Before this change | 32.84 s | 100.91 GB | Success with low headroom |
| Before this change, repeat | N/A | 102.63 GB | CUDA OOM |
| With this change | 10.64 s | 75.91 GB (70.70 GiB) | Success |

Relative to the successful before-change observation, this reduced observed runtime by 67.6% and peak device use by 25.0 GB (24.8%).

The experiment used one SPMD worker on SF1000 TPC-H Q1 with S3 input, pinned memory and a zero-byte initial pool, four streaming threads, a 4.2-GB Parquet target/pass limit, eight concurrent I/O tasks, and periodic spill checking disabled.

This is targeted at preventing `MemoryReserveOrWait` from directly evicting queued streaming data, it does not imply that reservation-aware availability accounting is incorrect or prescribe broader spill-policy defaults.

### DIFF
--- a/cpp/src/streaming/core/memory_reserve_or_wait.cpp
+++ b/cpp/src/streaming/core/memory_reserve_or_wait.cpp
@@ -4,7 +4,6 @@
  */
 
 #include <algorithm>
-#include <atomic>
 #include <memory>
 #include <mutex>
 #include <ranges>
@@ -203,36 +202,6 @@ coro::task<void> MemoryReserveOrWait::periodic_memory_check() {
         RAPIDSMPF_EXPECTS(err, "cannot spawn push-into-queue task");
     };
 
-    // Helper that spills until `headroom` bytes are reservable, swallowing a spill
-    // function that throws and logging it once per task. Returns false only when a
-    // non-blocking spill found the spill lock held.
-    bool spill_failure_logged = false;
-    auto make_headroom = [&](std::size_t headroom, bool blocking) -> bool {
-        // Only device memory is supported, see `spill_to_make_headroom()`.
-        if (mem_type_ != MemoryType::DEVICE) {
-            return true;
-        }
-        try {
-            auto const target = safe_cast<std::int64_t>(headroom);
-            if (blocking) {
-                br_->spill_manager().spill_to_make_headroom(target);
-                return true;
-            }
-            return br_->spill_manager().try_spill_to_make_headroom(target).has_value();
-        } catch (...) {
-            if (!std::exchange(spill_failure_logged, true)) {
-                try {
-                    logger_->warn(
-                        "a spill function threw while making headroom for ",
-                        format_nbytes(headroom)
-                    );
-                } catch (...) {  // NOLINT(bugprone-empty-catch)
-                }
-            }
-            return true;
-        }
-    };
-
     // RAII helper that releases `periodic_task_running_` when this task exits without
     // reaching one of the `co_return` paths below, such as on an exception. Those paths
     // release the flag under the same lock acquisition that observes the empty request
@@ -256,9 +225,6 @@ coro::task<void> MemoryReserveOrWait::periodic_memory_check() {
 
     while (true) {
         auto last_reservation_success = Clock::now();
-        // Spilling is attempted once per timeout window, re-armed by each served
-        // request.
-        bool spill_attempted = false;
         while (true) {
             // Exit if no more pending requests remain.
             {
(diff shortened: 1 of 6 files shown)


cpp/src/streaming/core/memory_reserve_or_wait.cpp
```diff
@@ -283,18 +249,9 @@ coro::task<void> MemoryReserveOrWait::periodic_memory_check() {
...
-                    auto const target = std::ranges::min_element(
-                        reservation_requests_, std::less<>{}, &Request::net_memory_delta
-                    );
-                    auto const headroom = target->size;
-                    // Spill without holding the mutex, the spill functions take the
-                    // buffer resource's lock.
-                    lock.unlock();
-                    spill_attempted = make_headroom(headroom, /* blocking = */ false);
-                }
+                // Nothing currently fits. Preserve resident data while ordinary
```
madsbk, line 249: Need to update the docs, now that we don't spill anymore.

cpp/src/streaming/core/memory_reserve_or_wait.cpp
```diff
@@ -203,36 +202,6 @@ coro::task<void> MemoryReserveOrWait::periodic_memory_check() {
...
-            auto const target = safe_cast<std::int64_t>(headroom);
-            if (blocking) {
-                br_->spill_manager().spill_to_make_headroom(target);
-                return true;
-            }
-            return br_->spill_manager().try_spill_to_make_headroom(target).has_value();
-        } catch (...) {
-            if (!std::exchange(spill_failure_logged, true)) {
-                try {
-                    logger_->warn(
```
madsbk, line 225: The `logger_` is never used anymore, should we make a breaking change and remove `std::shared_ptr<Logger>` from the ctor?

cpp/tests/streaming/test_memory_reserve_or_wait.cpp
```diff
@@ -355,13 +357,30 @@ TEST_P(StreamingMemoryReserveOrWait, SpillingUnblocksWaiter) {
...
+        }
+    }(mrow));
+    std::thread thd(run_actor_network, std::move(actors));
+
+    // Let the poller observe the pending request several times while the timeout
+    // remains far away. The current eager #1164 path spills here; the desired path
+    // leaves the request pending.
+    auto const deadline = Clock::now() + std::chrono::seconds(1);
+    while (mrow.size() < 1 && spills.amounts().empty() && Clock::now() < deadline) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
madsbk: I really dislike these non-deterministic tests. Can we avoid this, even if it means slightly less coverage?

cpp/src/streaming/core/memory_reserve_or_wait.cpp
pentschev: Yes, removed now. There were no external users (in the cudf repository) of that, so I think that breaking change is fine.

cpp/src/streaming/core/memory_reserve_or_wait.cpp
pentschev: Done.

cpp/tests/streaming/test_memory_reserve_or_wait.cpp
pentschev: Agreed. I've changed to a condition-driven test instead of time-driven, which is not strictly deterministic but I think matches your expectation. Let me know if you think otherwise.

pentschev (commented):
Thanks @madsbk , I've addressed your comments, would you mind having another look?

wence- (approved):
I think this looks fine: we need to understand more holistically the interaction of the spilling decisions and consequences on memory usage/live set, but we can do so separately.

PR #1191, opened 2026-09-01
Add C++ API reference to Sphinx docs

bdice:
## Description

Expose the librapidsmpf C++ API in the Sphinx documentation through Breathe, organized by subsystem. Normalize Doxygen 1.9.1 XML that Breathe cannot parse and resolve generated C++ references so the documentation builds without warnings.

## API impact

No library API changes; this only changes generated documentation.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -11,7 +11,10 @@
 from __future__ import annotations
 
 import datetime
+import glob
 import os
+import re
+import xml.etree.ElementTree as ET
 
 from enum import IntEnum, IntFlag
 from typing import Any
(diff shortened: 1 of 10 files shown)


wence- (approved):

vyasr (approved):
The source changes look fine. The helper functions in conf.py look like they were copied over from cudf/rmm and it would probably be good to validate that 1) they do everything that rpaidsmpf needs, and 2) they don't do anything extra, but neither is critical for the first pass PR. We can always improve in a follow-up. Unfortunately I cannot actually review the preview right now due to our build settings so I suggest we just merge and check the results once they're published.

PR #1192, opened 2026-09-01
Remove obsolete devcontainer dependency list

bdice:
Remove the redundant `devcontainers` dependency file list. Development containers can use the default `all` key now that the Ray and PyTorch dependencies resolve together.

Depends on rapidsai/devcontainers#761.


gforsyth (approved):
Appoving so this can be merged once https://github.com/rapidsai/devcontainers/pull/761 is in

PR #1193, opened 2026-09-01
Add memory limiting capability to host memory resource

nirandaperera:
Pageable-host allocations were not tracked, and pinned-pool limits were not reflected in reservation accounting, allowing reservations beyond the configured pool maximum. This adds the host-memory accounting and limit handling needed for disk-spill decisions.

- Generalize the resource adaptor to track pageable-host allocations and include them in statistics.
- Add an absolute `spill_host_limit`, unbounded by default, and derive the pageable-host budget after accounting for bounded pinned memory.
- Fix pinned-memory configuration by wiring `pinned_max_pool_size` into reservation limits, supporting `disabled` as unbounded, and rejecting zero or incompatible limits.
- Update Python bindings, documentation, and C++/Python test coverage.

This is related to #1170. IMO it would be a nice to have change in general. 

## Option Defaults 

| Option | Default | Effect |
|:---:|:---:|---|
| `pinned_memory` | false | No pinned pool. PINNED_HOST limit is 0. |
| `pinned_initial_pool_size` | 0% | Used only if pinned is on. |
| `pinned_max_pool_size` | 80% | Used only if pinned is on. 80% of per-GPU host memory. |
| `spill_host_limit` | disabled | Unbounded pageable HOST. |

## Effect on PINNED_HOST and HOST 

Following table shows the effect on memory limits based on different option settings. 

| `pinned_memory` | `pinned_max_pool_size` | `spill_host_limit` | `PINNED_HOST` limit | `HOST` limit | Checks |
|:---:|:---:|:---:|:---:|:---:|---|
| off | - | disabled | `0` | unlimited |  |
| off | - | set | `0` | set |  |
| on | disabled | disabled | unlimited | unlimited |  |
| on | disabled | set | unlimited | set |  |
| on | set | disabled | set | unlimited | `pinned_max <= numa_host` |
| on | set | set | set | set | `pinned_max <= numa_host`, then `spill_host_limit <= total_numa_host - pinned_max` |

- `numa_host` = host memory of current numa node 
- `total_numa_host` = host memory of current numa nodes (governed by numa policy setting)
- If `pinned_max_pool_size` is set, it should be greater-than zero

## Indicative values 

|  | GB200 | DGX B200 | g7e.8xlarge |
|:---:|:---:|:---:|:---:|
| host mem/node | 960GB | 2TB | 256 GB |
| GPUs/node | 4 | 8 | 1 |
| host mem/GPU | 240GB | 256GB | 256GB |
|80% host mem/GPU | 192GB | 204GB | 204GB |

Ideally, `pinned_max_pool_size + spill_host_limit <= 80% host_mem_per_gpu`


cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -630,6 +630,22 @@ static_assert(StatisticsProvider<BufferResource>);
...
+/**
+ * @brief Parse the `spill_host_limit` parameter from configuration options.
+ *
+ * The limit must be an absolute byte count. Disabled values produce an
+ * unbounded pageable-host budget. This limit is independent of
+ * `pinned_max_pool_size`. When both limits are bounded,
+ * `BufferResource::from_options()` rejects configurations where their sum
+ * exceeds the summed host memory of the nodes in the calling thread's memory
+ * policy. The pinned maximum is also constrained by the host memory of its
+ * NUMA node.
```
pentschev: So now users have two limits they have to independently set not to cause a host OOM? If so, I don't think this is a good UX, it will be difficult for users to ensure they're both within safe limits.

Also, while the sum of the two is in theory correct, in practice this is an unfeasible limit, since OS alone takes non-negligible amounts of memory. I don't know what would be a safe value, but perhaps looking at the amount of _free_ memory when the process is starting minus some percentage is more reasonable.

cpp/src/memory/buffer_resource.cpp
```diff
@@ -100,12 +107,47 @@ std::shared_ptr<BufferResource> BufferResource::from_options(
...
+
+        if (host_limit != unbounded && pinned_limit != unbounded) {
+            auto const numa_host = safe_cast<std::int64_t>(
+                get_numa_node_host_memory(pinned_pool_properties->numa_id)
+            );
+            RAPIDSMPF_EXPECTS(
+                pinned_limit <= numa_host,
+                "pinned_max_pool_size exceeds NUMA node host memory",
+                std::invalid_argument
+            );
```
pentschev: Could we move the `pinned_limit <= numa_host` validation outside this combined bounded-host/bounded-pinned condition, and mirror that change in the Python `from_options()` path? As written, `pinned_max_pool_size` can be set above its NUMA node’s host memory whenever `spill_host_limit` is left at its default `"disabled"`. That contradicts the new configuration documentation and accepts an incompatible pool configuration, only the combined pageable+pinned budget check needs both limits to be bounded. Please add also coverage for a bounded oversized pinned limit with `spill_host_limit=disabled`.

docs/source/configuration.md
```diff
@@ -125,6 +126,19 @@ rapidsmpf::config::Options options{rapidsmpf::config::get_environment_variables(
...
+  - **Default**: disabled (unbounded)
+  - **Description**: Soft upper limit on pageable host memory configured for
+    RapidsMPF, independent of `pinned_max_pool_size`. When both limits are
+    bounded, their sum cannot exceed the summed host memory of the nodes in the
+    calling thread's memory policy; the pinned maximum also cannot exceed its
+    NUMA node's host memory. This is a coarse validation because other processes
+    and allocations may consume the same host memory. It accepts absolute byte
+    counts (for example, `"10GiB"` or `"512MB"`). Percentages are not supported
+    because the appropriate host-memory share depends on the job's process and
+    NUMA topology.
```
pentschev: This looks closer to proper docstrings than that from `host_limit_from_options`. However, those same issues are still relevant, perhaps taking into account the free memory is more realistic. I also don't understand how depending on job's processes/NUMA topology voids the potential for using percentages.

python/rapidsmpf/rapidsmpf/memory/buffer_resource.pyx
```diff
@@ -277,16 +279,44 @@ cdef class BufferResource:
         # means pinned host memory is disabled.
         cdef optional[cpp_PinnedPoolProperties] props = \
             pinned_pool_properties_from_options(options._handle)
+        unbounded = 2**63 - 1
```
pentschev:
```python
from libc.stdint cimport INT64_MAX, int64_t
...
cdef int64_t unbounded = INT64_MAX
```

pentschev (changes requested):

cpp/include/rapidsmpf/memory/buffer_resource.hpp
nirandaperera: I see your point. 
One issue I had was, what should be the max limit for HOST memory? 
In multi GPU nodes, I think its correct to limit PINNED to `host_memory_per_gpu` because a pool can only be assigned to a single numa ID. But HOST can span across numa nodes. 
But since we are running with `rrun` (or with `apply_bindings`), I think the processes will have `MPOL_BIND` set to proper numa nodes. 
So, in that case, `HOST + PINNED < get_numa_node_host_memory(get_current_numa_node())` IINM. 

Getting free memory is a reasonable idea. let me look at that

cpp/src/memory/buffer_resource.cpp
nirandaperera: @pentschev I am not sure if I understood you correctly. `spill_host_limit`=="disabled" implies that host memory is unbounded. if `pinned_pool_max_size` is not set, but `spill_host_limit` is set, we cant reliably ensure `host_limit <= total_host - pinned_limit` during BR creation.

nirandaperera:
@pentschev I thought the `spill_host_limit` value representation a bit more. 

I think we can not reliably use a % of available/ free memory. It may give use an earlier signal before an OOM, but
- This changes the current default behavior 
- In a multiprocess run, % of free memory is always a moving target during startup, unless some process determines this and broadcasts. 

So, I decided to leave `spill_host_limit` as a value-only option (no percentages), that defaults to `disabled` (to preserve the current default behavior). I updated the PR description to explain the all option permutations and the checks that were put in place.

pentschev:
> * In a multiprocess run, % of free memory is always a moving target during startup, unless some process determines this and broadcasts.

Free memory is a moving target at any point in time, I don't get it why taking a snapshot once we hit the point where we need to configure it is not as reliable as we could make it, presumably after that point no external applications will be running.

> So, I decided to leave `spill_host_limit` as a value-only option (no percentages), that defaults to `disabled` (to preserve the current default behavior). I updated the PR description to explain the all option permutations and the checks that were put in place.

I'm also not sure how a value-only makes anything more reliable than percentages, but ok. We should at minimum document what we _think_ are best practices here. Obviously 100% isn't, a relative 99% (that now is up to the user to measure) isn't either, but what is? The fact that we don't know is to me telling that users will have a hard time deciding that and are highly likely to make bad and uninformed decisions, that's why I thought a reasonable initial default would be something like "90% of free NUMA-node memory divided by the number of ranks on that NUMA node".

cpp/src/memory/buffer_resource.cpp
pentschev: Sorry, I was referring to two separate validations:

1. Whenever `pinned_pool_max_size` is explicitly bounded, validate that it does not exceed the selected NUMA node’s host memory. This check is independent of `spill_host_limit`: even when pageable host spilling is unlimited, a pinned-pool cap larger than the backing NUMA node is not a valid configuration.
2. Only when both `pinned_pool_max_size` and `spill_host_limit` are bounded, validate their combined budget against the host memory available under the current NUMA policy.

I agree that when `pinned_pool_max_size` is unset/unbounded, we cannot perform the second check, even if `spill_host_limit` is bounded. That is not what I was asking for. The missing case was specifically: bounded `pinned_pool_max_size` + `spill_host_limit=disabled` should still reject a pinned limit larger than its NUMA node.

PR #1195, opened 2026-09-03
Add disk-backed buffer storage

### LINKED ISSUE #1184: Add a disk buffer

nirandaperera:
Add KvikIO-backed disk storage so buffers can move between addressable memory and per-process spill files through the existing reservation and copy APIs.

- Add KvikIO build, packaging, and environment dependencies.
- Introduce `DiskResource`, `DiskBuffer`, and `disk_spill_dir` configuration.
- Extend `MemoryType`, `Buffer`, and `BufferResource` with unbounded disk reservations and synchronous memory-to-disk copies.
- Add disk-copy statistics, documentation, Python enum exposure, and test coverage.

Closes #1184 
Depends on #1186 
Related #1170


cpp/include/rapidsmpf/disk/disk_buffer.hpp
```diff
@@ -0,0 +1,125 @@
...
+ * `from_buffer` and `restore` are synchronous: they block until the disk
+ * transfer completes.
+ */
+class DiskBuffer {
+  public:
+    /**
+     * @brief Move constructor.
+     *
+     * Transfers ownership of the backing file. The moved-from object is empty
+     * (`path()` is empty and `size()` is zero).
```
pentschev:
```suggestion
     * Transfers ownership of the backing file.
```

pentschev: Seems unnecessarily verbose docstring for a mover.

cpp/include/rapidsmpf/disk/disk_buffer.hpp
```diff
@@ -0,0 +1,125 @@
...
+     * @brief Restore bytes from a disk handle into a newly allocated `Buffer`.
+     *
+     * Takes ownership of @p source, allocates from @p reservation on @p stream,
+     * reads the backing file with blocking disk I/O, and records the fill as a
+     * stream-ordered write on the destination buffer. The backing file is
+     * deleted when @p source is destroyed.
+     *
+     * @param source Disk-resident handle whose bytes are restored.
+     * @param reservation Memory reservation covering at least `source->size()`
+     * bytes.
```
pentschev, line 97: Why _at least_, and not _exactly_?

cpp/src/disk/disk_buffer.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+    if (nbytes == 0) {
+        return make_disk_buffer(0);
+    }
+
+    RAPIDSMPF_EXPECTS(
+        source->is_latest_write_done(),
+        "cannot write buffer to disk with pending stream-ordered writes",
+        std::logic_error
+    );
```
pentschev: Could we validate `source->is_latest_write_done()` before calling `create_unique_path()`? As written, a buffer with pending stream-ordered writes throws here after `path` has already been created, and thus not cleaned.

cpp/include/rapidsmpf/disk/disk_resource.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     * @brief Whether the transfer has completed without consuming the result.
+     *
+     * @return True if the I/O finished and `get()` has not been called yet.
+     *
+     * @note This is a point-in-time check and is subject to TOCTOU races:
+     *       another thread may call `get()` after this returns `true`. Like
+     *       `std::future`, concurrent `is_ready()`/`get()` from multiple
+     *       threads is undefined behavior.
```
pentschev:
```suggestion
     * @note This is a point-in-time check and is subject to TOCTOU races:
     * another thread may call `get()` after this returns `true`. Like
     * `std::future`, concurrent `is_ready()`/`get()` from multiple
     * threads is undefined behavior.
```

pentschev: Style.

cpp/include/rapidsmpf/disk/disk_resource.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     * @brief Wait for the transfer, release backend resources, and return the
+     * byte count.
+     *
+     * @return Number of bytes transferred. The caller must check this against
+     *         the requested size.
+     *
+     * @note Like `std::future::get()`, this is not thread-safe. Calling get()
+     *       concurrently from multiple threads is undefined behavior.
```
pentschev:
```suggestion
     * @return Number of bytes transferred. The caller must check this against
     * the requested size.
     *
     * @note Like `std::future::get()`, this is not thread-safe. Calling get()
     * concurrently from multiple threads is undefined behavior.
```

cpp/include/rapidsmpf/disk/disk_resource.hpp
```diff
@@ -0,0 +1,187 @@
...
+     * @brief Whether the transfer has completed without consuming the result.
+     *
+     * @return True if the I/O finished and `get()` has not been called yet.
+     *
+     * @note This is a point-in-time check and is subject to TOCTOU races:
+     *       another thread may call `get()` after this returns `true`. Like
+     *       `std::future`, concurrent `is_ready()`/`get()` from multiple
+     *       threads is undefined behavior.
+     */
+    [[nodiscard]] virtual bool is_ready() const = 0;
```
pentschev: This seems to have been added by this PR (not included in #1186 that introduced this class) and is only used by `DiskFutureIsReadyBeforeGet`, do we really need it just to satisfy testing?

cpp/src/memory/buffer_resource.cpp
```diff
@@ -47,6 +53,9 @@ BufferResource::BufferResource(
     }
     RAPIDSMPF_EXPECTS(stream_pool_ != nullptr, "the stream pool pointer cannot be NULL");
     RAPIDSMPF_EXPECTS(statistics_ != nullptr, "the statistics pointer cannot be NULL");
+    RAPIDSMPF_EXPECTS(
+        disk_resource_ != nullptr, "the disk resource pointer cannot be NULL"
+    );
```
pentschev: Why not? Isn't that effectively forcing us to enable disk spilling, even when that's not wanted?

cpp/src/memory/buffer_resource.cpp
```diff
@@ -71,13 +81,18 @@ std::shared_ptr<BufferResource> BufferResource::create(
         pinned_mr = PinnedMemoryResource{*pinned_pool_properties};
     }
 
+    // create a dir for each pid under the spill directory
+    std::shared_ptr<disk::DiskResource> disk_res{
+        new disk::DiskResource{std::move(spill_directory) / std::to_string(::getpid())}
+    };
```
pentschev: And also forcing creating a spilling directory?

cpp/include/rapidsmpf/disk/disk_buffer.hpp
nirandaperera: We can have a larger reservation which can be passed on to several allocations.

cpp/src/disk/disk_buffer.cpp
nirandaperera: good catch. thanks

cpp/src/memory/buffer_resource.cpp
nirandaperera: fair point. I will guard this with an opt in option

ci/build_wheel_librapidsmpf.sh
```diff
@@ -15,7 +15,7 @@ rapids-dependency-file-generator \
   --output requirements \
   --file-key "py_build_${package_name}" \
   --file-key "py_rapids_build_${package_name}" \
-  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION};cuda_suffixed=true" \
+  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION};cuda_suffixed=true;use_cuda_wheels=true" \
```
nirandaperera, line 18: `use_cuda_wheels=true` without this, kvikio cuda13 libs are being installed for cuda12 builds

madsbk: why do we need this now?

nirandaperera: For some reason, libkvikio cuda13 wheels gets installed in cuda12 env without this. Previously, we didnt have kvikio in the rapidsmpf buildl

nirandaperera:
I reorg'ed the PR into 3 commits to make this a little more easy to review 
- [0ba3a2a](https://github.com/rapidsai/rapidsmpf/pull/1195/commits/0ba3a2a7747f7bcaa82f8774bdbf040179dc74fd) build changes related to kvikio 
- [b393b41](https://github.com/rapidsai/rapidsmpf/pull/1195/commits/b393b417e3e7be12a226a338ade67eb1cf23d87f) new DISK mem type 
- [dd702eb](https://github.com/rapidsai/rapidsmpf/pull/1195/commits/dd702eb01e04646a402c4c422b32a7c2f6ff0226) disk resource, buffer, testing etc

nirandaperera:
@jameslamb I closed #1186 in favor of this PR. This includes all the build/cmake changes.  Could you please take a look? all changes related to build should be in https://github.com/rapidsai/rapidsmpf/pull/1195/commits/0ba3a2a7747f7bcaa82f8774bdbf040179dc74fd commit

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -592,6 +603,19 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
      */
     std::shared_ptr<Statistics> statistics() const noexcept;
 
+    /**
+     * @brief Disk I/O resource and spill directory configuration.
+     *
+     * Returned as a `std::shared_ptr`; `DiskBuffer`s keep a copy so the disk
+     * resource outlives those buffers if this `BufferResource` is destroyed.
+     *
```
madsbk:
```suggestion
```

cpp/include/rapidsmpf/memory/memory_type.hpp
```diff
@@ -16,17 +16,26 @@ namespace rapidsmpf {
...
 };
 
 /// @brief All memory types sorted in decreasing order of preference.
-constexpr std::array<MemoryType, 3> MEMORY_TYPES{
+constexpr std::array<MemoryType, 4> MEMORY_TYPES{
+    {MemoryType::DEVICE, MemoryType::PINNED_HOST, MemoryType::HOST, MemoryType::DISK}
+};
+
+/// @brief Memory types that support `BufferResource::reserve`.
+constexpr auto RESERVABLE_MEMORY_TYPES = MEMORY_TYPES;
```
madsbk: I don't think we need RESERVABLE_MEMORY_TYPES. It is misleading, since neither PINNED_HOST nor DISK is guaranteed to be reservable, and it is a literal alias of MEMORY_TYPES with one use.

cpp/include/rapidsmpf/memory/memory_type.hpp
```diff
@@ -16,17 +16,26 @@ namespace rapidsmpf {
...
+
+/// @brief Memory types that expose an addressable host or device pointer.
```
madsbk: Add a note that this array is ordered

cpp/src/disk/disk_resource.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+#include <kvikio/compat_mode.hpp>
+#include <kvikio/file_handle.hpp>
+
+#include <rapidsmpf/config.hpp>
+#include <rapidsmpf/disk/disk_resource.hpp>
+#include <rapidsmpf/error.hpp>
+#include <rapidsmpf/utils/string.hpp>
+
+namespace rapidsmpf::disk {
```
madsbk: I don't think we need a new namespace? We don't have one for host, pinned-host, or device?

cpp/src/disk/disk_resource.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+namespace {
+
+std::size_t wait_io(kvikio::FileHandle& file, std::future<std::size_t> io) {
```
madsbk: No need for a `wait_io`, `kvikio::FileHandle` already close the file on exit.

cpp/src/memory/buffer.cpp
```diff
@@ -125,9 +158,79 @@ void Buffer::rebind_stream(cuda::stream_ref new_stream) {
...
-    std::visit([&](auto&& storage) { storage->set_stream(new_stream); }, storage_);
+    std::visit(
+        [&](auto&& storage) {
+            if constexpr (!std::is_same_v<std::decay_t<decltype(storage)>, DiskBufferT>) {
+                storage->set_stream(new_stream);
+            }
+        },
+        storage_
+    );
+}
```
madsbk, line 159: I don't think we should change the stream semantics for disk buffers. The thing that is genuinely different about disk is that it has no data pointer, which `ADDRESSABLE_MEMORY_TYPES` already captures.

Being synchronous does not conflict with the stream contract. A blocking call trivially satisfies it: just record the event after the call returns. From the user's perspective, the stream associated with a disk-backed `Buffer` should be handled the same way as for a device or host buffer.

Concretely, record `latest_write_event_` in the disk constructor as the others do, give `DiskBuffer` a no-op `set_stream()` so `rebind_stream()` no longer needs the `if constexpr`, and narrow the disk branches in `buffer_copy()` to only choosing how the bytes are moved. Event recording and statistics can then remain common to all paths.

Keeping a wall-clock duration for copy timing seems right, since `StreamOrderedTiming` would be noisy around host-blocking work. But that is an argument about how we time the copy, not about whether a disk `Buffer` should behave like a `Buffer`.

cpp/src/memory/buffer.cpp
```diff
@@ -125,9 +158,79 @@ void Buffer::rebind_stream(cuda::stream_ref new_stream) {
...
+    std::size_t size,
+    std::ptrdiff_t destination_offset,
+    std::ptrdiff_t source_offset
+) {
+    RAPIDSMPF_EXPECTS(
+        source.disk_resource() != nullptr, "DiskBuffer has no DiskResource"
+    );
+
+    auto const start = Clock::now();
+    destination.write_access([&](std::byte* destination_data, cuda::stream_ref) {
```
madsbk: `copy_from_disk()` should use `exclusive_data_access()` rather than `write_access()`, which is what the blocking MPI and UCXX paths use.

Once disk buffers have the same stream semantics as the other buffer types (see my stream-semantics comment), all that is needed here is the pointer. `exclusive_data_access()` is therefore a better fit, and it makes the latest-write precondition explicit and checked rather than relying on the caller to have already host-waited.

It should also use an unlock guard so that a short read does not leave the buffer locked.

cpp/src/memory/buffer.cpp
```diff
@@ -152,12 +255,33 @@ void buffer_copy(
...
+        return;
+    }
+    if (src_is_disk) {
+        dst.latest_write_event().host_wait();
+        auto const duration = copy_from_disk(
+            dst, *src.get_storage<Buffer::DiskBufferT>(), size, dst_offset, src_offset
+        );
+        statistics->record_copy(src.mem_type(), dst.mem_type(), size, duration);
+        return;
+    }
```
madsbk, line 279: This can also be simplified once disk buffers have the same stream semantics as the other buffer types (see my stream-semantics comment). Both branches end in the same `record_copy(...); return;` pair as the stream path below, so the tail could be shared rather than written three times.

They are also already inconsistent about the event: `src_is_disk` records `dst`'s event as a side effect of `write_access`, while `dst_is_disk` records nothing, and the disk constructor does not record either, so a disk `Buffer`'s `latest_write_event_` is never recorded at all. It goes unnoticed because synchronizing an unrecorded event is a no-op.

The timing would still fork between `StreamOrderedTiming` and the wall-clock `Duration`, which seems right.

madsbk (changes requested):
First pass, overall looks

ci/build_wheel_librapidsmpf.sh
madsbk: Alright, we need someone from `packaging-codeowners` to review it.

wence-: @rapidsai/packaging-codeowners

cpp/include/rapidsmpf/disk/disk_buffer.hpp
```diff
@@ -0,0 +1,87 @@
...
+    DiskBuffer& operator=(DiskBuffer&& other) = delete;  ///< Not move-assignable.
+    DiskBuffer(DiskBuffer const&) = delete;  ///< Not copyable.
+    DiskBuffer& operator=(DiskBuffer const&) = delete;  ///< Not copy-assignable.
+
+    /**
+     * @brief Current size of the backing file.
+     *
+     * @return Backing file size in bytes.
+     */
+    [[nodiscard]] std::uintmax_t file_size() const;
```
wence-: Why is this not just "size"? Is somehow the size of a buffer different from the size of its backing file?

cpp/include/rapidsmpf/disk/disk_resource.hpp
```diff
@@ -0,0 +1,143 @@
...
+     * @param file_offset Byte offset within the file.
+     * @return Number of bytes transferred. The caller must check this against
+     *         @p size.
+     */
+    [[nodiscard]] std::size_t write(
+        std::filesystem::path const& path,
+        void const* data,
+        std::size_t size,
+        MemoryType mem_type,
+        std::size_t file_offset = 0
```
wence-: Surely `offset_t`?

cpp/include/rapidsmpf/memory/buffer.hpp
```diff
@@ -362,22 +404,34 @@ class Buffer {
...
+ * @brief Copy data between buffers.
  *
- * @note The copy is stream-ordered on @p dst's stream, correct cross-stream ordering
- * between @p src's stream and @p dst's stream is provided automatically.
+ * @note Copies between in-memory buffers are stream-ordered on @p dst's stream with
+ * automatic cross-stream ordering between @p src and @p dst. Copies involving disk
+ * storage are synchronous and block until the transfer completes; neither @p src nor
+ * @p dst is consumed.
  *
  * Copies @p size bytes from @p src, starting at @p src_offset, into @p dst at
```
wence-, line 481: Like Mads, I think we should make a disk buffer behave "as if" stream-ordered.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -592,6 +603,19 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
...
+     * @return Shared pointer to the disk resource, or `nullptr` if no spill
+     * directory was configured.
+     */
+    [[nodiscard]] std::shared_ptr<disk::DiskResource> disk_resource() const {
```
wence-: nit, why not `shared_ptr<...> const&`?

cpp/src/disk/disk_buffer.cpp
```diff
@@ -0,0 +1,53 @@
...
+    if (!path_.empty()) {
+        std::error_code ec;
+        std::filesystem::remove(path_, ec);
+        path_.clear();
+    }
+}
+
+DiskBuffer::~DiskBuffer() {
+    deallocate();
+}
```
wence-, line 67: discussion: I think this is likely the wrong model.

Suppose we have a shared file system that all (or some subset of the processes) can see. If we want to "send" a `DiskBuffer` between two processes then we "only" need to fdatasync on the sender, and inform the receiver of the path.

But the model here doesn't really permit that, because the sender is not allowed to forget about the `DiskBuffer` until the receiver has used it (which is an unbounded time).

pentschev (changes requested):
Please address the ignored reviews from https://github.com/rapidsai/rapidsmpf/pull/1186 in here.

ci/build_wheel_librapidsmpf.sh
gforsyth: Is there a CI run where this happened?  This should not be required

nirandaperera: Let's see how this goes https://github.com/rapidsai/rapidsmpf/pull/1195/commits/fe5d7f5bc3c12ebf304d68d6dc1ca0f8b4bad77b

cpp/src/memory/buffer.cpp
nirandaperera: Okay, I will add a no-op `set_stream()` to disk buffer

cpp/src/memory/buffer.cpp
nirandaperera: fair point.

ci/build_wheel_librapidsmpf.sh
nirandaperera: @gforsyth Here we go https://github.com/rapidsai/rapidsmpf/actions/runs/35112527309/job/104850157795?pr=1195

gforsyth: Ahh, ok, understood.  You have a specific `cuda-toolkit` metapackage requirement and we need the `use_cuda_wheels=true` so we can remove it for DLFW builds.  Thanks for testing that for me.  You can revert that commit and I will approve

cpp/src/memory/buffer.cpp
nirandaperera: I think we can ommit event record for disk buffer, isnt it? Unrecordered event is always ready IINM.

madsbk: True, and that is why it works today. But it makes the invariant hold by accident rather than by construction, and it leaves disk as the one buffer type where `latest_write_event()` means nothing. 

The cost is a single `cudaEventRecord` per buffer, and in exchange the disk branches in `buffer_copy` can share the tail with the others instead of each ending in their own `record_copy(...); return;`. I would rather have one rule for all buffer types than one type with a footnote.

cpp/src/memory/buffer.cpp
nirandaperera: @madsbk I think I addressed your comments. Disk buffers now has a no-op `set_stream(...)` and copy from disk uses `exclusive_data_access` with a RAII. I skipped using a single tail for `src_is_disk`/`dst_is_disk` branches, because I felt it didnt make a lot of difference. Happy to circle back on it.

nirandaperera:
> Please address the ignored reviews from #1186 in here.

@pentschev sorry about that. I addressed your comment now

cpp/include/rapidsmpf/disk/disk_buffer.hpp
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * @brief Disk resource used for backing-file I/O.
+     *
+     * @return Shared pointer to the disk resource.
+     */
+    [[nodiscard]] std::shared_ptr<DiskResource> const& disk_resource() const noexcept {
+        return disk_;
+    }
+
+    ///@brief No-op stream rebind.
```
madsbk: Is should not be a no-op, it must associate the buffer with a new stream.
```suggestion
    /**
     * @brief Set the associated CUDA stream.
     *
     * This function only updates the buffer's associated CUDA stream. It does not
     * perform any stream synchronization or establish ordering guarantees between
     * the previous stream and @p stream. Callers must ensure that any work
     * previously enqueued on the old stream that may access the buffer has completed
     * or is otherwise properly synchronized before switching streams.
     *
     * @param stream The new CUDA stream.
     */
```

cpp/include/rapidsmpf/disk/disk_resource.hpp
```diff
@@ -0,0 +1,142 @@
...
+#include <rapidsmpf/config.hpp>
+#include <rapidsmpf/memory/back_ref_mixin.hpp>
+#include <rapidsmpf/memory/memory_type.hpp>
+
+namespace rapidsmpf {
+
+class BufferResource;
+
+/**
+ * @brief Non-stream-ordered disk I/O for host or device byte buffers.
```
madsbk: Semantically, I think this should stream-ordered.

cpp/include/rapidsmpf/disk/disk_resource.hpp
```diff
@@ -0,0 +1,142 @@
...
+     * @return Path to the reserved empty file.
+     */
+    [[nodiscard]] std::filesystem::path create_unique_path() const;
+
+    /**
+     * @brief Write bytes to a file and block until the transfer completes.
+     *
+     * @param path File path.
+     * @param data Host or device pointer to the source bytes. Must remain
+     *        valid until this call returns.
```
madsbk:
```suggestion
     * @param data Host or device pointer to the source bytes.
```

cpp/src/disk/disk_buffer.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+namespace rapidsmpf {
+
+DiskBuffer::DiskBuffer(std::shared_ptr<DiskResource> disk) : disk_{std::move(disk)} {
+    RAPIDSMPF_EXPECTS(disk_ != nullptr, "disk resource cannot be null");
+    path_ = disk_->create_unique_path();
+}
+
+DiskBuffer::DiskBuffer(DiskBuffer&& other) noexcept
+    : disk_{std::move(other.disk_)}, path_{std::move(other.path_)} {}
```
madsbk:
```suggestion
DiskBuffer::DiskBuffer(DiskBuffer&& other) noexcept
    : disk_{std::move(other.disk_)}, path_{std::exchange(other.path_, {})} {}
```

cpp/src/disk/disk_buffer.cpp
```diff
@@ -0,0 +1,53 @@
...
+            std::runtime_error
+        );
+    }
+    return ret;
+}
+
+void DiskBuffer::deallocate() noexcept {
+    if (!path_.empty()) {
+        std::error_code ec;
+        std::filesystem::remove(path_, ec);
```
madsbk, line 56: `deallocate()` captures the `error_code` from `remove` but never inspects it, so a failed removal leaks the backing file silently, and `path_` is cleared afterwards so nothing can retry or report it.

Could `DiskResource` hold a `std::shared_ptr<Logger>`?

cpp/src/disk/disk_resource.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include <kvikio/file_handle.hpp>
+
+#include <rapidsmpf/config.hpp>
+#include <rapidsmpf/disk/disk_resource.hpp>
+#include <rapidsmpf/error.hpp>
+#include <rapidsmpf/utils/string.hpp>
+
+namespace rapidsmpf {
+
+std::filesystem::path DiskResource::create_unique_path() const {
```
madsbk, line 37: Nothing ever removes this dir, a normal exit leaves an empty directory behind, and a crash leaves the spill files in it too.

Removing the directory when the `BufferResource` is destroyed would at least stop a long-lived node accumulating one per process. Leftovers from crashed processes are a harder problem and fine to punt on.

madsbk (changes requested):

cpp/src/memory/buffer.cpp
madsbk: I would still like the event recorded on the disk paths, even if the tail refactor waits.

The point of `latest_write_event()` is that it answers "when is the latest write to this buffer complete?" for every buffer, without the caller needing to know the memory type. Right now the disk paths answer that correctly only because the I/O is synchronous. Nothing is released early, but the invariant depends on how the copy happens rather than on the `Buffer` contract, and there is no indication to a future reader that disk is the exception.

KvikIO also has a stream-ordered API (`read_async()` and `write_async()`, taking a `CUstream` and returning a `StreamFuture`) that we might want later, which would make some disk copies genuinely asynchronous. Not a reason on its own, but it is the kind of change that stays local if the event semantics are already right.

The cost is one `cudaEventRecord` per copy. So I would record

```cpp
dst.latest_write_event_.record(dst.stream());
```

on both disk branches, with or without the shared tail refactor.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -606,13 +606,10 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
...
      * @brief Disk I/O resource and spill directory configuration.
      *
-     * Returned as a `std::shared_ptr`; `DiskBuffer`s keep a copy so the disk
-     * resource outlives those buffers if this `BufferResource` is destroyed.
-     *
      * @return Shared pointer to the disk resource, or `nullptr` if no spill
      * directory was configured.
      */
-    [[nodiscard]] std::shared_ptr<disk::DiskResource> disk_resource() const {
+    [[nodiscard]] std::shared_ptr<DiskResource> disk_resource() const {
```
wence-: return const ref?

cpp/include/rapidsmpf/disk/disk_buffer.hpp
nirandaperera: I had this as `size()` before. But, there is a slight nuance with the other buffer types. file_size() is dynamic, while host/ device buffers have a fixed size during construction. So, I just wanted to make that distinction.

cpp/include/rapidsmpf/disk/disk_resource.hpp
nirandaperera: let me add a stream param to the API. I agree with you here

cpp/include/rapidsmpf/disk/disk_resource.hpp
nirandaperera: this size_t is in kvikio. But I agree, lets make it std::ptrdiff_t

cpp/src/disk/disk_buffer.cpp
nirandaperera: @madsbk this is good point. I think lets write to `std::err` for the time being. `DiskResource` is created during BufferResource construction. Currently, we dont take the logger there.

cpp/src/disk/disk_buffer.cpp
nirandaperera: TBH this is something I didnt think about.
I think, later on, we can introduce a `path release()` method. At that point, the `this->path_` gets cleared, and passed on to a shared-fs-communicator (maybe?). Once the path is informed to the receiver, the communicator drops it from its ledger, and the receivers' communicator takes ownership of that path. Then during recv, create a DiskBuffer from the path and hand-off.

cpp/include/rapidsmpf/disk/disk_buffer.hpp
wence-: I think this speaks to an incorrect model. I think that a disk buffer should be a fixed size at creation time and we should know that size. What am I missing?

cpp/src/disk/disk_resource.cpp
nirandaperera: done

cpp/src/disk/disk_buffer.cpp
nirandaperera: Done

cpp/src/memory/buffer.cpp
nirandaperera: disk buffer now carries a stream ref.

cpp/src/memory/buffer.cpp
nirandaperera: @madsbk I believe I have addressed this now. Please take another look.

cpp/include/rapidsmpf/disk/disk_buffer.hpp
nirandaperera: I previously had a `size_` class member. I removed it, because I thought the contract of the `DiskBuffer` was mainly to hold-on to the `path`. Size is only determined during `write`.
Let me restore the `size_` class variable, and maybe validate write against the set size

cpp/include/rapidsmpf/memory/buffer_resource.hpp
nirandaperera: Done

cpp/include/rapidsmpf/memory/buffer.hpp
nirandaperera: This is done now.

cpp/include/rapidsmpf/memory/memory_type.hpp
nirandaperera: removed

cpp/include/rapidsmpf/memory/memory_type.hpp
nirandaperera: done

cpp/src/disk/disk_resource.cpp
nirandaperera: removed `disk` ns

cpp/src/disk/disk_resource.cpp
nirandaperera: inlined furture get and removed close

cpp/include/rapidsmpf/memory/buffer_resource.hpp
nirandaperera: done

cpp/include/rapidsmpf/disk/disk_resource.hpp
```diff
@@ -0,0 +1,147 @@
...
+class BufferResource;
+
+/**
+ * @brief Disk I/O for host or device byte buffers.
+ *
+ * Uses KvikIO with CompatMode::AUTO (GDS when available, POSIX/compat otherwise).
+ *
+ * Callers must synchronize any CUDA stream that produced or consumes a device
+ * pointer before calling write() or read(). KvikIO is not asked to synchronize
+ * the default stream (`sync_default_stream=false`).
```
madsbk: To be clear on what I am after: I want the API to be stream-ordered, even though the I/O underneath is currently blocking. Those are not in conflict. A blocking call trivially satisfies a stream-ordered contract, the synchronization just has to happen inside rather than being required of the caller.

As it stands the `stream` parameter is `[[maybe_unused]]` in both `read()` and `write()`, and the class doc still says "Callers must synchronize any CUDA stream that produced or consumes a device pointer before calling `write()` or `read()`". So the parameter is accepted and ignored, and the responsibility stays with the caller, which is what the `host_wait()` calls in `copy_to_disk()` and `copy_from_disk()` are doing.

The same thing has since appeared on `DiskBuffer`, which now stores a `stream_` with a `set_stream()` that assigns it and a `stream()` that nothing reads outside the test asserting the setter works. So there are now three stream-shaped pieces of API that do nothing: the two `[[maybe_unused]]` parameters here and the stream on `DiskBuffer`. That is what makes me want to settle the contract rather than keep adding the shape of one.

I would rather `DiskResource` owned that: synchronize the given stream before the transfer, and document the operation as ordered on it. Callers then do not need to know that disk behaves differently from any other memory type, and the "callers must synchronize" line can go, along with "Disk I/O is intentionally outside the MemoryType / BufferResource taxonomy", which is no longer true of the code either way.

The payoff is that if we move to KvikIO's `read_async()` and `write_async()`, or any other stream-ordered API, the signature and the contract stay exactly as they are and only the body changes.

One thing to note: synchronizing the whole stream inside `DiskResource` is coarser than the current `latest_write_event().host_wait()`, which waits on a single event rather than draining the stream. Both are correct, it is just marginally more synchronization. If that turns out to matter we could take an event instead, but on a path that is already doing blocking disk I/O I would expect the difference to be noise.

madsbk (changes requested):

cpp/include/rapidsmpf/disk/disk_resource.hpp
nirandaperera: @madsbk sorry I misunderstood your earlier comments. I think I have addressed this now 5718d1144c794e415b933bece956cb8b2f5b9df5

cpp/src/memory/buffer.cpp
```diff
@@ -271,24 +265,36 @@ void buffer_copy(
...
+        },
+        timing
+    );
 }
 
+
 }  // namespace rapidsmpf
+
+// TODO: currently disk read/write is synchronous. Use stream ordered timing, once disk
+// resource supports async read/write.
```
madsbk: Move to buffer_copy?

cpp/src/memory/buffer.cpp
```diff
@@ -151,23 +186,54 @@ void buffer_copy(
...
     src.latest_write_event().stream_wait(dst.stream());
-    StreamOrderedTiming timing{dst.stream(), statistics};
-    dst.write_access([&](std::byte* dst_data, cuda::stream_ref stream) {
-        RAPIDSMPF_CUDA_TRY(cuda_memcpy_async(
-            dst_data + dst_offset, src.data() + src_offset, size, stream
-        ));
-    });
+
+    // TODO: currently disk read/write is synchronous. Use stream ordered timing once
+    // disk resource supports async read/write.
```
madsbk: Now that `DiskResource` syncs the stream, I think we can just use `StreamOrderedTiming` everywhere and drop the wall-clock path. This TODO is already satisfied by your own change, it does not need to wait for async I/O.

As a bonus, `stream_delay` becomes meaningful, how deep the queue was at submission, instead of the hardcoded `0` the wall-clock overload writes.

That would let `record_disk_copy`, the `Duration` overload of `record_copy` (whose only caller is this lambda), and the TODO all go. The cost is two `cudaLaunchHostFunc` per disk copy, single-digit microseconds against milliseconds of I/O.

madsbk (changes requested):

cpp/src/memory/buffer.cpp
nirandaperera: Yes, we can, but then wouldnt we unnecessarily schedule 2 host CBs? I understand this will need to be the case, when things are truly stream ordered. But now, we will be calculating read/ write throughput with 2 host CB registering and invocations. Wouldnt that be noisy?

nirandaperera: NVM. I think I see your point. I changed this now to use `StreamOrderedTiming`. [fca9aa6](https://github.com/rapidsai/rapidsmpf/pull/1195/commits/fca9aa6d1027f59b109e517974ad86912b339a16)

cpp/include/rapidsmpf/memory/buffer.hpp
```diff
@@ -362,23 +406,35 @@ class Buffer {
...
 /**
- * @brief Asynchronously copy data between buffers.
+ * @brief Copy data between buffers.
  *
- * @note The copy is stream-ordered on @p dst's stream, correct cross-stream ordering
- * between @p src's stream and @p dst's stream is provided automatically.
+ * @note Copies between in-memory buffers are stream-ordered on @p dst's stream with
+ * automatic cross-stream ordering between @p src and @p dst. Copies involving disk
+ * storage are synchronous and block until the transfer completes; neither @p src nor
+ * @p dst is consumed.
```
madsbk: Revert back the original now?

cpp/include/rapidsmpf/memory/buffer.hpp
```diff
@@ -214,6 +243,13 @@ class Buffer {
         return latest_write_event_;
     }
 
+    /**
+     * @brief Record a write on the buffer's associated CUDA stream.
+     *
+     * Updates the event returned by `latest_write_event()`.
+     */
+    void record_write_event();
```
madsbk: Does it need to be public? 
`buffer_copy` is already a friend, so it could sit in the private section. As public API it lets any caller claim a write that did not happen, and `latest_write_event()` is only useful if it is truthful.

cpp/src/disk/disk_resource.cpp
```diff
@@ -0,0 +1,155 @@
...
+        );
+    }
+    return path_template;
+}
+
+std::size_t DiskResource::write(
+    std::filesystem::path const& path,
+    void const* data,
+    std::size_t size,
+    [[maybe_unused]] MemoryType mem_type,
```
madsbk, line 87: Do we need `mem_type` here and in DiskResource::read?

cpp/src/disk/disk_resource.cpp
```diff
@@ -0,0 +1,155 @@
...
+std::filesystem::path DiskResource::create_unique_path() const {
+    std::error_code ec;
+    std::filesystem::create_directories(dir_, ec);
+    RAPIDSMPF_EXPECTS(
+        !ec,
+        "failed to create directory " + dir_.string() + ": " + ec.message(),
+        std::runtime_error
+    );
+
+    auto path_template = (dir_ / "XXXXXX").string();
```
madsbk, line 62: Should we use a prefix like `rapidsmpf-` ?

cpp/src/memory/buffer.cpp
```diff
@@ -151,23 +187,73 @@ void buffer_copy(
...
+    } else {
+        dst.write_access([&](std::byte* dst_data, cuda::stream_ref stream) {
+            RAPIDSMPF_CUDA_TRY(cuda_memcpy_async(
+                dst_data + dst_offset, src.data() + src_offset, size, stream
+            ));
+        });
+    }
+    statistics->record_copy(src.mem_type(), dst.mem_type(), size, std::move(timing));
     // after the dst.write_access(), its last_write_event is recorded on dst.stream(). So,
     // we need the src.stream() to wait for that event.
```
madsbk:
```suggestion
    // Every branch above records dst's latest-write event on dst.stream(). Make
    // src.stream() wait for it, so src is not reused before the copy completes.
```

cpp/src/disk/disk_resource.cpp
```diff
@@ -0,0 +1,155 @@
...
+#include <rapidsmpf/disk/disk_resource.hpp>
+#include <rapidsmpf/error.hpp>
+#include <rapidsmpf/utils/misc.hpp>
+#include <rapidsmpf/utils/string.hpp>
+
+namespace rapidsmpf {
+
+DiskResource::~DiskResource() noexcept {
+    std::error_code ec;
+    std::filesystem::remove(dir_, ec);
```
madsbk, line 54: The destructor assumes this `DiskResource` owns `dir_` exclusively, but the directory is keyed only on the PID, so two `BufferResource`s created in the same process with the same `spill_directory` get two `DiskResource`s pointing at the same path.

madsbk (changes requested):

cpp/src/disk/disk_resource.cpp
nirandaperera: This is something I saw in cucascade disk resource impl. It maybe important to know the accessibility of `*data`. But I dont have strong feelings about it. WDYT?

cpp/src/disk/disk_resource.cpp
nirandaperera: This is already inside the `rapidsmpf_spill_dir`, so, I think this is fine.

cpp/src/disk/disk_resource.cpp
nirandaperera: Good point. This disallowed creating 2 disk resources pointing to the same dir. Fixed this now 99691d5090921eec83f61fef1f5e36e82b26ebf0

cpp/include/rapidsmpf/memory/buffer.hpp
nirandaperera: done

cpp/include/rapidsmpf/memory/buffer.hpp
nirandaperera: done

cpp/src/disk/disk_resource.cpp
madsbk: right, good point

cpp/src/disk/disk_resource.cpp
madsbk: I would say, let's add it later if needed.

cpp/include/rapidsmpf/disk/disk_resource.hpp
```diff
@@ -0,0 +1,145 @@
...
+ * @brief Disk I/O for host or device byte buffers.
+ *
+ * Uses KvikIO with CompatMode::AUTO (GDS when available, POSIX/compat otherwise).
+ *
+ * Each resource atomically creates and exclusively owns its directory. The directory
+ * is removed when the resource is destroyed.
+ *
+ * `BufferResource` owns a `std::shared_ptr<DiskResource>` when a spill
+ * directory is configured; `DiskBuffer`s hold additional copies so the
+ * resource outlives those buffers.
```
madsbk, line 31:
```suggestion
```

cpp/tests/test_disk_resource.cpp
```diff
@@ -0,0 +1,405 @@
...
+        {},
+        std::chrono::milliseconds{1},
+        std::make_shared<StreamPool>(16),
+        Statistics::disabled(),
+        disk_dir.path()
+    );
+
+    EXPECT_EQ(br->disk_resource()->directory().parent_path(), disk_dir.path());
+}
+
```
madsbk, line 378: Is redundant 
```suggestion
```

cpp/tests/test_buffer_resource.cpp
```diff
@@ -945,6 +1042,156 @@ TEST(BufferResource, DeviceMrIsAddressableByMemoryRecorder) {
...
+            GTEST_SKIP() << "Pinned memory resources are not supported on this system";
+        }
+        auto pinned_pool_properties = is_pinned_memory_resources_supported()
+                                          ? PinnedPoolProperties{}
+                                          : PinnedMemoryDisabled;
+        br_ = BufferResource::create(
+            rmm::mr::get_current_device_resource_ref(),
+            std::move(pinned_pool_properties),
+            {},
+            std::chrono::milliseconds{1},
```
madsbk, line 1060: Don't think we need spilling?
```suggestion
            std::nullopt,
```

cpp/tests/test_buffer_resource.cpp
```diff
@@ -922,6 +927,98 @@ TEST(RmmResourceAdaptor, EqualityAcrossCopiesAndAccessPaths) {
 
 // Guarantee that when stats enabled, br->device_mr() reference gets properly casted to an
 // RmmResourceAdaptor and used by the memory recorder.
+namespace {
+
+std::filesystem::path disk_test_dir() {
```
madsbk, line 932: Can we use `TempDir` instead, as `test_disk_resource.cpp` does?

cpp/src/memory/buffer_resource.cpp
```diff
@@ -241,7 +266,10 @@ std::unique_ptr<Buffer> BufferResource::make_buffer(
     std::size_t size, cuda::stream_ref stream, MemoryReservation& reservation
 ) {
     auto const mem_type = reservation.mem_type_;
-    StreamOrderedTiming timing{stream, statistics_};
+    // disk buffer creation is not stream ordered, so we disable statistics for it
```
madsbk, line 269:
```suggestion
    // A disk buffer is created with mkstemp, which never touches the stream, so a
    // stream-ordered timing would measure stream backlog rather than the creation.
```

cpp/src/statistics.cpp
```diff
@@ -731,27 +750,18 @@ void Statistics::record_copy(
...
-                .base = base,
-                .nbytes = base + "-bytes",
-                .time = base + "-time",
-                .stream_delay = base + "-stream-delay",
-            };
-        }
-        return ret;
-    }();
+    if (mem_type == MemoryType::DISK) {
+        return;  // disk allocation are not stream ordered
```
madsbk, line 754:
```suggestion
        // Creating a disk buffer is a host-side mkstemp, so there is no stream-ordered
        // duration to pair with the byte count. Skip the whole record rather than
        // registering a MemoryThroughput entry with no time.
        return;
```

madsbk (approved):
Good work @nirandaperera, I only have a few nits left.

cpp/tests/test_buffer_resource.cpp
nirandaperera: Done

cpp/src/disk/disk_resource.cpp
nirandaperera: removed

nirandaperera:
Thanks @madsbk. Merging on green CI

gforsyth (approved):

pentschev (approved):

PR #1196, opened 2026-09-03
Add disk spilling support to the shuffler

### LINKED ISSUE #1185: Use disk buffer in the shuffler path

nirandaperera:
Enable the shuffler to spill buffers through host tiers to disk while preserving safe communication and extraction behavior.
- Add non-throwing, spill-aware reservation APIs and deferred receive allocation.
- Make spill destinations configurable, defaulting to pinned host, host, and disk.
- Preserve disk-resident extracted data while restoring outgoing disk buffers before sending.
- Add coverage for reservations, deferred allocation, and disk spilling.

Depends on #1195 
Closes #1185 
Related #1170


cpp/src/shuffler/shuffler.cpp
```diff
@@ -338,11 +347,20 @@ void Shuffler::insert(std::unordered_map<PartID, PackedData>&& chunks) {
...
-            chunk.set_data_buffer(br_->move(chunk.release_data_buffer(), reservation));
+            Rank const dst_rank =
+                partition_owner(comm_, chunk.part_id(), total_num_partitions);
+            if (dst_rank == comm_->rank()) {
+                // Local received chunks may fall back to disk when host is full.
+                chunk.spill_from_device(*br_);
+            } else {
+                // Outgoing chunks stay in host-tier memory.
+                auto reservation =
+                    br_->reserve_or_fail(chunk.data_size(), SPILL_TARGET_MEMORY_TYPES);
```
pentschev: But what if indeed fails? Not ideal, but we should probably allow spilling to disk even if suboptimal instead of failing, we may hit this in an unlucky situation where a failure is worse than being slow, I think.

cpp/src/shuffler/chunk.cpp
```diff
@@ -138,13 +146,65 @@ bool Chunk::validate_format(std::vector<std::uint8_t> const& serialized_buf) {
...
+    auto const size = disk_data_->size();
+    constexpr std::array restore_mem_types{
+        MemoryType::DEVICE, MemoryType::PINNED_HOST, MemoryType::HOST
+    };
+    auto reservation = br.try_reserve_or_spill(size, restore_mem_types);
+    RAPIDSMPF_EXPECTS(
+        reservation.has_value(),
+        "failed to reserve memory to restore a disk-resident chunk after spilling",
+        std::runtime_error
+    );
```
pentschev: Could we retain the final `AllowOverbooking::YES` attempt for pinned/pageable host memory here? `try_reserve_or_spill()` only permits non-overbooked host reservations, so a disk-resident chunk now throws during `extract()` whenever device spilling cannot free enough space and the host limit is full, even though the API documentation says: "Tries a non-overbooking reservation of device, pinned host, then pageable host memory. If none succeed, retries pinned then pageable host with overbooking so extract can progress.". This is especially reachable with disk spilling enabled and the host limit set to zero, the current tests make device memory available before extraction, so they do not cover this pressure case.

cpp/include/rapidsmpf/memory/buffer_resource.hpp
```diff
@@ -431,6 +437,130 @@ class BufferResource : public std::enable_shared_from_this<BufferResource> {
...
+     *
+     * Tries each requested memory type in preference order. If no reservation is
+     * immediately available, spills only enough device memory to satisfy the retained
+     * device reservation, retrying up to @p num_spill_retries times. Only device memory
+     * is spilled, because spilling other memory types is not supported yet.
+     *
+     * @param size The size of the memory to reserve.
+     * @param mem_types Range of memory types to consider.
+     * @param num_spill_retries Maximum number of spill attempts. Only applies to device
+     * memory.
```
pentschev: Seems a bit arbitrary. Since this is blocking should we use a time range instead of a number of attempts?

nirandaperera:
@pentschev I resolved all your comments because rebased this on top of #1195. The API has changed significantly.

nirandaperera:
@pentschev Please refer https://github.com/rapidsai/rapidsmpf/pull/1196/changes/06795768535558cdab298da695b92478d3182bd5 commit. The rest is captured in #1195

PR #1203, opened 2026-09-08
CUDA context in PausableThreadLoop's spawned thread

### LINKED ISSUE #1202: Shuffling with Medium Amount Spilling Failure
When running the nightly spill shuffle benchmark on a dgxh100:

> /raid/bzaitlen/miniforge3/envs/2026-09-04-cudf-polars/bin/rrun -n 8 --bind-to cpu --bind-to memory -x UCX_MAX_RNDV_RAILS=1 -x 'UCX_TLS=^sm' -x UCX_LOG_LEVEL=warn -x UCX_RNDV_PIPELINE_ERROR_HANDLING=y -x UCX_PROTO_ENABLE=y -x UCX_RNDV_FRAG_WORKER_MAX_MEM=1G -x UCX_RNDV_FRAG_MEM_TYPES=cuda /raid/bzaitlen/miniforge3/envs/2026-09-04-cudf-polars/bin/libcudf_streaming_bench_shuffle -C ucxx -w 3 -r 10 -g -s -x -l 40960 -p 50 -o 4 -c 4 -n 67108864 -x -m async

We get an error with latest on main of both rapidsmpf/cudf:

> [7:PRINT:0:2026-09-08 18:22:33.816472384] elapsed: 1.15 min | local throughput: 739.23 MiB/s | global throughput: 5.78 GiB/s (warmup run)
[6:PRINT:0:2026-09-08 18:22:33.816518297] elapsed: 1.15 min | local throughput: 739.60 MiB/s | global throughput: 5.78 GiB/s (warmup run)
[0:PRINT:0:2026-09-08 18:22:33.816482422] elapsed: 1.15 min | local throughput: 739.06 MiB/s | global throughput: 5.77 GiB/s (warmup run)
[5:PRINT:0:2026-09-08 18:22:33.816490212] elapsed: 1.15 min | local throughput: 739.25 MiB/s | global throughput: 5.78 GiB/s (warmup run)
[3:PRINT:0:2026-09-08 18:22:33.816512146] elapsed: 1.15 min | local throughput: 739.12 MiB/s | global throughput: 5.77 GiB/s (warmup run)
[2:PRINT:0:2026-09-08 18:22:33.816521759] elapsed: 1.15 min | local throughput: 739.05 MiB/s | global throughput: 5.77 GiB/s (warmup run)
[1:PRINT:0:2026-09-08 18:22:33.816483186] elapsed: 1.15 min | local throughput: 739.02 MiB/s | global throughput: 5.77 GiB/s (warmup run)
[4:PRINT:0:2026-09-08 18:22:33.816621603] elapsed: 1.15 min | local throughput: 739.60 MiB/s | global throughput: 5.78 GiB/s (warmup run)
terminate called after throwing an instance of 'cuda::__4::cuda_error'
  what():  ''include/rapids/cuda/__driver/driver_api.h:171 invalid device context(201): Failed to allocate memory from a memory pool
terminate called after throwing an instance of 'cuda::__4::cuda_error'
  what():  ''include/rapids/cuda/__driver/driver_api.h:171 invalid device context(201): Failed to allocate memory from a memory pool
(END)

Reverting both #1162 #1164  resolves the issue.   This is obviously a problem we should fix but I would prioritize this for after the 26.10 release

quasiben:
This PR was largely generated by Claude

I think this resolves https://github.com/rapidsai/rapidsmpf/issues/1202.  

What I think is happening is that rapidsmpf's PausableThreadLoop spawns a background thread that never establishes a CUDA ctx. Agent seems to believe that we never really hit this problem problem as most of the time we aren't spilling loads of data.  Issue #1202 specifically tests spilling loads of data.  Additionally, I asked the agent to also write a test which specifically attempts to spill data using Periodic Spill Check.  Test fails without the patch included in this PR

I find the documentation here quite verbose but left it for now to provide additional ctx -- heh

### DIFF
--- a/cpp/src/pausable_thread_loop.cpp
+++ b/cpp/src/pausable_thread_loop.cpp
@@ -1,5 +1,5 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 
--- a/cpp/src/pausable_thread_loop.cpp
+++ b/cpp/src/pausable_thread_loop.cpp
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <cuda_runtime_api.h>
+
 #include <rapidsmpf/error.hpp>
 #include <rapidsmpf/pausable_thread_loop.hpp>
 
@@ -10,6 +12,12 @@ namespace rapidsmpf::detail {
 
 PausableThreadLoop::PausableThreadLoop(std::function<void()> func, Duration sleep) {
     thread_ = std::thread([this, f = std::move(func), sleep]() {
+        // Establish a CUDA context on this thread before any user code runs. A
+        // freshly spawned std::thread does not inherit the constructing thread's
+        // CUDA context, and low-level driver-API calls (unlike the CUDA Runtime
+        // API) do not lazily establish one on first use.
+        RAPIDSMPF_CUDA_TRY(cudaFree(nullptr));
+
         while (true) {
             // wait until the thread is not paused
             state_.wait(State::Paused, std::memory_order_acquire);
--- a/cpp/src/pausable_thread_loop.cpp
+++ b/cpp/src/pausable_thread_loop.cpp
@@ -12,11 +12,7 @@ namespace rapidsmpf::detail {
 
 PausableThreadLoop::PausableThreadLoop(std::function<void()> func, Duration sleep) {
     thread_ = std::thread([this, f = std::move(func), sleep]() {
-        // Establish a CUDA context on this thread before any user code runs. A
-        // freshly spawned std::thread does not inherit the constructing thread's
-        // CUDA context, and low-level driver-API calls (unlike the CUDA Runtime
-        // API) do not lazily establish one on first use.
-        RAPIDSMPF_CUDA_TRY(cudaFree(nullptr));
+        bool is_thread_initialized = false;
 
         while (true) {
             // wait until the thread is not paused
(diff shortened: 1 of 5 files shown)


nirandaperera:
@quasiben Hmm. I've seen this before when running some test cases with gtest filter. Eg https://github.com/rapidsai/rapidsmpf/blob/main/cpp/tests/test_host_buffer.cpp#L325-L327

But I thought `cudaFree(nullptr)` is a process-wide call. So, ideally, we should do this before creating rapidsmpf objects like communicator, buffer resource, etc.

nirandaperera:
@quasiben can you check if calling `cudaFree(nullptr)` at the very beginning of the `main` resolves #1202 ?
Somewhere [here](https://github.com/NVIDIA/cudf/blob/main/cpp/libcudf_streaming/benchmarks/bench_shuffle.cpp#L466)

pentschev:
> But I thought `cudaFree(nullptr)` is a process-wide call. So, ideally, we should do this before creating rapidsmpf objects like communicator, buffer resource, etc.

It creates a CUDA context for the main application, yes, but new threads do not inherit the CUDA context from the parent thread, which is a problem if that thread uses the CUDA driver API which, unlike the runtime API, does not ensure CUDA device/context are set. This change ensures the pausable thread sets the CUDA context and device appropriately, and [we already have precedent for that for the progress thread](https://github.com/rapidsai/rapidsmpf/blob/effb3db53d45acef33952751fc094de31dbea1a2/cpp/src/progress_thread.cpp#L31-L33).

pentschev (approved):
This is a small enough change for a real issue that was surfaced due to changes in 26.10, therefore I have retargeted it to `release/26.10`.

cpp/src/pausable_thread_loop.cpp
```diff
@@ -40,6 +44,17 @@ PausableThreadLoop::PausableThreadLoop(std::function<void()> func, Duration slee
                 return;
             }
 
+            if (!is_thread_initialized) {
```
nirandaperera, line 47: Since `ProgressThread` uses a `PausableThreadLoop` we can remove the initialized check there?

nirandaperera:
> > But I thought `cudaFree(nullptr)` is a process-wide call. So, ideally, we should do this before creating rapidsmpf objects like communicator, buffer resource, etc.
> 
> It creates a CUDA context for the main application, yes, but new threads do not inherit the CUDA context from the parent thread, which is a problem if that thread uses the CUDA driver API which, unlike the runtime API, does not ensure CUDA device/context are set. This change ensures the pausable thread sets the CUDA context and device appropriately, and [we already have precedent for that for the progress thread](https://github.com/rapidsai/rapidsmpf/blob/effb3db53d45acef33952751fc094de31dbea1a2/cpp/src/progress_thread.cpp#L31-L33).

Thanks @pentschev. TIL

cpp/src/pausable_thread_loop.cpp
pentschev: You're right, only in reverse order: we can remove the check from `ProgressThread` and keep this one here. Done in https://github.com/rapidsai/rapidsmpf/pull/1203/commits/2a714f24e25414c2e7030bb3b0b31fdb091f6914 .

nirandaperera (approved):

pentschev:
Thanks @nirandaperera !

PR #1207, opened 2026-09-09
Add basic metrics to channels

### LINKED ISSUE #1200: Add monitoring hook to `Channel.send` / `Channel.recv`
In https://github.com/NVIDIA/cudf/pull/23179, we're adding Quent tracing to cudf-polars. Our best way to accurately measure the total input / output bytes to a cudf-polars Operator is by measuring the sizes of the `Message`s that flow through `Channel.send` (output) and `Channel.recv` (input).

Today, this requires a new class in cudf-polars that composes `Channel`, along with the `ActorTracer` that records the statistics in cudf-polars. In addition to the instrumented send/recv methods:

```python
    async def recv(self, context: Context) -> Message[T] | None:
        message = await self._channel.recv(context)
        if message is not None and self._tracer is not None:
            self._tracer.input_bytes += _message_size(message)
        return message
```

we override all the other methods to provide the same interface as `Channel`.

To make this slightly easier, `Channel` could have a pair of optional properties `on_send` and `on_recv`. These would be (Python) callables that take two arguments: `context: Context, message: Message[T]`.

However, `send` and `recv` are async functions. Whoever defines & registers these functions need to be really careful to not block the event loop. Maybe we shouldn't give users that footgun.

wence-:
This allows us to track message count and message volume.

By separating the send and receive side of the channel we can track if any message was spilled in the channel.

- Closes #1200


cpp/include/rapidsmpf/streaming/core/channel.hpp
```diff
@@ -53,6 +55,22 @@ class Channel {
     friend Context;
 
   public:
+    /**
+     * @brief A snapshot of metrics recorded on a channel.
+     *
+     * @note Byte counts for message volume rely on the message having an accurate content
+     * description.
+     */
+    struct MetricsSnapshot {
```
TomAugspurger, line 64: Thinking through the names here, and trying to align with https://github.com/rapidsai/rapidsmpf/blob/main/docs/source/statistics.md.

- The `shuffle-payload-recv` statistics uses `recv` instead of `receive[s]`
- Most of those are roughly `<object>-<property>` e.g. `alloc-device`-`bytes`. Maybe we should use `sent_bytes` and `received_bytes`?

And [other systems](https://opentelemetry.io/docs/specs/semconv/general/naming/) I've used have more recommendations but a full consistency check is out of scope for now.

cpp/include/rapidsmpf/streaming/core/channel.hpp
```diff
@@ -53,6 +55,22 @@ class Channel {
...
+        std::array<std::uint64_t, MEMORY_TYPES.size()>
```
TomAugspurger, line 65: An array, rather than a struct, seems fine here since this won't change often (despite a PR currently being open adding a disk tier).

Edit: oh and it seems like the python bindings even convert it to a dict, and that code should be future-compatible.

TomAugspurger (approved):
Looks good, thanks. Just a question on the names in the `MetricsSnapshot` but the proposed ones are fine too.

cpp/include/rapidsmpf/streaming/core/channel.hpp
wence-: went with {send/recv}_{count/bytes}

python/rapidsmpf/rapidsmpf/streaming/core/channel.pyx
```diff
@@ -332,6 +333,53 @@ cdef extern from * nogil:
...
+
+    @property
+    def send_bytes(self):
+        """The total byte count of sent messages, stratified by memory type."""
+        return {
+            t: cpp_metrics_send_bytes(self._handle, <cpp_MemoryType>(t))
+            for t in MemoryType
+        }
+
+    @property
```
nirandaperera, line 369: Nit. should these be, `cached_property`s instead?

cpp/tests/streaming/test_channel.cpp
```diff
@@ -40,6 +40,10 @@ TEST_F(StreamingChannel, DataRoundTripWithoutMetadata) {
     run_actor_network(std::move(actors));
 
     ASSERT_EQ(outputs.size(), num_messages);
+    auto metrics = ch->metrics();
+    EXPECT_EQ(metrics.send_count, num_messages);
+    EXPECT_EQ(metrics.recv_count, num_messages);
+    EXPECT_EQ(metrics.spillable_count, 0);
```
nirandaperera: nit. should we also accumulate `send_bytes` and `recv_bytes`? I think it should be `num_messages * sizeof(int)` isnt it?

cpp/include/rapidsmpf/streaming/core/channel.hpp
```diff
@@ -53,6 +55,23 @@ class Channel {
...
+            send_bytes{};  ///< Number of bytes sent, stratified by memory type.
+        std::array<std::uint64_t, MEMORY_TYPES.size()>
+            recv_bytes{};  ///< Number of bytes received, stratified by memory type.
+        std::uint32_t send_count{};  ///< Number of times send was called on the channel.
+        std::uint32_t
+            recv_count{};  ///< Number of times receive was called on the channel.
+        std::uint32_t spillable_count{};  ///< Number of spillable messages sent.
```
nirandaperera: I'm a bit on the fence on naming here. Personally, send/ recv implies communication. should we use ingress/ egress instead?

nirandaperera (commented):
Just some nitpiks

cpp/include/rapidsmpf/streaming/core/channel.hpp
wence-: This mirrors the names we have for pushing data into/pulling data out of the channel.

python/rapidsmpf/rapidsmpf/streaming/core/channel.pyx
wence-: No, because it doesn't matter, and your can't put them on cdef classes without some jiggery-pokery

cpp/tests/streaming/test_channel.cpp
wence-: Done. One has to give the message a content_description.

madsbk (approved):

PR #1214, opened 2026-09-14
Add statistics to `MemoryReserveOrWait`

madsbk:
`MemoryReserveOrWait` had no statistics wiring at all, so none of the memory demand arriving from actors was visible. This adds enough to answer one question: **when a reservation has to wait, does waiting ever work, or does it just run out the timeout?**

### Statistics

All prefixed `reserve-<memtype>-`, recorded once per request rather than per poll of the admission loop.

| Stat | Formatter | Meaning |
|---|---|---|
| `wait-avoided` | HitRate | Of all requests, how many the fast path satisfied without waiting |
| `wait-timeout` | HitRate | Of the requests that waited, how many ran out `memory_reserve_timeout` |
| `wait-satisfied-time` | Duration | How long requests waited before a release satisfied them |
| `wait-timeout-time` | Duration | How long requests waited before the timeout fired |
| `waiting-requests` | Gauge | Peak number of requests waiting at once |
| `request-bytes` | Bytes | Requested bytes |
| `overbook-bytes` | Bytes | Bytes by which `reserve_or_wait_or_overbook()` exceeded the limit |

The two hit rates nest, so `wait-timeout` is measured over the requests that `wait-avoided` missed. Reading them together gives the full funnel without any arithmetic.

Nothing is recorded when statistics are disabled, and a memory type that is never reserved produces no report entries at all, so the report stays as short as the run warrants.

### What it shows

PDS-H SF10 q9 with `spill_device_limit=512MiB` on my local workstation:

```
reserve-device-overbook-bytes:       19.41 GiB | avg 1.62 GiB
reserve-device-request-bytes:        20.49 GiB | avg 299.75 MiB
reserve-device-wait-avoided:         55/70 (hits/lookups)
reserve-device-wait-satisfied-time:  24.09 ms | avg 8.03 ms
reserve-device-wait-timeout:         12/15 (hits/lookups)
reserve-device-wait-timeout-time:    1.36 s | avg 113.11 ms
reserve-device-waiting-requests:     max 3 | avg 1.2 (15 samples)
```

Of 70 requests, 55 never waited. Of the 15 that did, 12 ran out the timeout and only 3 were satisfied by a release, and those three were satisfied within 8 ms on average while the other twelve each burned the full timeout to end up overbooking anyway. The run overbooked 19.41 GiB against a 512 MiB limit.

### New `Statistics::Formatter::Gauge`

The one change outside the reservation path. For quantities that rise and fall, where the running total is meaningless and the peak is the point. `Default` would have rendered `waiting-requests` above as `18 (count 15)`, a sum of queue sizes. `Gauge` renders `max 3 | avg 1.2 (15 samples)`, suppressing the average for a single sample as `Bytes` and `Duration` already do. Appended before the `_Count` sentinel so serialised formatter values are unchanged.


cpp/src/streaming/core/memory_reserve_or_wait.cpp
```diff
@@ -123,7 +161,11 @@ MemoryReserveOrWait::reserve_or_wait_or_overbook(
 ) {
     auto ret = co_await reserve_or_wait(size, net_memory_delta);
     if (ret.size() < size) {
-        co_return br_->reserve(mem_type_, size, AllowOverbooking::YES);
+        auto overbooked = br_->reserve(mem_type_, size, AllowOverbooking::YES);
+        if (overbooked.second > 0) {
+            record_stat("overbook-bytes", static_cast<double>(overbooked.second));
+        }
```
pentschev, line 174: Could we record only the incremental overbooking introduced by this request, for example `std::min(size, overbooked.second)`, or rename/document this as accumulated deficit samples? `BufferResource::reserve()` returns the total deficit after the reservation, including prior outstanding overbooking.

cpp/src/streaming/core/memory_reserve_or_wait.cpp
```diff
@@ -102,6 +138,8 @@ coro::task<MemoryReservation> MemoryReserveOrWait::reserve_or_wait(
     }
     lock.unlock();
 
+    record_stat("waiting-requests", static_cast<double>(waiting_requests));
```
pentschev, line 144: Could we either sample this gauge when requests leave `reservation_requests_` as well, or describe its average as "queue depth observed on request arrival"? Currently values are recorded only after insertions, the removals in the satisfied and timeout paths record nothing.

cpp/src/streaming/core/memory_reserve_or_wait.cpp
madsbk: Good catch, done

cpp/src/streaming/core/memory_reserve_or_wait.cpp
madsbk: Took the documentation option.

Sampling on removal would still not give the mean number of concurrent waiters, since that needs time weighting rather than event sampling. The maximum is unaffected either way, as the queue only grows at insertion.

docs/source/statistics.md
```diff
@@ -9,6 +9,13 @@ This table gives an overview of the different statistics collected.
...
 | `event-loop-total` | Time spent in in the background `ProgressThread` event-loop. |
 | `recv-into-host-memory` | Data received directly into host memory rather than device memory, due to memory pressure at receive time. |
+| `reserve-{memtype}-wait-avoided` | Reservation requests that `MemoryReserveOrWait` satisfied at once, over all requests. A miss means the request had to wait. |
+| `reserve-{memtype}-wait-timeout` | Waiting requests that ran out `memory_reserve_timeout`, over the requests that had to wait. A miss means memory was released in time. |
+| `reserve-{memtype}-wait-satisfied-time` | Time requests spent waiting before a reservation release satisfied them. |
+| `reserve-{memtype}-wait-timeout-time` | Time requests spent waiting before the progress timeout fired. |
+| `reserve-{memtype}-waiting-requests` | Requests waiting concurrently. Recorded each time a request starts waiting, not sampled over time, so the maximum is exact but the mean only covers those peaks and overstates the typical number of waiters. |
```
pentschev, line 16: How about describing the mean simply as "queue depth when a request starts waiting", without saying it necessarily overstates the typical number of waiters? Because these samples are not time-weighted, the bias direction depends on the workload: a high queue depth that persists for a long time can make this arrival-only mean understate the time-weighted mean, while quickly draining queues can make it overstate it.

cpp/tests/streaming/test_memory_reserve_or_wait.cpp
```diff
@@ -561,3 +582,139 @@ TEST_P(StreamingMemoryReserveOrWait, ReserveMemoryHelperDefaultOverbookingDisabl
...
+        br_stats
+    };
+
+    coro::sync_wait([](MemoryReserveOrWait& mrow) -> Actor {
+        auto [res, overbooked_bytes] = co_await mrow.reserve_or_wait_or_overbook(10, 0);
+        EXPECT_EQ(res.size(), 10);
+        EXPECT_EQ(overbooked_bytes, 10);
+    }(mrow));
+
+    EXPECT_EQ(stats->get_stat("reserve-device-overbook-bytes").value(), 10.0);
```
pentschev, line 699: Could we extend this test to issue a second overbooked reservation while the first remains outstanding and assert the accumulated statistic? With one request, `overbooked.second` and `std::min(size, overbooked.second)` are identical, so this test also passed with the previous double-counting implementation and does not protect the fix.

pentschev (approved):
Minor asks, otherwise LGTM.

docs/source/statistics.md
madsbk: It now says:
```
| `reserve-{memtype}-waiting-requests` | Requests waiting concurrently. Recorded each time a request starts waiting, not sampled over time, so the maximum is exact while the mean is the queue depth seen when a request starts waiting. |
```

cpp/tests/streaming/test_memory_reserve_or_wait.cpp
madsbk: done

madsbk:
thanks @pentschev

PR #1218, opened 2026-09-15
Include rmm/cuda_stream_view.hpp in the CUDA stream tests.

miscco:
The test still names rmm::cuda_stream_view, so it needs that header after the stream_ref migration.


madsbk (approved):

wence-:
This needs to go to release/26.10, I think?

wence-:
FWIW, if everything is building fine without this include on release/26.10 then we don't need to include it I suppose?

PR #1220, opened 2026-09-16
Record how long buffer data stays spilled

madsbk:
This PR adds one statistic, `buffer-spilled-time`: how long a spill kept device memory free, from freeing the spilled buffer to allocating its replacement. Data that is never unspilled is not counted.

`copy-device-to-host` already reports what spilling costs, in bytes and in copy time. Nothing reports what it buys. Fewer bytes moved is not on its own an improvement, since it can be had by spilling large things rarely and badly, and the spills that remain holding their memory longer is what says they were the right ones. This is that side of the ledger.

A `SpillTrackToken` follows the data, opened and closed by `BufferResource::move()`, which is the only path that hands device capacity back. A plain `buffer_copy()` leaves its source allocated, as `PackedData::copy()` relies on, so it is not a spill. Both ends are measured host-side to match the memory reservation accounting, from deallocating the source to allocating its replacement, rather than from the stream-ordered copies that run between them.

### Follow-up

cudf-streaming needs a companion change, because its dominant spill never goes through `BufferResource`. `table_chunk::copy` packs a live chunk with `cudf::pack` and the pinned memory resource, writing device to host in one step, so no source `Buffer` ever exists. The companion change opens the token there, passes it to the `record_copy` call already being made, and hands it to `BufferResource::move`. Until it lands, those spills are not counted.

This says how long the room was free, not whether the spill needed to happen. That question is being answered separately by the `ProgressMonitor` statistics, which record what the rest of the system was doing each time `MemoryReserveOrWait` ran out its timeout and forced a spill. The two meet only in the periodic spill thread, which spills on a timer with no request behind it, so it has no wait to instrument and this is the only statistic that covers it.

### What the report looks like

```
 - buffer-spilled-time:                 7.30 ms | avg 3.65 ms
 - copy-device-to-pinned_host:          488.28 MiB | 104.93 ms | 4.54 GiB/s | avg-stream-delay 20.50 us
 - copy-pinned_host-to-device:          488.28 MiB | 42.46 ms | 11.23 GiB/s | avg-stream-delay 52.81 us
```

Two spills, each freeing its memory for about 3.7ms while its own round trip took about 74ms. Raising `memory_reserve_timeout` from 1ms to 100ms on the same query turns that into `296.51 ms | avg 148.26 ms`, so the same spills hold their memory for twice their round trip instead of a twentieth of it.

### Measurements

On my local workstation, PDS-H at SF50 with a 10% device limit, two ranks, median of three iterations.

| query | wall time | spills unspilled | mean time spilled | longest time spilled | mean size spilled |
|---|---|---|---|---|---|
| 21 | 5.66s | 14 | 2414ms | 5566ms | 487 MiB |
| 18 | 3.98s | 9 | 1915ms | 3741ms | 1264 MiB |
| 9 | 3.78s | 17 | 1061ms | 3723ms | 785 MiB |
| 8 | 2.34s | 16 | 909ms | 2246ms | 589 MiB |
| 7 | 1.76s | 22 | 606ms | 1533ms | 283 MiB |
| 5 | 1.67s | 10 | 571ms | 1366ms | 738 MiB |
| 3 | 1.03s | 8 | 205ms | 487ms | 421 MiB |
| 1 | 1.68s | 5 | 274ms | 449ms | 1168 MiB |

Query 1 is the only query in the set with no join, so nothing accumulates and no spill has long-lived state to sit behind.


TomAugspurger:
Maybe this would be answered by looking at the code, but I don't see how we can judge whether some spill was "wasted" just by comparing the spill/unspill time with the duration it was spilled... I reread the "shorter than their own spill-unspill round trip" bit, but I don't see a sequence of operations where it's clearly wasted.

For example:

1. initiate device-to-host spill
2. spill completes / free device memory
3. initiate host-to-device unspill
4. unspill completes

IIUC, the spill/unspill round-trip is $t_4 - t_1$, and the time spent spilled is $t_3 - t_2$. I'd think that $t_4 - t_1$ would always be larger.

Am I misunderstanding the definition of these concepts?

madsbk:
> Maybe this would be answered by looking at the code, but I don't see how we can judge whether some spill was "wasted" just by comparing the spill/unspill time with the duration it was spilled... I reread the "shorter than their own spill-unspill round trip" bit, but I don't see a sequence of operations where it's clearly wasted.
> 
> For example:
> 
>     1. initiate device-to-host spill
> 
>     2. spill completes / free device memory
> 
>     3. initiate host-to-device unspill
> 
>     4. unspill completes
> 
> 
> IIUC, the spill/unspill round-trip is t 4 − t 1 , and the time spent spilled is t 3 − t 2 . I'd think that t 4 − t 1 would always be larger.
> 
> Am I misunderstanding the definition of these concepts?

Good question. The comparison was never against $t_4 - t_1$. It was against $(t_2 - t_1) + (t_4 - t_3)$, the two copy durations summed, which leaves out $t_3 - t_2$, the time the data sits at rest. You are right that $t_4 - t_1$ is always larger, but that is not what the round trip stood for here.

So the predicate was $t_3 - t_2 < (t_2 - t_1) + (t_4 - t_3)$, and that can go either way.

That said, your question made me rethink the wasted term itself. My thinking was that a spill whose copies cost more than the time the data spent off device had been wasted, but that is not a good definition, since in principle any amount of time spilled could be useful.

So I have dropped wasted from this PR and deferred it to a follow-up, which will look at whether the demand was durable enough to enable real progress, or whether waiting would have been preferable.

I have updated the PR description.

cpp/src/statistics.cpp
```diff
@@ -723,6 +728,41 @@ void Statistics::record_copy(
...
+    }
+    if (src == MemoryType::DEVICE && dst != MemoryType::DEVICE) {
+        // The data is being spilled. The device memory is not free until the copy has
+        // run, so the interval starts in the sink rather than here.
+        spill_token->opened = true;
+        return [spill_token](Duration, Statistics&) {
+            spill_token->since = Clock::now();
+            spill_token->freed.store(true, std::memory_order_release);
+        };
+    }
```
pentschev: Could we open the token only for an ownership-transferring relocation, rather than every device-to-host copy? `buffer_copy()` is non-destructive meaning its device source remains allocated, and callers such as `PackedData::copy()` deliberately retain that source. As written, an ordinary copy can later produce `buffer-spilled-time` even though no device memory was released. Perhaps the relocation path should explicitly open the token.

cpp/src/statistics.cpp
```diff
@@ -723,6 +728,41 @@ void Statistics::record_copy(
...
+        return [spill_token](Duration back, Statistics& statistics) {
+            if (!spill_token->freed.load(std::memory_order_acquire)) {
+                // The copy that spilled the data has yet to report, which can happen
+                // when the two copies use different streams, leaving no point to
+                // measure the interval from.
+                return;
+            }
+            Duration const window{Duration{Clock::now() - spill_token->since} - back};
+            statistics.add_duration_stat(
+                "buffer-spilled-time", std::max(Duration::zero(), window)
```
pentschev: Should this resolve the interval once both callbacks have reported instead of returning when `freed` is false? The cross-stream dependency orders the copies, but not necessarily their timing callbacks. The inbound callback can therefore permanently omit the sample, or observe `freed` only after the inbound copy began and record a clamped zero. Shared start/end state that records once both timestamps are available would avoid scheduling-dependent results.

cpp/src/statistics.cpp
```diff
@@ -723,6 +728,41 @@ void Statistics::record_copy(
...
+            );
```
pentschev: Could we close this interval when the device reservation/allocation is obtained, or revise the statistic’s documented meaning? `BufferResource::move()` allocates the destination device buffer before `buffer_copy()` begins, while this calculation ends the interval at the inbound copy’s stream-ordered start. It can therefore count allocation-to-copy queueing time even though the replacement allocation is already consuming the supposedly free capacity.

cpp/src/memory/buffer.cpp
```diff
@@ -168,7 +167,20 @@ void buffer_copy(
     // after the dst.write_access(), its last_write_event is recorded on dst.stream(). So,
     // we need the src.stream() to wait for that event.
     dst.latest_write_event().stream_wait(src.stream());
-    statistics->record_copy(src.mem_type(), dst.mem_type(), size, std::move(timing));
+
+    // The spill token follows the data.
+    std::shared_ptr<SpillTrackToken> spill_token;
+    if (dst.mem_type() == MemoryType::DEVICE) {
+        spill_token = std::move(src.spill_track_token_);  // An unspill, so it closes.
```
nirandaperera: Couldnt D2D buffer copy be a call to copy a messages? I think `dst== DEVICE` alone constitutes a spill. `src` should not be DEVICE?

cpp/src/memory/buffer.cpp
```diff
@@ -168,7 +167,20 @@ void buffer_copy(
...
+    } else if (src.mem_type() == MemoryType::DEVICE) {
+        dst.spill_track_token_ = std::make_shared<SpillTrackToken>();  // A spill.
+        spill_token = dst.spill_track_token_;
+    } else {
+        dst.spill_track_token_ = std::move(src.spill_track_token_);
```
nirandaperera: So, `spill_token` is null in this case. and then `[PINED, HOST] <-> [PINNED, HOST]` copies omit spill token during `record_copy`?

cpp/src/memory/buffer.cpp
nirandaperera: I am a bit split on this `std::move`. Now, after the `buffer_copy`, `src` state changes. What happens if I want to make a second copy? What is the spill track token's semantic on the second copy of a `src` buffer?

cpp/src/statistics.cpp
```diff
@@ -723,6 +728,41 @@ void Statistics::record_copy(
...
+            return nullptr;
+        }
+        return [spill_token](Duration back, Statistics& statistics) {
+            if (!spill_token->freed.load(std::memory_order_acquire)) {
+                // The copy that spilled the data has yet to report, which can happen
+                // when the two copies use different streams, leaving no point to
+                // measure the interval from.
+                return;
+            }
+            Duration const window{Duration{Clock::now() - spill_token->since} - back};
```
nirandaperera: can `window < 0` here? I think we shouldnt accumulate `0` via `std::max`. That incorrectly increments the duration stats count, and dilutes mean.

nirandaperera (commented):
I am a little skeptical about this change TBH. I am wondering if we should thrashing events alongside this (ie, same buffer going through several disk -> host -> disk -> host ... rounds)

cpp/src/memory/buffer.cpp
nirandaperera, line 57: On a separate note, I think we should remove `mem_type` arg in this ctr. we only use `device_buffer` for DEVICE type only. Pinned buffers always go into `Buffer` as a `HostBuffer`.

madsbk: Sound right, but it is a breaking change better done separately.

cpp/src/statistics.cpp
madsbk: Good point, fixed. `buffer_copy()` no longer touches tokens at all. Opening and closing now happens in `BufferResource::move()`, which takes its source by value and releases it, so nothing can open a spill without freeing something. `PackedData::copy()` and the other direct `buffer_copy()` callers keep their source and are no longer counted. Covered by `ACopyThatKeepsItsSourceIsNotASpill`.

cpp/src/statistics.cpp
madsbk: Agreed with the problem, though the fix went the other way. Both ends are now host-side calls made in order on one thread, so there are no callbacks to race, nothing to drop and nothing to clamp. The sink, the `freed` flag and the `StreamOrderedTiming` sink parameter are all gone.

cpp/src/statistics.cpp
madsbk: Agreed, fixed, and you were right on either reading of the timeline, since the allocation precedes the copy in both host order and stream order. The interval now closes at `make_buffer()`.

cpp/src/memory/buffer.cpp
madsbk: Right, the condition was under-specified. It is now structural rather than conditional: `BufferResource::move()` returns early when source and destination share a memory type, so a device to device relocation never reaches the token at all.

cpp/src/memory/buffer.cpp
madsbk: They did, and the asymmetry was confusing. `record_copy` no longer takes a token in any case. A host to host relocation hands the token to the destination and records nothing, which is now the only thing it could do.

cpp/src/memory/buffer.cpp
madsbk: Agreed, and @pentschev raised the same thing from the other direction. `buffer_copy()` no longer touches tokens, so copying a buffer twice is unremarkable and neither copy is a spill. Only `move()` opens one, and it consumes its source, so a second copy cannot arise.

cpp/src/statistics.cpp
madsbk: Agreed on both counts. Both ends are now host calls ordered on one thread, so there is no clamp and no zero sample. You were right about the failure mode in general: I hit exactly it elsewhere, opening tokens for zero-size relocations, which added near-zero samples and pulled the mean down.

One thing your comment does not fix, which is worth recording. `Clock` is `std::chrono::high_resolution_clock`, whose `is_steady` is false on our toolchain, where it aliases `system_clock`. A clock step could therefore still produce a negative interval. That affects every duration statistic in the library rather than this one, so it seems like separate work.

madsbk:
> I am a little skeptical about this change TBH. I am wondering if we should thrashing events alongside this (ie, same buffer going through several disk -> host -> disk -> host ... rounds)

Good idea, but harder than it looks.

Counting rounds needs the data to stay identifiable across an unspill, and it does not. The token rides on a `Buffer`, and the unspilled `Buffer` is consumed, unpacked into a libcudf table whose memory RapidsMPF does not own. The next spill packs a fresh one, so every round would look like the first.

madsbk:
Thanks @nirandaperera and @pentschev. Your comments made me question the stream-ordered tracking altogether, and I no longer think it should be stream ordered, both for usefulness and for simplicity.

CUDA allocation and deallocation are stream ordered, but RMM's accounting is not. It records at the host call, and `memory_available()` is `limit - current_allocated()`, so every reservation and every spill trigger is checked against that number. Measuring against the stream would describe when the bytes physically become reusable, which nothing consults.

The two are ordered consistently anyway, since `buffer_copy()` makes the source's stream wait for the copy before the source is released, so a consumer never receives bytes the GPU is still reading.

Taking the accounting window also made the code much simpler :)

docs/source/statistics.md
```diff
@@ -7,6 +7,7 @@ This table gives an overview of the different statistics collected.
 | --- | --- |
 | `alloc-{memtype}` | Bytes allocated via `BufferResource::allocate()`, broken down by memory type (`device`, `pinned_host`, `host`). Shows total bytes, total time, allocation throughput, and average stream delay. |
 | `copy-{src}-to-{dst}` | Amount of data copied between memory types by RapidsMPF. `{src}` and `{dst}` are `device`, `pinned_host`, or `host`. Shows total bytes, total copy time, throughput, and average stream delay (time between CPU submission and GPU execution of the copy). |
+| `buffer-spilled-time` | How long spilled data remains spilled until unspilled back into device memory. |
```
pentschev, line 10: Can we clarify that this duration runs from the host-side deallocation accounting of the source device buffer to the host-side allocation of its replacement? The implementation intentionally no longer uses stream-ordered copy boundaries, but the PR description still says it measures from spill-copy completion to unspill-copy start and references the removed `StreamOrderedTiming` sink.

pentschev (approved):
One more doc improvement ask, otherwise LGTM. Thanks Mads.

docs/source/statistics.md
madsbk: Right, the PR description was stale on both counts and is now updated.

I would rather keep it out of the docs table though, since that is a one-line overview per statistic. `SpillTrackToken` documents the implementation detail, which is where anyone needing it will be looking.

cpp/include/rapidsmpf/statistics.hpp
```diff
@@ -30,6 +31,29 @@
...
+ *  - Opened when a relocation frees a device buffer.
+ *  - Carried along by further relocations, so one token follows the data.
+ *  - Closed when a relocation allocates a device buffer again, recording the interval.
+ *
+ * A copy that keeps its source frees nothing and opens no token, and data freed while
+ * spilled is never closed and never recorded.
+ *
+ * **Stream ordering.** The interval is not stream ordered. CUDA allocation and
+ * deallocation are, but RMM's accounting is not, and the accounting is what
+ * reservations are checked against and what triggers spilling. So a spill is measured
```
nirandaperera: I am a little skeptical about this statement. Driver immediately accounts for async allocs and deallocs. So, in that sense, both rmm accounting and driver accounting happens at the same time. In rapidsmpf, the cuda events, host callbacks, etc are actually temporally tied to stream execution order IINM.

cpp/src/memory/buffer_resource.cpp
```diff
@@ -291,23 +298,51 @@ std::unique_ptr<Buffer> BufferResource::move(
...
+    auto token = std::move(buffer->spill_track_token_);
+
+    // Closes the spill: `make_buffer` takes the capacity back here.
+    auto ret = make_buffer(nbytes, buffer->stream(), reservation);
+    if (tracked && reservation.mem_type_ == MemoryType::DEVICE && token != nullptr) {
+        statistics_->add_duration_stat(
+            "buffer-spilled-time", Duration{Clock::now() - token->since}
+        );
+        token.reset();
+    }
```
nirandaperera, line 331: are `token` and `statistics` mutually exclusive? ie. if stats is disabled, would `token` be always nullptr? That the impression I got with L339. But I feel like that relationship needs to be explicitly checked.

cpp/tests/test_buffer_resource.cpp
```diff
@@ -943,3 +944,178 @@ TEST(BufferResource, DeviceMrIsAddressableByMemoryRecorder) {
...
+        };
+    }
+
+    /// @brief The number of samples recorded under `name`.
+    std::size_t samples(std::string const& name = "buffer-spilled-time") const {
+        try {
+            return stats->get_stat(name).count();
+        } catch (std::out_of_range const&) {
+            return 0;
+        }
```
nirandaperera: Nit. Maybe querying if a stat is available, something we can cheaply provide in `Statistics` obj?

nirandaperera (commented):
just some comments, but overall I like this approach

cpp/include/rapidsmpf/statistics.hpp
madsbk: You are right. It is not stream ordered versus unordered, it is a choice of which clock to measure on. Reworded to claim only what I can back:

> **Stream ordering.** The interval is not stream ordered. Both ends are taken at the host calls that move `BufferResource::memory_available()`, which is what reservations are checked against and what triggers spilling, rather than at the stream positions where the copies run.

cpp/src/memory/buffer_resource.cpp
madsbk: Not quite, so the question is a fair one. A token is only opened when statistics are enabled, but one can also arrive through the device-buffer overload of `move()`, which cudf-streaming uses and which never consults statistics. So a token can exist while statistics are off.

The close is still correct, because `add_duration_stat` is a no-op when disabled, but that was doing unstated work. Added a comment there rather than a check, since a check would only duplicate what `add_duration_stat` already does.

cpp/tests/test_buffer_resource.cpp
madsbk: Agreed, added `Statistics::has_stat()`. Exceptions were the only way to ask, which is a gap in the API rather than just a test wart. The test helper is now a one-liner.

cpp/src/memory/buffer_resource.cpp
nirandaperera: fair enough

nirandaperera (approved):
LGTM. Thanks @madsbk

PR #1222, opened 2026-09-18
Remove explicit GB300 CI runners

pentschev:
GB300 CI runners have been removed, see https://github.com/rapidsai/shared-workflows/pull/640 .


bdice (approved):
Thanks. If this affects gh200 as well, you may need to filter those jobs similarly.

pentschev:
> Thanks. If this affects gh200 as well, you may need to filter those jobs similarly.

Unlikely, but we'll have to wait and see.

msarahan (approved):

